PR #13, opened 2018-08-28
Added '/' route to API

hrishiballal:
Based on discussion in #12 , creating a new "default" route to the API


hi-kevin:
Can you add to the testStatus unit test in storage_api_test.py? I think if you copy lines 39-41 and change /status to /, that should test all of the code.

hrishiballal:
OK thanks for this, I will send on a second pull request.

PR #17, opened 2018-09-10
Bug fixes for slippy, dates, and OAuth key

hi-kevin:
Added uss_metadata.py to separate logic from data structure for better maintenance,
Updated docker script to 0.0.0.0 out local servers as required for Zookeeper,
Slippy errors fixe, input standardized to CSV of lat,lon, and output more JSON friendly,
Date parsing is more robust, handling timezone aware and timezone unaware dates,
Updated OAuth public key parsing to handle the way they are formatted as environment variables,
Updated version to 1.0.1 to cover changes in slippy api, Swaggerhub updates in progress.

### DIFF
--- a/datanode/docker/datanode_entrypoint.sh
+++ b/datanode/docker/datanode_entrypoint.sh
@@ -34,7 +34,11 @@ echo "syncLimit=$ZOO_SYNC_LIMIT" >> "$ZOO_CONFIG"
 echo "maxClientCnxns=$ZOO_MAX_CLIENT_CNXNS" >> "$ZOO_CONFIG"
 
 for server in $ZOO_SERVERS; do
-    echo "$server" >> "$ZOO_CONFIG"
+    if [[ $server = "server.${ZOO_MY_ID}"* ]]; then
+        echo "$server" | sed 's/=[0-9.]*:/=0.0.0.0:/' >> "$ZOO_CONFIG"
+    else
+        echo "$server" >> "$ZOO_CONFIG"
+    fi
 done
 
 # Write myid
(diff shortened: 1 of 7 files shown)


datanode/src/storage_api_test.py
```diff
@@ -101,35 +101,35 @@ def testSlippyConversionWithValidData(self):
     r = self.app.get('/slippy/11?coords=1,1')
     self.assertEqual(r.status_code, 200)
     j = json.loads(r.data)
-    self.assertEqual(j['data']['tiles'][0], [11, 1029, 1018])
-    self.assertEqual(j['data']['links'][0],
-                    'http://tile.openstreetmap.org/11/1029/1018.png')
-    r = self.app.get('/slippy/10?coords=37.203335,-80.599481')
-    r = self.app.get('/slippy/10?coords=37.203335,-80.599481')
+    self.assertEqual(11, j['data']['grid_cells'][0]['zoom'], 11)
```
BenjaminPelletier: I think we want better assert failure messages here (third argument to assertEqual), or omit the third argument.

datanode/docker/datanode_entrypoint.sh
```diff
@@ -34,10 +34,14 @@ echo "syncLimit=$ZOO_SYNC_LIMIT" >> "$ZOO_CONFIG"
 echo "maxClientCnxns=$ZOO_MAX_CLIENT_CNXNS" >> "$ZOO_CONFIG"
 
 for server in $ZOO_SERVERS; do
-    echo "$server" >> "$ZOO_CONFIG"
+    if [[ $server = "server.${ZOO_MY_ID}"* ]]; then
+        echo "$server" | sed 's/=[0-9.]*:/=0.0.0.0:/' >> "$ZOO_CONFIG"
```
BenjaminPelletier: Seems like this only replaces IP-specified servers with 0.0.0.0.  How about
```echo "$server" | sed 's/server\.\([0-9]\+\)=[^:]*:/server.\1=0.0.0.0:/'```

datanode/src/storage_api_test.py
hi-kevin: Roger. Typo.

PR #31, opened 2018-10-04
Update docker configuration and enable gunicorn

### LINKED ISSUE #30: sync_token is not supported in headers on master branch
Commit 5e28b35ffd13220aa311d595024a24c6874700fb added support for sync_token in the tcl4 branch, but that support has not yet been incorporated into the master branch.

BenjaminPelletier:
This PR primarily updates the docker configuration to a much more powerful and robust architecture, but one that must be brought up with docker-compose (orchestrating several containers) rather than a single container.  In the new architecture, only an nginx reverse proxy is exposed to the outside world and it proxies HTTP requests to the storage API in a separate container, as well as acting as an SSL terminator proxying HTTPS requests to the same storage API.  Zookeeper is run in a separate container using an unmodified image.

The storage API only serves unprotected HTTP (and has had its SSL option removed), but it is hidden inside the container system and can only interact with outside world via the nginx reverse proxy.  The Zookeeper container's election ports are open to the outside world, but its API port (2181) is only exposed to other containers.

By default, the data node will use the test_* certificate whose private key is included in the repository.  Doing this will print a big WARNING message that this is an unsafe practice as doing this is intended only as a demonstration of capabilities.  The expectation is that production users will use the docker-compose_localssl.yaml override to provide their own certificate.

Fixes #30.

From the commit message:
----
Update docker configuration to bring up the data node via docker-compose
so that each component lives in its own container rather than in a
single container.

Switch from Flask to gunicorn WSGI server; this necessitates ensuring
startup logic in storage_api.py is executed when main() is not used and
it is invalid to extract options from command line arguments.

Add an nginx reverse proxy to the data node to serve HTTPS, which allows
the SSL options to be removed from storage_api.py.

Add a Postman collection for testing of a local data node.

Fix a few typos and details, and get sync_token from header for issue #30.

Update all documentation as needed.


datanode/docker/Dockerfile_reverseproxy
```diff
@@ -0,0 +1,46 @@
...
+#     -e STORAGE_API_SERVER="storage_api_ip" -e STORAGE_API_PORT="8121" \
+#     -p 8121:8121 -p 8120:8120 -d interussplatform/reverse_proxy
+
+# Required environment variables:
+#   STORAGE_API_SERVER: Address of the storage API service.
+#   STORAGE_API_PORT: Port on which the storage API service is listening.
+
+# Optional environment variables:
+#   INTERUSS_API_PORT_HTTP: Port on which the InterUSS API will be served via HTTP (default 8121).
+#   INTERUSS_API_PORT_HTTPS: Port on which the InterUSS API will be served via HTTPS (default 8120).
```
hi-kevin: Can we reverse these? 8120 should be http, 8121 for https. This aligns to lower number, lower functionality, and maps to the current deployment.

Also, how would you turn OFF http? not saying it is a requirement, but if someone wanted to run only a SSL server, is it possible only with envoirnmental variables?

datanode/docker/Dockerfile_reverseproxy
```diff
@@ -0,0 +1,46 @@
...
+
+FROM nginx
+
+COPY nginx_entrypoint.sh .
+COPY test_cert.pem /etc/ssl/certs/cert.pem
+COPY test_key.pem /etc/ssl/private/key.pem
+COPY test_signatures .
+
+ENV INTERUSS_API_PORT_HTTP 8121
+ENV INTERUSS_API_PORT_HTTPS 8120
```
hi-kevin: As above, 8120 http, 8121 https.

datanode/docker/Dockerfile_storageapi
```diff
@@ -32,19 +32,21 @@
...
 #   INTERUSS_API_SERVER: IP address that will respond to InterUSS API requests (default 0.0.0.0)
 #   INTERUSS_VERBOSE: True for more log messages from the InterUSS API (default true)
+#   INTERUSS_TESTID: Force testing mode with test data located in specific test id
 
 FROM python:2
 
 WORKDIR /usr/src/app
 
-RUN pip install --no-cache-dir python-dateutil kazoo flask PyJWT djangorestframework cryptography pyopenssl
+RUN pip install --no-cache-dir python-dateutil kazoo flask PyJWT djangorestframework cryptography pyopenssl pytz gunicorn
```
hi-kevin: Can we get rid of flask? What about pyopenssl?
If we can get rid of them, don't forget to delete the import statements too.

datanode/tools/calculate_datanode_metrics.bash
```diff
@@ -1,7 +1,7 @@
 #/bin/bash
```
hi-kevin: This script doesn't work in a Docker config (nor would it work with gunicorn), correct? Let's get rid of it and we can have a separate issue for adding the metrics we need, even in a Docker config.

datanode/docker/Dockerfile_reverseproxy
BenjaminPelletier: Yep, will reverse.

There isn't a way to turn off HTTP in this PR, but I'm intending to add an option to have nginx issue 301 redirects on the HTTP endpoint.  It's not on this PR because this PR is already pretty big.

datanode/docker/Dockerfile_reverseproxy
BenjaminPelletier: Done.

datanode/tools/calculate_datanode_metrics.bash
BenjaminPelletier: Deleted.

PR #34, opened 2018-10-12
Additional slippy conversions for path and polygon

hi-kevin:
This creates a slippy utility for point, path and polygon to correctly identify the tiles impacted at the specified zoom level, including in cases where a vertex extends over more than two tiles. This fixes #33. 

This also brings forward some bug fixes from TCL4 to support various OAuth in the header and makes unit tests a little more robust.

PR #36, opened 2018-10-15
slippy non-breaking api changes to support path/polygon

hi-kevin:
This fixes issue #34 , as it allows an optional parameter to /slippy to support path/polygon (defaulted to the current capability of point). 

This will be published as 1.1.0 once the multigrid operations are added, but for now is version 1.0.2.004 internally. Swaggerhub draft changes for 1.1.0 are here:
https://app.swaggerhub.com/apis/InterUSS_Platform/data_node_api/1.1.0#/

### DIFF
--- a/datanode/src/storage_api_test.py
+++ b/datanode/src/storage_api_test.py
@@ -86,14 +86,14 @@ class InterUSSStorageAPITestCase(unittest.TestCase):
     endpoint = 'https://utmalpha.arc.nasa.gov//fimsAuthServer/oauth/token?grant_type=client_credentials'
     headers = {'Authorization': 'Basic ' + os.environ.get('FIMS_AUTH', '')}
     r = requests.post(endpoint, headers=headers)
-    self.assertEqual(r.status_code, 200)
+    self.assertEqual(200, r.status_code)
     token = r.json()['access_token']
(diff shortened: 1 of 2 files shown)


datanode/src/storage_api_test.py
```diff
@@ -113,17 +113,17 @@ def testSlippyConversionWithInvalidData(self):
...
     r = self.app.get('/slippy/11?coords=1,1')
-    self.assertEqual(r.status_code, 200)
+    self.assertEqual(200, r.status_code)
     j = json.loads(r.data)
     self.assertEqual(11, j['data']['grid_cells'][0]['zoom'])
     self.assertEqual(1029, j['data']['grid_cells'][0]['x'])
     self.assertEqual(1018, j['data']['grid_cells'][0]['y'])
     self.assertEqual('http://tile.openstreetmap.org/11/1029/1018.png',
                      j['data']['grid_cells'][0]['link'])
-    r = self.app.get('/slippy/10?coords=37.408959,-122.053834')
```
BenjaminPelletier: nit: might want to also keep this test for backwards compatibility checking

hi-kevin: Test for those lat/lon is still there, just with the point parameter. The non "point" parameter is tested above as well.

Issue #35, opened 2018-10-15
Add endpoint for GET/PUT/DELETE to multiple grids at once


hi-kevin:
**Is your feature request related to a problem? Please describe.**
Currently, the USS must decide which grids are affected by their operation, GET data from each, deconflict, then PUT to each grid. This may be difficult to identify the grids affected and handle the case of a write conflict.

**Describe the solution you'd like**
A single endpoint to support multiple grids. 

Note, if a USS has concerns about providing details of their operations to the platform, they can obfuscate their position (Iarger), provide a bounding box, or use the single grid endpoints.

**Additional context**
This endpoint must handle the rollback if a write conflict happens, and not lock itself nor leave hanging data if there is a failure in writing. It must also provide an easy to use data model for iterating through deconfliction.

Remember, to best support collaboration among USSs, there are two key tenets to the InterUSS Platform:
1. The InterUSS Platform is focused on facilitating communication amongst actively operating USSs.
2. No details about UAS operations are stored or processed on the InterUSS Platform. 

Regardless of the new features and changes brought to the platform via the open source process, these two tenets must be enforced to ensure protection of operator and consumer privacy and ensure the competition between USSs for drone services remains intact.

hi-kevin:
Fixed in the latest release!

PR #38, opened 2018-10-19
API Changes to support multi-grid GET/PUT/DEL

hi-kevin:
Added endpoint /GridCellsMetaData to support multiple grid changes via a set of points, a path or an enclosed polygon. This fixes Issue #35 .

Swaggerhub is updated as well as the server that drives it, available here to test:
https://app.swaggerhub.com/apis/InterUSS_Platform/data_node_api/1.1.0

### DIFF
--- a/datanode/src/slippy_util.py
+++ b/datanode/src/slippy_util.py
@@ -41,6 +41,8 @@ from shapely.geometry import Polygon
 logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.INFO)
 log = logging.getLogger('InterUSS_DataNode_SlippyUtil')
 
+TILE_LIMIT = 50  # Number of tiles to allow for multi get/put/del
+
 
 def validate_slippy(z, x, y, raise_error=False):
   """Validates slippy tile ranges.
@@ -258,6 +260,9 @@ def _calculate_tiles_from_bounding_box(zoom, bbox):
     x, y = convert_point_to_tile(zoom, lat, lon)
     if (x, y) not in result:
       result.append((x, y))
+      if len(result) > TILE_LIMIT:
+        raise OverflowError('Limit of %d tiles impacted exceeded'
+                            % (TILE_LIMIT))
     last = (lat, lon)
   if not result:
     raise ValueError('No tiles found for path coordinates')
--- a/datanode/src/slippy_util.py
+++ b/datanode/src/slippy_util.py
@@ -187,29 +187,6 @@ def convert_polygon_to_tiles(zoom, points):
   return _convert_shape_to_tiles(zoom, points, is_poly=True)
 
 
-def convert_points_to_geojson(points, headers=True, name='slippy-out'):
-  """ Converts a set of lat/lon points into a path in geojson format."""
-  result = ''
-  if headers:
-    result += """{
-      "type": "FeatureCollection",
-      "features": ["""
-  result += """{
-      "type": "Feature",
-      "properties": {"name" : "%s" },
-      "geometry": {
-        "type": "Polygon",
-        "coordinates": [ [ """ % name
-  for lat, lon in points:
-    result += "[ %.5f, %.5f ]," % (lon, lat)
-  result = result[:-1] + """] ]
-        }
-      }"""
-  if headers:
-    result +=' ] }'
-  return result
-
-
 ######################################################################
 ################       INTERNAL FUNCTIONS    #########################
 ######################################################################
--- a/datanode/src/storage_api.py
+++ b/datanode/src/storage_api.py
@@ -512,7 +512,7 @@ def _DeleteGridCellsMetaData(zoom, tiles, uss_id):
   log.info('Grid cells metadata delete instantiated for %s, %sz, %s...',
            uss_id, zoom, str(tiles))
   if uss_id:
-    result = wrapper.delete_multi(uss_id, zoom, tiles)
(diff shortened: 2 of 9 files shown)


datanode/src/slippy_util.py
```diff
@@ -41,6 +41,8 @@
 logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.INFO)
 log = logging.getLogger('InterUSS_DataNode_SlippyUtil')
 
+TILE_LIMIT = 50  # Number of tiles to allow for multi get/put/del
```
BenjaminPelletier: Do we foresee a use case for this many tiles?  If operations can only reasonably be as big as a single tile, then we'd want this limit to be 4.  If they might be slightly larger than a tile, then 9 might be appropriate.  50 seems very large.

datanode/src/storage_api.py
```diff
@@ -380,6 +405,149 @@ def _DeleteGridCellMetaData(zoom, x, y, uss_id):
...
+        status.HTTP_400_BAD_REQUEST,
+      'message':
+        """uss_id must be provided in the request to
+          delete a USS from a GridCell."""
+    }
+  return result
+
+
+def _ConvertCSVtoTiles(zoom):
+  """Converts an CSV of coords into slippy tile format at the specified zoom
```
BenjaminPelletier: I'd suggest making explicit that the `coords` and `coord_type` pseudo-arguments come from global request parameters (renaming function as above would probably be sufficient IMO).

datanode/src/storage_api.py
```diff
@@ -195,6 +181,48 @@ def GridCellMetaDataHandler(zoom, x, y):
...
 
+@webapp.route(
+  '/GridCellsMetaData/<zoom>',
+  methods=['GET', 'PUT', 'POST', 'DELETE'])
+def GridCellsMetaDataHandler(zoom):
+  """Handles the web service request for multi-grid operations.
+
+  Args:
+    zoom: zoom level in slippy tile format
+    x: x tile number in slippy tile format
```
BenjaminPelletier: Are `x` and `y` accepted as arguments at all here?

datanode/src/storage_api.py
```diff
@@ -195,6 +181,48 @@ def GridCellMetaDataHandler(zoom, x, y):
...
+    y: y tile number in slippy tile format
+    OAuth access_token as part of the header
+    Plus posted webargs:
+      coords: csv of lon,lat,long,lat,etc.
+      coord_type: (optional) type of coords - point (default), path, polygon
+      and additional as needed for PUT/POST and DELETE methods (see below)
```
BenjaminPelletier: What is "see below" referring to?

datanode/src/storage_api.py
```diff
@@ -380,6 +405,149 @@ def _DeleteGridCellMetaData(zoom, x, y, uss_id):
...
+      'code':
+        status.HTTP_400_BAD_REQUEST,
+      'message':
+        """uss_id must be provided in the request to
+          delete a USS from a GridCell."""
+    }
+  return result
+
+
+def _ConvertCSVtoTiles(zoom):
```
BenjaminPelletier: Maybe name this `_GetTilesFromRequest` since it doesn't actually accept a CSV to convert?

datanode/src/storage_api.py
```diff
@@ -380,6 +405,149 @@ def _DeleteGridCellMetaData(zoom, x, y, uss_id):
...
+
+  Returns:
+    200 and a new composite token if updated successfully,
+    409 if there is a locking conflict that could not be resolved, or
+    the other nominal 4xx error codes as necessary.
+  """
+  log.info('Grid cells metadata submit instantiated for %s at %sz, %s...',
+           uss_id, zoom, str(tiles))
+  sync_token = _GetRequestParameter('sync_token', None)
+  if not sync_token and 'sync_token' in request.headers:
```
BenjaminPelletier: It might be a good idea to move the header check into _GetRequestParameter so we don't have to remember to repeat it whenever we want to get the sync token.

datanode/src/storage_api_test.py
```diff
@@ -223,96 +221,96 @@ def testIncorrectPutsOnGridCells(self):
     j = json.loads(result.data)
     s = j['sync_token']
     self.assertEqual(404, self.app.put(
-        '/GridCellMetaDatas/1/1/1',
```
BenjaminPelletier: It's not important, but isn't the original 4-space indent following a paren the standard rather than the 2-spaces this PR changes it to?

datanode/src/slippy_util.py
hi-kevin: For the main platform, zoom level 14 or 15 is expected. That gives a width of < 1km. 45 degree angle of a 10km flight might be near 40? Performance-wise, 50 seems manageable in web timeouts, since it is committed as a transaction.

datanode/src/storage_api.py
hi-kevin: Nope, fixed.

datanode/src/storage_api.py
hi-kevin: Done.

datanode/src/storage_api.py
hi-kevin: Moving it to GetRequestParameter would cause all request parameters to be allowed in the header, and we have only specifically allowed access_token and sync_token.

Issue #39, opened 2018-10-22
TCL4 branch:  UVRs should be stored in grid


nasajoey:
**Problem**
To support future operations, there is a gap related to knowing about any dynamic airspace restrictions (to be known as UAS Volume Reservations or UVRs).

**Proposed Solution**
_This is a TCL4 problem/solution._  We should explore storing UVRs directly in the grid.  This would allow a new entrant to a given grid to immediately have access to the state of that airspace.  UVRs should only be writeable by USSs with an enhanced role, thus we'd have to check for a different scope in the access_token provided.  Only the USS that writes the UVR would have rights to modify or delete.  We may want a feature that automatically expires/deletes UVRs after some time (like within an hour of the UVR's end time), but we may want to wait on that and track as an additional github issue.

**Describe alternatives you've considered**
Alternatively, one could force enhanced USSs to serve these data to other USSs upon request.  There are issues with this approach.  First and foremost, how do we guarantee that such a USS is always available in a given grid?  And how are they easily identified from other USSs in the grid (some new metadata tag?)?

**Additional context**
Will need this sooner than later.  We can maybe target Sprint 4, but would take it whenever it might be available.

The model written to grid may leverage the existing UVR schema:  https://github.com/nasa/utm-apis/blob/3189acdb3ce13dd64258cecb0e4f3c40c158a202/utm-domains/utm-domain-commons.yaml#L1270

myspri:
@BenjaminPelletier, we have updated UVR model 
https://github.com/nasa/utm-apis/blob/v4-draft/utm-domains/utm-domain-commons.yaml

This is usually announced in the uss_implementor's channel.
Going forward, we will give you advance notice that will allow our models to be in sync before the announcement goes out.

Can you please update your model based on this?

We have been trying to codegen from your version and have encountered a problem with "allOf" attribute. We are currently blocked. I believe the v4-draft does not include this attribute.

Hopefully, this problem will go away after you sync up with our version.

drsmit22:
I have also noticed a "timestamp" field being added to UASVolumeReservation when it is being returned from the grid that is causing trouble.

issmith1:
If the grid has old prior-version models that are returned via GET, that can happen.   We need to clean out incompat models.

BenjaminPelletier:
Please open a new issue for each new issue.  This issue has already been resolved.

BenjaminPelletier:
[...]

Issue #20, opened 2018-09-13
Not all status codes in storage_interface.py can be returned by Flask


BenjaminPelletier:
Flask's abort method only recognizes some of the HTTP codes, and storage_interface.py uses codes that abort does not recognize.  When one of these unrecognized codes is passed to abort, it causes an HTTP 500 due to a LookupError in werkzeug.  HTTP 424 should be replaced as a return code in storage_interface.py.

hrishiballal:
According to Flask API spec the closest one that I can see is [HTTP 412](https://www.flaskapi.org/api-guide/status-codes/) "Precondition failed". Would that work? I can make a PR to return 412 with a note that it is really 424.

hi-kevin:
Actually, this issue should revamp the storage_interface error handling. It currently uses HTTP codes but should have it's own enum, and the API lauerr should decide how this maps to http codes. There is a TODO in the storage interface already. Let's use this issue to do it properly.

hi-kevin:
This has been patched in the TCL4 branch. This issue will remain open to create an error enum for storage_interface and update the API to match enum to HTTP error codes that Flask supports.

Issue #19, opened 2018-09-13
The InterUSS DataNode occasionally returns a 500 HTTP status code.


vbaskara:
2018-09-12 15:36:05 [pool-0] INFO gov.nasa.utm.ussn.ussapiservice.ConcurrencySender:93 Making a rest call http://35.193.160.127:8121/GridCellOperator/9/73/191 GET <{Content-Type=[application/json]}>
2018-09-12 15:36:05 [pool-0] ERROR gov.nasa.utm.ussn.ussapiservice.ConcurrencySender:100 
HttpStatusCodeException code:500 body:<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>500 Internal Server Error</title>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error and was unable to complete your request.  Either the server is overloaded or there is an error in the application.</p>

}

BenjaminPelletier:
This particular 500 was being caused by an ImmatureSignatureError in JWT while validating the access token.  It was corrected on the tcl4 branch in commit 3fff76100a33e0cd1a83bd3b6aaac18b4f9e1483

hi-kevin:
Closing, as it has been corrected, and issue #20 will address refactoring all error handling.

PR #44, opened 2018-10-29
Add in-grid UVR support

BenjaminPelletier:
This PR adds the concept of a [UVR](https://github.com/nasa/utm-apis/blob/3189acdb3ce13dd64258cecb0e4f3c40c158a202/utm-domains/utm-domain-commons.yaml#L1270) and the ability to store them in the grid per issue #39.  The UVRs in this PR are specifically matched to NASA specifications.  The intention going forward is that we will evaluate the pros and cons of this means of sharing flight restriction information before developing something more appropriate for the core InterUSS Platform.  These changes will not be merged to master as-is.

```uvrs.py``` encapsulates a UVR itself, which is essentially a duck-typed dict containing the valid parameters, plus a timestamp when emplaced in the grid (but differing timestamps do not change equality of UVRs).  Adherence to the spec is ensured here; ```Uvr``` instances are always in conformance.

These UVR objects are added to ```USSMetadata``` in ```ussmetadata.py```.  Some formatting is pulled out into ```formatting_utils.py``` for reuse and the UVRs are stored in memory as ```Uvr``` instances while automatically converting to JSON-serializable nested dicts upon ```to_json()```.  Avoidance of ```message_id``` collision is ensured here on a per-cell basis.

UVR grid emplacement is added to ```storage_interface.py```, which prohibits removal of UVRs whose spec differs from already-emplaced UVRs.  This means that no UVR may be removed unless the entire UVR matches the spec with which it was emplaced, which means the coordinate bounds will match, which means a UVR cannot be removed unless it is removed from exactly the set of grid cells  in which it was emplaced.  This prevents partial erroneous UVR removal.

Finally, the UVR/zoom/message_id endpoint is added in ```storage_api.py```.  This is where the inputs are validated including matching ```message_id``` in URL and UVR, matching USS ID in ```access_token``` to ```originator_id``` in UVR, and validating the incoming UVR JSON against the UVR spec.

Test cases are added to both ```storage_interface_test.py``` and ```storage_api_test.py``` which share some routines factored out into ```test_utils.py```.

### CODE REVIEW

File: datanode/src/storage_api.py [hunk 1]
```diff
@@ -81,7 +83,10 @@
```
hi-kevin: Add 4.0.3.009, as this adds to the API but is non breaking.

datanode/src/storage_api.py
```diff
@@ -292,17 +297,86 @@ def GridCellsOperationDataHandler(zoom, gufi):
...
+  if not uss_id:
+    return {
+      'status': 'fail',
+      'code': status.HTTP_400_BAD_REQUEST,
+      'message': 'uss_id must be provided in the request to modify a UVR.'}
+
+  uvr_json = _GetRequestParameter('uvr', None)
+  if not uvr_json:
+    return {
+      'status': 'error',
```
hi-kevin: Status should be "fail" if the problem is with the data submitted. https://labs.omniti.com/labs/jsend

datanode/src/storage_api.py
```diff
@@ -292,17 +297,86 @@ def GridCellsOperationDataHandler(zoom, gufi):
...
+  except ValueError as e:
+    abort(status.HTTP_400_BAD_REQUEST, 'Error parsing UVR JSON: ' + e.message)
+
+  try:
+    uvr = uvrs.Uvr(unvalidated_uvr)
+
+    if uvr['message_id'] != message_id:
+      raise ValueError('message_id "%s" in request does not match message_id '
+                       '"%s" in UVR' % (message_id, uvr['message_id']))
+    if uvr['originator_id'] != uss_id:
```
hi-kevin: uss_id is created by the OAuth provider, originator_id is created by the USS, and I don't think we want the USS to specify it. In gridcell metadata, we specify a request object that does not contain the fields we write (version, uss_id, timestamp, etc.).

datanode/src/storage_api.py
```diff
@@ -292,17 +297,86 @@ def GridCellsOperationDataHandler(zoom, gufi):
...
+      raise OverflowError('Limit of %d tiles impacted exceeded (%d)'
+                          % (UVR_TILE_LIMIT, len(tiles)))
+  except (KeyError, ValueError, TypeError) as e:
+    abort(status.HTTP_400_BAD_REQUEST, e.message)
+  except OverflowError as e:
+    abort(status.HTTP_413_REQUEST_ENTITY_TOO_LARGE, e.message)
+  if request.method in ('PUT', 'POST'):
+    result = wrapper.insert_uvr(zoom, tiles, uvr)
+  elif request.method == 'DELETE':
+    result = wrapper.delete_uvr(zoom, tiles, uvr)
```
hi-kevin: Shouldn't delete only require a UVR identifier? This makes the API straight forward with nothing in the body, and we want to make removal as easy as possible (consider failover, expiration, etc.)

datanode/src/storage_interface.py
```diff
@@ -586,6 +629,58 @@ def delete_multi_operation(self, z, grids, uss_id, gufi):
...
+          if not slippy_util.validate_slippy(z, x, y):
+            raise ValueError('Invalid slippy grids for lookup')
+          (content, metadata) = self._get_raw(z, x, y)
+          if metadata:
+            m = uss_metadata.USSMetadata(content)
+            removed_uvr = m.remove_uvr(uvr['message_id'])
+            if removed_uvr:
+              diff = uvrs.diff(removed_uvr, uvr)
+              if diff:
+                return self._format_status_code_to_jsend(
```
hi-kevin: Same comment as above, Delete should probably just be by ID and USS_ID.

datanode/src/uss_metadata.py
```diff
@@ -248,13 +286,44 @@ def remove_operation(self, uss_id, gufi):
...
+    """Inserts a UVR, failing if another UVR with the same ID already exists.
+
+    Args:
+      uvr: UVR instance
+
+    Raises:
+      ValueError: if a UVR already exists with the new UVR's message_id
+    """
+
+    if uvr['message_id'] in [existing['message_id'] for existing in self.uvrs]:
```
hi-kevin: This behavior is different from girdcell operator or operation, we overwrite with PUT (TCL4 didn't want to support PATCH). Can we do the same here?

datanode/src/storage_api.py
BenjaminPelletier: This was modeled off of lines 516, 610, 728, and 819 (PUT Operator(s) or Operation(s) fails with 'error' if missing fields).  Should I change those as well for consistency?

datanode/src/storage_api.py
BenjaminPelletier: I think we had previously discussed that a USS, even one with UVR write scope, should not be allowed to delete a UVR emplaced by by a different USS.  If originator_id is not uss_id, does that mean we also need an additional ```emplaced_by_uss_id``` field to be stored in the grid cell?

datanode/src/storage_api.py
BenjaminPelletier: No, at very minimum it needs the geographic extents of the UVR to find the appropriate grid cells.  We could have the coordinates as a PUT field like GridCellsOperator, but 1) that requires the client to convert between coordinates formats and 2) makes the likelihood of incomplete delete much higher.

If an emplacing client has lost the full UVR definition but still has the ```message_id``` and at least one grid cell containing it, they can easily retrieve the UVR and then parrot it back for deletion.  If they don't know at least one grid cell, then the UVR is essentially orphaned -- it's the same problem as with orphaned Operators.  If we require just the coordinates (in a different format) in the DELETE request, a client who has lost the full UVR definition will still need to query the grid cell to look up the coordinates, and additionally they will have to perform the conversion.

datanode/src/storage_interface.py
BenjaminPelletier: Deletion requires at least coordinates in addition to ```message_id```.  If the deleting client can store coordinate data about the UVR in addition to its ID, why not just store the entire UVR?  Requiring the full UVR definition, and requiring it to match, ensures that a UVR cannot be partially emplaced in the grid.

datanode/src/uss_metadata.py
BenjaminPelletier: I don't think we want to do that because part of the UVR itself is its coordinate outline.  What happens if a client tries to update a UVR in grid cells (0,0) and (0,1) with a new UVR of the same ID in grid cells (0,0) and (1,0)?  Currently, the client only needs to send the UVR itself and coordinates are extracted from it; with an update, we could no longer extract coordinates for the old UVR to be replaced from the update request so we would need to add an additional coordinates input parameter.  But, then that parameter could conflict with the UVR itself, and what should happen in that case?  There probably shouldn't be an error because that would be inconsistent with the behavior during update where error checking cannot be guaranteed.  But it also seems very odd to allow clients to update a UVR when we know coordinates specifications don't match.  I think the exact-UVR-match approach is a clean way to ensure UVR consistency in the grid while adding little to no overhead to what the client would already need to do to delete UVRs (and perhaps less overhead since full UVR specification eliminates the need to specify coordinates or grid cells separately).

datanode/src/storage_api.py
hi-kevin: Ah, I see. I think we'll make everything more consistent with #19 and #20.

datanode/src/storage_api.py
hi-kevin: I was saying originator_id is uss_id, and the USS does not specify it (it comes from their OAuth token).

datanode/src/storage_api.py
hi-kevin: Roger, makes sense.

datanode/src/uss_metadata.py
hi-kevin: Roger. We'll address it specifically as we support PATCH, as that should do the changes to the grid if the boundary changes.

datanode/src/uss_metadata.py
```diff
@@ -248,13 +286,44 @@ def remove_operation(self, uss_id, gufi):
...
+      raise ValueError('UVR with message_id %s already exists; it must be '
+                       'removed before inserting a new UVR with that ID' %
+                       (uvr['message_id']))
+
+    # Now add the new record
+    uvr['timestamp'] = format_utils.format_ts()
+    self.uvrs.append(uvr)
```
hi-kevin: I believe you need to update the version here. Can you also add a test to verify the version increases on insert and remove?

BenjaminPelletier: Thanks for catching that.  Previous codepath always went through remove_uvr and I didn't fix the version update when I changed it to not go through remove_uvr.  I've added a test that was failing, and then I fixed the test as suggested.

Issue #56, opened 2018-12-01
Create an API update process


BenjaminPelletier:
When the API is updated, the stored data format sometimes needs to change.  Because the InterUSS Platform usually uses multiple nodes, the nodes cannot be guaranteed to be using the same storage API version, especially during an update process.  Furthermore, when one node at a time is updated, any data in the grid will be retained through the update even after all nodes are updated.  This can cause problems when a new version of the storage API is implicitly expected to load data written by an old version of the storage API.  We should document the way upgrades are expected to work.

My suggestion is that we define two versions: the API version, and the GridCellMetaData version.  The GridCellMetaData version will be stored with each GridCellMetaData in the grid, and strictly define the version schema as [branch, major, minor, patch, build] where, e.g., tcl4.0.3.010 is tcl branch, major version 4, minor version 0, patch version 3, build 010.  The GridCellMetaData version would not be visible to API consumers; it would be added before writing to Zookeeper and stripped before returning the result to an API consumer.  Then, when the API reads GridCellMetaData, it checks the version of the stored data against its version.  If the version of the data in the grid differs from the version of the API reading that data:
* If the branch differs, the API indicates an error prompting a grid data purge or manual in-place update.
* If the major version is smaller than this API, the API indicates an error prompting grid data purge or manual in-place update.
* If the major version is larger than this API, the API indicates an error prompting an API update.
* If the minor version is larger than this API, the API indicates an error prompting an API update.
* If the minor version is smaller than this API, the API imports the older data format.  When a write is performed by a client, the API overwrites the old-format data with new-format data.
* If just the patch and/or build versions differ, the API reads the data normally.  It will update the version when writing, but otherwise assumes:
  * APIs both older and newer in patch.build will be able to read the GridCellMetaData structures this API writes.
  * This API will be able to read GridCellMetaData written by APIs both older and newer in patch.build.

Note that this is entirely separate from the API version, which describes how the clients interact with the server API.  It is possible to change how clients interact with the API without changing the GridCellMetaData format, and vice versa (though often the two will probably change at the same time).

This will mean that un-upgraded servers at the minor+ version level will clearly indicate non-functionality, the reason for that non-functionality, and the way to fix the non-functionality in the event they attempt to interact with data from a newer API version.  And, most importantly, it will protect against unexpectedly invalid data in the grid from an out-of-sync server.

PR #58, opened 2018-12-04
Improve validation and error messages

BenjaminPelletier:
This PR addresses a number of issues in the TCL4 branch as well as fixing and improving some tests that weren't testing the intended features.  Major improvements include improved validation of both operators and operations, and more informative error messages when requests can't be fulfilled.

### DIFF
--- a/datanode/docker/Dockerfile_storageapi
+++ b/datanode/docker/Dockerfile_storageapi
@@ -16,7 +16,7 @@
 
 # To build an image from this Dockerfile:
 # docker image build -f Dockerfile_storageapi -t interussplatform/storage_api:tcl4 ../src
-# docker tag interussplatform/storage_api:tcl4 interussplatform/storage_api:tcl4.1.0.011
+# docker tag interussplatform/storage_api:tcl4 interussplatform/storage_api:tcl4.1.1.012
 
 # To run this image:
 # docker run \
--- a/datanode/src/storage_api.py
+++ b/datanode/src/storage_api.py
@@ -92,7 +92,7 @@ import uvrs
 # VERSION = 'TCL4.0.3.009'  # Added support for UVRs
 # VERSION = 'TCL4.0.3.010'  # Fixed backwards compatibility issue
 # VERSION = 'TCL4.1.0.011'  # Updated UVR schema (not API-compatible with previous UVR schema)
-VERSION = 'TCL4.1.0.012' # Improved error messages
+VERSION = 'TCL4.1.1.012' # Improved validation and error messages
 
 # Initialize everything we need
 TESTID = None
--- a/datanode/src/storage_api.py
+++ b/datanode/src/storage_api.py
@@ -57,14 +57,7 @@ import slippy_util
 import uvrs
 
 # VERSION HISTORY - BRANCH.MAJOR.MINOR.BUILD
-# Where BRANCH is the specific version for this branch, or numeric for the mainline branch,
-#   changing only on new branches or fundamental changes to the mainline branch.
-# and MAJOR changes when there is major funcationality change that either directly
-#   or eventually depracates functionality in previous major versions.
-# and MINOR changes when there is a non-breaking API change (new fields or methods) and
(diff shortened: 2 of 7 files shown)


datanode/docker/Dockerfile_storageapi
```diff
@@ -16,7 +16,7 @@
 
 # To build an image from this Dockerfile:
 # docker image build -f Dockerfile_storageapi -t interussplatform/storage_api:tcl4 ../src
-# docker tag interussplatform/storage_api:tcl4 interussplatform/storage_api:tcl4.1.0.011
+# docker tag interussplatform/storage_api:tcl4 interussplatform/storage_api:tcl4.1.1.012
```
hi-kevin: I really don't like using the tags and updating every time (including updating the startup script on the nodes). Can we do a latest tag and a version tag? That way everything points to latest and you can manually go to a specific version if needed.

datanode/src/storage_api.py
```diff
@@ -91,7 +91,8 @@
 # VERSION = 'TCL4.0.2.008'  # sync with master branch for multi-grid and docker updates
 # VERSION = 'TCL4.0.3.009'  # Added support for UVRs
 # VERSION = 'TCL4.0.3.010'  # Fixed backwards compatibility issue
-VERSION = 'TCL4.1.0.011'  # Updated UVR schema (not API-compatible with previous UVR schema)
+# VERSION = 'TCL4.1.0.011'  # Updated UVR schema (not API-compatible with previous UVR schema)
+VERSION = 'TCL4.1.1.012' # Improved validation and error messages
```
hi-kevin: For versions, I would have expected build .011 to have changed the minor version number, and .012 to not change the minor version number. No change required since this version would be the same, but make sure the swaggerhub version number matches (without the build #). 

There are instructions up top: "MINOR changes when there is a non-breaking API change (new fields or methods)".

datanode/src/updating.md
```diff
@@ -0,0 +1,11 @@
+# Updating version info
+
+When making changes, version information needs to be updated in several places.
```
hi-kevin: From storage_api, with formatting and spelling corrections:
**VERSION HISTORY - BRANCH.MAJOR.MINOR.BUILD**
Where BRANCH is the specific version for this branch, or numeric for the mainline branch,
changing only on new branches or fundamental changes to the mainline branch;
and MAJOR changes when there is major functionality change that either directly or eventually deprecates functionality in previous major versions;
and MINOR changes when there is a non-breaking API change (new fields or methods) and resets to 0 when a MAJOR change takes place;
and BUILD continuously increments with every release, regardless of other version number changes.

datanode/docker/Dockerfile_storageapi
BenjaminPelletier: Why not?  A tag is just tiny metadata that names a particular docker layer stack, and it seems like it should be very useful to be able to go back to any previously-used docker image.  Keeping every version fixes the "well it worked fine before; put it back like it was" problem.  The image is still built and published with the tag "interussplatform/storage_api:tcl4"; this line merely also publishes it as "interussplatform/storage_api:tcl4.1.1.012" (and actually, it doesn't publish; that requires an undocumented docker push since only the interussplatform docker hub admins can do that).  This is how most docker repositories are done; for instance at https://hub.docker.com/_/zookeeper/, the tags "latest", "3.4", and "3.4.13" all point to the same image.  Later, "3.4.13" will still point to the same image, but "3.4" will point to, perhaps, "3.4.14".

datanode/src/storage_api.py
BenjaminPelletier: We need to define exactly what "breaking the API" means (#56).  In this case, the API will return UVRs that are incompatible with UVRs from the previous version.  A client designed to validate the old UVR format will be broken by this API change.

datanode/src/updating.md
BenjaminPelletier: I've moved the versioning information into updating.md and added a note pointing there from storage_api.py; will merge and deploy later today unless there are any other issues.

PR #68, opened 2018-12-13
Update TCL4 branch to sync UVR schema and fix small issues

BenjaminPelletier:


assets/swaggerhub_api.yaml
```diff
@@ -1176,15 +1194,17 @@ definitions:
       min_altitude:
         description: >-
           The UVR does not affect flights below this altitude
-        allOf:
-          - $ref: '#/definitions/Altitude'
+        # Note that this description does not render properly.  See the link
```
punamverma: Could you also change the type for message_id in UVR? To
      message_id:
        type: string
        format: uuid
        description: >-
          id.
        maxLength: 36
        minLength: 36

drsmit22: In addition to message_id, the permitted_gufis field still references UUID. If you're changing message_id could make the same change similarly.

BenjaminPelletier:
@punamverma, what command are you using for your codegen?  This, for example, seems to work fine as-is: `docker run --rm -v ${PWD}:/local swaggerapi/swagger-codegen-cli generate -i /local/swaggerhub_api.yaml -l java -o /local/out/java`

punamverma:
@BenjaminPelletier Codegen generates a new class Uuid, which is not very useful. Whereas, we want to use the existing java.util.UUID which is created by codegen when you use -

type: string
format: uuid

BenjaminPelletier:
This is [a bug](https://github.com/swagger-api/swagger-codegen/issues/8994) in the codegen CLI.  I'll  hack around it by putting the object type definition directly in the field, but repeating the content of a definition like that is generally suboptimal API design.

assets/swaggerhub_api.yaml
BenjaminPelletier: Done and done.

punamverma:
@BenjaminPelletier Are you able to codegen? I am getting "failed to read resource listing".

[main] INFO io.swagger.parser.Swagger20Parser - reading from https://github.com/wing-aviation/InterUSS-Platform/blob/b2441a0d612ed1f6cd390666b44547456d5fe913/assets/swaggerhub_api.yaml
[main] ERROR io.swagger.parser.SwaggerCompatConverter - failed to read resource listing
com.fasterxml.jackson.core.JsonParseException: Unexpected character ('<' (code 60)): expected a valid value (number, String, array, object, 'true', 'false' or 'null')
 at [Source: java.io.StringReader@5f8edcc5; line: 7, column: 2]

BenjaminPelletier:
The URL you listed is to the GitHub page describing the swaggerhub_api.yaml resource; you need to link to the raw content itself: https://raw.githubusercontent.com/wing-aviation/InterUSS-Platform/b2441a0d612ed1f6cd390666b44547456d5fe913/assets/swaggerhub_api.yaml

punamverma:
Thanks @BenjaminPelletier for catching it. In a hurry, I missed that. 
It looks good to me for codegen now. 
If @drsmit22 has any comments.

PR #77, opened 2019-03-04
Augment status endpoint

BenjaminPelletier:
This PR improves the `status` endpoint to include more information that is likely to indicate why a node is in poor health including disk usage (if logs have filled up the disk) and connection to the Zookeeper server.

### DIFF
--- a/datanode/src/storage_api.py
+++ b/datanode/src/storage_api.py
@@ -87,50 +87,68 @@ webapp = Flask(__name__)  # Global object serving the API
 ################    WEB SERVICE ENDPOINTS    #########################
 ######################################################################
 
+@webapp.route('/status/webserver', methods=['GET'])
+def WebserverStatus():
+  log.debug('Webserver status handler instantiated...')
+  return _FormatResult({
+    'status': 'success',
+    'message': 'OK',
+  })
+
+
 @webapp.route('/', methods=['GET'])
 @webapp.route('/status', methods=['GET'])
 def Status():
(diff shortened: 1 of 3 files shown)


datanode/src/storage_api.py
```diff
@@ -89,11 +90,47 @@
 @webapp.route('/', methods=['GET'])
 @webapp.route('/status', methods=['GET'])
 def Status():
-  # just a quick status checker, not really a health check
```
hi-kevin: We discussed adding another endpoint that does a lot less (like the current status) so we can treat it like a heartbeat and hit it every few seconds. The /status we may hit every 5-10 minutes or so.

datanode/src/storage_api.py
```diff
@@ -89,11 +90,47 @@
...
+    zk_status = type(e).__name__ + (' ' + msg if msg else '')
+    result = None
+  if result:
+    if result['status'] == 'success':
+      zk_status = 'ok'
+    else:
+      zk_status = result['message']
+
+  return _FormatResult({
+      'status': 'success',
```
hi-kevin: Should status here be changed to capture the exception, or even change the status if it goes below a certain percentage?

BenjaminPelletier: Sounds good; done.  HTTP code is still 200 in all cases to indicate that the request for status succeeded, but the JSend changes status to 'error' in off-nominal cases.

Issue #73, opened 2019-01-31
Arguments via one source are ignored when other sources are present


BenjaminPelletier:
**Describe the bug**
Similar to #54, specifying some arguments in one format and other arguments in another format does not work even when the argument values do not conflict.

**To Reproduce**
1. Send a proper POST request to, e.g., /GridCellOperators with `coords` and `coord_type` arguments specified in the URL and all other arguments in a JSON body
1. Observe error "CSV of coordinates must be a string in lat,lon...format."
This happens because if a "higher-priority" means of specifying arguments is present (in this case, JSON body is higher-priority than URL arguments), the arguments specified by every other means are ignored entirely (in this case, the URL arguments are ignored)

**Expected behavior**
When arguments are provided in mixed formats, they should still be combined and used unless the values of the arguments are different between formats.

PR #78, opened 2019-03-07
Add custom/area-based authorization

BenjaminPelletier:
This PR factors out _ValidateAccessToken into an `authorization` module which allows not only the existing solution of providing a single, global public key, but also a custom authorization configuration with multiple overlapping authorization providers and complex applicability specifications.

Additional `introspect` endpoints are added to help troubleshoot any issues that may arise due to complex custom configurations.

The way options are loaded in `storage_api.py` is modified somewhat to always use the ParseOptions method, which uses environment variables as defaults.  This happens to solve questions of whether a command line argument or environment variable would take precedence, but its primary purpose is to ensure that a global `auth` variable in `storage_api.py` is instantiated regardless of the run mode of the `storage_api.py` webserver.

### DIFF
--- a/datanode/src/authorization.py
+++ b/datanode/src/authorization.py
@@ -1,0 +1,105 @@
+"""The InterUSS Platform Data Node authorization tools.
+
+Copyright 2018 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import logging
+import os
+import sys
+from flask import abort
+from flask import Flask
+from flask import jsonify
+from flask import request
+import jwt
+from rest_framework import status
+
+TESTID = None
+
+logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.INFO)
+log = logging.getLogger('InterUSS_DataNode_Authorization')
+
+
+def SetTestId(testid):
+  global TESTID
+  TESTID = testid
+  log.info('Authorization set to test mode with TESTID=%s' % TESTID)
+
+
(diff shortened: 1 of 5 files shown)


datanode/src/authorization.py
```diff
@@ -0,0 +1,361 @@
...
+    for authority in authorities:
+      try:
+        jwt.decode(token, authority.public_key, algorithms='RS256')
+        valid = True
+        break
+      except jwt.InvalidSignatureError:
+        # Access token signature not valid for this public key, but might be
+        # valid for a different public key.
+        pass
+    if not valid:
```
hi-kevin: 140-164 is all for an invalid token? Do we need all of this different logic for different error messages, or can we just say the key is not valid? Just maintenance issue.

datanode/src/storage_api.py
```diff
@@ -112,6 +111,55 @@ def Introspect():
   })
 
 
+@webapp.route('/introspect/<zoom>/<x>/<y>', methods=['GET'])
```
hi-kevin: Don't forget to update Swaggerhub.

datanode/src/storage_api.py
```diff
@@ -76,12 +76,11 @@
 # VERSION = 'PublicPortal1.1.1.006'  # Added public portal support
 VERSION = 'PublicPortal1.1.1.007'  # Fixed multi-cell bug
```
hi-kevin: New minor version for non-breaking API additions?

datanode/src/authorization.py
BenjaminPelletier: Correct, 140-164 is all just to provide an informative error message to the client.  Some of the largest maintenance burdens for the InterUSS Platform so far have been unclear or misleading error messages (leading to difficult troubleshooting with partners; see, e.g., #73), so I think it's worth specifying exactly why a token has failed to validate.

datanode/src/storage_api.py
BenjaminPelletier: Done.

datanode/src/storage_api.py
BenjaminPelletier: Done.

BenjaminPelletier:
The publicportal branch will soon become an obsolete reference and it makes sense to make this functionality available to those who would like to refer to it.  It was first-pass reviewed back in March and I believe that clears the bar for a reference branch we do not intend to deploy again.

PR #83, opened 2019-07-02
Replace zookeeper with cockroachdb for testing purposes

vosst:
This PR aims to replace zookeeper for cockroachdb for testing purposes. This is a WIP, and meant to gather data and feedback. Important remarks:
- I started over with a very simple database schema, pretty much emulating a KV-store.
- Table creation is done in the ctor of `USSMetadataManager`. Not optimal, but I was hesitant to introduce full-blown schema migration support in the scope of this PR. 
- Tile coordinates still use slippy, and the original pathing is kept in place.
- While zookeeper distinguishes transaction ids and versions, I naively used the transaction timestamp for both in the case of cockroach. Tests seem to be happy, not sure if calling code expects integers, though. If so, we could consider hashing the transaction timestamps (slightly ugly and not monotonically increasing) or introduce a version column.
- In the case of zookeeper, a query for a tile that hasn't been written, yet, still returns a transaction id/version. For cockroach, I adjusted the code to map this case to `None`, adjusting supporting code (in particular hashing multiple transaction ids) to handle the case. Tests seem happy, but feedback is very much appreciated :smile:
- I did not look into optimizing the fast-paths introduced for zookeeper, yet. In particular, the code executes a get before a set and returns early if the relevant tile has been changed in the meantime. We might be able to squeeze out some performance by avoiding the roundtrip to the DB and just rely on our upsert statement.

@BenjaminPelletier mentioning you explicitly to get the PR on your radar.

### DIFF
--- a/datanode/src/storage_interface.py
+++ b/datanode/src/storage_interface.py
@@ -14,7 +14,7 @@ information interface to translate grid cell USS information into the correct
 data storage format, and an information consistency store to ensure data is up
 to date.
 
-This module is the information interface to Zookeeper.
+This module is the information interface to Cockroach.
 
 
 Copyright 2018 Google LLC
@@ -41,13 +41,8 @@ import slippy_util
 
 import uvrs
 
-# Kazoo is the zookeeper wrapper for python
-from kazoo.client import KazooClient
-from kazoo.exceptions import BadVersionError
-from kazoo.exceptions import NoNodeError
-from kazoo.exceptions import RolledBackError
-from kazoo.handlers.threading import KazooTimeoutError
-from kazoo.protocol.states import KazooState
+# psycopg2 is a postgres database driver for python's DB API.
+import psycopg2
 
 # logging is our log infrastructure used for this application
 logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.INFO)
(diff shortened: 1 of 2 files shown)


datanode/src/storage_interface.py
```diff
@@ -58,14 +53,47 @@
...
 GRID_PATH = USS_BASE_PREFIX
 MAX_SAFE_INTEGER = 9007199254740991
 DELETE_ATTEMPTS = 3
 OK_STATUS = {200, 204}
 
+SQL_CREATE_TABLE = """
+  CREATE TABLE IF NOT EXISTS
+    grid (
+      tile STRING PRIMARY KEY NOT NULL,
+      uss_metadata STRING,
```
steeling: do we want this to be a string, now that we have access to a relational db?

datanode/src/storage_interface.py
```diff
@@ -80,34 +108,30 @@ def __init__(self, connectionstring=DEFAULT_CONNECTION, testgroupid=None):
...
+    log.debug('Creating metadata manager object and connecting to cockroach...')
     try:
-      if set(BAD_CHARACTER_CHECK) & set(connectionstring):
-        raise ValueError
-      self.zk = KazooClient(hosts=connectionstring, timeout=CONNECTION_TIMEOUT)
-      self.zk.add_listener(self.zookeeper_connection_listener)
-      self.zk.start()
+      self.conn = psycopg2.connect(connectionstring)
+      with self.conn.cursor() as curs:
+        curs.execute(SQL_CREATE_TABLE)
```
steeling: should we pull table creation into a separate file?

We can skip this and clean this up later for the reference implementation

vosst: Absolutely, I think I noted that down in the PR description. We should handle schema migrations properly for the reference implementation. For the scope and purpose of this PR, I'm inclined to leave it in the `__init__` unless someone has strong objections.

datanode/src/storage_interface.py
vosst: I think, to begin with: yes. Going forward, I would expect the database layout to evolve, and the entities to be managed as first class citizens in a relational model.

datanode/src/storage_interface.py
```diff
@@ -80,34 +108,30 @@ def __init__(self, connectionstring=DEFAULT_CONNECTION, testgroupid=None):
 
     Args:
       connectionstring:
-        Zookeeper connection string - server:port,server:port,...
+        cokcroach connection string - following postgres URL syntax
```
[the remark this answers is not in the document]
vosst: Fixed, thanks :smile:

vosst:
> This looks great; I successfully tested using psycopg2-binary and the docker command provided in storage_interface_test. I think this is good to commit as-is (optionally fixing the single small typo noted).
> 

Thanks for the testing feedback, typo is fixed.

> For a follow-up PR, the Zookeeper sync_tokens are 64-bit ints and even if clients shouldn't be assuming that sync_tokens are ints, they may make that assumption any way. For the purposes of this test, I think it would be good to convert outgoing sync_tokens to ints, but that could be via CockroachDB versions, or taking the lower 64 bits of the timestamp, or something else.

Yup, also interesting in terms of implications for future versions. We should probably state that version tokens are essentially to be treated as opaque blobs.

vosst:
@BenjaminPelletier Can you handle the merge?

PR #120, opened 2019-10-01
Feature/protos

steeling:
Migrate to more sane proto names. Deprecate generating the protos from the API spec


BenjaminPelletier:
This pull request has a huge number of commits, likely because of the merge commits f533631 and 9896f34 -- please rebase the new commits appropriately so that the pull request only has the new changes.  Also, let's expand the pull request description to note why the original proto names were chosen and why the updated names are better (surely existing proto names aren't actually insane).  Same with generating protos from the API spec -- let's explain why it is how it is and why the new way is better.

Makefile
```diff
@@ -6,7 +6,7 @@ interuss:
 
 .PHONY: format
 format:
-	clang-format -style=file -i pkg/dssproto/dss.proto
+	clang-format -style=file -i pkg/dss_v1/dss.proto
```
vosst: nit: Any chance we could avoid `_` in package names?

pkg/dss/geo/s2_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"testing"
 
 	"github.com/steeling/InterUSS-Platform/pkg/dss/geo/testdata"
-	dspb "github.com/steeling/InterUSS-Platform/pkg/dssproto"
+	dspb "github.com/steeling/InterUSS-Platform/pkg/dss_v1"
```
vosst: While you are here, you might want to adjust `dspb` -> `dss_v1`.

Makefile
steeling: done

pkg/dss/geo/s2_test.go
steeling: done

steeling:
orbiting this PR since git history got borked after the remote switch. Using the new orphaned branch at https://github.com/interuss/dss/pull/121/

PR #121, opened 2019-10-02
Use more sane proto names

steeling:
Use more standard proto names. Previous names were generated from the HTTP spec, so it's descriptors placed assumptions on the HTTP method (ie, get, update, etc). gRPC does not have this, so using more descriptive (and shorter) names is easier to work with


datanode/centos/README.md
```diff
@@ -0,0 +1,123 @@
+# InterUSS Platform evaluation on CentOS
```
BenjaminPelletier: It doesn't seem like this file should have been added.  Same with many others.

PR #123, opened 2019-10-08
Feature/e2e

charlie-pisuraj:
Patched in @steeling pull request from his branch and extended it with scripts to start the E2E tests, building and starting docker containers for all components of the DSS and launching the pytests in its own container.


config/test-certs/temp.go
```diff
@@ -0,0 +1,72 @@
+package main
```
steeling: Let's rename this file.. gen-cert or something

steeling: Add package comment for what this file does

config/test-certs/temp.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func pemBlockForKey(priv interface{}) *pem.Block {
+	switch k := priv.(type) {
+	case *rsa.PrivateKey:
+		return &pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(k)}
+	case *ecdsa.PrivateKey:
+		b, err := x509.MarshalECPrivateKey(k)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Unable to marshal ECDSA private key: %v", err)
```
steeling: switch to log.Fatalf

docker_e2e.sh
```diff
@@ -0,0 +1,91 @@
...
+	local-dummy-oauth \
+	-private_key_file /app/test.key
+
+echo " -------------- PYTEST -------------- "
+sleep 5
+echo "Building Integration Test container"
+docker build -q --rm -f monitoring/prober/Dockerfile monitoring/prober -t e2e-test
+
+echo "Finally Begin Testing"
+docker run --link dummy-oauth-for-testing:oauth \
```
steeling: How will we capture test output to know if the tests failed/succeeded?

steeling: @vosst can we hook this into the existing integration checks?

charlie-pisuraj: we might be able to export the test result out to xml or other formats http://doc.pytest.org/en/latest/usage.html#creating-junitxml-format-files

cmds/dummy-oauth/main.go
```diff
@@ -0,0 +1,76 @@
...
+
+var (
+  address    = flag.String("addr", ":8085", "address")
+  keyFile    = flag.String("private_key_file", "config/test-certs/oauth.key", "oauth private key file")
+  privateKey *rsa.PrivateKey
+)
+
+// TODO(steeling): add other parameters so we can test expired tokens, invalid tokens, etc.
+func getToken(w http.ResponseWriter, r *http.Request) {
+  fmt.Println(r)
```
vosst: Seems like `httputil.DumpRequest` is a better fit here.

cmds/dummy-oauth/main.go
```diff
@@ -0,0 +1,76 @@
...
+  w.Header().Set("Content-Type", "application/json")
+  params := r.URL.Query()
+
+  aud := ""
+  auds := params["intended_audience"]
```
vosst: nit: a `var`-block would help with readability.

cmds/dummy-oauth/main.go
```diff
@@ -0,0 +1,76 @@
...
+  if len(auds) == 1 {
+    aud = auds[0]
+  }
+
+  scope := ""
+  scopes := params["scope"]
```
vosst: Same nit for a `var`-block.

cmds/dummy-oauth/main.go
```diff
@@ -0,0 +1,76 @@
...
+
+  // Sign and get the complete encoded token as a string using the secret
+  // Ignore the error, it will fail the test anyways if it is not nil.
+  tokenString, err := token.SignedString(privateKey)
+  if err != nil {
+    http.Error(w, err.Error(), 500)
+    return
+  }
+
+  json.NewEncoder(w).Encode(map[string]string{
```
vosst: Would be nice to check for the error here and at least print it.

cmds/dummy-oauth/main.go
```diff
@@ -0,0 +1,76 @@
...
+  if err != nil {
+    return nil, err
+  }
+  return jwt.ParseRSAPrivateKeyFromPEM(bytes)
+}
+
+func main() {
+  flag.Parse()
+  var err error
+  privateKey, err = readPrivateKey()
```
vosst: I would prefer if we could avoid the global variable `privateKey` and instead have a proper handler along the lines of:
```
func createGetTokenHandler(*rsa.PrivateKey) http.Handler {
  ...
}
```

cmds/dummy-oauth/main.go
```diff
@@ -0,0 +1,76 @@
...
+  if err != nil {
+    log.Fatal(err)
+  }
+  http.HandleFunc("/token", getToken)
+  log.Fatal(http.ListenAndServe(*address, nil))
+
```
vosst: nit: stray whitespace.

docker_e2e.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
```
vosst: Mind moving this file into a subdirectory `test`?

vosst: Please add `set -eo pipefail` such that test failures bubble up as proper exit codes.

docker_e2e.sh
vosst: > @vosst can we hook this into the existing integration checks?

Yes we can. Let's get this PR in and tackle the integration in a follow-up PR.

cmds/dummy-oauth/main.go
steeling: fmt.infof.

charlie-pisuraj: infof? I don't see it here https://golang.org/pkg/fmt/?m=all

config/test-certs/gen-cert.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	pubKey := publicKey(priv)
+
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, pubKey, priv)
+	if err != nil {
+		log.Fatalf("Failed to create certificate: %s", err)
+	}
+
+	out := &bytes.Buffer{}
+	pem.Encode(out, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
```
vosst: You could just pass `os.Stdout` instead of the buffer to both `pem.Encode` invocations.

charlie-pisuraj:
I think next time merging might be a better option since rebasing made 12 which is crazy noisy

monitoring/prober/conftest.py
```diff
@@ -38,6 +57,7 @@ def __init__(self, token_endpoint, service_account_json):
     credentials = service_account.Credentials.from_service_account_file(
         service_account_json).with_scopes(['email'])
     oauth_session = google_requests.AuthorizedSession(credentials)
+    oauth_session = requests.Session()
```
BenjaminPelletier: Should the previous line be removed given this line?  Since this line is in ServiceAccountAuthAdapter, does authorization against a service account still work with this line added?

monitoring/prober/test_isa_expiry.py
```diff
@@ -11,7 +11,7 @@ def test_create(session, isa1_uuid):
   time_end = time_start + datetime.timedelta(seconds=5)
 
   resp = session.put(
-      '/identification_service_areas/{}'.format(isa1_uuid),
+      '/v1/dss/identification_service_areas/{}'.format(isa1_uuid),
```
BenjaminPelletier: I would recommend not changing all of these paths to require the /v1/dss prefix as it reduces testing flexibility -- a system that includes /v1/dss can be easily tested with the previous version, but a system that does not include /v1/dss cannot be tested with this version.  The API path prefixes (/dss, /uss) are primarily meant to indicate which role would be hosting them, and not as a strict requirement on path configuration for all implementations.

test/docker_e2e.sh
```diff
@@ -0,0 +1,107 @@
...
+echo "Starting cockroachdb with admin port on :8080"
+docker run -d --rm --name dss-crdb-for-debugging \
+	-p 26257:26257 \
+	-p 8080:8080 \
+	cockroachdb/cockroach:v19.1.2 start \
+	--insecure > /dev/null
+
+sleep 5
+echo " ------------ GRPC BACKEND ---------------- "
+# building grpc backend docker
```
BenjaminPelletier: nit: comment probably not needed

monitoring/prober/conftest.py
charlie-pisuraj: This line needs to be removed, its a stray

monitoring/prober/test_isa_expiry.py
charlie-pisuraj: should we pass this as a flag to the test? since we need this to prefix every requests as specified in the proto

cmds/dummy-oauth/main.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		var (
+			scope  string   = ""
+			scopes []string = params["scope"]
+		)
+		if len(scopes) == 1 {
+			scope = scopes[0]
+		}
+		token := jwt.NewWithClaims(jwt.SigningMethodRS256, jwt.MapClaims{
```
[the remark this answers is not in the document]
charlie-pisuraj: Issue created https://github.com/interuss/dss/issues/127

BenjaminPelletier:
LGTM :)  Thanks for all the work here and sorry for the difficulties; we'll see if we can make things a little easier going forward.

PR #124, opened 2019-10-09
finish readme

steeling:
* Edit readme to better reflect options for running a DSS. 
* Fix some inconsistencies in the docs
* Fix make-certs.py to use combined certs so we don't need to share private keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,92 +1,90 @@
 # USS to USS Communication and Synchronization Wrapper
 
 This repository contains a simple and open API used for separate UAS Service
-Suppliers (USS) to communicate during UAS operations, known as the InterUSS
-Platform™. This flexible and distributed system is used to connect multiple USSs
-operating in the same general area to share safety information while protecting
+Suppliers (USS) to communicate during UAS operations, known as a Discovery and
+Synchronization Service (DSS) as described in the ASTM remote ID standard.
+This flexible and distributed system is used to connect multiple USSs
+operating in the same general area to share information while protecting
 operator and consumer privacy. The system is focused on facilitating
 communication amongst actively operating USSs with no details about UAS
-operations stored or processed on the InterUSS Platform.
+operations stored or processed in the DSS.
 
-### Main Features
 
-*   Simple, open, flexible, and scalable information interface and Optimistic
-    Concurrency Control to simplify USS to USS communication and control race
-    conditions.
-*   Effective information sharing to enable effective, deterministic conflict
-    detection for flight planning across multiple overlapping 4D volumes from
-    multiple USSs.
-*   Effective reconstitution of flight volumes when restarting a USS or adding a
-    USS to an existing environment.
+## Architecture
 
-### Main Assumptions
 
-*   Trust each USS to deconflict with known flight data.
-*   Auditing is available, as all USSs can verify the authorship and addition of
-    erroneous conflictions.
-*   Deconfliction is simply defined as no overlapping flights in time and
-    volume.
+See the [API specification here](https://tiny.cc/dssapi_rid)
 
-For the API specification and online test area, see
-https://app.swaggerhub.com/apis/InterUSS_Platform/data_node_api.
+### HTTP Frontend
 
-## Sequence for USS information exchange
+Serves as an Https gateway to the business logic, translating between Http Request
+and gRPC to allow users to communicate to the DSS via simple Https calls. This
+code is currently generated via
+[grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway), and does not do
+much other than translation.
 
-![Simple Sequence](assets/USS0.png)
+### gRPC Backend
 
-When a USS wants to plan a flight, the “planning USS” performs the following
-steps:
 
-1.  Discover - Communicates with the InterUSS platform to discover what other
-    USSs have an active operation in the specific area of flight and how to
-    contact them. Flight details are not stored in the InterUSS Platform for
-    both minimizing data bandwidth and maximizing privacy.
-2.  Sync - If there are other USSs with active operations, the planning USS
-    contacts the other USS(s) with active operations in the area and retrieves
-    the operational information necessary to de-conflict the operation.
-3.  Plan & Verify - Once the operation is safely planned and deconflicted, the
-    USS updates the InterUSS platform with their contact information for other
-    USSs to use for future discovery. This update provides final verification
-    that the planning was done with the latest operational picture.
-4.  Delete - Once the operation is completed, the USS removes their information
-    as they do not need to be contacted in the future since the USS no longer
-    has operations in that specific area of flight. All of these operations are
-    secured using a standard OAuth communications flow, authorized by a
-    governmental authority or, in the absence of an authority in certain parts
-    of the world, industry provided OAuth solutions decided by the InterUSS
-    Technical Steering Committee.
+Component responsible for all the business logic as well as authentication. This
+backend talks directly to CockroachDB.
 
-![Simple Sequence](assets/USS1.png)
 
-For an additional examples, this [Sequence Diagram](assets/USS2.png) shows a
-more complex operation with three USSs, two trying to plan at the same time.
-This [Sequence Diagram](assets/USS3.png) shows a USS affecting multiple grids.
-This [Sequence Diagram](assets/USS4.png) shows another USS updating one of
-multiple grids during an update. And finally, this
-[Sequence Diagram](assets/USS5.png) shows different examples of protection with
-concurrent updates.
+### CockroachDB (CRDB)
+
+Individual CockRoach DB instances hosting sharded data. More information about 
+CockRoachDB 
+[here](https://www.cockroachlabs.com/docs/stable/architecture/overview.html).
 
 ## Directories of Interest:
-* `config/` has all of the configuration required to deploy a DSS instance to a kubernetes cluster. The README in that directory contains more information.
-* `pkg/` contains all of the source code for the DSS. See the README in that directory for more information.
-* `cmds/` contains entry points and docker files for the actual binaries (the `http-gateway` and `grpc-backend`)
-* 
+*   `config/` has all of the configuration required to deploy a DSS instance
+    (typically to a kubernetes cluster). The README in that directory contains
+    more information.
+*   `pkg/` contains all of the source code for the DSS. See the README in that
+    directory for more information.
+*   `cmds/` contains entry points and docker files for the actual binaries (the
+    `http-gateway` and `grpc-backend`)
 
 ## Notes
 
-* Currently this branch only supports the Remote ID API's. 
-* The current implementation relies on CockroachDB for data storage and synchronization between DSS participants. It is recommended to read up on CockroachDB for performance characteristics and operational caveats. We list some of the caveats that we've run into below:
+*   Currently this branch only supports the Remote ID API's. 
+*   The current implementation relies on CockroachDB for data storage and
+    synchronization between DSS participants. It is recommended to read up on 
+    CockroachDB for performance characteristics and operational caveats. We list
+    some of the caveats that we've run into below:
 
 ### CockroachDB Notes
-* CockroachDB (CRDB) currently uses certificates to authenticate clients and node to node communication. All of Node certs, client certs, and even a CA cert are all generated through the cockroach cli. The CA certs must be custom generated since the Node certs are require to have CN=node, which no public CA will sign.
-* CockroachDB allows public CA certs to be concatenated, to allow for certificate rotation. We abuse this to allow each DSS participant to bring their own CA cert.
-* CockroachDB nodes join the cluster virally. That is, each node keeps state on all the other nodes, and if a node connects to a new node, it will learn about the entire cluster through a gossip protocol.
-* Each CRDB node must be uniquely addressable and routable.
-  * Because of this, we expect each node to have it's own static IP and/or publicly resolvable hostname. In the future, we likely want to explore the use of Istio, which allows multi cluster private service discovery, layers on top it's own TLS so that we could run CRDB in insecure mode, and not worry about the certificate problems (Istio can use standard CA's), and would reduce latency by removing the extra hop of the CRDB per-node Loadbalancer we are currently using.
-* CRDB splits up data based on the locality string. Data replication strategy is a database variable. The CRDB nodes will traverse the key,values of the locality flag to determine how to divy up the replicas. This means there could be mroe than 7 participants of the DSS, with only 3 or 5 (or 7) replicas, and each participant would simply recieve a shard of a single replica.
-* CRDB clients talk to any CRDB node which will proxy traffic to the correct node(s).
-* There is an admin UI for CRDB (default port 8080). This should be locked down to internal traffic only.
-* The cluster init command must *only* ever be run against one node in the cluster. It seeds the data directories, and if it is run against another node it won't be able to join the cluster, or will create some splits within the cluster. The command is not dangerous once a node has joined a cluster that has been initiated, but it is possible a node's data directory gets destroyed so it should be avoided.
+*   CockroachDB (CRDB) currently uses certificates to authenticate clients and
+    node to node communication. All of Node certs, client certs, and even a CA
+    cert are all generated through the cockroach cli. The CA certs must be
+    custom generated since the Node certs are require to have CN=node, which no
+    public CA will sign.
+*   CockroachDB allows public CA certs to be concatenated, to allow for
+    certificate rotation. We abuse this to allow each DSS participant to bring
+    their own CA cert.
+*   CockroachDB nodes join the cluster virally. That is, each node keeps state
+    on all the other nodes, and if a node connects to a new node, it will learn
+    about the entire cluster through a gossip protocol.
+*   Each CRDB node must be uniquely addressable and routable.
+    *   Because of this, we expect each node to have it's own static IP and/or
+    publicly resolvable hostname. In the future, we likely want to explore the
+    use of service mesh frameworks.
+*   CRDB splits up data based on the locality string. Data replication strategy
+    is a database variable. The CRDB nodes will traverse the key,values of the
+    locality flag to determine how to divy up the replicas. This means there
+    could be more than 7 participants of the DSS, with only 3 or 5 (or 7)
+    replicas, and each participant would simply receive a shard of a single
+    replica.
+*   CRDB clients talk to any CRDB node which will proxy traffic to the correct
+    node(s).
+*   There is an admin UI for CRDB (default port 8080). This should be locked
+    down to internal traffic only.
+*   The cluster init command must *only* ever be run against one node in the
+    cluster. It seeds the data directories, and if it is run against another
+    node it won't be able to join the cluster, or will create some splits within
+    the cluster. The command is not dangerous once a node has joined a cluster
+    that has been initiated, but it is possible a node's data directory gets
+    destroyed so it should be avoided.
 
 
 ## Contribution
--- a/README.md
+++ b/README.md
@@ -32,13 +32,13 @@ backend talks directly to CockroachDB.
 
 ### CockroachDB (CRDB)
 
-Individual CockRoach DB instances hosting sharded data. More information about 
-CockRoachDB 
+Individual CockroachDB instances hosting sharded data. More information about 
+CockroachDB 
 [here](https://www.cockroachlabs.com/docs/stable/architecture/overview.html).
 
 ## Directories of Interest:
 *   `config/` has all of the configuration required to deploy a DSS instance
-    (typically to a kubernetes cluster). The README in that directory contains
+    (typically to a Kubernetes cluster). The README in that directory contains
     more information.
 *   `pkg/` contains all of the source code for the DSS. See the README in that
     directory for more information.
--- a/config/README.md
+++ b/config/README.md
@@ -1,10 +1,24 @@
 # Deploying a DSS
 
+
+
+## Glossary
+
+*   DSS cluster - an entire synchronized DSS, typically operated by multiple
+    organizations
+*   DSS node - a single logical replica in a synchronized cluster.
+
+
 ## Preface
+
 
 This doc provides a well-lit path for deploying the DSS and its dependencies (
 namely CockroachDB) on Kubernetes. This is not a requirement, and a DSS node
-can join a cluster as long as it meets the following requirements:
+can join a cluster as long as it meets the requirements below.
+
+
+## CockroachDB requirements
+
 
 *   Every CockroachDB node must advertise a unique and routable address.
     *   It is preferred to use domain names with unique prefixes and homogenous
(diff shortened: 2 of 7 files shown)


config/README.md
```diff
@@ -1,102 +1,103 @@
-# Multi-region cockroachdb setup
+# Deploying a DSS
+
+## Preface
+
+This doc provides a well-lit path for deploying the DSS and its dependencies (
+namely CockroachDB) on Kubernetes. This is not a requirement, and a DSS node
```
BenjaminPelletier: Let's pick a consistent terminology for referring to a single DSS instance/node.  I think "DSS instance" is generally more common that "DSS node", "instance" helps differentiate between CRDB nodes and DSS instances, and "DSS instance" is used above, so I would recommend "DSS instance" over "DSS node".  If we really meant to say "CRDB node" here rather than DSS node (since a DSS instance only joins a cluster via its CRDB nodes), let's clarify that.

config/README.md
```diff
@@ -1,102 +1,103 @@
...
+can join a cluster as long as it meets the following requirements:
+
+*   Every CockroachDB node must advertise a unique and routable address.
+    *   It is preferred to use domain names with unique prefixes and homogenous
+        suffixes, ie: 0.c.dss.interussplatform.com.
+    *   This allows wildcard usage in the CRDB certificates.
+*   Every DSS node should run a minimum of 3 CockroachDB nodes.
```
BenjaminPelletier: Let's elaborate on the reasoning behind this recommendation so that others can identify when (if ever) it might not apply.

config/README.md
```diff
@@ -1,102 +1,103 @@
...
+*   At least 3 CockroacbDB addresses must be shared with all participants.
+    *   If not using the recommended hostname prefix above, every cockroachDB
+        hostname must be shared with every participant.
+*   Every DSS node must supply and share their CockroachDB public certificate
+*   All CockroachDB nodes must be run in fully secure mode.
```
BenjaminPelletier: Can we be more specific here to make it easier to verify (non-)compliance?  Is there a single specific flag, for instance, that implementers must include or exclude?

config/README.md
```diff
@@ -1,102 +1,103 @@
...
+
+*   Every CockroachDB node must advertise a unique and routable address.
+    *   It is preferred to use domain names with unique prefixes and homogenous
+        suffixes, ie: 0.c.dss.interussplatform.com.
+    *   This allows wildcard usage in the CRDB certificates.
+*   Every DSS node should run a minimum of 3 CockroachDB nodes.
+*   At least 3 CockroacbDB addresses must be shared with all participants.
+    *   If not using the recommended hostname prefix above, every cockroachDB
+        hostname must be shared with every participant.
+*   Every DSS node must supply and share their CockroachDB public certificate
```
BenjaminPelletier: To whom via what means?  If we don't have a protocol for automated sharing yet and the recipients are a well-defined limited set, we can just say they will be shared out-of-band, but we should make these steps as directly actionable as possible.

config/README.md
```diff
@@ -1,102 +1,103 @@
...
 
-### Google Container Registry
+The grpc-backend and http-gateway are the 2 main binaries for processing DSS 
+requests. These are both built and pushed to a docker registry of your choice.
+You can easily find out how to push to a docker registry through a quick search.
+All major cloud providers have a docker registry service, or you can set up your
+own.
 
-List existing images:
+TODO(): The linux foundation should provide a default docker registry, with
```
BenjaminPelletier: Let's not put TODOs in READMEs

config/README.md
```diff
@@ -1,102 +1,103 @@
...
+*   Cockroachdb
+    *   Required to generate new CockroachDB certificates.
 *   Google Cloud SDK (if deploying on GCP)
-*   [Optional] Golang. Recommended to understand go, and the go toolchain.
+    * Required if deploying to Google Cloud.
+*   Golang.
+    *   Required if developing the DSS codebase.
 
 
 ## Building Docker images
```
BenjaminPelletier: This heading should probably now be just "Docker images" with a "Building Docker images" subheading directly preceding the "Set environment variable..." line since the first paragraph here does not relate to building images.

config/README.md
```diff
@@ -1,102 +1,103 @@
...
 
 
 ## Creating a new Kubernetes cluster
 
-You need to create a Kubernetes cluster to run the cockroachdb instance and the
-gRPC and HTTP servers.  You can do this on any supported
-[cloud provider](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/)
-or even on your own infrastructure.  Instructions for GCP are given below,
-consult the Kubernetes documentation for other providers.
+This secion discusses deploying a kubernetes service, although you can deploy
```
BenjaminPelletier: section

Kubernetes (capitalized)

config/README.md
```diff
@@ -1,102 +1,103 @@
-# Multi-region cockroachdb setup
+# Deploying a DSS
+
+## Preface
+
+This doc provides a well-lit path for deploying the DSS and its dependencies (
+namely CockroachDB) on Kubernetes. This is not a requirement, and a DSS node
+can join a cluster as long as it meets the following requirements:
```
BenjaminPelletier: Since these requirements are referred to from below, let's put them in their own addressable section.  So on this line, "...meets the requirements below." and then add a "## CRDB requirements" heading just below.

config/README.md
```diff
@@ -1,102 +1,103 @@
...
 
 ## Creating a new Kubernetes cluster
 
-You need to create a Kubernetes cluster to run the cockroachdb instance and the
-gRPC and HTTP servers.  You can do this on any supported
-[cloud provider](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/)
-or even on your own infrastructure.  Instructions for GCP are given below,
-consult the Kubernetes documentation for other providers.
+This secion discusses deploying a kubernetes service, although you can deploy 
+a DSS node however you like as long as it meets the above requirements. You can
```
BenjaminPelletier: Let's mention exactly which requirements.  If the "CRDB requirements" section comment is adopted above, we can say "...meets the CRDB requirements above."

config/README.md
```diff
@@ -1,102 +1,103 @@
...
-## Creating a new cockroachdb cluster
-
-1.  Create 5 static IP addresses.  How you do this depends on your cloud
-    provider.  3 IPs will be used for the 3 individual cockroachdb nodes, 1 IP
-    will be used to load balance amongst all the cockroachdb nodes, and 1 IP
-    will be used for the HTTPS frontend.  The HTTPS frontend IP will be used on
-    a Kubernetes Ingress, and if you're using Google Cloud it needs to be
-    created as a "Global" IP address.
+1.  Create your static IP addresses: 1 for each of your CockroachDB nodes (min
+    of 3) and 1 for the HTTPS Gateway's Ingress.  How you do this depends on
```
BenjaminPelletier: ingress (capitalization)

config/README.md
```diff
@@ -1,102 +1,103 @@
...
+    your cloud provider. Note: if you're using Google Cloud the HTTPS gateway
```
BenjaminPelletier: Gateway (capitalization consistency with above)

config/README.md
```diff
@@ -1,102 +1,103 @@
...
+    Ingress needs to be created as a "Global" IP address.
```
BenjaminPelletier: ingress (capitalization)

config/README.md
```diff
@@ -1,102 +1,103 @@
...
     at the top.
 
 1.  Use the `make-certs.py` script in this directory to create certificates for
     the new cockroachdb cluster:
 
-        ./make-certs.py $NAMESPACE \
-            [--node-address <ADDRESS> <ADDRESS> <ADDRESS> ...]
-            [--ca-certs-dir <CA_CERTS_DIR>]
+        ./make-certs.py [--node-address <ADDRESS> <ADDRESS> <ADDRESS> ...]
+            [--ca-cert-to-join <CA_CERT_FILE>]
```
BenjaminPelletier: This is in the "Creating a new CockroachDB cluster" section; how is the CA_CERT_FILE generated/obtained if not joining an existing cluster?

config/README.md
```diff
@@ -107,14 +108,15 @@ and makes all future kubecfg commands target this cluster.
 Follow the steps above for creating a new cockroachdb cluster, but with the
```
BenjaminPelletier: CockroachDB (capitalization) -- let's make sure there capitalization is consistent throughout the document.

config/README.md
```diff
@@ -123,13 +125,15 @@ your local machine using the kubectl command:
...
+Pick a username and create an account:
+
+    kubectl -n dss-main exec cockroachdb-0 -ti -- \
         ./cockroach --certs-dir ./cockroach-certs \
         user set $USERNAME --password
 
 ### Access the web UI
 
-    kubectl -n $NAMESPACE port-forward cockroachdb-0 8080
+    kubectl -n dss-main port-forward cockroachdb-0 8080
```
BenjaminPelletier: All DSS operators will have at least two DSS instances (prod and test environment) per DSS0210; shouldn't we keep $NAMESPACE to facilitate this?  If not, how would the two different DSS instances be addressed without $NAMESPACE?

config/build.sh
```diff
@@ -12,8 +12,8 @@ set -e -x
 VERSION="$(date -u +%Y-%m-%d)-$(git rev-parse --short HEAD)"
 
 cd "${BASEDIR}"
-docker build -f cmds/http-gateway/Dockerfile  . -t gcr.io/$CLOUD_PROJECT/http-gateway:$VERSION
-docker build -f cmds/grpc-backend/Dockerfile  . -t gcr.io/$CLOUD_PROJECT/grpc-backend:$VERSION
+docker build -f cmds/http-gateway/Dockerfile  . -t $DOCKER_URL:$VERSION
```
BenjaminPelletier: Shouldn't there be an "http-gateway" suffix here?  And "grpc-backend" suffix on the next line?

README.md
```diff
@@ -58,35 +57,81 @@ steps:
...
 
-For an additional examples, this [Sequence Diagram](assets/USS2.png) shows a
-more complex operation with three USSs, two trying to plan at the same time.
-This [Sequence Diagram](assets/USS3.png) shows a USS affecting multiple grids.
-This [Sequence Diagram](assets/USS4.png) shows another USS updating one of
-multiple grids during an update. And finally, this
-[Sequence Diagram](assets/USS5.png) shows different examples of protection with
-concurrent updates.
+
+## Architecture
```
BenjaminPelletier: Nearly everything above this point should be removed, or updated to reflect what is in this repo (though an update could happen in a follow-up PR).

config/README.md
steeling: args in brackets are optional, so if not joining a cluster you wouldn't specify it at all

config/README.md
steeling: that's a bad practice since kubernetes namespaces are not fully isolated. They should be completely separate clusters

README.md
steeling: A lot of it looks like it still applies.. I've removed the sequence diagram for now though. I think I'll leave the general intro to somebody else though and try to stick to the technical components :)

config/README.md
BenjaminPelletier: Let's make it clear in the sub-bullet below that --ca-cert-to-join should be omitted if not joining an existing cluster.

Under what conditions would the user not specify --node-address?

config/README.md
```diff
@@ -1,135 +1,139 @@
...
 
-Create a new cluster in the given zone:
+### Deploying with Kubernetes
 
-    gcloud --project $CLOUD_PROJECT container clusters create $CLUSTER_NAME --zone $ZONE
-
-Fetch credentials for the cluster.  This populates your \~/.kube/config file
-and makes all future kubecfg commands target this cluster.
-
-    gcloud --project $CLOUD_PROJECT container clusters get-credentials $CLUSTER_NAME --zone $ZONE
```
BenjaminPelletier: Let's add a step to create the Kubernetes cluster and note that a separate cluster should be used for each DSS instance.  This content can be re-added as a GCP-specific sub-bullet under that note.

config/README.md
BenjaminPelletier: Ok.  What was the original purpose of $NAMESPACE?  Are we just fixing the prior bad practice?

config/build.sh
```diff
@@ -12,8 +12,8 @@ set -e -x
...
 
 cd "${BASEDIR}"
-docker build -f cmds/http-gateway/Dockerfile  . -t gcr.io/$CLOUD_PROJECT/http-gateway:$VERSION
-docker build -f cmds/grpc-backend/Dockerfile  . -t gcr.io/$CLOUD_PROJECT/grpc-backend:$VERSION
+docker build -f cmds/http-gateway/Dockerfile  . -t $DOCKER_URL/http-gateway:$VERSION
+docker build -f cmds/grpc-backend/Dockerfile  . -t $DOCKER_URL/grpc-backend:$VERSION
 
-docker push gcr.io/$CLOUD_PROJECT/http-gateway:$VERSION
-docker push gcr.io/$CLOUD_PROJECT/grpc-backend:$VERSION
+docker push $DOCKER_URL/http-gateway:$VERSION
```
BenjaminPelletier: Let's make sure to exercise build.sh (and the other scripts) to ensure they still work after the changes.  We probably didn't do that as of last revision as the HTTP gateway wouldn't have deployed correctly without the suffixes.

config/values.yaml.template
```diff
@@ -50,7 +48,7 @@ JWTPublicCert: <FILENAME>
 
 
 # Optionally edit the values below.
-
+Namespace: dss-main
```
BenjaminPelletier: Let's add a warning that the certs scripts and some README commands will not work properly if this value is not "dss-main".

config/README.md
steeling: it was an artifact of the original cockroachdb docs, that put the database in a separate namespace. These systems are so interrelated I don't think there's benefit in putting each in its own namespace.

config/build.sh
steeling: done!

config/README.md
steeling: done

README.md
steeling: done!

config/README.md
steeling: done

config/README.md
steeling: which bullet point are specifically commenting on here?

config/README.md
steeling: https://kubernetes.io/docs/concepts/services-networking/ingress/ capitalized in kube docs

config/README.md
steeling: https://kubernetes.io/docs/concepts/services-networking/ingress/ capitalized in kube docs

config/README.md
steeling: from past teams i think we should avoid the word "instance" at all costs, as it's probably one of the most ambiguous and over used words in software engineering :)

Added a glossary to explain the terminology i'm using in this doc.

config/README.md
steeling: added a bullet point, planning on adding the functionality in the near future

config/README.md
steeling: done

config/README.md
steeling: done

config/README.md
steeling: done

config/README.md
steeling: done

config/README.md
steeling: done

config/README.md
steeling: done

config/README.md
steeling: done

config/README.md
steeling: added for the min CRDB nodes

config/README.md
```diff
@@ -1,135 +1,169 @@
...
 
-Build a new image:
+*   Every CockroachDB node must advertise a unique and routable address.
+    *   It is preferred to use domain names with unique prefixes and homogenous
+        suffixes, ie: 0.c.dss.interussplatform.com.
+    *   This allows wildcard usage in the CRDB certificates.
+*   Every DSS node should run a minimum of 3 CockroachDB nodes, which ensures
+    enough nodes are always to support failovers and gradual rollouts.
+*   At least 3 CockroacbDB addresses must be shared with all participants.
+    *   If not using the recommended hostname prefix above, every cockroachDB
```
BenjaminPelletier: CockroachDB

config/README.md
```diff
@@ -1,135 +1,169 @@
...
+Download & install the following tools to your workstation:
 
-## Creating a new Kubernetes cluster
+*   The [helm client](https://helm.sh/docs/using_helm/#installing-the-helm-client)
+    *   Required if deploying using the defined Kubernetes templates.
+*   kubectl
+    *   Required if deploying with Kubernetes.
+*   docker
+    *   Required if building new images of the DSS.
+*   Cockroachdb
```
BenjaminPelletier: CockroachDB or cockroach

config/make-certs.py
```diff
@@ -40,20 +50,16 @@ def node_certs_dir(self):
 def parse_args():
     parser = argparse.ArgumentParser(
         description='Creates certificates for a new Cockroachdb cluster')
```
BenjaminPelletier: CockroachDB

README.md
```diff
@@ -1,92 +1,90 @@
...
+
+
+Component responsible for all the business logic as well as authentication. This
+backend talks directly to CockroachDB.
+
+
+### CockroachDB (CRDB)
+
+Individual CockRoach DB instances hosting sharded data. More information about 
+CockRoachDB
```
BenjaminPelletier: CockroachDB

config/README.md
```diff
@@ -1,135 +1,169 @@
...
 
-*   helm
-*   kubectl
-*   docker
-*   cockroachdb
-*   Google Cloud SDK (if deploying on GCP)
-*   [Optional] Golang. Recommended to understand go, and the go toolchain.
+## Glossary
 
+*   DSS cluster - an entire synchronized DSS, typically operated by multiple
```
BenjaminPelletier: As below, let's keep our terminology consistent with the standard.  So, "DSS Region - The geographic scope supported by a set of Discovery and Synchronization Service instances." (definition from standard) and "DSS instance - a single deployed implementation of the DSS interface which synchronizes with other DSS instances in its DSS Region." (not replica because there can be different implementations)

README.md
```diff
@@ -1,92 +1,90 @@
...
+CockRoachDB 
+[here](https://www.cockroachlabs.com/docs/stable/architecture/overview.html).
 
 ## Directories of Interest:
-* `config/` has all of the configuration required to deploy a DSS instance to a kubernetes cluster. The README in that directory contains more information.
-* `pkg/` contains all of the source code for the DSS. See the README in that directory for more information.
-* `cmds/` contains entry points and docker files for the actual binaries (the `http-gateway` and `grpc-backend`)
-* 
+*   `config/` has all of the configuration required to deploy a DSS instance
+    (typically to a kubernetes cluster). The README in that directory contains
```
BenjaminPelletier: Kubernetes (capitalization)

config/README.md
```diff
@@ -1,135 +1,169 @@
...
+## Deploying the DSS on Kubernetes
+
+This secion discusses deploying a Kubernetes service, although you can deploy 
+a DSS node however you like as long as it meets the CockroachDB requirements
+above. You can do this on any supported
+[cloud provider](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/)
+or even on your own infrastructure.  Consult the Kubernetes documentation for
+which ever provider you choose.
+
+1.  Create a new Kubernetes cluster as mentioned above. We recommend a new
```
BenjaminPelletier: Where above is this mentioned?  It seems like original lines 45-62 would be useful to include, but are not actually included currently.

config/README.md
```diff
@@ -1,135 +1,169 @@
...
+    cluster for each logically separate DSS.
```
BenjaminPelletier: for each DSS instance

config/README.md
```diff
@@ -1,135 +1,169 @@
...
 
-    gcloud --project $CLOUD_PROJECT container images list-tags gcr.io/$CLOUD_PROJECT/http-gateway
 
-Build a new image:
+*   Every CockroachDB node must advertise a unique and routable address.
+    *   It is preferred to use domain names with unique prefixes and homogenous
+        suffixes, ie: 0.c.dss.interussplatform.com.
+    *   This allows wildcard usage in the CRDB certificates.
+*   Every DSS node should run a minimum of 3 CockroachDB nodes, which ensures
+    enough nodes are always to support failovers and gradual rollouts.
```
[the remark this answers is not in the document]
steeling: yup, fixed

README.md
steeling: done

README.md
steeling: done

config/README.md
steeling: I'm going to push back on the term instance, cause when I hear instance I think of the entire cluster, whereas somebody else might think of a node -- ie: it's much too ambiguous

I think DSS Node is a good term, as well as cluster. DSS region to me also means the serviced region, so would like to pick something more specific there as well.

Also this terminology more closely aligns with that of CockroachDB, so a +1 for cluster and Node.

config/README.md
steeling: The link directly above is for creating a kubernetes cluster

config/README.md
steeling: switched to DSS node

config/README.md
steeling: FYI https://www.cockroachlabs.com/docs/stable/architecture/overview.html#glossary

config/README.md
steeling: done

config/README.md
steeling: "Make sure you supply the following flags if joining an existing cluster".. i tried making this as clear as possible but open to suggestions if you have another way of wording it.

It's only required if joining an existing cluster, otherwise you omit it

README.md
```diff
@@ -27,66 +27,81 @@ operations stored or processed on the InterUSS Platform.
...
+### gRPC Backend
+
+
+Component responsible for all the business logic as well as authentication. This
+backend talks directly to CockroachDB.
+
+
+### CockroachDB (CRDB)
+
+Individual CockRoach DB instances hosting sharded data. More information about
```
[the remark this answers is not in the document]
steeling: i think i commented a few minutes before pushing

config/README.md
BenjaminPelletier: How about "Supply the following flags only if joining an existing cluster"?

But is that actually true?  If I am creating a cluster and I don't specify the --node-address flag, how will the addresses of my nodes be included in the node certificates?  And will another party be able to join my cluster later if I don't specify the addresses of my nodes in my nodes' certificates?

steeling: done. yes, it uses service names for internal discovery

config/README.md
steeling: it was fixed in the following commit

config/README.md
steeling: done

config/README.md
BenjaminPelletier: I think we may want to take another look at this section in the future regarding which node addresses need to be specified when, but I think that can be addressed outside this PR.

config/make-certs.py
BenjaminPelletier: We can address this in a later PR

config/README.md
BenjaminPelletier: We can address this in a later PR

PR #131, opened 2019-10-15
Add missing Test Cases

charlie-pisuraj:
Token Validation
	Get ISA with huge area
	Delete ISA with bad version
	Get Subscription with huge area
	Put Subscription on huge area
	Put too many Subscriptions on an area

	This commit exposes failures from the following tests:

	test_get_isa_by_search_huge_area
	Note: Error message is correct only code is wrong
		Expect code:	413
		Got code: 	400
	test_put_isa_with_read_only_scope_token
		Expect code:	403
		Got code: 	200
	test_get_isa_without_scope
		Expect code:	403
		Got code: 	200
	test_get_sub_by_searching_huge_area
	Note: Error message is correct only code is wrong
		Expect code:	413
		Got code: 	400

### DIFF
--- a/monitoring/prober/common.py
+++ b/monitoring/prober/common.py
@@ -1,23 +1,25 @@
 MAX_SUB_PER_AREA = 10
 
+MAX_SUB_TIME_HRS = 24
+
 DATE_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
 
 VERTICES = [
     {
-        'lat': 130.6205,
-        'lng': -23.6558
+        'lng': 130.6205,
+        'lat': -23.6558
     },
     {
-        'lat': 130.6301,
-        'lng': -23.6898
+        'lng': 130.6301,
+        'lat': -23.6898
     },
     {
-        'lat': 130.6700,
-        'lng': -23.6709
+        'lng': 130.6700,
+        'lat': -23.6709
     },
     {
-        'lat': 130.6466,
-        'lng': -23.6407
+        'lng': 130.6466,
+        'lat': -23.6407
     },
 ]
 
@@ -26,20 +28,20 @@ GEO_POLYGON_STRING = ','.join(
 
 HUGE_VERTICES = [
     {
-        'lat': 130,
-        'lng': -23
+        'lng': 130,
+        'lat': -23
     },
     {
-        'lat': 130,
-        'lng': -24
+        'lng': 130,
+        'lat': -24
     },
     {
-        'lat': 132,
-        'lng': -24
+        'lng': 132,
+        'lat': -24
     },
     {
-        'lat': 132,
-        'lng': -23
+        'lng': 132,
+        'lat': -23
     },
 ]
 
--- a/monitoring/prober/common.py
+++ b/monitoring/prober/common.py
@@ -1,3 +1,5 @@
+MAX_SUB_PER_AREA = 10
+
 DATE_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
 
 VERTICES = [
@@ -21,3 +23,25 @@ VERTICES = [
 
 GEO_POLYGON_STRING = ','.join(
     '{},{}'.format(x['lat'], x['lng']) for x in VERTICES)
+
+HUGE_VERTICES = [
+    {
+        'lat': 130,
+        'lng': -23
+    },
+    {
+        'lat': 130,
+        'lng': -24
+    },
+    {
+        'lat': 132,
+        'lng': -24
+    },
+    {
+        'lat': 132,
+        'lng': -23
+    },
+]
+
+HUGE_GEO_POLYGON_STRING = ','.join(
+    '{},{}'.format(x['lat'], x['lng']) for x in HUGE_VERTICES)
--- a/monitoring/prober/test_isa_simple.py
+++ b/monitoring/prober/test_isa_simple.py
@@ -13,143 +13,156 @@ import common
 
 
 def test_isa_does_not_exist(session, isa1_uuid):
-  resp = session.get('/identification_service_areas/{}'.format(isa1_uuid))
-  assert resp.status_code == 404
-  assert resp.json()['message'] == 'resource not found: {}'.format(isa1_uuid)
+    resp = session.get("/identification_service_areas/{}".format(isa1_uuid))
+    assert resp.status_code == 404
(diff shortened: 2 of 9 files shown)


charlie-pisuraj:
When you get a chance can I get a review @vosst @steeling for this.
As discussed offline 2 of the above test with token scope should be resolved with another API renaming

monitoring/prober/common.py
```diff
@@ -21,3 +23,25 @@
 
 GEO_POLYGON_STRING = ','.join(
     '{},{}'.format(x['lat'], x['lng']) for x in VERTICES)
+
+HUGE_VERTICES = [
+    {
+        'lat': 130,
```
BenjaminPelletier: Did this work before and does it work now?  130 degrees is an invalid latitude so it seems like we should return 400 regardless of anything else.  Presumably the latitude and longitude values are just swapped in the test data.  Same issue with VERTICES.

monitoring/prober/test_subscription_validation.py
```diff
@@ -0,0 +1,113 @@
...
+                'time_start': time_start.strftime(common.DATE_FORMAT),
+                'time_end': time_end.strftime(common.DATE_FORMAT),
+            },
+            'callbacks': {
+                'identification_service_area_url': 'https://example.com/foo'
+            },
+        })
+    all_resp.append(resp.status_code == 429)
+
+  assert any(all_resp)
```
BenjaminPelletier: Probably better to assert `resp.status_code == (429 if index == common.MAX_SUB_PER_AREA else 200)` on each loop iteration because we want to make sure the first MAX_SUB_PER_AREA calls succeed in addition to the final one failing.  If this test passes with the current code then something is wrong because common.VERTICES has latitudes that are greater than 90 degrees.

monitoring/prober/test_isa_simple.py
```diff
@@ -107,6 +112,24 @@ def test_get_isa_by_search_latest_time_excluded(session, isa1_uuid):
...
+  resp = session.get('/identification_service_areas'
+                     '?area={}'.format(common.GEO_POLYGON_STRING))
+  assert resp.status_code == 200
+  assert isa1_uuid in [x['id'] for x in resp.json()['service_areas']]
+
+
+def test_get_isa_by_search_huge_area(session, isa1_uuid):
+  resp = session.get('/identification_service_areas'
+                     '?area={}'.format(common.HUGE_GEO_POLYGON_STRING))
+  assert resp.status_code == 413
```
BenjaminPelletier: If this test passes currently, we should be suspicious of something -- HUGE_GEO_POLYGON_STRING has latitudes greater than 90 degrees, so we should return 400 before HUGE_VERTICES is fixed.

monitoring/prober/test_isa_validation.py
```diff
@@ -21,24 +21,7 @@ def test_isa_huge_area(session, isa1_uuid):
           'extents': {
               'spatial_volume': {
                   'footprint': {
-                      'vertices': [
-                          {
-                              'lat': 130,
```
BenjaminPelletier: Something is fishy here...5034048745.815665km² (the "expected" error message below) is substantially larger than the surface area of the earth so this seems like an invalid error message (true error message should be something about invalid latitudes with current input).  Let's make sure to add an issue to figure out what was going on, add a test to reproduce, and fix the problem.

monitoring/prober/test_isa_simple.py
charlie-pisuraj: This test sort of pass in a way. It returns the correct message indicating the area is too large. But the code returned is incorrect. So that means the server knows how to respond but just uses the wrong code

monitoring/prober/common.py
charlie-pisuraj: This worked. I found it weird too that all the vertices used throughout all the test are not on the surface of the earth, but I thought it might have been intentional. If you think we should go with a proper one then I can change it

monitoring/prober/test_subscription_validation.py
charlie-pisuraj: Through out other tests subscriptions are created in the same Area, that is the reason why I used 'any' instead of specific index, but what I can do is to make sure I clean up other test making sure there are zero subs until it gets to this one

charlie-pisuraj: I its probably easier if I just picked a different location, so we don't have to worry about cleaning it up if we add more test in the future.

monitoring/prober/common.py
BenjaminPelletier: Yes, there are two issues: with lat=130, this test should return 400.  So if it returned 413 with lat=130, we should add another test case (probably in a different PR) that reveals that flaw (the flaw of not returning 400 for invalid latitudes).  Also, we should fix VERTICES and HUGE_VERTICES to have valid lat/lngs (probably by switching the lat and lng labels) to make these tests behave as intended.

monitoring/prober/test_subscription_validation.py
BenjaminPelletier: That sounds good -- in general, we should make sure that tests don't interfere with each other.  It seems like the easiest way to accomplish that would be to have each test clean up after itself, but ensuring that every area-using test has its own area would work too.

monitoring/prober/common.py
charlie-pisuraj: Other tests that uses  VERTICES that create areas returns 200, that means there is something wrong with the happy case too, we should not be able to create ISA or subs that is not on the surface of the earth

monitoring/prober/test_isa_simple.py
BenjaminPelletier: Yeah, there are two things wrong here.  First, lat=130 (invalid area) should return 400, not 413.  Second, valid areas that are too large should return 413, not 400.  In a different PR, we should add a test with lat=130 to reveal the additional flaw.

monitoring/prober/test_subscription_validation.py
```diff
@@ -0,0 +1,113 @@
...
+              'time_end': time_end.strftime(common.DATE_FORMAT),
+          },
+          'callbacks': {
+              'identification_service_area_url': 'https://example.com/foo'
+          },
+      })
+  assert resp.status_code == 400
+
+
+def test_create_too_man_subs(session):
```
steeling: too_many*

PR #134, opened 2019-10-16
[docker] Unify Dockerfile

vosst:


vosst:
> I see that this reduces the number of builds in Azure Pipelines, but is that important? It seems like an advantage to have each image be exclusively for a particular task rather than having a single image that can do either of two tasks. But if we do accept a polymorphic docker image, this PR should also change docker_e2e.sh appropriately.

A single image makes it easier to atomically upgrade both entities. While the gateway and the backend are separate in principle, they are nevertheless tightly coupled. Aiming to reflect this in the primary artifact produced from the code.

vosst:
> > I see that this reduces the number of builds in Azure Pipelines, but is that important? It seems like an advantage to have each image be exclusively for a particular task rather than having a single image that can do either of two tasks. But if we do accept a polymorphic docker image, this PR should also change docker_e2e.sh appropriately.
> 
> A single image makes it easier to atomically upgrade both entities. While the gateway and the backend are separate in principle, they are nevertheless tightly coupled. Aiming to reflect this in the primary artifact produced from the code.

e2e tests are updated.

test/docker_e2e.sh
```diff
@@ -28,18 +28,20 @@ docker run -d --rm --name dss-crdb-for-debugging \
 	--insecure > /dev/null
 
 sleep 5
-echo " ------------ GRPC BACKEND ---------------- "
-echo "Building grpc-backend Container"
-docker build -q --rm -f cmds/grpc-backend/Dockerfile . -t local-grpc-backend
+echo " -------------- BOOTSTRAP ----------------- "
+echo "Building local container for testing"
+docker build -q --rm . -t local-image
```
BenjaminPelletier: nit: maybe local-interuss-dss-image to avoid collisions

PR #135, opened 2019-10-16
[interfaces] Pull in API specs from upstream

vosst:


pkg/dssproto/dss.proto
```diff
@@ -211,13 +211,18 @@ message LatLngPoint {
...
+  // The new or updated extents of the Identification Service Area.
+  // 
+  // Omitted if Identification Service Area was deleted.
+  Volume4D extents = 1;
+
   // Identification Service Area that the notifying client changed or created.
   // 
   // If this field is populated, the Identification Service Area was created or updated.  If this field is not populated, the Identification Service Area was deleted.
-  IdentificationServiceArea service_area = 1;
+  IdentificationServiceArea service_area = 2;
```
BenjaminPelletier: Not a problem for this PR, but we'll have to note that gRPC message indices aren't guaranteed to stay constant -- that's atypical for a gRPC approach but I think is acceptable here because gRPC is just behind the HTTP bridge.

Makefile
```diff
@@ -29,8 +29,12 @@ pkg/dssproto/dss.pb.gw.go: pkg/dssproto/dss.proto
 	protoc -I/usr/local/include -I.   -I$(GOPATH)/src   -I$(GOPATH)/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@v1.9.6/third_party/googleapis   --grpc-gateway_out=logtostderr=true,allow_delete_body=true:. pkg/dssproto/dss.proto
 
 pkg/dssproto/dss.proto: install-proto-generation
-	go run github.com/NYTimes/openapi2proto/cmd/openapi2proto -spec api.yaml -annotate -out pkg/dssproto/dss.proto -tag dss -indent 2 -package dssproto
-
+	go run github.com/NYTimes/openapi2proto/cmd/openapi2proto \
+		-spec interfaces/uastech/standards/remoteid/canonical.yaml -annotate \
```
steeling: Might be worth doing in this PR, or we can come back to it.. but if we add "operationId" to each endpoint we can get cleaner names. You can see how I did it previously here https://github.com/interuss/dss/blob/af89392461e083477be37bf9af77de1023ac526d/api.yaml. Feel free to defer to a later PR though.

pkg/dssproto/dss.proto
steeling: This actually is a breaking change. So previous versions will not be compatible with this version. It's early days, so we could agree to take some down time, but I think we should probably just keep the indices constant.

vosst: Note that this file is auto-generated. I *think* we might have to adjust the upstream API to alter the member ordering. Will investigate how the indices are determined.

steeling: At some point we also might want to consider abandoning auto generation for reasons like this

Makefile
BenjaminPelletier: operationId tags will need to go in the standards repo which means those changes can't be in this PR.  That sounds like a good idea, but they shouldn't go in canonical.yaml -- instead, we should kick off the optionally-augmented API by branching canonical.yaml to augmented.yaml (or something like that).

vosst: +1, ping when done and I'll adjust this PR.

BenjaminPelletier: operationId tags added to newly-created `augmented.yaml`.

pkg/dssproto/dss.proto
BenjaminPelletier: A one-time breaking change should be ok for this PR because there won't be any impact.  We can discuss further following this PR.

pkg/dssproto/dss.proto
```diff
@@ -513,7 +518,7 @@ message Volume4D {
   google.protobuf.Timestamp time_start = 3;
 }
 
-service DiscoveryAndSynchronizationService {
+service DiscoveryAndSynchronizationServiceUSSDSSUSSUSSService {
```
BenjaminPelletier: augmented.yaml has the USSDSSUSSUSS removed from title.

Makefile
vosst: Pulled in, thanks.

pkg/dssproto/dss.proto
```diff
@@ -76,6 +76,14 @@ message CreateIdentificationServiceAreaParameters {
   string flights_url = 2;
 }
```
steeling: FYI: A lot of the top level proto definitions are unused, ie: verticalAccuracy. Just wanted to highlight, you're call on how you want to address that

pkg/dssproto/dss.proto
```diff
@@ -514,94 +519,94 @@ message Volume4D {
...
 service DiscoveryAndSynchronizationService {
-  // /dss/identification_service_areas/{id}/{version}
+  // /dss/identification_service_areas/{id}
   // 
-  // Delete an Identification Service Area.  USSs should not delete Identification Service Areas before the end of the last managed flight plus the retention period.
-  rpc DeleteV1DssIdentificationServiceAreasIdVersion(DeleteV1DssIdentificationServiceAreasIdVersionRequest) returns (DeleteIdentificationServiceAreaResponse) {
+  // Create a new Identification Service Area.  This call will fail if an Identification Service Area with the same ID already exists.
+  // 
+  // The DSS assumes the USS has already added the appropriate retention period to operation end time in `time_end` field before storing it.
+  rpc CreateIdentificationServiceArea(CreateIdentificationServiceAreaRequest) returns (PutIdentificationServiceAreaResponse) {
```
steeling: Can we rename this to put? This can also be an udpate

Issue #89, opened 2019-08-21
Conformance/E2E tests


steeling:
We should add E2E conformance tests to validate code changes at the API level

vosst:
@charlie-pisuraj @steeling with the tests being integrated into CI, do you think we can close out this ticket?

charlie-pisuraj:
@vosst I'm still working on other parts of the conformance test that tests the interoperability the DSS, the plan is to spin up 2 copies of the DSS and run through a series of tests to ensure that both  work together properly

charlie-pisuraj:
This was merged in to master, we can close this issue

PR #144, opened 2019-10-22
Interoperability Test Suite

charlie-pisuraj:
A test suite used to for testing the interoperability between DSS based on the ASTM standards.
Ensuring that every action taken in one DSS are reflected in the other DSSs

#89

### DIFF
--- a/test/interoperability/clients.py
+++ b/test/interoperability/clients.py
@@ -1,0 +1,95 @@
+import requests
+import time
+from enum import Enum
+from google.auth.transport import requests as google_requests
+from google.oauth2 import service_account
+from typing import Optional, Dict, List, Any
+import urllib
+
+
+class AuthType(Enum):
+    NONE = 0
+    SVC_ACC = 1
+    PASSWORD = 2
+
+
+SCOPES = [
+    "dss.write.identification_service_areas",
+    "dss.read.identification_service_areas",
+]
+
+
+class OAuthClient:
+    def __init__(
+        self,
+        endpoint: str,
+        auth_type: AuthType,
+        service_account_json: Optional[str] = "",
+        username: Optional[str] = "",
+        password: Optional[str] = "",
+        client_id: Optional[str] = "",
+    ):
+        self._endpoint = endpoint
+        self._token_cache = {}
+        self._req_params = {}
+        self.req = requests.Session()
+
+        if auth_type is AuthType.SVC_ACC:
+            credentials = service_account.Credentials.from_service_account_file(
+                service_account_json
+            ).with_scopes(["email"])
+            self.req = google_requests.AuthorizedSession(credentials)
+            self._req_params = {"grant_type": "client_credentials"}
+        elif auth_type is AuthType.PASSWORD:
+            self._req_params = {
+                "grant_type": "password",
+                "username": username,
+                "password": password,
+                "client_id": client_id,
+            }
+        self.parameterized_url = False
+
+    def _isExpired(self, token: Dict[str, Any]) -> bool:
+        expiration = token.get("expire_time")
+        if not expiration:
+            return False
+        return expiration > time.time()
+
+    def getToken(self, scopes_list: List[str], audience: str) -> str:
+        scopes = " ".join(scopes_list)
+        token = self._token_cache.get((scopes, audience))
+        if token is None or self._isExpired(token):
+            self._token_cache[(scopes, audience)] = self._issueToken(scopes, audience)
+        return self._token_cache[(scopes, audience)].get("access_token", "")
+
+    def _issueToken(self, scopes, audience) -> Dict[str, Any]:
+        data = {"scope": scopes, "intended_audience": audience}
+        data.update(self._req_params)
+        if self.parameterized_url:
+            param_str = "?" + "&".join([f"{key}={val}" for key, val in data.items()])
+            response = self.req.post((self._endpoint + param_str))
+        else:
+            response = self.req.post(self._endpoint, data=data)
+        response.raise_for_status()
+        return response.json()
+
+
+class DSSClient(requests.Session):
+    def __init__(self, host: str, oauth_client: OAuthClient):
+        super().__init__()
+        self._host = host
+        self._oauth_client = oauth_client
+        self.scope: List[str] = []
+        self.intended_audience: str = ""
+        self.scope: List[str] = [
+            "dss.write.identification_service_areas",
+            "dss.read.identification_service_areas",
+        ]
+        self.intended_audience = urllib.parse.urlparse(host).hostname
+
+    def prepare_request(self, request, **kwargs) -> requests.request:
+        token = self._oauth_client.getToken(self.scope, self.intended_audience)
+        if request.url.startswith("/"):
+            request.url = self._host + request.url
+        request.headers["Authorization"] = f"Bearer {token}"
+        return super().prepare_request(request, **kwargs)
--- a/test/interoperability/clients.py
+++ b/test/interoperability/clients.py
@@ -3,20 +3,14 @@ import time
 from enum import Enum
 from google.auth.transport import requests as google_requests
 from google.oauth2 import service_account
-from typing import Optional, Dict, List, Any
+from typing import Optional, Dict, List, Any, Tuple
 import urllib
 
 
 class AuthType(Enum):
     NONE = 0
-    SVC_ACC = 1
+    SERVICE_ACCOUNT = 1
     PASSWORD = 2
-
-
-SCOPES = [
-    "dss.write.identification_service_areas",
-    "dss.read.identification_service_areas",
-]
 
 
 class OAuthClient:
@@ -30,11 +24,11 @@ class OAuthClient:
         client_id: Optional[str] = "",
     ):
         self._endpoint = endpoint
-        self._token_cache = {}
-        self._req_params = {}
+        self.token_response_cache: Dict[Tuple[str, str], Dict[str, Any]] = {}
+        self._req_params: Dict[str, str] = {}
         self.req = requests.Session()
 
-        if auth_type is AuthType.SVC_ACC:
+        if auth_type is AuthType.SERVICE_ACCOUNT:
             credentials = service_account.Credentials.from_service_account_file(
                 service_account_json
             ).with_scopes(["email"])
@@ -47,31 +41,45 @@ class OAuthClient:
                 "password": password,
                 "client_id": client_id,
             }
+        elif auth_type is AuthType.NONE:
+            # No special setup requred
+            pass
+        else:
+            # Something unknown was passed in
+            raise ("Unknown OAuth authentication Type")
         self.parameterized_url = False
 
     def _isExpired(self, token: Dict[str, Any]) -> bool:
-        expiration = token.get("expire_time")
+        expiration = token.get("expires_in")
+        request_time = token["request_time"]
         if not expiration:
             return False
-        return expiration > time.time()
+        return time.time() > (request_time + expiration)
 
     def getToken(self, scopes_list: List[str], audience: str) -> str:
         scopes = " ".join(scopes_list)
-        token = self._token_cache.get((scopes, audience))
+        token = self.token_response_cache.get((scopes, audience))
         if token is None or self._isExpired(token):
-            self._token_cache[(scopes, audience)] = self._issueToken(scopes, audience)
-        return self._token_cache[(scopes, audience)].get("access_token", "")
+            self.token_response_cache[(scopes, audience)] = self._issueToken(scopes, audience)
+        return self.token_response_cache[(scopes, audience)].get("access_token", "")
 
     def _issueToken(self, scopes, audience) -> Dict[str, Any]:
         data = {"scope": scopes, "intended_audience": audience}
         data.update(self._req_params)
+        request_time = time.time()
+        # Parameteterized URL is required to work with the Dummy Oauth Server
+        # as it currently doesn't read from parameters from request body
+        # TODO(charlie-pisuraj): Make Dummy OAuth Read params from request body
         if self.parameterized_url:
             param_str = "?" + "&".join([f"{key}={val}" for key, val in data.items()])
             response = self.req.post((self._endpoint + param_str))
         else:
+            # methods is a list of Tuples
             response = self.req.post(self._endpoint, data=data)
         response.raise_for_status()
-        return response.json()
+        token = response.json()
+        token["request_time"] = request_time
+        return token
 
 
 class DSSClient(requests.Session):
@@ -79,7 +87,6 @@ class DSSClient(requests.Session):
         super().__init__()
         self._host = host
         self._oauth_client = oauth_client
-        self.scope: List[str] = []
         self.intended_audience: str = ""
         self.scope: List[str] = [
             "dss.write.identification_service_areas",
--- a/test/interoperability/interop.py
+++ b/test/interoperability/interop.py
@@ -13,13 +13,18 @@ from typing import Dict
 def parseArgs() -> argparse.Namespace:
     parser = argparse.ArgumentParser(description="Test Interoperability of DSSs")
     parser.add_argument("OAuth", help="URI to the OAuth Server.")
+
+    # When using Password OAuth flow, Username, Password, and Clients-id are
+    # necessary for authentication
     parser.add_argument("--username", help="Username used to get OAuth Token")
     parser.add_argument("--password", help="Password used to get OAuth Token")
     parser.add_argument(
-        "--clients-_id",
+        "--client-id",
         help="Client ID used to get OAuth Token, used with Username and Password",
     )
 
+    # When using Service Account OAuth flow, only the Service Account JSON File
+    # is required to request Token.
     parser.add_argument(
         "--service-account",
         "--svc",
@@ -37,9 +42,9 @@ def main() -> int:
     args = parseArgs()
 
     if args.service_account:
-        oauth_client = clients.OauthClient(
+        oauth_client = clients.OAuthClient(
             args.OAuth,
-            clients.AuthType.SVC_ACC,
+            clients.AuthType.SERVICE_ACCOUNT,
             service_account_json=args.service_account,
         )
     elif args.username:
@@ -64,8 +69,8 @@ def main() -> int:
     tests = InterOpTestSuite(dss_clients)
     tests.startTest()
 
-
     return os.EX_OK
+
 
 if __name__ == "__main__":
     sys.exit(main())
--- a/test/interoperability/interop.py
+++ b/test/interoperability/interop.py
@@ -1,0 +1,71 @@
+#!env/bin/python3
+
+import os
+import sys
+import argparse
+import clients
+import datetime
+import uuid
+from interop_test_suite import InterOpTestSuite
+from typing import Dict
+
+
+def parseArgs() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description="Test Interoperability of DSSs")
+    parser.add_argument("OAuth", help="URI to the OAuth Server.")
+    parser.add_argument("--username", help="Username used to get OAuth Token")
+    parser.add_argument("--password", help="Password used to get OAuth Token")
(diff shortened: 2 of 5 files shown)


charlie-pisuraj:
Sorry for the large PR, I can refactor this if you feel its too ugly

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+                extra_subs == set()
+            ), f"{dss} returned too many Subscriptsions, extra: {extra_subs}"
+
+    def testStep5(
+        self, dss_map: Dict[str, clients.DSSClient], primary_dss: str, **kwargs
+    ) -> None:
+        """Can modify ISA in primary DSS, ISA modification triggers 
+        subscription notification requests"""
+        isa_1_uuid = self.context["isa_1_uuid"].key
+        resp = dss_map[primary_dss].get(f"/identification_service_areas/{isa_1_uuid}")
```
BenjaminPelletier: We should retain `version` from a previous call; a compliant DSS is not required to implement this endpoint.

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+        put_resp = dss_map[primary_dss].put(
+            f"/identification_service_areas/{isa_1_uuid}/{data['version']}",
+            json={
+                "extents": {
+                    "spatial_volume": {
+                        "footprint": {"vertices": VERTICES},
+                        "altitude_lo": 20,
+                        "altitude_hi": 400,
+                    },
+                    "time_start": data["time_start"],
```
BenjaminPelletier: nit: This argument can be omitted and it will be unchanged

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+                    "time_end": time_end.strftime(DATE_FORMAT),
+                },
+                "flights_url": "https://example.com/dss",
```
BenjaminPelletier: Same URL comment as above

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+        assert updated_data["service_area"]["time_end"] == time_end.strftime(
+            DATE_FORMAT
+        ), f"Unsuccessful Update; no change to end time"
+
+    def testStep6(
+        self, dss_map: Dict[str, clients.DSSClient], primary_dss: str, **kwargs
+    ) -> None:
+        """Can delete all Subscription in primary DSS"""
+        dss = dss_map[primary_dss]
+        for name, context in self.context.items():
```
BenjaminPelletier: This is confusing; should it be "self.contexts" if each value is a context?

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+            if name.startswith("sub_1_"):
+                resp = dss.delete(f"/subscriptions/{context.key}/")
```
BenjaminPelletier: This should fail; no `version` parameter is supplied in the path

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+        all_dss = [primary_dss] + all_other_dss
+        for dss in all_dss:
+            resp = dss_map[dss].get(f"/subscriptions?area={GEO_POLYGON_STRING}")
+            assert (
+                resp.status_code == 200
+            ), f"Expecting code 200, found {resp.status_code}"
+
+            data = resp.json()
+            assert (
+                data["subscriptions"] == []
```
BenjaminPelletier: To be as robust as possible, we actually only care that none of the known-deleted Subscriptions are included.  There could perhaps be pre-existing Subscriptions.

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+            resp = dss_map[dss].put(
+                f"/subscriptions/{sub_2_uuid}",
+                json={
+                    "extents": {
+                        "spatial_volume": {
+                            "footprint": {"vertices": VERTICES},
+                            "altitude_lo": 20,
+                            "altitude_hi": 400,
+                        },
+                        "time_start": time_start.strftime(DATE_FORMAT),
```
BenjaminPelletier: Same comment as above

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+                        "time_end": time_end.strftime(DATE_FORMAT),
+                    },
+                    "callbacks": {
+                        "identification_service_area_url": "https://example.com/foo"
```
BenjaminPelletier: Same comment as above

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+                    },
+                },
+            )
+            assert resp.status_code == 200, f"Failed to Insert Subscription to {dss}"
+            data = resp.json()
+            isa_ids = [isa["id"] for isa in data["service_areas"]]
+            assert (
+                self.context["isa_1_uuid"].key not in isa_ids
+            ), f"{dss} return ISA_1 when creating Subscription"
```
BenjaminPelletier: Let's make it clear why this is bad (e.g., "...returned expired ISA_1...")

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+        resp = dss_map[primary_dss].put(
+            f"/identification_service_areas/{self.context['isa_2_uuid'].key}",
+            json={
+                "extents": {
+                    "spatial_volume": {
+                        "footprint": {"vertices": VERTICES},
+                        "altitude_lo": 20,
+                        "altitude_hi": 400,
+                    },
+                    "time_start": time_start.strftime(DATE_FORMAT),
```
BenjaminPelletier: Same comment as above

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+                    "time_end": time_end.strftime(DATE_FORMAT),
+                },
+                "flights_url": "https://example.com/dss",
```
BenjaminPelletier: Same comment as above

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+            },
+        )
+        assert (
+            resp.status_code == 200
+        ), f"Failed to insert ISA to {primary_dss}. Error: {resp.json()['error']}"
+        subscribers = resp.json()["subscribers"]
+
+        assert (
+            subscribers == []
```
BenjaminPelletier: To be as robust as possible, we actually only care that none of the known-expired Subscriptions are included. There could perhaps be pre-existing Subscriptions.

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+        ), f"Expecting empty Subscribers list, found {len(subscribers)} subscribers"
```
BenjaminPelletier: Let's be very clear about why this is bad (e.g., "found [...] subscribers to notify even though they were expired")

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+        all_dss = [primary_dss] + all_other_dss
+        all_sub_2 = set()
+        for index in range(len(all_dss)):
+            all_sub_2.add(self.context[f"sub_2_{index}_uuid"].key)
+        for index, dss in enumerate(all_dss):
+            resp = dss_map[dss].get(f"/subscriptions?area={GEO_POLYGON_STRING}")
+            assert resp.status_code == 200, f"{dss} failed to get SUB_2 by area"
+
+            subs = resp.json()["subscriptions"]
+            assert subs == [], (
```
BenjaminPelletier: Same comment as above

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+                f"Expecting empty Subscriptions list, "
+                f"found {len(subs)} Subscriptions from {dss}"
```
BenjaminPelletier: Same comment as above

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+            sub_2_uuid = self.context[f"sub_2_{index}_uuid"].key
+            resp = dss_map[dss].get(f"/subscriptions/{sub_2_uuid}")
+            assert (
+                resp.status_code == 404
+            ), f"Expecting code 404, found {resp.status_code}"
+
+    def testStep15(
+        self, dss_map: Dict[str, clients.DSSClient], primary_dss: str, **kwargs
+    ) -> None:
+        """ISA deletion triggers does not trigger subscription
```
BenjaminPelletier: This was a typo in the guidance document; remove "triggers"

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+        notification requests for expired Subscriptions"""
+        isa_3_uuid = self.context["isa_3_uuid"].key
+        resp = dss_map[primary_dss].delete(
+            f"/identification_service_areas/{isa_3_uuid}/"
+        )
+        assert resp.status_code == 200, f"Failed to delete ISA_3 from {primary_dss}"
+
+        subs = resp.json()["subscribers"]
+        assert subs == [], (
```
BenjaminPelletier: Same comment as above

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+            resp = dss_map[dss].put(
+                f"/subscriptions/{sub_3_uuid}",
+                json={
+                    "extents": {
+                        "spatial_volume": {
+                            "footprint": {"vertices": VERTICES},
+                            "altitude_lo": 20,
+                            "altitude_hi": 400,
+                        },
+                        "time_start": time_start.strftime(DATE_FORMAT),
```
BenjaminPelletier: Same comment as above

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+                        "time_end": time_end.strftime(DATE_FORMAT),
+                    },
+                    "callbacks": {
+                        "identification_service_area_url": "https://example.com/foo"
```
BenjaminPelletier: Same comment as above

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+                    },
+                },
+            )
+            assert resp.status_code == 200, f"Failed to Insert Subscription to {dss}"
+            data = resp.json()
+            isa_ids = [isa["id"] for isa in data["service_areas"]]
+            assert (
+                self.context["isa_3_uuid"].key not in isa_ids
+            ), f"{dss} returned ISA_3 when creating Subscription"
```
BenjaminPelletier: Let's be extra clear why this is bad

test/interoperability/clients.py
```diff
@@ -0,0 +1,95 @@
...
+        token = self._token_cache.get((scopes, audience))
+        if token is None or self._isExpired(token):
+            self._token_cache[(scopes, audience)] = self._issueToken(scopes, audience)
+        return self._token_cache[(scopes, audience)].get("access_token", "")
+
+    def _issueToken(self, scopes, audience) -> Dict[str, Any]:
+        data = {"scope": scopes, "intended_audience": audience}
+        data.update(self._req_params)
+        if self.parameterized_url:
+            param_str = "?" + "&".join([f"{key}={val}" for key, val in data.items()])
```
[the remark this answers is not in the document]
charlie-pisuraj: I was debating this, Either I change the the Dummy OAuth Server to read from the request body or do this. I can make the changes to the dummy-oauth server when I fix this

test/interoperability/clients.py
```diff
@@ -0,0 +1,95 @@
...
+        response.raise_for_status()
+        return response.json()
+
+
+class DSSClient(requests.Session):
+    def __init__(self, host: str, oauth_client: OAuthClient):
+        super().__init__()
+        self._host = host
+        self._oauth_client = oauth_client
+        self.scope: List[str] = []
```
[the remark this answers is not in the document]
charlie-pisuraj: I left this opened so that if we want to add a test to use read only / invalid scopes in the request we can modify it before making the call.

test/interoperability/interop.py
```diff
@@ -0,0 +1,71 @@
...
+from typing import Dict
+
+
+def parseArgs() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description="Test Interoperability of DSSs")
+    parser.add_argument("OAuth", help="URI to the OAuth Server.")
+    parser.add_argument("--username", help="Username used to get OAuth Token")
+    parser.add_argument("--password", help="Password used to get OAuth Token")
+    parser.add_argument(
+        "--clients-_id",
```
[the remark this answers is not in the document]
charlie-pisuraj: it was a typo

test/interoperability/interop.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+    return parser.parse_args()
+
+
+def main() -> int:
+    args = parseArgs()
+
+    if args.service_account:
+        oauth_client = clients.OauthClient(
```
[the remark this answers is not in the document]
charlie-pisuraj: Service account flow was not tested yet, will need to setup one that mimics the setup we had for the demo

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+
+    def _getTests(self) -> Dict[str, object]:
+        methods = inspect.getmembers(TestSteps, predicate=inspect.isfunction)
+        result: Dict[str, object] = {}
+        ordered_result = collections.OrderedDict()
+        for name, obj in methods:
+            if name.startswith("testStep"):
+                result[name] = obj
+        # sort by name
+        for i in range(1, len(result) + 1):
```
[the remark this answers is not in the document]
charlie-pisuraj: So this would end up sorting like this 
testStep1
testStep10
testStep11
.....
testStep2
.....
Since I know that it follows the numbering scheme, I went with this approach. I can write a lambda to make it sort numerically

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+        resp = dss_map[primary_dss].put(
+            f"/identification_service_areas/{self.context['isa_1_uuid'].key}",
+            json={
+                "extents": {
+                    "spatial_volume": {
+                        "footprint": {"vertices": VERTICES},
+                        "altitude_lo": 20,
+                        "altitude_hi": 400,
+                    },
+                    "time_start": time_start.strftime(DATE_FORMAT),
```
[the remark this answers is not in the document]
charlie-pisuraj: so we should remove all start times and have the DSS itself populate it?

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+import collections
+import time
+import logging
+
+from typing import Dict, Any, List
+
+logging.basicConfig(level=logging.INFO)
+LOG = logging.getLogger(__name__)
+
+DATE_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
```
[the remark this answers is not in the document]
charlie-pisuraj: Just a heads up this format is used across all of out tests. Do you want to change it there too?

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+                        "altitude_lo": 20,
+                        "altitude_hi": 400,
+                    },
+                    "time_start": time_start.strftime(DATE_FORMAT),
+                    "time_end": time_end.strftime(DATE_FORMAT),
+                },
+                "flights_url": "https://example.com/dss",
+            },
+        )
+        assert (
```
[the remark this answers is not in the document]
charlie-pisuraj: So i most cases the subsequent test depends on the action of the previous tests, if we don't abort the next test will likely fail. What we could do is to clean up the current test loop and move on and permute to the next DSS as the primary DSS.

test/interoperability/interop_test_suite.py
charlie-pisuraj: Really? I thought it was required based on this https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/BenjaminPelletier/DiscoveryAndSynchronization/rid/api/dss.yaml#tag/dss/paths/~1v1~1dss~1identification_service_areas~1{id}/get

test/interoperability/interop_test_suite.py
charlie-pisuraj: I'll call it "entity" then

test/interoperability/interop_test_suite.py
charlie-pisuraj: is this case to make it the same as other tests, we should probably check for specific subscribers, since we do have their ID

test/interoperability/clients.py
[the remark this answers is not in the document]
BenjaminPelletier: Ah, makes sense -- let's leave a comment explaining why the parameters need to be in the URL rather than the request's params.

test/interoperability/clients.py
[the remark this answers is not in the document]
BenjaminPelletier: Not sure I follow; line 84 overwrites anything line 82 does so that line 82 doesn't do anything, right?

test/interoperability/interop_test_suite.py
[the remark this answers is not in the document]
BenjaminPelletier: Eh...I suppose it doesn't matter too much in either place (here or elsewhere).  Probably not a bad idea to have different formats different places to make sure all of them parse appropriately.

test/interoperability/interop_test_suite.py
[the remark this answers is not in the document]
BenjaminPelletier: Ah, right.  If leaving as-is, let's at least comment on TestSteps that only methods following the convention testStep%d will be run as test steps.

test/interoperability/interop_test_suite.py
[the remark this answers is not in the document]
BenjaminPelletier: I think so, yes, except for ones not in the immediate future

test/interoperability/clients.py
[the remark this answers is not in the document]
charlie-pisuraj: my bad, this line was a typo

test/interoperability/interop.py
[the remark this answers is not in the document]
charlie-pisuraj: Ok Tested now and works just fine, will ensure the Password flow has no issues too

test/interoperability/interop_test_suite.py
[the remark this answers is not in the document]
charlie-pisuraj: ok if thats the case, then I can just use the isoformat call here on datetime since it formats it exactly as you ask

test/interoperability/interop_test_suite.py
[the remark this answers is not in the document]
charlie-pisuraj: Created #151

test/interoperability/interop_test_suite.py
```diff
@@ -0,0 +1,537 @@
...
+        primary_dss: str,
+        all_other_dss: List[str],
+    ) -> None:
+        """Can query all Subscriptions in area from all DSSs."""
+        all_dss = [primary_dss] + all_other_dss
+        all_sub_1 = set()
+        for index in range(len(all_dss)):
+            all_sub_1.add(self.context[f"sub_1_{index}_uuid"].key)
+        for index, dss in enumerate(all_dss):
+            resp = dss_map[dss].get(f"/subscriptions?area={GEO_POLYGON_STRING}")
```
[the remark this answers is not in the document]
charlie-pisuraj: Created #152

test/interoperability/interop_test_suite.py
charlie-pisuraj: Created #153

BenjaminPelletier:
AzurePipelines /run

Issue #127, opened 2019-10-11
Make DSS reject token with missing 'iss' and 'exp' field


charlie-pisuraj:
Currently the DSS accepts token that are missing the 'iss' and 'exp' field. The [standard requires](https://tiny.cc/dssapi_rid) that the DSS must reject the request with incomplete information. This issue was found when reviewing PR #123 

**To Reproduce**
1. Generate a token with a dummy oauth server with only 'scope', 'sub', and '
aud'
2. Make a request to the DSS with generated token
3. DSS accepts token and return valid result

**Expected behavior**
Status code 4xx should be returned along with error message

PR #146, opened 2019-10-23
Add Check for Token Issuer and Expiration

charlie-pisuraj:
Makes the DSS reject token with missing Expiration and Issuer field #127 
In addition the standard mentions that the token cannot be valid for more than 1 hour into the future so we also check it and reject token are are given too wide of a window

### DIFF
--- a/cmds/dummy-oauth/main.go
+++ b/cmds/dummy-oauth/main.go
@@ -34,32 +34,27 @@ func createGetTokenHandler(privateKey *rsa.PrivateKey) http.Handler {
 		params := r.URL.Query()
 
 		var (
-			aud  string   = ""
-			auds []string = params["intended_audience"]
+			aud         string   = ""
+			auds        []string = params["intended_audience"]
+			scope       string   = ""
+			scopes      []string = params["scope"]
+			issuer      string   = ""
+			issuers     []string = params["issuer"]
+			expireTime  int64    = time.Now().Add(time.Hour).Unix()
+			expireTimes []string = params["expire"]
 		)
 		if len(auds) == 1 {
 			aud = auds[0]
 		}
 
-		var (
-			scope  string   = ""
-			scopes []string = params["scope"]
-		)
 		if len(scopes) == 1 {
 			scope = scopes[0]
 		}
 
-		var (
-			issuer  string   = ""
-			issuers []string = params["issuer"]
-		)
 		if len(issuers) == 1 {
 			issuer = issuers[0]
 		}
-		var (
-			expireTime  int64    = time.Now().Add(time.Hour).Unix()
-			expireTimes []string = params["expire"]
-		)
+
 		if len(expireTimes) == 1 {
 			parsedTime, err := strconv.ParseInt(expireTimes[0], 10, 64)
 			if err != nil {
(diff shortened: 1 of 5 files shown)


cmds/dummy-oauth/main.go
```diff
@@ -46,9 +48,29 @@ func createGetTokenHandler(privateKey *rsa.PrivateKey) http.Handler {
...
 		}
+
+		var (
+			issuer  string   = ""
+			issuers []string = params["issuer"]
+		)
+		if len(issuers) == 1 {
+			issuer = issuers[0]
+		}
+		var (
```
steeling: let's put all of the var blocks at the top of this function. it's getting a little long, and there's no early return so no extra cost in creating them

pkg/dss/auth/auth.go
```diff
@@ -42,6 +42,14 @@ func (m *missingScopesError) Error() string {
 	return strings.Join(m.s, ", ")
 }
 
+type tokenExpireError struct {
```
steeling: nit: how about 2 vars: tokenExpired, and tokenExpiredTooFar, ie:

tokenExpired = errors.New("the token is already expired")

pkg/dss/auth/auth.go
```diff
@@ -229,9 +237,31 @@ func (a *Authorizer) AuthInterceptor(ctx context.Context, req interface{}, info
...
 
+func (a *Authorizer) checkExpired(claimedExpireTime int64) error {
+	now := time.Now()
+	if claimedExpireTime < now.Unix() {
+		return &tokenExpireError{msg: "Token Expired or Expiration missing"}
+	}
+
+	if claimedExpireTime > now.Add(time.Hour).Unix() {
+		return &tokenExpireError{
+			msg: "Token expiration time is too far in the furture, Max token duration is 1 Hour",
```
steeling: @BenjaminPelletier can you comment on this piece with respect to the standards?

BenjaminPelletier: One hour in the future is from the YAML text for AuthFromAuthorizationAuthority describing required fields in the JWT: "exp, with a time no further than 1 hour in the future."  So, this looks like a good test to me.

charlie-pisuraj:
All comments resolved.

pkg/dss/auth/auth.go
```diff
@@ -229,9 +247,32 @@ func (a *Authorizer) AuthInterceptor(ctx context.Context, req interface{}, info
...
+	}
+
+	if err := a.checkExpired(claims.ExpiresAt); err != nil {
+		return nil, dsserr.Unauthenticated(fmt.Sprintf("%v", err))
+	}
+
 	return handler(ContextWithOwner(ctx, models.Owner(claims.Subject)), req)
 }
 
+func (a *Authorizer) checkExpired(claimedExpireTime int64) error {
```
steeling: Can we add the checkExpired to the claims.Valid() ?
https://github.com/interuss/dss/blob/60d63fb8dc0eee095d688d4d9427961dfbce15f1/pkg/dss/auth/claims.go#L38

This is how it is done in jwt.Valid, so it makes sense to put it there.

Also a nit: we can use their expiredAt if you choose too https://github.com/dgrijalva/jwt-go/blob/master/claims.go#L69

steeling:
LGTM

charlie-pisuraj:
I don't understand whats failing, I ran the test locally and it all passed
the logs pasted is just the logging error we expect from it.

PR #147, opened 2019-10-23
Feature/certs

steeling:
Separate out the CA cert for easier rotation and adding partner operators


config/apply-certs.sh
```diff
@@ -8,17 +8,29 @@ DIR="$(pwd)"
...
 JWT_PUBLIC_CERTS_DIR="$DIR/jwt-public-certs"
+UPLOAD_CA_KEY=false
 
 # Delete previous secrets in case they have changed.
 kubectl create namespace "$NAMESPACE" || true
 
 kubectl delete secret cockroachdb.client.root || true
 kubectl delete secret cockroachdb.client.root --namespace "$NAMESPACE" || true
 kubectl delete secret cockroachdb.node --namespace "$NAMESPACE" || true
+kubectl delete secret cockroachdb.node --namespace "$NAMESPACE" || true
```
BenjaminPelletier: Looks like this line is a duplicate of the previous one?

config/build.sh
```diff
@@ -15,3 +15,4 @@ cd "${BASEDIR}"
 docker build . -t $DOCKER_URL/dss:$VERSION
 
 docker push $DOCKER_URL/dss:$VERSION
+echo $DOCKER_URL/dss:$VERSION
```
BenjaminPelletier: nit: Maybe add a prefix mentioning what's being done (e.g., `echo Pushing docker image $DOCKER_URL/dss:$VERSION`)

config/make-certs.py
```diff
@@ -128,6 +133,9 @@ def main():
         '--ca-key', cr.ca_key_file]
         + node_addresses)
 
+    subprocess.check_call(['rm', os.path.join(cr.node_certs_dir, 'ca.crt')])
```
BenjaminPelletier: Maybe os.remove instead of shelling out?

config/templates/cluster-init-secure.yaml
```diff
@@ -16,7 +16,14 @@ spec:
         imagePullPolicy: IfNotPresent
         volumeMounts:
         - name: client-certs
```
BenjaminPelletier: Just to verify, should this name be duplicated for both items?  (same question elsewhere as well)

jgrossmac:
Looks good to me other than a few duplicates mentioned by @BenjaminPelletier

config/templates/cluster-init-secure.yaml
steeling: ya to mount multiple volumes to the same directory you have to call out each file specifically

config/make-certs.py
steeling: done

config/build.sh
steeling: done

config/apply-certs.sh
steeling: yup, done

PR #148, opened 2019-10-23
[logging] Add support for HTTP request tracing

vosst:
This PR adds support for tracing incoming requests to the HTTP gateway. To this end, the change drops gorilla's `mux` in favor of two thin middleware layers to support a health endpoint and a logging layer that conservatively logs relevant aspects of incoming requests. Logging middleware instantiation is guarded by a CLI flag.


pkg/logging/http.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (w *tracingResponseWriter) Header() http.Header {
+	return w.next.Header()
+}
+
+func (w *tracingResponseWriter) Write(data []byte) (int, error) {
+	// While this looks slightly ugly, it follows the specification
+	// for an http.ResponseWriter according to:
+	//   https://golang.org/pkg/net/http/#ResponseWriter
+	if w.statusCode == 0 {
```
steeling: This doesn't get handled internally?

vosst: Nope, it is handled by `next`, but that does not propagate back up again.

Issue #154, opened 2019-10-28
DSS should return error code on subscriptions with too long timespan (>24hrs)


charlie-pisuraj:
**Describe the bug**
Currently the DSS automatically truncates the requested subscription time span to 24hrs when the request end time is more than 24hrs after the requested start time. Although this behavior is compliant with the standard; it is confusing as the requester gets a success and may believe they have successfully registered a subscription lasting longer than 24hrs. We should change the behavior to returning an error code

**To Reproduce**
1. Create Subscription with duration of more than 24hrs
2. Return code is 200
3. Get Subscription by ID
4. Subscription end time is 24hrs from start time instead of requested value

PR #155, opened 2019-10-29
Reject Subscription PUTS if longer than 24hrs

charlie-pisuraj:
Currently the DSS accepts Subscription PUTs even though the subscription duration is longer than 24hrs, but truncates it internally to 24hrs; which could be confusing to the caller. This checks for the duration and returns an Error with appropriate error message 

Issue: #154

### DIFF
--- a/pkg/dss/server.go
+++ b/pkg/dss/server.go
@@ -298,16 +298,6 @@ func (s *Server) createOrUpdateSubscription(
 	if extents == nil {
 		return nil, dsserr.BadRequest("missing required extents")
 	}
-	var (
-		// if TimeEnd is not supplied default values to 0
-		// later in the code it will defaulted to 1 day after TimeStart
-		endTime   = time.Unix(extents.TimeEnd.GetSeconds(), 0)
-		startTime = time.Unix(extents.TimeStart.GetSeconds(), 0)
-		diff      = endTime.Sub(startTime)
-	)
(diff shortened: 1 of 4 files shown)


pkg/dss/server.go
```diff
@@ -298,6 +298,14 @@ func (s *Server) createOrUpdateSubscription(
 	if extents == nil {
 		return nil, dsserr.BadRequest("missing required extents")
 	}
+	var (
+		endTime   = time.Unix(extents.TimeEnd.GetSeconds(), 0)
+		startTime = time.Unix(extents.TimeStart.GetSeconds(), 0)
+		diff      = endTime.Sub(startTime)
+	)
+	if diff.Hours() > 24 {
```
BenjaminPelletier: Rejecting Subscription requests longer than 24 hours is fine, as is truncating them to 24 hours.  To use the rejection approach, let's remove [the now-redundant truncation code](https://github.com/interuss/dss/blob/976a7f16a9006c6e15b17cf21cfe37267f491bbb/pkg/dss/models/subscriptions.go#L137) and let's declare the "24" magic number as a constant

charlie-pisuraj: I've exported models.MaxSubscriptionDuration and removed the truncation clause of the check, but kept the rest so that it could default to 1Day when end time is not supplied

PR #159, opened 2019-11-05
Fix issue with allowing too many subs

charlie-pisuraj:
1. SQL MAX statement returns NULL in the event there is no data, which triggers an error unable to convert nil to int even though there is simply no subscriptions

2. Test Subscription was too short lived, the cumulative time it takes to create 11 subscriptions sometimes take longer than the test subscription's life which expires and allows new subs to be created 

This now fixes the test_create_too_many_subs test flakiness

### DIFF
--- a/pkg/dss/cockroach/subscriptions.go
+++ b/pkg/dss/cockroach/subscriptions.go
@@ -129,7 +129,7 @@ func (c *Store) fetchMaxSubscriptionCountByCellAndOwner(
 	ctx context.Context, q queryable, cells s2.CellUnion, owner models.Owner) (int, error) {
 	var query = `
     SELECT
-      MAX(subscriptions_per_cell_id)
+      IFNULL(MAX(subscriptions_per_cell_id), 0)
     FROM (
       SELECT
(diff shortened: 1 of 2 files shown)


pkg/dss/cockroach/subscriptions.go
```diff
@@ -258,6 +258,8 @@ func (c *Store) InsertSubscription(ctx context.Context, s models.Subscription) (
 		count, err := c.fetchMaxSubscriptionCountByCellAndOwner(ctx, tx, s.Cells, s.Owner)
 		if err != nil {
 			c.logger.Warn("Error fetching max subscription count", zap.Error(err))
```
steeling: Looks good, we can remove the logger.warn since the error will bubble up and be logged anyways

steeling:
let's also remove the flag.parse that's causing the tests to not run so we can make sure everything is working

charlie-pisuraj:
@steeling it looks like it runs just fine in the Azure pipelines, I can try to figure this out and remove it in the next PR.

PR #163, opened 2019-11-08
Deletion of ISA and SUB were permitted w/o version

charlie-pisuraj:
Interoperability test exposed that this was allowed in the DSS
	while the specs states that Version is a required field
	New behivior is to return Bad Request Error with proper message

### DIFF
--- a/pkg/dss/models/models.go
+++ b/pkg/dss/models/models.go
@@ -6,13 +6,6 @@ import (
 	"time"
 )
 
-const (
-	// Convert updatedAt to a string, why not make it smaller
-	// WARNING: Changing this will cause RMW errors
-	// 32 is the highest value allowed by strconv
-	versionBase = 32
-)
-
 type (
 	ID      string
 	Owner   string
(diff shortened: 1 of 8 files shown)


pkg/dss/models/models.go
```diff
@@ -29,9 +30,12 @@ func (owner Owner) String() string {
 	return string(owner)
 }
 
-func VersionFromString(s string) (*Version, error) {
+func VersionFromString(s string, required bool) (*Version, error) {
```
vosst: I'd rather prefer:
```
type EmptyVersionPolicy int

const (
  EmptyVersionPolicyRequireNonEmpty EmptyVersionPolicy = 0
  EmptyVersionPolicyRelaxed EmptyVersionPolicy                 = 1
)

func VersionFromString(s string, EmptyVersionPolicy) (*Version, error)
```

charlie-pisuraj:
let me rebase this, I think it should immediately pass the test since the scope bug fix was already merged to master

charlie-pisuraj:
uh oh I chose the wrong copy during the merge conflict, I need to fix this again

PR #164, opened 2019-11-08
Fix local test failure with double flag.Parse Call

charlie-pisuraj:
This flag.Parse is not necessary as go test already calls flags.Parse once, calling it again causes failure on local tests


charlie-pisuraj:
same here will have to rebase onto new master

PR #179, opened 2019-12-05
Namespace Flags, notes, and README formatting

jgrossmac:
- Allow users to be able to more easily specify a different namespace
- README whitespace cleanup
- Added some notes around the jsonnet config file


build/README.md
```diff
@@ -140,27 +128,23 @@ which ever provider you choose.
     Kubernetes cluster containing the certificates and keys generated in the
     previous step.
 
-        ./apply-certs.sh <cluster_context>
+        ./apply-certs.sh <cluster_context> <namespace>
```
BenjaminPelletier: If namespace is an optional argument to make-certs, it should be an optional argument here too (and both should use the same default value)

build/make-certs.py
```diff
@@ -10,18 +10,18 @@
 
 class CockroachCluster(object):
 
-    def __init__(self, cluster, ca_cert_to_join=None):
+    def __init__(self, cluster, ca_cert_to_join=None, namespace=None):
```
BenjaminPelletier: It seems like we never omit the `namespace` argument and a value of None doesn't really make sense when constructing `node_addresses`; let's just make this argument required with no default value.

build/deploy/examples/minimum.jsonnet
```diff
@@ -3,19 +3,20 @@ local dss = import '../../deploy/dss.libsonnet';
 local metadataBase = import '../../deploy/metadata_base.libsonnet';
 
 local metadata = metadataBase {
-  namespace: 'dss-main',
+  namespace: 'namespace',
```
BenjaminPelletier: The default namespace argument value in make-certs.py is 'dss-main' -- the value here should match if the namespace argument is omitted from make-certs.  We should probably also add a callout comment that this should match whatever was provided to make-certs and apply-certs, if that is true.

build/apply-certs.sh
```diff
@@ -1,18 +1,22 @@
 #!/bin/bash
 
 if [ "$1" == "" ]; then
-  echo "usage: ./apply-certs.sh <cluster_context>"
+  echo "usage: ./apply-certs.sh <cluster_context> <namespace>"
   exit 1
 fi
 
+if [ "$2" == "" ]; then
+  echo "usage: ./apply-certs.sh <cluster_context> <namespace>"
```
BenjaminPelletier: If namespace is an optional argument to make-certs, it should be an optional argument here too (and both should use the same default value)

build/README.md
```diff
@@ -1,118 +1,106 @@
...
+    - If you want to be able to join other clusters, you'll need static IP's
+      for each of your CockroachDB nodes (min of 3).
+    - [OPTIONAL] 1 for a loadbalancer across your CockroachDB nodes to provide
+      a single join target for other users.
 
 1.  Use the `make-certs.py` script in this directory to create certificates for
     the new CockroachDB cluster:
 
         ./make-certs.py --cluster <cluster_context>
+            [--namespace <NAMESPACE>]
```
BenjaminPelletier: Let's document how a user might choose a reasonable value for NAMESPACE

build/deploy/examples/minimum.jsonnet
```diff
@@ -3,19 +3,20 @@ local dss = import '../../deploy/dss.libsonnet';
...
   clusterName: 'your_cluster_context',
   cockroach+: {
     hostnameSuffix: 'db.your_hostname_suffix.com',
     locality: 'your_unique_locality',
     nodeIPs: ['0.0.0.0', '1.1.1.1', '2.2.2.2'],
     balancedIP: '3.3.3.3',
-    shouldInit: true,
+    shouldInit: true, //Set to false if joining a cluster
+    JoinExisting: ['0.db.westus.example.com', '1.db.westus.example.com', '2.db.westus.example.com' ],
```
BenjaminPelletier: Let's add a comment noting what the user should replace these with

build/README.md
jgrossmac: See bulletpoint number 2 above, and let me know if you feel that is sufficient.

build/README.md
jgrossmac: Opting for making it required throughout.

build/apply-certs.sh
jgrossmac: Opting to make it required throughout per `make-certs.py` comment.

build/make-certs.py
jgrossmac: Done.

build/deploy/examples/minimum.jsonnet
jgrossmac: done.

build/deploy/examples/minimum.jsonnet
jgrossmac: Done.

build/README.md
BenjaminPelletier: Looks good above.  But brackets indicate optional parameters -- let's remove them for consistency with the new requirement in make-certs.py.

build/README.md
```diff
@@ -1,113 +1,110 @@
...
 above. You can do this on any supported
 [cloud provider](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/)
-or even on your own infrastructure.  Consult the Kubernetes documentation for
+or even on your own infrastructure. Consult the Kubernetes documentation for
 which ever provider you choose.
 
 1.  Create a new Kubernetes cluster as mentioned above. We recommend a new
     cluster for each DSS instance.
 
+1.  Create a namespace for your Kubernetes cluster in which the DSS services will reside. This can be achieved with `kubectl create ns <NAMESPACE>`. Alternatively, you may want deploy in a namespace that already exists, so keep track of the namespace for later steps.
```
BenjaminPelletier: Let's make a suggestion for the value of NAMESPACE.  Like `your-organization` or `dss-prod` or whatever an example of the appropriate class of names is.

build/README.md
BenjaminPelletier: Looks good, but let's maintain stylistic consistency and put `<NAMESPACE>` in all caps.

build/README.md
```diff
@@ -180,33 +173,26 @@ following differences:
...
 your local machine using the kubectl command:
 
-
 ### Create a user account
 
-
 Pick a username and create an account:
 
-    kubectl -n dss-main exec cockroachdb-0 -ti -- \
+    kubectl -n $NAMESPACE exec cockroachdb-0 -ti -- \
```
BenjaminPelletier: Let's maintain consistency and either change this to `<NAMESPACE>` to match above, or have the user set a $NAMESPACE variable above and carry $NAMESPACE all the way through (I'm indifferent to which approach is best, but we should stick to just one).

build/README.md
```diff
@@ -180,33 +173,26 @@ following differences:
...
         ./cockroach --certs-dir ./cockroach-certs \
         user set $USERNAME --password
 
-
 ### Access the web UI
 
+    kubectl -n $NAMESPACE port-forward cockroachdb-0 8080
```
BenjaminPelletier: Same as above; <NAMESPACE> or $NAMESPACE throughout

build/apply-certs.sh
```diff
@@ -1,18 +1,22 @@
 #!/bin/bash
 
 if [ "$1" == "" ]; then
-  echo "usage: ./apply-certs.sh <cluster_context>"
+  echo "usage: ./apply-certs.sh <cluster_context> <namespace>"
```
BenjaminPelletier: For stylistic consistency, let's make this `<NAMESPACE>`

build/apply-certs.sh
BenjaminPelletier: For stylistic consistency, let's make this `<NAMESPACE>`

build/deploy/examples/minimum.jsonnet
```diff
@@ -3,19 +3,20 @@ local dss = import '../../deploy/dss.libsonnet';
 local metadataBase = import '../../deploy/metadata_base.libsonnet';
 
 local metadata = metadataBase {
-  namespace: 'dss-main',
+  namespace: 'your_namespace', //Same namespace used in the make-certs.py and apply-certs.sh scripts
```
BenjaminPelletier: nit: Let's cue the user into where to find this value just a little more with
`//Same <NAMESPACE> used in the make-certs.py and apply-certs.sh scripts`

build/README.md
jgrossmac: IMO capitalizing it doesn't mesh stylistically with `<cluster_context>`.

build/README.md
jgrossmac: I don' think the namespace matters TBH.  They might want to deploy to a pre-existing namespace that does not follow any sort of convention, and it would work perfectly fine.

build/README.md
BenjaminPelletier: I think the most important thing is to have the same thing represented the same way all the way through, so `<NAMESPACE>` or $NAMESPACE or `<namespace>` are all fine, but it should be just one of those.  I'm not particularly bothered by the capitalization of cluster_context not matching because it's a different kind of entity (folder name rather than arbitrary name), but I'd be happy capitalizing it to match.  Definitely `<cluster_context>` should match `<your_cluster_context>` above, but that's a pre-existing issue.

build/README.md
BenjaminPelletier: That's fine, but let's tell the user that so they don't spend unnecessary mental energy trying to figure out whether/when it matters and/or what it should be.  How about adding at the end:

The namespace [affects DNS addresses](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#namespaces-and-dns), but the value chosen is not particularly important for these instructions.  [One convention](https://kubernetes.io/docs/tasks/administer-cluster/namespaces/#subdividing-your-cluster-using-kubernetes-namespaces) might be to choose a namespace of `production` to differentiate from `development`.

PR #180, opened 2019-12-11
trace requests by default

steeling:


BenjaminPelletier:
I don't think we should enable this by default because any implementation that stores logs would be in violation of DSS0040 with this flag turned on unless we redacted the 4D volumes.

PR #187, opened 2020-01-22
fix scrape path &  scheme for crdb & jsonnet

charlie-pisuraj:


cloud_build.json
```diff
@@ -0,0 +1,23 @@
+{
+    "steps": [
```
steeling: what is this file for? is it supposed to be part of this change?

charlie-pisuraj:
@steeling, I've disabled scrapping annotations for grpc backend and http backend since there is no endpoint to hit, please re-review

build/deploy/grpc-backend.libsonnet
```diff
@@ -6,7 +6,7 @@ local volumes = import 'volumes.libsonnet';
     service: base.Service(metadata, 'grpc-backend') {
       app:: 'grpc-backend',
       port:: metadata.backend.port,
-      enable_monitoring:: true,
+      enable_monitoring:: false,
```
steeling: you can just omit these since it defaults to false in the base

charlie-pisuraj: I really prefer making it explicitly clear that monitoring is off, an when the time comes when we want to turn it on we know where it goes, more importantly the user of this repo will know by just searching "monitoring"

Issue #176, opened 2019-12-04
More descriptive error messages on oauth validation


steeling:
"invalid token" is not super descriptive, let's put the reason why it is failing validation.

PR #188, opened 2020-01-31
Add more descriptive error messages for oauth validation

raulamit:
Fix for https://github.com/interuss/dss/issues/176

I see I have different indentations from yours. Is there a specific guide/file where we have project specific rules?


steeling:
gofmt takes care of that formatting... looks like you're indentation is the right one :)

PR #189, opened 2020-02-06
Feature/prometheus configs

charlie-pisuraj:
as part of deploying the DSS you will also get Prometheus and Grafana now


build/deploy/grafana.libsonnet
```diff
@@ -0,0 +1,108 @@
...
+
+{
+	all(metadata) : {
+		configMap: base.ConfigMap(metadata, 'grafana-datasources') {
+			metadata: {
+				name: 'grafana-datasources',
+				namespace: metadata.namespace,
+			},
+			data: {
+				'prometheus.yaml': '{\n    "apiVersion": 1,\n    "datasources": [\n        {\n           "access":"proxy",\n            "editable": true,\n            "name": "prometheus",\n            "orgId": 1,\n            "type": "prometheus",\n            "url": "http://prometheus-service.' + metadata.namespace + '.svc:8080",\n            "version": 1\n        }\n    ]\n}',
```
steeling: note: you can put this in a yaml file and use the std lib's parseyaml to import it
https://qbec.io/reference/jsonnet-native-funcs/#parseyaml

charlie-pisuraj: Tanka feature was removed
https://github.com/grafana/tanka/blob/master/pkg/jsonnet/importer.go#L43

reverted to inline yaml

PR #192, opened 2020-02-18
add more metadata to prometheus and grafana

charlie-pisuraj:


build/deploy/prometheus.libsonnet
```diff
@@ -126,7 +126,8 @@ local base = import 'base.libsonnet';
       },
     },
     service: base.Service(metadata, 'prometheus-service') {
-      metadata: {
+      app:: 'prometheus-service',
```
steeling: this should be prometheus-server

build/deploy/prometheus.libsonnet
```diff
@@ -126,7 +126,8 @@ local base = import 'base.libsonnet';
...
+      metadata+: {
```
steeling: you're not adding anything new to metadata here.. let's delete the whole thing

build/deploy/prometheus.libsonnet
```diff
@@ -126,7 +126,8 @@ local base = import 'base.libsonnet';
...
         name: 'prometheus-service',
         namespace: metadata.namespace,
         annotations: {
```
steeling: below, under spec, if you do spec+: you'll get the selector for free. 

How did you pick the nodeport value? maybe we can rework base, to check if it's a nodeport and instead of taking a name for the port it takes a nodeport

build/deploy/grafana.libsonnet
```diff
@@ -12,7 +12,7 @@ local base = import 'base.libsonnet';
 			},
 		},
 		deployment: base.Deployment(metadata, 'grafana') {
-			metadata: {
```
steeling: same everywhere, just delete the metadata tuple

build/deploy/prometheus.libsonnet
charlie-pisuraj: wait something was wrong, I definitely added it wrong here

charlie-pisuraj: I have scrapping annotation below

build/deploy/prometheus.libsonnet
charlie-pisuraj: node port value was based on tutorials I found.

build/deploy/grafana.libsonnet
```diff
@@ -82,9 +74,8 @@ local base = import 'base.libsonnet';
 			},
 		},
 		service: base.Service(metadata, 'grafana') {
-			metadata: {
-				name: 'grafana',
-				namespace: metadata.namespace,
+			app:: 'grafana',
+			metadata+: {
```
steeling: you still don't need this if you set the port

build/deploy/grafana.libsonnet
```diff
@@ -82,9 +74,8 @@ local base = import 'base.libsonnet';
...
 				annotations: {
 					'prometheus.io/scrape': 'true',
 					'prometheus.io/port': '3000',
```
steeling: again, below if you set spec+: then you can remove selector, and the ports array is close to being set, if you want to override the base.Service object to account for this case

build/deploy/prometheus.libsonnet
```diff
@@ -61,7 +61,7 @@ local base = import 'base.libsonnet';
       ],
     },
     configMap: base.ConfigMap(metadata, 'prometheus-server-conf') {
-      metadata: {
+      metadata+: {
```
steeling: you don't need this

build/deploy/prometheus.libsonnet
```diff
@@ -70,7 +70,7 @@ local base = import 'base.libsonnet';
       },
     },
     deployment: base.Deployment(metadata, 'prometheus-deployment') {
-      metadata: {
+      metadata+: {
```
steeling: same

build/deploy/prometheus.libsonnet
steeling: you don't need that, take a look at base.Service

PR #194, opened 2020-02-21
Convert Prometheus to statefulset + other fixes

charlie-pisuraj:
Prometheus is now a statefulset with persistent volumes which
	means we will not lose our metrics on prometheus restarts


build/deploy/prometheus.libsonnet
```diff
@@ -65,23 +65,55 @@ local base = import 'base.libsonnet';
...
       },
     },
-    deployment: base.Deployment(metadata, 'prometheus-deployment') {
-      spec: {
+    statefulset: base.StatefulSet(metadata, 'prometheus-server') {
+      metadata+: {
```
steeling: let's get rid of this on the first pass

PR #202, opened 2020-03-11
Stalesync

anichawla:
Adding the configuration for Probot to automatically mark and subsequently remove the stale issues and PRs from the repository.


anichawla:
Before applying these changes, we need to install the "stale" application in the github repo. Presently, I don't have access to install that application.

PR #204, opened 2020-03-14
Fix LB's not forwarding traffic to downed nodes that can't reconnect …

steeling:
…to the cluster, because other LB's won't forward trafffic


steeling:
azp run

PR #210, opened 2020-03-23
Feature/federated prometheus

charlie-pisuraj:
Enables federation of prometheus cluster if the user enables exposing prometheus externally


build/deploy/metadata_base.libsonnet
```diff
@@ -43,4 +43,9 @@
       dest: error 'must specify destination email',
     },
   },
+  prometheus: {
+    external: false,
```
steeling: can we make this variable more descriptive?

build/deploy/metadata_base.libsonnet
```diff
@@ -43,4 +43,9 @@
...
+    IP: '',  // Leaving blank means your cloud provider will assign an ephemeral IP
```
steeling: can you add a comment on what populating it does?

build/deploy/metadata_base.libsonnet
```diff
@@ -43,4 +43,9 @@
...
+    whitelist_ip_ranges: [],  // Empty list means firewall rules are open
```
steeling: add a comment meaning anyone can access via public internet... in fact let's have an assertion somewhere that checks to see if we are making it external then this is not empty

build/deploy/metadata_base.libsonnet
steeling: maybe, "exposed"

build/deploy/metadata_base.libsonnet
charlie-pisuraj: So we don't use this field at all unless, you want to expose prometheus externally, there is another service attached to the pod for intra cluster communication

PR #223, opened 2020-04-18
Add proto for the new Strategic deconfliction spec

raulamit:
The proto has been generated from open api 2 spec: https://raw.githubusercontent.com/raulamit/DiscoveryAndSynchronization/feature/deconfliction/api/swagger.yml

### DIFF
--- a/pkg/dssproto/dss.proto
+++ b/pkg/dssproto/dss.proto
@@ -67,6 +67,14 @@ enum VerticalAccuracy {
   V_A1M = 7;
 }
 
+enum OperationState {
+  ACCEPTED = 0;
+  ACTIVATED = 1;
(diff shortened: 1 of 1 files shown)


BenjaminPelletier:
Thanks for doing this import!  While we used my repo for sketches of what the full deconfliction API might look like while we were discussing remote ID, that version is now far outdated.  We'll need to maintain full compatibility to the ASTM standard remote ID API (uastech/standards) while adding in support for the draft ASTM UTM API (current version at astm-utm/Protocol).  I think the best way to do this will be to have both repos as subrepos, generate two sets of protos/Go files, and import both of them into the main project.  Thomas and I started working on that last week and hope to have an initial set of changes in the next couple days.

PR #224, opened 2020-04-20
Split out protos

vosst:
This change prepares splitting out `gRPC` services and types per API in anticipation of the new UTM APIs landing in a follow-up PR.


pkg/dssproto/rid.pb.go
```diff
@@ -0,0 +1,4852 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.21.0
+// 	protoc        v3.6.1
+// source: pkg/dssproto/rid.proto
+
+package dssproto
```
BenjaminPelletier, line 7: Should we change this to dssproto_rid or something like that to preemptively make it easier to deal with object naming conflicts between different APIs? (perhaps something for a later PR)

vosst: Agreed. I would also consider naming clashes a code smell and investigate case by case.

BenjaminPelletier:
The google-cloud-build check was unsuccessful because lint doesn't like the package location github.com/golang/protobuf/proto preferring google.golang.org/protobuf/proto but when that change is made, everything breaks.  Ignoring the lint failure for this PR merge and adding #225.

PR #229, opened 2020-04-21
make required audiences a list, so we can support migrations.

steeling:
Also remove the keyguard since the code is already threadsafe


pkg/dss/auth/auth.go
```diff
@@ -140,19 +139,18 @@ func (r *JWKSResolver) ResolveKey(ctx context.Context) (interface{}, error) {
...
+	requiredAudiences map[string]bool
 }
 
 // Configuration bundles up creation-time parameters for an Authorizer instance.
 type Configuration struct {
 	KeyResolver       KeyResolver         // Used to initialize and periodically refresh keys.
 	KeyRefreshTimeout time.Duration       // Keys are refreshed on this cadence.
 	RequiredScopes    map[string][]string // RequiredScopes are enforced if not nil.
-	RequiredAudience  string              // RequiredAudience is enforced if not empty.
+	RequiredAudiences []string            // RequiredAudiences is enforced if not empty.
```
BenjaminPelletier, line 153: Enforced always, right?  It seems like we get an error if `acceptable_jwt_audiences` isn't specified

pkg/dss/auth/auth.go
```diff
@@ -212,20 +212,18 @@ func (a *Authorizer) AuthInterceptor(ctx context.Context, req interface{}, info
 
 	claims := claims{}
 
-	a.keyGuard.RLock()
```
BenjaminPelletier, line 215: Does the keyGuard relate to the audiences change?  Why did we need it before but not need it now?

cmds/grpc-backend/main.go
```diff
@@ -129,7 +130,7 @@ func RunGRPCServer(ctx context.Context, address string) error {
 			KeyResolver:       keyResolver,
 			KeyRefreshTimeout: *keyRefreshTimeout,
 			RequiredScopes:    dssServer.AuthScopes(),
-			RequiredAudience:  *jwtAudience,
+			RequiredAudiences: strings.Split(*jwtAudiences, ","),
```
BenjaminPelletier, line 133: nit: Seems like this is better named "AcceptableAudiences" (mirroring the name of the flag) since no one audience is required unless there is exactly one audience

pkg/dssproto/dss.pb.go
```diff
@@ -1,30 +1,36 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.21.0
```
BenjaminPelletier, line 3: This is a huge change to dss.pb.go.  This isn't small enough to be included with an audiences list PR; let's break it out separately.

Makefile
```diff
@@ -50,7 +50,7 @@ pkg/dssproto/dss.pb.gw.go: pkg/dssproto/dss.proto
 
 pkg/dssproto/dss.proto: install-proto-generation
 	go run github.com/NYTimes/openapi2proto/cmd/openapi2proto \
-		-spec interfaces/uastech/standards/remoteid/augmented.yaml -annotate \
+		-spec api/v1/remoteid.yaml -annotate \
```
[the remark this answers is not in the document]
vosst: +1, I would rather prefer the submodule approach.

pkg/dss/auth/auth.go
```diff
@@ -140,19 +139,18 @@ func (r *JWKSResolver) ResolveKey(ctx context.Context) (interface{}, error) {
...
 type Authorizer struct {
-	logger           *zap.Logger
-	keyGuard         sync.RWMutex
-	key              interface{}
-	requiredScopes   map[string][]string
-	requiredAudience string
+	logger            *zap.Logger
+	key               interface{}
+	requiredScopes    map[string][]string
+	requiredAudiences map[string]bool
```
vosst, line 145: nit: `map[string]struct{}` is typically used for set-like semantics iirc.

pkg/dss/auth/auth_test.go
```diff
@@ -109,18 +109,19 @@ func TestRSAAuthInterceptor(t *testing.T) {
...
 		},
 		KeyRefreshTimeout: 1 * time.Millisecond,
+		RequiredAudiences: []string{""},
 	})
 
 	require.NoError(t, err)
 
-	for _, test := range authTests {
+	for i, test := range authTests {
 		_, err := a.AuthInterceptor(test.ctx, nil, &grpc.UnaryServerInfo{},
```
vosst, line 118: Why not do `t.Run(strconv.Itoa(i), t.Run(...)`. With that, you can avoid manually printing the index in the error messages.

pkg/dss/auth/auth.go
steeling: We didn't need it before since we were only ever setting it in a single thread. @vosst to verify

pkg/dss/auth/auth.go
steeling: ehh I agree, but tbh I'm not a huge fan of those semantics unless we really need to squeeze out some performance gains. In reality people won't be specifying any huge number of maps, and I think this is both easier to read and easier to use ie: 

if aud["key"] {...}

 vs

 if _, ok := aud["key"]; ok {...}

BenjaminPelletier: FWIW, [the Go blog](https://blog.golang.org/maps) mentions (endorses?) the "`map[ValueType]bool` -> set" pattern

BenjaminPelletier: (my weak recommendation is to keep as-is (`map[string]bool`), but defer to both of you with better Go experience)

pkg/dssproto/dss.pb.go
steeling: done

pkg/dss/auth/auth.go
steeling: I don't think we need it, but the data race code throws up a red flag since we can access it independently of setting it. so adding it back

PR #230, opened 2020-04-22
Make staticcheck happy

steeling:


vosst:
Investigating the azure pipelines issue. @steeling could you push an empty commit to force the pipeline again?

PR #231, opened 2020-04-22
Re add the external label for federation

charlie-pisuraj:
This was accidentally removed with https://github.com/interuss/dss/commit/502b13129a011ab956cd48dae3c26a91761257b1


vosst:
@charlie-pisuraj mind pushing an empty commit to trigger CI? I disabled CI builds for forks (by accident). My bad!

charlie-pisuraj:
For some reason I can't do an empty commit, it won't sign it, @steeling mentioned that we can close the and reopen the PR and it should trigger the CI or commit a comment in it

PR #234, opened 2020-04-22
Add OAuth validation endpoint

steeling:
Add a validation endpoint, and an auxilliary server for additional endpoints from the standard.


pkg/README
```diff
@@ -1,20 +0,0 @@
-# New Go Implementation of the Remote ID portion of the DSS.
```
BenjaminPelletier, line 1: Let's update this README rather than delete -- we don't want to regress on amount of documentation.

api/v1/auxilliary.yaml
```diff
@@ -0,0 +1,59 @@
+---
+openapi: 3.0.2
+info:
+  title: DSSAuxService
+  version: 0.8.0
+  description: Discovery and Synchronization Service Auxilliary API's
```
BenjaminPelletier, line 6: 'Auxiliary' is spelled with one l

api/v1/auxilliary.yaml
```diff
@@ -0,0 +1,59 @@
...
+      type: object
+security:
+- AuthFromAuthorizationAuthority:
+  - dss.read.position_reporting_entities
+  - dss.write.position_reporting_entities
+tags:
+- name: dss
+  description: Auxilliary endpoints exposed by the DSS server.
+x-global-options:
+  go_package: pkg/api/v1//auxpb
```
BenjaminPelletier, line 59: nit: Is this necessary given that this option isn't specified in the RID API and we specify the package explicitly in the Makefile?  It seems like the API spec should be independent of implementation

api/v1/auxilliary.yaml
```diff
@@ -0,0 +1,59 @@
+---
+openapi: 3.0.2
+info:
+  title: DSSAuxService
+  version: 0.8.0
```
BenjaminPelletier, line 5: This API isn't tied to remote ID API (right?), so it should probably be versioned independently; 0.1.0 or 1.0.0 maybe?

pkg/dssproto/README.md
```diff
@@ -1,16 +0,0 @@
-All of these files are generated using
```
BenjaminPelletier, line 1: Let's move this README to /pkg/api and update it to explain how the files in that folder's subfolders are generated -- we don't want to regress on amount of documentation.

api/v1/auxilliary.yaml
BenjaminPelletier, line 59: There's a double slash here -- intentional?

api/v1/auxilliary.yaml
```diff
@@ -0,0 +1,59 @@
+---
```
BenjaminPelletier, line 1: Let's put auxiliary.yaml in /interfaces because that's where the other interface is (and where other interfaces will go).

steeling: done

api/v1/auxilliary.yaml
steeling: done

api/v1/auxilliary.yaml
steeling: done

api/v1/auxilliary.yaml
steeling: it's not necessary but there's a warning when generating it. SInce this part is specific I figured I'd include it.

Not super necessary for the upstream standards api

steeling: fixed now, thanks

pkg/README
steeling: This readme is super old. All of the content here has been captured in the newer readme in the root of the directory

steeling: root of the repo*

pkg/dssproto/README.md
steeling: done

PR #235, opened 2020-04-23
Make scripts friendlier

BenjaminPelletier:
This PR generally makes scripts a little more friendly by printing more useful/actionable error messages, fixing cross-platform issues so that they work on Mac as well as Linux, and generating an additional debugging artifact.

### DIFF
--- a/test/docker_e2e.sh
+++ b/test/docker_e2e.sh
@@ -8,7 +8,7 @@ touch $RESULTFILE
 cat /dev/null > $RESULTFILE
 
 OS=$(uname)
-if [[ $OS == "Darwin" ]]; then 
+if [[ $OS == "Darwin" ]]; then
 	# OSX uses BSD readlink
 	BASEDIR="$(dirname $0)/.."
 else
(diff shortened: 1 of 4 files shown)


test/docker_e2e.sh
```diff
@@ -62,8 +62,8 @@ docker run -d --rm --name dss-crdb-for-debugging \
 
 sleep 5
 echo " -------------- BOOTSTRAP ----------------- "
-echo "Building local container for testing"
-docker build -q --rm . -t local-interuss-dss-image
```
charlie-pisuraj, line 66: This does print a lot of stuff, I find it too verbose

BenjaminPelletier: Yeah, I'm piping the output into a log file rather than silencing it entirely so that build issues can be quickly debugged; I agree we don't want to dump it to stdout here.

BenjaminPelletier: I'm assuming that piping the `docker build` output to `grpc-backend-build.log` is sufficient to address the verbosity concern, but will open another PR to adjust if not.

PR #236, opened 2020-04-27
Add experimental UTM API support

vosst:


pkg/dss/server.go
```diff
@@ -44,21 +43,27 @@ func (s *Server) AuthScopes() map[string][]string {
...
-	}
-}
 
-// Validate will exercise validating the Oauth token
-func (a *AuxServer) ValidateOauth(ctx context.Context, req *auxpb.ValidateOauthRequest) (*auxpb.ValidateOauthResponse, error) {
-	owner, ok := auth.OwnerFromContext(ctx)
-	if !ok {
-		return nil, dsserr.PermissionDenied("missing owner from context")
+		// TODO: replace with correct scopes
+		"DeleteConstraintReference": {ReadISAScope}, //{ConstraintManagementScope},
```
steeling, line 48: should we namespace the API's, so that we can match here? ie: v1/dss is currently remote id, but we can choose v1/utm for the utm stuff?

BenjaminPelletier: I think that would be an API change, right?  We would want to raise that issue at [the API repo](https://github.com/astm-utm/Protocol) and then flow it back into here.  That API may be "locked" enough for the current project that we wouldn't get to pull in those changes until after mid-May, but we should definitely make all these good suggestions so there is an opportunity to include them in the upcoming standard itself.

PR #242, opened 2020-05-01
Fixing templated variables, and queries

charlie-pisuraj:


steeling:
is it possible to have a toggle on what we can group by on the grafana dashboard itself? So that I have an option to view by 

Additionally can we also export an environment variable, such that I can view Prod, autopush, or external staging. We could have this latter variable be a required variable, and always only show a single environment at a time. We can also do this in a later CL

charlie-pisuraj:
> is it possible to have a toggle on what we can group by on the grafana dashboard itself? So that I have an option to view by
> 
> Additionally can we also export an environment variable, such that I can view Prod, autopush, or external staging. We could have this latter variable be a required variable, and always only show a single environment at a time. We can also do this in a later CL

There is this "grouping" thing for variable its currently experimental I'll give it a try but it would be easier to simply have the originating prometheus export the environment then we can group later.

PR #257, opened 2020-05-06
Add support for per-server auth

vosst:


pkg/dss/auth/scopes.go
```diff
@@ -0,0 +1,34 @@
...
+
+// String returns the string representation of s.
+func (s Scope) String() string {
+	return string(s)
+}
+
+// MergeOperationsAndScopes merges a and be together.
+func MergeOperationsAndScopes(requiredScopes ...map[Operation][]Scope) map[Operation][]Scope {
+	if len(requiredScopes) == 0 {
+		return nil
```
BenjaminPelletier, line 22: nit: Why not just return an empty map rather than nil?

pkg/dss/scd/server.go
```diff
@@ -22,6 +22,12 @@ type Server struct {
 	Timeout time.Duration
 }
 
+// AuthScopes returns a map of endpoint to required Oauth scope.
+func (a *Server) AuthScopes() map[auth.Operation][]auth.Scope {
+	// TODO: Fill in details here.
```
BenjaminPelletier, line 27: Since we're deleting the existing scope declarations from rid/server.go, can we add them here in this PR so they're not lost?  At a minimum, I'd move the existing declarations over here and comment them all out so this TODO can be more easily fulfilled.

pkg/dss/auth/scopes.go
```diff
@@ -0,0 +1,34 @@
+package auth
+
+// Operation models the name of an operation.
+type Operation string
```
BenjaminPelletier, line 4: Let's document the formatting required here -- maybe something like "grpc.UnaryServerInfo.FullMethod of operation, generally formatted like `/{package-qualified service name}/{handler name}`"

pkg/dss/auth/auth_test.go
```diff
@@ -130,40 +130,40 @@ func TestRSAAuthInterceptor(t *testing.T) {
...
 	}}
 
 	var tests = []struct {
 		info   *grpc.UnaryServerInfo
-		scopes map[string]struct{}
+		scopes map[Scope]struct{}
 		want   error
 	}{
 		{
 			&grpc.UnaryServerInfo{FullMethod: "/dss/syncservice/PutFoo"},
```
BenjaminPelletier, line 143: Let's take a note to make these FullMethod names more similar to real FullMethod values (like, `/dss.SyncService/PutFoo` I think)

pkg/dss/rid/server.go
```diff
@@ -29,35 +29,19 @@ type Server struct {
...
+		"/dsspb.DiscoveryAndSynchronizationService/DeleteIdentificationServiceArea":  {writeISAScope},
+		"/dsspb.DiscoveryAndSynchronizationService/GetIdentificationServiceArea":     {readISAScope},
+		"/dsspb.DiscoveryAndSynchronizationService/SearchIdentificationServiceAreas": {readISAScope},
+		"/dsspb.DiscoveryAndSynchronizationService/UpdateIdentificationServiceArea":  {writeISAScope},
+		"/dsspb.DiscoveryAndSynchronizationService/CreateSubscription":               {writeISAScope},
+		"/dsspb.DiscoveryAndSynchronizationService/DeleteSubscription":               {writeISAScope},
+		"/dsspb.DiscoveryAndSynchronizationService/GetSubscription":                  {readISAScope},
+		"/dsspb.DiscoveryAndSynchronizationService/SearchSubscriptions":              {readISAScope},
+		"/dsspb.DiscoveryAndSynchronizationService/UpdateSubscription":               {writeISAScope},
+		"/dsspb.DiscoveryAndSynchronizationService/ValidateOauth":                    {writeISAScope},
```
BenjaminPelletier, line 44: Is this the right service name?  Seems like this should be moved to aux/server.go's AuxScopes()

pkg/dss/auth/scopes.go
vosst: Semantics are the same mostly, except for inserting into a `nil` map. I don't have a strong opinion here, will adjust as you proposed.

pkg/dss/scd/server.go
vosst: Done.

pkg/dss/rid/server.go
vosst: Nice catch. Fixed.

PR #260, opened 2020-05-07
Introduce scd cockroach store

vosst:
This PR:
  * introduces an implementation of `scd.Store` talking to a `cockroach` DB, drawing heavy inspiration from the existing `rid.Store` CRDB implementation (read as: copy'n'pasted mostly and adjusted)
  * adjusts models as required by the differing versioning schemes in `rid` and `scd`
  * adjusts and occasionally simplifies the `scd.Server` implementation
  * provides plenty of future opportunity to consolidate types and methods across `rid` and `scd`

### DIFF
--- a/pkg/dss/scd/cockroach/doc.go
+++ b/pkg/dss/scd/cockroach/doc.go
@@ -1,3 +1,3 @@
-// Package cockroach provides an implementation of a dss.Store on top of a
+// Package cockroach provides an implementation of a scd.Store on top of a
 // cockroach DB instance.
 package cockroach
--- a/pkg/dss/scd/cockroach/doc.go
+++ b/pkg/dss/scd/cockroach/doc.go
@@ -1,0 +1,3 @@
+// Package cockroach provides an implementation of a dss.Store on top of a
+// cockroach DB instance.
+package cockroach
--- a/pkg/dss/scd/cockroach/store.go
+++ b/pkg/dss/scd/cockroach/store.go
@@ -1,0 +1,124 @@
+package cockroach
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+
+	"github.com/dpjacques/clockwork"
+	"go.uber.org/zap"
+)
+
+var (
+	// DefaultClock is what is used as the Store's clock, returned from Dial.
+	DefaultClock = clockwork.NewRealClock()
+)
+
+// Store is an implementation of dss.Store using
+// Cockroach DB as its backend store.
+type Store struct {
+	*sql.DB
+	logger *zap.Logger
+	clock  clockwork.Clock
+}
+
+// Dial returns a Store instance connected to a cockroach instance available at
+// "uri".
+// https://www.cockroachlabs.com/docs/stable/connection-parameters.html
+func Dial(uri string, logger *zap.Logger) (*Store, error) {
+	db, err := sql.Open("postgres", uri)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Store{
+		DB:     db,
+		logger: logger,
+		clock:  DefaultClock,
+	}, nil
+}
+
+// BuildURI returns a cockroachdb connection string from a params map.
+func BuildURI(params map[string]string) (string, error) {
+	an := params["application_name"]
+	if an == "" {
+		an = "dss"
+	}
+	h := params["host"]
+	if h == "" {
+		return "", errors.New("missing crdb hostname")
+	}
+	p := params["port"]
+	if p == "" {
+		return "", errors.New("missing crdb port")
+	}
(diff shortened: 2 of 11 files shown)


pkg/dss/scd/cockroach/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package cockroach provides an implementation of a dss.Store on top of a
```
BenjaminPelletier, line 1: scd.Store?

pkg/dss/scd/server.go
```diff
@@ -169,41 +170,23 @@ func (a *Server) PutSubscription(ctx context.Context, req *scdpb.PutSubscription
...
 		BaseURL:              params.UssBaseUrl,
 		NotifyForOperations:  params.NotifyForOperations,
 		NotifyForConstraints: params.NotifyForConstraints,
-		ImplicitSubscription: implicitSubscription,
 	}
 	//TODO: Set StartTime, EndTime, AltitudeHi, AltitudeLo, DependentOperations
 
 	// Store Subscription model
-	sub, err := a.Store.InsertSubscription(ctx, sub, owner) //TODO: This should be UpsertSubscription
+	sub, err := a.Store.InsertSubscription(ctx, sub) //TODO: This should be UpsertSubscription
```
BenjaminPelletier, line 189: Per comment, this should really be UpsertSubscription because PUT either creates or updates -- it looks like some of the logic for updating was removed above.

pkg/dss/scd/cockroach/store.go
```diff
@@ -0,0 +1,124 @@
...
+type Store struct {
+	*sql.DB
+	logger *zap.Logger
+	clock  clockwork.Clock
+}
+
+// Dial returns a Store instance connected to a cockroach instance available at
+// "uri".
+// https://www.cockroachlabs.com/docs/stable/connection-parameters.html
+func Dial(uri string, logger *zap.Logger) (*Store, error) {
```
steeling, line 29: It might be nice to allow the caller to use one connection for both scd and rid. We could do something like have a Store.WithDB() or something

pkg/dss/scd/cockroach/store.go
```diff
@@ -0,0 +1,124 @@
...
+// *and* the database with this manual approach here.
+func (s *Store) Bootstrap(ctx context.Context) error {
+	const query = `
+	CREATE TABLE IF NOT EXISTS scd_subscriptions (
+		id UUID PRIMARY KEY,
+		owner STRING NOT NULL,
+		version INT4 NOT NULL DEFAULT 0,
+		url STRING NOT NULL,
+		notification_index INT4 DEFAULT 0,
+		notify_for_operations BOOL DEFAULT false,
```
steeling, line 101: super nit: "include_operations" makes me think less than "notify_for_..."

pkg/dss/scd/cockroach/subscriptions.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
+
+	// Next: update the notification_index of each one and return the rest of the
+	// data.
+	var updateQuery = fmt.Sprintf(`
+			UPDATE
+				scd_subscriptions
```
steeling, line 130: FYI: Not for this review, but I found a library called squirrel which is nice https://github.com/Masterminds/squirrel

pkg/dss/scd/cockroach/doc.go
BenjaminPelletier: Fixed.

pkg/dss/scd/server.go
BenjaminPelletier: Will defer this to another PR.

pkg/dss/scd/cockroach/store.go
BenjaminPelletier: Added #261

pkg/dss/scd/cockroach/store.go
BenjaminPelletier: This is based on [the API](https://github.com/astm-utm/Protocol/blob/d7e55fff0b4c55203c20e9408643f20d1cd6435a/utm.yaml#L503) -- a Subscription with this flag set will trigger notifications for changes to Operations in the airspace.

pkg/dss/scd/cockroach/subscriptions.go
BenjaminPelletier: Added #262

pkg/dss/scd/cockroach/store.go
BenjaminPelletier: Given nit status, I'll push this until later if we want to open an issue.

PR #267, opened 2020-05-12
Add conversion functions for SCD

vosst:
This PR provides conversion functions for SCD models <-> DSS models.

### DIFF
--- a/pkg/dss/geo/s2.go
+++ b/pkg/dss/geo/s2.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/golang/geo/s1"
 	"github.com/golang/geo/s2"
 	"github.com/interuss/dss/pkg/dss/models"
 )
@@ -25,6 +26,7 @@ const (
 	maxLat                  = 90.0
 	minLng                  = -180.0
 	maxLng                  = 180.0
+	radiusEarthMeter        = 6371010.0
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/dss/geo/s2.go
```diff
@@ -84,10 +91,40 @@ func Volume3DToCellIDs(v3 *models.Volume3D) (s2.CellUnion, error) {
 	if v3 == nil {
 		return nil, errBadCoordSet
 	}
-	if v3.PolygonFootprint == nil {
+
+	switch t := v3.Footprint.(type) {
```
BenjaminPelletier, line 95: Rather than switching here, could we have ToCellIDs be a member function of the Geometry interface?

pkg/dss/geo/s2.go
```diff
@@ -120,24 +157,34 @@ func loopAreaKm2(loop *s2.Loop) float64 {
...
 func Covering(points []s2.Point) (s2.CellUnion, error) {
-	loop := s2.LoopFromPoints(points)
-	if loopAreaKm2(loop) <= maxAllowedAreaKm2 {
-		return RegionCoverer.Covering(loop), nil
+	cu, err := CoveringForLoop(s2.LoopFromPoints(points))
+	switch err.(type) {
+	case nil:
+		return cu, nil
+	case *ErrAreaTooLarge:
+		// Empty on purpose.
```
BenjaminPelletier, line 165: Let's make this clearer; something like "Area may be too large because vertices were wound in the opposite direction; check below"

pkg/dss/geo/s2.go
vosst: I started out with that approach, but found that it breaks encapsulation quite a lot and I would rather revisit the idea in a follow-up PR if that's okay with you.

pkg/dss/geo/s2.go
vosst: cross-check: You mean that we should log the fact that we are retrying with reversed order?

BenjaminPelletier: Nah, just the comment (for developer readability, why is an error handler empty?)

vosst: Done

steeling:
Can we move the conversion files from the proto dirs to the model dirs? this breaks integrations with bazel based systems.

steeling:
@vosst @BenjaminPelletier 

Can we move the conversion files from the proto dirs to the model dirs? this breaks integrations with bazel based systems.

PR #270, opened 2020-05-12
Introduce operation store to scd

vosst:


pkg/dss/scd/store.go
```diff
@@ -2,14 +2,34 @@ package scd
...
+
+	// DeleteOperation deletes the operation identified by "id" and owned by "owner".
+	// Returns the deleted Operation and all Subscriptions affected by the delete.
+	DeleteOperation(ctx context.Context, id scdmodels.ID, owner dssmodels.Owner) (*scdmodels.Operation, []*scdmodels.Subscription, error)
+
+	// UpsertOperation inserts or updates an operation using key as a fencing
+	// token. If operation does not reference an existing subscription, an
+	// implicit subscription with parameters notifySubscriptionForConstraints
+	// and subscriptionBaseURL is created.
+	UpsertOperation(ctx context.Context, operation *scdmodels.Operation, key []scdmodels.OVN, notifySubscriptionForConstraints bool, subscriptionBaseURL string) (*scdmodels.Operation, []*scdmodels.Subscription, error)
```
BenjaminPelletier, line 25: I don't think we want to put this much application logic in the storage layer.  I'd think the storage layer would just accept an Operation to upsert and wouldn't care about the OVNs (application layer would check the OVNs before calling the storage upsert).  Likewise, I would just have the application layer query for Subscriptions in the Operation's Volume4D (an existing storage method) rather than embedding that application logic in the storage layer.  I'd also have the application layer create the implicit Subscription.

pkg/dss/scd/store.go
```diff
@@ -2,14 +2,34 @@ package scd
...
+
+	// SearchOperations returns all operations ownded by "owner" in "cells".
+	SearchOperations(ctx context.Context, cells s2.CellUnion, altitudeLower *float64, altitudeUpper *float64, earliest *time.Time, latest *time.Time) ([]*scdmodels.Operation, error)
```
BenjaminPelletier, line 28: Could we add a dss.models.Geometry interface implementation that has a CellUnion and then have SearchOperations just accept a dss.models.Volume4D?

pkg/dss/scd/store.go
```diff
@@ -2,14 +2,34 @@ package scd
...
-// Store abstracts interactions with a backing data store.
-type Store interface {
+// OperationStore abstracts operation-specific interactions with the backing data store.
+type OperationStore interface {
+	// GetOperation returns the operation identified by "id".
+	GetOperation(ctx context.Context, id scdmodels.ID) (*scdmodels.Operation, error)
+
+	// DeleteOperation deletes the operation identified by "id" and owned by "owner".
+	// Returns the deleted Operation and all Subscriptions affected by the delete.
+	DeleteOperation(ctx context.Context, id scdmodels.ID, owner dssmodels.Owner) (*scdmodels.Operation, []*scdmodels.Subscription, error)
```
BenjaminPelletier, line 19: I know this is how the Subscription storage API methods are constructed, but I think it's probably a better idea to reduce the application logic in the storage layer, so the application layer should check the owner rather than the storage layer.  Concretely, I suggest removing the owner argument.

This also leads into the question of how we want to handle deletion requests for non-existent Operations.  I suggest the application layer should query the Operation first, then check whether it exists and who the owner is, then delete the Operation.  With this pattern, there is certainly no need for the owner argument.

pkg/dss/scd/cockroach/operations.go
```diff
@@ -0,0 +1,31 @@
...
+
+var (
+	errNotImplemented = errors.New("not implemented")
+)
+
+func (s *Store) GetOperation(ctx context.Context, id scdmodels.ID) (*scdmodels.Operation, error) {
+	return nil, errNotImplemented
+}
+
+func (s *Store) DeleteOperation(ctx context.Context, id scdmodels.ID, owner dssmodels.Owner) (*scdmodels.Operation, []*scdmodels.Subscription, error) {
```
steeling, line 21: Delete should take a version as well

pkg/dss/scd/store.go
vosst: The problem then would be that the service logic needs to learn about transactions, and the storage layer would have to provide abstractions of those to ensure proper isolation of concurrent operations. That being said, we could probably move creation of implicit subscriptions to the service layer, but I would rather like to keep the check for fencing tokens in the storage layer.

pkg/dss/scd/store.go
vosst: Yup, I'll rework the interfaces to that extent.

pkg/dss/scd/models/operations.go
```diff
@@ -0,0 +1,37 @@
...
+// Operation models an operation.
+type Operation struct {
+	ID             ID
+	Version        Version
+	OVN            OVN
+	Owner          dssmodels.Owner
+	StartTime      *time.Time
+	EndTime        *time.Time
+	AltitudeLower  *float32
+	AltitudeUpper  *float32
```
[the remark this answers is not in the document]
vosst: Yup, see my comment further below.

pkg/dss/scd/models/operations.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"time"
+
+	"github.com/golang/geo/s2"
+	dssmodels "github.com/interuss/dss/pkg/dss/models"
+)
+
+// Aggregates constants for operations.
+const (
+	OperationStateUnknown       OperationState = ""
```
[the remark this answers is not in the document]
vosst: I expect them to be used, yes.

pkg/dss/scd/store.go
vosst: As pointed out before, we designed the storage layer interfaces to allow for richer storage layer implementations such that we avoid exposing the notion of transactions to service code.

vosst: That would introduce a roundtrip to the storage layer for hardly any gain from my POV. In particular, we can easily check if the number of affected rows in the database is 0 and return an error in that case.

pkg/dss/scd/cockroach/operations.go
vosst: I looked through the API spec, and didn't see the version in the request parameters, hence why I didn't add it to the store.

steeling: @BenjaminPelletier  we should be including the version here for deletions

PR #271, opened 2020-05-13
add unit tests for subscriptions in scd server

sanandk:
Commented out failing tests for now.


BenjaminPelletier:
To fix the GCP check, see the results of the command in the "Lint checks" section [here](https://github.com/interuss/dss/blob/master/test/README.md).

PR #274, opened 2020-05-13
Introduce operations tables

vosst:


pkg/dss/scd/cockroach/store.go
```diff
@@ -121,6 +121,7 @@ func (s *Store) Bootstrap(ctx context.Context) error {
 	CREATE TABLE IF NOT EXISTS scd_operations (
 		id UUID PRIMARY KEY,
 		owner STRING NOT NULL,
+		state STRING NOT NULL CHECK (state IN ('Accepted', 'Activated', 'NonConforming', 'Contingent', 'Ended')),
```
BenjaminPelletier, line 124: We don't want to store state -- it is use only for determining how to treat the key.

vosst: Done

PR #276, opened 2020-05-13
Add CRDB-based scd.Store implementation

vosst:


pkg/dss/scd/models/operations.go
```diff
@@ -35,3 +36,39 @@ type Operation struct {
...
+			o.StartTime = old.StartTime
+		}
+	} else {
+		// If setting the StartTime explicitly ensure it is not too far in the past.
+		if now.Sub(*o.StartTime) > maxClockSkew {
+			return dsserr.BadRequest("Operation time_start must not be in the past")
+		}
+	}
+
+	// If EndTime was omitted default to the existing ISA's EndTime.
```
BenjaminPelletier, line 58: Let's update "ISA"

pkg/dss/scd/models/operations.go
```diff
@@ -35,3 +36,39 @@ type Operation struct {
...
+	if o.EndTime == nil && old != nil {
+		o.EndTime = old.EndTime
+	}
+
+	// EndTime cannot be omitted for new ISAs.
```
BenjaminPelletier, line 63: Update ISA

pkg/dss/scd/models/operations.go
```diff
@@ -35,3 +36,39 @@ type Operation struct {
 	Cells          s2.CellUnion
 	SubscriptionID ID
 }
+
+// AdjustTimeRange adjusts the time range to the max allowed ranges on an
+// Operation.
+func (o *Operation) AdjustTimeRange(now time.Time, old *Operation) error {
+	if o.StartTime == nil {
+		// If StartTime was omitted, default to Now() for new ISAs or re-
```
BenjaminPelletier, line 44: I think we actually don't want to default to now for Operations; we should just reject if the start time isn't specified.

pkg/dss/scd/models/operations.go
```diff
@@ -35,3 +36,39 @@ type Operation struct {
...
+		// use the existing time of existing ISAs.
+		if old == nil {
+			o.StartTime = &now
+		} else {
+			o.StartTime = old.StartTime
+		}
+	} else {
+		// If setting the StartTime explicitly ensure it is not too far in the past.
```
BenjaminPelletier, line 52: The start time for Operations can be in the past

pkg/dss/scd/models/operations.go
vosst: Done.

pkg/dss/scd/models/operations.go
vosst: Done.

pkg/dss/scd/models/operations.go
vosst: Done.

pkg/dss/scd/models/operations.go
vosst: Done.

PR #280, opened 2020-05-14
Delete implicit subscriptions

vosst:


pkg/dss/scd/cockroach/operations.go
```diff
@@ -270,6 +278,10 @@ func (s *Store) DeleteOperation(ctx context.Context, id scdmodels.ID, owner dssm
 		return nil, nil, multierr.Combine(err, tx.Rollback())
 	}
 
+	if _, err := tx.ExecContext(ctx, deleteImplicitSubscriptionQuery, old.SubscriptionID, owner); err != nil {
```
BenjaminPelletier, line 281: I think we can delete the implicit Subscription before searching for Subscriptions to return because the caller shouldn't have to notify themself that they deleted their Operation in response to deleting their Operation.

pkg/dss/scd/cockroach/operations.go
```diff
@@ -233,8 +233,16 @@ func (s *Store) DeleteOperation(ctx context.Context, id scdmodels.ID, owner dssm
...
+		`
+		deleteImplicitSubscriptionQuery = `
+			DELETE FROM
+				scd_subscriptions
+			WHERE
+				id = $1
+			AND
+				owner = $2
+			AND
+				implicit = true
```
BenjaminPelletier, line 245: We also need to check that the only dependent Operation is the one we're deleting.  A client could create a new Operation which triggers an implicit Subscription creation, then create another new Operation pointing to the implicit Subscription, then delete the first Operation.  In that case, the Subscription would still have a dependent Operation so it shouldn't be deleted.

pkg/dss/scd/cockroach/operations.go
vosst: I have been pondering on that a little bit. From a client's perspective, it could very well be easier (as in "more uniformly") to just notify itself of the deleted operation. I don't have a strong opinion, but not returning a client's own subscription seems like special-casing without the requirement to do so.

pkg/dss/scd/cockroach/operations.go
vosst: Ah, nice catch.

pkg/dss/scd/cockroach/operations.go
BenjaminPelletier: I don't think I have a strong opinion, but my thought here is that at the time the response is returned to the client, the Subscription no longer exists.  I could buy that its last duty was to trigger this notification, but it seems not very useful to notify someone of something when they only want to know about things relevant to the Operation that no longer exists

vosst: ACK, let's keep it like it is right now and fix *if* we find problems. Right now, executing it here means less re-shuffling (somewhat arbitrary objective).

pkg/dss/scd/cockroach/operations.go
vosst: Fixed.

PR #291, opened 2020-05-19
Move proto conversions into common models package

steeling:
This reverts commit 0f3f789e2f7eb93b37387825e22aa5b05a303005.


BenjaminPelletier:
Can you document why these changes need to be reverted?

steeling:
> The DSS should use common geo business models that are not specific to RID. We needed to use geo operations for SCD and we don't want SCD to depend on RID protos for geo concepts. We also don't want to duplicate geo logic in RID and SCD.

Ya it's a bit of a mess right now. Let's keep the code duplication until we can refactor things to use common API's where possible.
 
> Also, it looks like there was a push directly to master without a PR; let's undo that and merge this PR through the normal merge process from where #290 left off when it's ready.

Yup, that's been reverted. I'd prefer to avoid deleting a commit (which I believe would require a force push). If you have another way happy to do that.

BenjaminPelletier:
> Let's keep the code duplication until we can refactor things to use common API's where possible.

I'd like to maintain or improve the current code quality by not introducing new code duplication in this PR.  Seems like the conversions.go files can be moved (and method names adjusted, as done here) without duplicating dss/models/geo.go into two different places, right?

steeling:
> > Let's keep the code duplication until we can refactor things to use common API's where possible.
> 
> I'd like to maintain or improve the current code quality by not introducing new code duplication in this PR. Seems like the conversions.go files can be moved (and method names adjusted, as done here) without duplicating dss/models/geo.go into two different places, right?

The issue is that the conversions are from specific protos.. so we would need a 

```
func GeoPolygonFromRIDProto(...) {...}
// and
func GeoPolygonFromSCDProto(...) {...}
```

At this point we've duplicated the logic anyways, so there's not much lost in duplicating the models.

BenjaminPelletier:
> At this point we've duplicated the logic anyways, so there's not much lost in duplicating the models.

Currently, there is only one set of logic for computing spatial characteristics like S2 coverings (in pkg/dss/models/geo.  This PR would copy/paste those 241 lines into two different places -- let's avoid that code duplication.  Spacetime volumes are a shared concept that it doesn't make sense to separate on a per-service basis, especially when we already have a single, common set of business objects for manipulating them.

We will need longish specific names (e.g., `GeoPolygonFromRIDProto` and `GeoPolygonFromSCDProto`); that is the down side of moving the conversions.go files from the context in which "RID" or "SCD" can be interpreted from the package in which the conversions reside, but I agree that is worth the change to make the codebase easier to use for some build systems.

steeling:
> > At this point we've duplicated the logic anyways, so there's not much lost in duplicating the models.
> 
> Currently, there is only one set of logic for computing spatial characteristics like S2 coverings (in pkg/dss/models/geo. This PR would copy/paste those 241 lines into two different places -- let's avoid that code duplication. Spacetime volumes are a shared concept that it doesn't make sense to separate on a per-service basis, especially when we already have a single, common set of business objects for manipulating them.
> 
> We will need longish specific names (e.g., `GeoPolygonFromRIDProto` and `GeoPolygonFromSCDProto`); that is the down side of moving the conversions.go files from the context in which "RID" or "SCD" can be interpreted from the package in which the conversions reside, but I agree that is worth the change to make the codebase easier to use for some build systems.

done

pkg/dss/models/geo.go
```diff
@@ -2,17 +2,27 @@ package models
...
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/golang/geo/s2"
+	"github.com/golang/protobuf/ptypes"
 	"github.com/interuss/dss/pkg/dss/geo"
+	"google.golang.org/protobuf/proto"
+
+	// TODO: all of the uses of ridpb should use protos that can be used
```
BenjaminPelletier, line 13: Let's remove this TODO as this project must accept whatever the standard APIs are -- we don't have the ability (as a code TODO) to change the RID and SCD object structures.  We can advocate for compatible object structures the the standard development, but the standards are consensus-based documents not controlled by this project.

steeling: Note that we could still use common protos while sticking to the standard's spec

PR #293, opened 2020-05-20
Pull everything out of the dss subdir; Split the server logic into per entity handlers

steeling:
DSS code lives both under and above the dss subdirectory; it's mostly confusing at this point so removing it here.

Splitting out the server logic in preparation for a large refactor that will split the concerns of the code by entity


steeling:
> While you are at it and moving stuff around:
> 
> * Unifying the naming of per-server handlers to `*_handler.go` for `scd` would be appreciated.

Done

> * Unifying the store layout for `rid` and introducing a subpackage `store` would be nice.

I have a larger refactor attempt going on at feature/repo-structure, which we can try to restructure this there :)

>   Otherwise LGTM.

PR #299, opened 2020-05-20
Add AirspaceConflictResponse implementation

BenjaminPelletier:
When attempting to PUT an Operation in strategic conflict detection, one possible error case is that the requesting client did not provide all required OVNs for the Operations and Constraints for which they are required to have retrieved information.  In this case, a specially-formatted error response is required (per API) that includes information about the Operations and Constraints for which the current OVN was not provided.  This PR adds that functionality by passing an AirspaceConflictResponse protobuf object in the Details field of the gRPC Status, and decoding that object into the response body in the HTTP gateway when appropriate.

This PR also improves testing documentation by stating the expected preconditions and mutations for each test step in strategic conflict detection Operations tests.


pkg/scd/errors/errors.go
```diff
@@ -0,0 +1,55 @@
...
+		if err != nil {
+			return nil, err
+		}
+		entityRef := &scdpb.EntityReference{
+			OperationReference: opRef,
+		}
+		response.EntityConflicts = append(response.EntityConflicts, entityRef)
+	}
+
+	serialized, err := proto.Marshal(response)
```
steeling, line 34: Let's use deterministic serialization: proto.MarshalOptions{Deterministic: true}.Marshal(response)

cmds/http-gateway/main.go
```diff
@@ -176,16 +179,37 @@ func myHTTPError(ctx context.Context, mux *runtime.ServeMux, marshaler runtime.M
 	}
 	w.Header().Set("Content-Type", contentType)
```
steeling, line 181: Fix func comment: "... only needed to add 1 extra Code to handle but since they didn't..."

cmds/http-gateway/main.go
```diff
@@ -176,16 +179,37 @@ func myHTTPError(ctx context.Context, mux *runtime.ServeMux, marshaler runtime.M
...
 
-	body := &errorBody{
-		Error:   s.Message(),
-		Message: s.Message(),
-		Code:    int32(s.Code()),
-		Details: s.Proto().GetDetails(),
-	}
+	var buf []byte
+	var merr error
+	if s.Code() == errors.MissingOVNs {
```
steeling, line 184: oof.. that sucks we have to do this.. might be worth filing a feature request to grpc-gateway folks. dug through their code a little bit, looks like they intentionally completely drop the response message.

cmds/http-gateway/main.go
```diff
@@ -176,16 +179,37 @@ func myHTTPError(ctx context.Context, mux *runtime.ServeMux, marshaler runtime.M
...
+			merr = errors.Internal("Missing Details from Status")
+		} else {
+			body, ok := s.Details()[0].(*scdpb.AirspaceConflictResponse)
+			if ok {
+				buf, merr = marshaler.Marshal(body)
+			} else {
+				grpclog.Infof("Cast to *scdpb.AirspaceConflictResponse failed from type %s", reflect.TypeOf(s.Details()[0]))
+				merr = errors.Internal("Unable to cast s.Details()[0] to *scdpb.AirspaceConflictResponse")
+			}
+		}
```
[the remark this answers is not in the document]
steeling: alternatively, we could just do another check on the path here: https://github.com/interuss/dss/blob/master/cmds/http-gateway/main.go#L81

and write our own func that calls out to the gprc-backend, which would allow us to keep the backend simple

pkg/scd/operations_handler.go
```diff
@@ -224,6 +225,19 @@ func (a *Server) PutOperationReference(ctx context.Context, req *scdpb.PutOperat
...
 
 	if err != nil {
+		if err == scderr.MissingOVNsInternalError() {
+			// The client is missing some OVNs; provide the pointers to the
+			// information they need
+			ops, err := a.Store.SearchOperations(ctx, uExtent, owner)
+			if err != nil {
+				return nil, err
+			}
+			intendedErr, unintentionalErr := scderr.MissingOVNsErrorResponse(ops)
```
steeling, line 235: nit: double errors are a little off-putting. Might be nice to do one error, and check if it's not a MissingOVNsError. Or could pass an error pointer and return an error, or could return a string to create a new error.

cmds/http-gateway/main.go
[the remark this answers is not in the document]
BenjaminPelletier: The last section in the first link looks like custom handling of the entire method.  I don't think we want to implement an entire custom handler for this entire method; we just want to augment returned content upon a certain error.  Even if we did write our own function to call out to the grpc-backend, the grpc-backend would still need to return different content in different circumstances.  We would have to write a new proto to handle that also -- it seems like this way is much easier.

I'll add a comment indicating when the error handler function is hit, and what is available at that time.

pkg/scd/operations_handler.go
```diff
@@ -224,6 +225,19 @@ func (a *Server) PutOperationReference(ctx context.Context, req *scdpb.PutOperat
 	}, key)
 
 	if err != nil {
+		if err == scderr.MissingOVNsInternalError() {
```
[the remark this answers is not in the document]
BenjaminPelletier: Will do.

pkg/scd/errors/errors.go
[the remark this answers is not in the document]
BenjaminPelletier: I don't think I follow; it seems like this particular line would be necessary in any case to send information back to the gateway, so generating the augmented error information will have to have the possibility of generating an error itself.

For better clarity, do you have a concrete suggestion here that will comply with the API we need to implement?

BenjaminPelletier: Will do.  Can we document the advantage/importance of this difference?

cmds/http-gateway/main.go
BenjaminPelletier: Will do.

cmds/http-gateway/main.go
BenjaminPelletier: I'm not sure what the feature request would be; when a gRPC function has an error, it returns a Status rather than the normal response.  It's basically a return type of `StatusOr<ResponseType>`.  It seems like of course the response message is dropped when there is an error; what else would make sense to do?

pkg/scd/operations_handler.go
BenjaminPelletier: Will change.

BenjaminPelletier: Done.

cmds/http-gateway/main.go
BenjaminPelletier: Done.

pkg/scd/operations_handler.go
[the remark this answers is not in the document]
BenjaminPelletier: Done.

cmds/http-gateway/main.go
steeling: ah, i thought it was sending both, nevermind then

cmds/http-gateway/main.go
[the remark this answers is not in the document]
steeling: yup, that was me going of grpc sending both status with a message

pkg/scd/operations_handler.go
```diff
@@ -223,6 +224,20 @@ func (a *Server) PutOperationReference(ctx context.Context, req *scdpb.PutOperat
...
+	if err == scderr.MissingOVNsInternalError() {
+		// The client is missing some OVNs; provide the pointers to the
+		// information they need
+		ops, err := a.Store.SearchOperations(ctx, uExtent, owner)
+		if err != nil {
+			return nil, err
+		}
+		err, success := scderr.MissingOVNsErrorResponse(ops)
+		if !success {
+			return nil, dsserr.Internal(fmt.Sprintf("failed to construct missing OVNs error message: %s", err))
```
steeling, line 236: do you mean to use this error in the Sprintf?

BenjaminPelletier: Yes; that indicates why we failed to construct the error we wanted to construct.

PR #304, opened 2020-05-21
Leverage inverted indexes on array in CRDB 20.1 to remove joins tables

steeling:
This PR *greatly* simplifies the DSS internal logic and structure by leveraging CRDB's new inverted indexes on arrays.

Instead of a joins table from the payload data to the cells, we can now store the cells directly and each entity, and query for intersection with other cells.

Migrations are intended to be handled by the operator.


steeling:
Note that we don't actually get to use the indexes yet due to lack of disjunction on Inverted indexes in CRBD https://github.com/cockroachdb/cockroach/issues/47340. This feature is slated for 20.2 which should arrive in ~6 months. 

Additionally with the current queries we use, we have the index on `end_time` and `owner` for Subscription Search, and the index on `end_time` for ISA's, so this wouldn't become an issue until we hit ~100k live ISA's or subscriptions. Until then, I anticipate an overall speed up of queries since we are dealing in single tables, and able to reduce the amount of transactions (and therefore contention) we have.

Required future improvements, if the disjunction does not make it through, can leverage the existing inverted index by "unfurling" the array like so:

`SELECT * FROM my_table WHERE array_column && ARRAY['sky', 'bird']`
becomes -> 

`SELECT * FROM my_table WHERE array_column @> ARRAY['sky'] UNION SELECT * FROM my_table WHERE array_column @> ARRAY['bird']`

PR #307, opened 2020-05-24
Introduce Transactor pattern to SCD

BenjaminPelletier:
This PR alters SCD's Store pattern to a Transactor which spawns transactions which provide the traditional Store interface.  In the future, we will want to pair down the logic in Store, but this PR "minimally" introduces the Transactor pattern without changing the Store interfaces.

This PR also deletes obsolete unit test sets that are already generally covered by integration tests and would take a very large amount of work to adapt to the Transactor pattern.


pkg/scd/store/cockroach/store.go
```diff
@@ -13,33 +15,85 @@ var (
...
+}
+
+// Implement store.Transaction interface
+func (t *Transaction) Rollback() error {
+	return t.tx.Rollback()
+}
+
+// Transactor is an implementation of scd.Transactor using
+// a CockroachDB database.
+type Transactor struct {
```
steeling, line 58: Have you taken a look at what I did in the RID refactor? This is very similar, and we should probably align the approaches.

In particular the other approach solves some of the following:

1. Protects against incorrectly nested transactions.
2. Uses the crdb-go library for retries which retries more types of retry-able errors.
3. Builds in timeouts for the retries
4. One struct instead of the 2 here (Store and Transactor).

Any reason in particular you're diverging from that setup here?

pkg/scd/store/store.go
```diff
@@ -12,146 +11,99 @@ import (
...
+// PerformOperationWithRetries creates a Transaction from the Transactor,
+// attempts to perform the provided action, and retries this process again if
+// it fails in a retryable way.
+func PerformOperationWithRetries(ctx context.Context, transactor Transactor, operation TransactionOperation, retries int) error {
+	var err error
+	for i := 0; i <= retries; i++ {
+		// Prepare a Store for `operation` to act on
+		tx, err := transactor.Transact(ctx)
+		if err != nil {
+			return err
```
steeling, line 85: All of the returns in this function are leaking transactions by not calling rollback

BenjaminPelletier: Oops, thanks; fixed.

pkg/scd/store/cockroach/store.go
BenjaminPelletier: This PR was started before 6997658070f2b98dbc100a665a9ce015c8759188 and the [documentation](https://github.com/interuss/dss/blob/master/pkg/rid/README.md#todo) seems out of sync if the TODOs being resolved already.

Re: 1, Could you elaborate on how/in what circumstances this implementation would have a problem with nested transactions?

Re: 3, Is there a structural reason why retries couldn't be added relatively easy to this architecture?

Re 4, It seems like a Repository from RID is an overloaded struct -- it represents both a repo outside a transaction and a repo inside a transaction.  When it is a repo outside a transaction, it seems dangerous to use most of its methods as the developer could easily not realize they are performing operations on the repo outside a transaction.  When a repo is inside a transaction, then the InTxnRetrier method doesn't really make sense to have, nor does Close.  The approach here (which is otherwise pretty similar, as you mentioned) fixes this issue by making the roles of the structs clear: the Transactor only has the equivalent of InTxnRetrier and Store only has the equivalent of the substantive operations.

pkg/scd/store/store.go
```diff
@@ -12,146 +11,99 @@ import (
...
-	subs := []*scdmodels.Subscription{
-		makeDummySubscription(scdmodels.ID("444eab15-8384-4e39-8589-5161689aee56")),
+// TransactionOperation is an application action involving one or more chained
+// Store actions joined by application logic.
+type TransactionOperation func(store Store) (err error)
+
+// PerformOperationWithRetries creates a Transaction from the Transactor,
+// attempts to perform the provided action, and retries this process again if
+// it fails in a retryable way.
+func PerformOperationWithRetries(ctx context.Context, transactor Transactor, operation TransactionOperation, retries int) error {
```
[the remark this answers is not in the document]
BenjaminPelletier: This PR attempts to only change the SCD Store interface structure; the internals are only changed to the extent necessary to use the new interface.  Let's defer implementation changes to after this PR.

pkg/scd/store/cockroach/store.go
```diff
@@ -13,33 +15,85 @@ var (
...
 )
 
-// Store is an implementation of dss.Store using
-// Cockroach DB as its backend store.
+// Store is an implementation of scd.Store using
+// a CockroachDB transaction.
 type Store struct {
-	*cockroach.DB
+	tx     *sql.Tx
+	ctx    context.Context
```
steeling, line 22: Please don't store contexts within structs, as it makes overriding contexts in subcalls much more difficult to reason about.
https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39
https://www.ardanlabs.com/blog/2019/09/context-package-semantics-in-go.html
http://p.agnihotry.com/post/understanding_the_context_package_in_golang/

pkg/scd/operations_handler.go
```diff
@@ -30,29 +31,42 @@ func (a *Server) DeleteOperationReference(ctx context.Context, req *scdpb.Delete
...
-	if err != nil {
-		return nil, err
-	}
-	if op == nil {
-		return nil, dsserr.Internal(fmt.Sprintf("DeleteOperation returned no Operation for ID: %s", id))
-	}
-	if subs == nil {
-		return nil, dsserr.Internal(fmt.Sprintf("DeleteOperation returned nil Subscriptions for ID: %s", id))
+	var response *scdpb.ChangeOperationReferenceResponse
+	action := func(store scdstore.Store) (err error) {
```
steeling, line 35: this doesn't need to be done in a txn

pkg/scd/operations_handler.go
```diff
@@ -67,23 +81,36 @@ func (a *Server) GetOperationReference(ctx context.Context, req *scdpb.GetOperat
 		return nil, dsserr.PermissionDenied("missing owner from context")
 	}
 
-	sub, err := a.Store.GetOperation(ctx, id)
-	if err != nil {
-		return nil, err
-	}
+	var response *scdpb.GetOperationReferenceResponse
+	action := func(store scdstore.Store) (err error) {
```
steeling, line 85: this doesn't need to be done in a txn

pkg/scd/store/cockroach/store.go
steeling: ah, took a closer look i see that nested txn's aren't possible.

Re 4: I disagree, in fact there's many queries here we wouldn't want to use transactions with, so it gives the developer the flexibility to do the right thing. Forcing transactions on the user is not what we want.

Splitting them out into 2 would be ok, but only for the reason of losing the Close and Begin functions within a transaction.

pkg/scd/operations_handler.go
BenjaminPelletier: When we simplify the Store APIs to do just simple CRUD (rather than packaging the application logic in the Store layer as is the case currently), changing an Operation Reference will involve retrieving the current version of that Operation Reference, making sure that all necessary preconditions (like owner, version) are satisfied, and then write the modified version back into the database.  The read must be performed in a transaction with the write.

pkg/scd/store/store.go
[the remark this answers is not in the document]
steeling: can we add a TODO then?

pkg/scd/operations_handler.go
steeling: actually it can be done in a single sql call, like how we do it in RID :)

Get's as well don't need to be run a txn. Giving us the flexibility to not run queries in a txn will help scale the DSS a lot as it will alleviate contention

BenjaminPelletier: Performing this action in a single SQL call embeds application logic in the Store layer that I think we generally want to avoid in favor of simple CRUD operations.  In this way, I think that RID's current approach of using two separate calls to [read](https://github.com/interuss/dss/blob/6ae81085411f569adc64dd2a79e525edd2162ce8/pkg/rid/application/isa.go#L87) and [modify](https://github.com/interuss/dss/blob/6ae81085411f569adc64dd2a79e525edd2162ce8/pkg/rid/application/isa.go#L136) an ISA is better than embedding application logic (checking and incrementing version, checking owner) in the Repo layer.

However, there appears to be a latent bug in [app.InsertISA](https://github.com/interuss/dss/blob/6ae81085411f569adc64dd2a79e525edd2162ce8/pkg/rid/application/isa.go#L86) where two racing calls could both pass their app-level version checks, but then one of the two racing modifications would hit the Repo-level version check and return a 500 rather than the desired 409 "old version" because the error returned by the query is returned directly to the gateway.

One way to fix this could be to eliminate the version check in app.InsertISA, instead relying on the Repo layer to perform the version check and return an appropriate error.  However, if we did that, we would definitely need to [document](https://github.com/interuss/dss/blob/0f305734354c8e8637ec65f75527b6688fe75bee/pkg/rid/repos/isa.go#L20) the required behavior of all Repo implementations (validating versions, how exactly to increment the version including the algorithm to generate the version, validate owner) as well as documenting exactly which intended errors might be returned (like the one that should trigger the 409).  All implementations, including mocks, would need to implement this logic.

A different way to fix the problem would be to simply place the initial GetISA in the same transaction as the later InsertISA.  In that case, there would be no need to perform the check in the query, and no need to document the required application logic to be implemented in all Repo implementations.  This option seems like a much simpler route to me.

steeling: We've already seen some issues due to contention. I think improving the user experience is more important than improving the developer experience. 

In regard to ensuring we return the correct version, we could check for a sql.ErrNoRows error if it is an update, and return a version mismatch (like we do for delete)

pkg/scd/store/cockroach/store.go
BenjaminPelletier: It seems unlikely we would want to override a context for a single query within a transaction, but will do since an explicit context parameter seems idiomatic to Go, [http/request](https://golang.org/src/net/http/request.go) notwithstanding.

vosst:
I'm late to the party, but I feel like this could land on master as a base for future iteration. @steeling any strong objections?

pkg/scd/store/store.go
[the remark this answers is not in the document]
BenjaminPelletier: Yes, will do.

pkg/scd/store/cockroach/store.go
```diff
@@ -13,33 +15,85 @@ var (
...
+// a CockroachDB transaction.
+type Transaction struct {
+	tx     *sql.Tx
+	logger *zap.Logger
+	ctx    context.Context
+	clock  clockwork.Clock
+}
+
+// Implement store.Transaction interface
+func (t *Transaction) Store() (scdstore.Store, error) {
```
steeling, line 37: can we have this return an interface, which will help for testing, and allow us to keep this modular

BenjaminPelletier: Could you elaborate?  `scdstore.Store` is an [interface](https://github.com/interuss/dss/pull/307/files#diff-2e50608ee73cee7175e96d52bd2bd2eeL51)

pkg/scd/store/store.go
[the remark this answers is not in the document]
BenjaminPelletier: Done (7560c8d)

pkg/scd/store/cockroach/store.go
BenjaminPelletier: Done (4b234b8).

pkg/scd/store/cockroach/store.go
steeling: ahh we're overloading store/cockroach/repo a lot.. there's a cockroach store which is a struct which I thought we were returning here

PR #311, opened 2020-05-28
Feature/transactor

steeling:
Add transactor pattern; fix up txn retries

### DIFF
--- a/pkg/rid/cockroach/store_test.go
+++ b/pkg/rid/cockroach/store_test.go
@@ -2,6 +2,7 @@ package cockroach
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"testing"
 	"time"
@@ -155,6 +156,10 @@ func TestGetVersion(t *testing.T) {
 	version, err = store.GetVersion(ctx)
 	require.NoError(t, err)
 	require.Equal(t, "v1", semver.Major(version))
+
+	_, err = store.db.ExecContext(ctx, `DROP TABLE cells_subscriptions;`)
+	require.NoError(t, err)
+
 }
 
 func TestTransactor(t *testing.T) {
@@ -168,14 +173,13 @@ func TestTransactor(t *testing.T) {
 	subscription1 := subscriptionsPool[1].input
 
 	txnCount := 0
-
-	store.InTxnRetrier(ctx, func(s1 repos.Repository) error {
+	err := store.InTxnRetrier(ctx, func(s1 repos.Repository) error {
 		// We should get to this retry, then return nothing.
 		if txnCount > 0 {
-			return nil
+			return errors.New("already failed")
 		}
 		txnCount++
-		store.InTxnRetrier(ctx, func(s2 repos.Repository) error {
+		err := store.InTxnRetrier(ctx, func(s2 repos.Repository) error {
 			subs, err := s1.SearchSubscriptions(ctx, subscription.Cells)
 			require.NoError(t, err)
 			require.Len(t, subs, 0)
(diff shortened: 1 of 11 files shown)


pkg/rid/cockroach/store_test.go
```diff
@@ -155,11 +150,92 @@ func TestGetVersion(t *testing.T) {
...
+	_, err = s2.InsertSubscription(ctx, subscription)
+	require.NoError(t, err)
+	// Query outside the txn
+	_, err = store.InsertSubscription(ctx, subscription1)
+	require.NoError(t, err)
+
+	subs, err = s2.SearchSubscriptions(ctx, subscription.Cells)
+
+	require.NoError(t, err)
+	require.Len(t, subs, 1)
```
steeling, line 233: @vosst do you know why this is passing? This test shouldn't pass IIUC

pkg/rid/cockroach/store_test.go
```diff
@@ -155,11 +150,92 @@ func TestGetVersion(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Len(t, subs, 0)
+
+	_, err = s2.InsertSubscription(ctx, subscription)
+	require.NoError(t, err)
+	// Query outside the txn
+	_, err = store.InsertSubscription(ctx, subscription1)
+	require.NoError(t, err)
+
+	subs, err = s2.SearchSubscriptions(ctx, subscription.Cells)
```
BenjaminPelletier, line 230: Shouldn't this fail since the database knows the transaction cannot be committed because the state of the database upon which the initial SearchSubscriptions was based is now invalid due to the Subscription insertion outside the transaction?  Or, at least tx.Commit() below should fail if this statement doesn't fail?

steeling: ya i was confused by this initially as well. Did some digging into the definition of serializable isolation, and a simple way to think of it is as: 2 transactions may not execute in parralel if there is not an order in which their results would be the same had they executed sequentially.

So basically because I'm not doing a secondary read within that transaction. Going to rewrite the tests so that they fail and we catch this.

pkg/rid/cockroach/store_test.go
steeling: nevermind.. see comment below

pkg/rid/cockroach/store_test.go
```diff
@@ -155,11 +150,115 @@ func TestGetVersion(t *testing.T) {
...
+func TestTransactor(t *testing.T) {
+	var (
+		ctx                  = context.Background()
+		store, tearDownStore = setUpStore(ctx, t)
+	)
+	require.NotNil(t, store)
+	defer tearDownStore()
+	subscription := subscriptionsPool[0].input
+	subscription1 := subscriptionsPool[1].input
+
```
charlie-pisuraj, line 174: nit: subscription1 and subscription2
to match the transactions

pkg/rid/cockroach/store_test.go
```diff
@@ -155,11 +150,115 @@ func TestGetVersion(t *testing.T) {
...
+// Test here for posterity to demonstrate transaction semantics
+func TestBasicTxn(t *testing.T) {
+	var (
+		ctx                  = context.Background()
+		store, tearDownStore = setUpStore(ctx, t)
+	)
+	require.NotNil(t, store)
+	defer tearDownStore()
+	subscription := subscriptionsPool[0].input
+	subscription1 := subscriptionsPool[1].input
```
charlie-pisuraj, line 225: same as above

PR #312, opened 2020-05-29
should always return by id, even if expired

steeling:
RMW breaks if we don't return by ID when expired.


pkg/rid/cockroach/store.go
```diff
@@ -84,7 +84,7 @@ func recoverRollbackRepanic(ctx context.Context, tx *sql.Tx) {
 // NewStore returns a Store instance connected to a cockroach instance via db.
 func NewStore(db *cockroach.DB, logger *zap.Logger) *Store {
 	return &Store{
-		ISA:          &ISAStore{db, DefaultClock, logger},
```
BenjaminPelletier, line 87: Why do we want to drop the clock specification here?  This seems disconnected from the rest of the PR

steeling: I think once we introduce GC we don't need a clock on the store at all, which should give a cleaner repo API. Hopefully dropping it now will discourage it's use until we do that.

PR #322, opened 2020-06-10
Fix grpc-backend crash triggered by query with degenerate polygon

### LINKED ISSUE #321: Issuing query request crashes grpc backend
Issuing a POST /operation_references/query request with the body below causes the grpc-backend to crash.

```json
{
  "area_of_interest": {
    "volume": {
      "outline_polygon": {
        "vertices": [
          {
            "lng": -122.45464324951172,
            "lat": 37.78943798147498
          },
          {
            "lng": -122.45464324951172,
            "lat": 37.78943798147498
          },
          {
            "lng": -122.45464324951172,
            "lat": 37.78943798147498
          }
        ]
      },
      "altitude_lower": {
        "value": 100,
        "reference": "W84",
        "units": "M"
      },
      "altitude_upper": {
        "value": 1000,
        "reference": "W84",
        "units": "M"
      }
    },
    "time_start": {
      "value": "2020-06-09T22:03:02.765Z",
      "format": "RFC3339"
    },
    "time_end": {
      "value": "2020-06-09T23:03:02.765Z",
      "format": "RFC3339"
    }
  }
}
```

BenjaminPelletier:
This PR fixes #321 where an S2 Loop would indicate area=0 yet result in a region covering of the entire earth.  This is probably a bug in S2 since even in degenerate cases, computed area and region covering should have consistent behavior.


pkg/geo/s2.go
```diff
@@ -119,11 +119,17 @@ func Covering(points []s2.Point) (s2.CellUnion, error) {
 // CoveringForLoop calculates an s2 cell covering for loop or returns an area if
 // the area covered by loop is too large.
 func CoveringForLoop(loop *s2.Loop) (s2.CellUnion, error) {
-	if loopAreaKm2(loop) > maxAllowedAreaKm2 {
```
steeling, line 122: might make more sense to clean this up in the above `Covering` function, ie:

```
func Covering(points []s2.Point) (s2.CellUnion, error) {
  loop := s2.LoopFromPoints(points)
  area := loop.Area() // change to area
  if areaTooLarge(area) {
    loop.Invert()
  }
 if areaTooLarge(area) {
   return nil, ErrTooLarge
 }
  if area <= 0 {
      pl := s2.Polyline(loop.Vertices())
		  return RegionCoverer.Covering(&pl), nil
   }
  return RegionCoverer.Covering(loop), nil
}
```

BenjaminPelletier: Done, and corrected a problem from #295 as well.

PR #323, opened 2020-06-11
Split out insert/update paths for RID

steeling:
Split out insert/update


pkg/rid/application/isa.go
```diff
@@ -46,103 +45,112 @@ func (a *app) SearchISAs(ctx context.Context, cells s2.CellUnion, earliest *time
...
+			return err
+		case old == nil:
+			return dsserr.NotFound(fmt.Sprintf("isa not found: %s", isa.ID))
+		case old.Owner != isa.Owner:
+			return dsserr.PermissionDenied(fmt.Sprintf("ISA is owned by %s", old.Owner))
+		case !old.Version.Matches(isa.Version):
+			return dsserr.VersionMismatch(fmt.Sprintf("old version for isa: %s", isa.ID))
+		}
+		// Validate and perhaps correct StartTime and EndTime.
+		if err := isa.AdjustTimeRange(a.clock.Now(), old); err != nil {
```
BenjaminPelletier, line 135: nit: We do this outside the transaction for InsertISA -- it would be nice to be consistent here, plus it's less stuff to do inside the transaction.

pkg/rid/application/subscription_test.go
```diff
@@ -77,10 +75,9 @@ func (store *subscriptionStore) DeleteSubscription(ctx context.Context, s *ridmo
 		delete(store.subs, s.ID)
 		return sub, nil
 	}
-	return nil, sql.ErrNoRows
```
BenjaminPelletier, line 80: Let's clarify the expected behavior of the interface indicating that nil should be returned when the thing wasn't found (rather than returning sql.ErrNoRows), and repeat for ISAs as well.  Are there any other behavioral requirements of the interface that aren't documented on the interface currently?

steeling: done

pkg/rid/application/isa.go
steeling: since this uses the old value in adjusttimerange, it needs to be done in the txn, unless we do the version check differently.

In another PR I think we can fix up the logic in adjusttimerange to do what you're saying though

PR #324, opened 2020-06-12
Use Volume4D in SearchSubscriptions

BenjaminPelletier:
This PR changes SearchSubscriptions to accept a Volume4D rather than a simple set of cells and owner.  This is done to move toward simpler Store operations that don't embed application logic (like checking the owner of Subscriptions) and therefore can be more easily combined into application-level operations and more easily mocked (while still testing all application logic).

A suite of Subscription query integration tests are added to verify functionality, and they also revealed an incorrect SQL query that has been corrected.

### DIFF
--- a/pkg/scd/store/cockroach/subscriptions.go
+++ b/pkg/scd/store/cockroach/subscriptions.go
@@ -485,25 +485,21 @@ func (c *Store) DeleteSubscription(ctx context.Context, id scdmodels.ID, owner d
 		return nil, err
 	}
 
-	if err := c.tx.Commit(); err != nil {
-		return nil, err
-	}
-
 	if rows == 0 {
 		return nil, dsserr.BadRequest("failed to delete implicit subscription with active operation")
 	}
 	return old, nil
 }
 
-// SearchSubscriptions returns all subscriptions in "cells".
-func (c *Store) SearchSubscriptions(ctx context.Context, cells s2.CellUnion, owner dssmodels.Owner) ([]*scdmodels.Subscription, error) {
+// Implements SubscriptionStore.SearchSubscriptions
+func (c *Store) SearchSubscriptions(ctx context.Context, v4d *dssmodels.Volume4D) ([]*scdmodels.Subscription, error) {
 	var (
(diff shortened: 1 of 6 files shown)


pkg/scd/store/cockroach/subscriptions.go
```diff
@@ -484,25 +484,21 @@ func (c *Store) DeleteSubscription(ctx context.Context, id scdmodels.ID, owner d
...
 	if rows == 0 {
 		return nil, dsserr.BadRequest("failed to delete implicit subscription with active operation")
 	}
 	return old, nil
 }
 
-// SearchSubscriptions returns all subscriptions in "cells".
-func (c *Store) SearchSubscriptions(ctx context.Context, cells s2.CellUnion, owner dssmodels.Owner) ([]*scdmodels.Subscription, error) {
+// Implements SubscriptionStore.SearchSubscriptions
+func (c *Store) SearchSubscriptions(ctx context.Context, v4d *dssmodels.Volume4D) ([]*scdmodels.Subscription, error) {
```
steeling, line 494: Given the RID split between API/App/Store, I think that the Cell logic would be best suited on the App level. What're your thoughts about leaving as is in the store level, and changing the app level to do the conversion (when scd has an app level)

BenjaminPelletier: A Volume4D is more than a 2D footprint; it contains time and altitude bounds.  Although our current implementation does not subdivide cells according to altitude, a future store implementation might.  The application-level need is to find Subscriptions within a Volume4D -- the fact that there is a cell index is a Store-level detail rather than something at the Application level.

steeling: sure, but we could change the signature, or add another method that accepts time bounds. I mostly believe the logic of converting to an s2 cell should be outside of the store level.

A good example of this is update ISA in RID, where we need to calculate the union of the old and new ISA to update the notification index, which would need to be done outside of teh store in the current architecture, and is easier to do with s2 cells than 4d volumes

That said, LGMT, we can always iterate

PR #327, opened 2020-06-16
Feature/ctx db

steeling:
Leverage contexts for passing the txn

### DIFF
--- a/cmds/grpc-backend/main.go
+++ b/cmds/grpc-backend/main.go
@@ -133,7 +133,7 @@ func RunGRPCServer(ctx context.Context, address string) error {
 
 	var (
 		dssServer = &rid.Server{
-			App:     application.NewFromTransactor(store, logger),
+			App:     application.NewFromRepo(store, logger),
 			Timeout: *timeout,
 		}
 		auxServer        = &aux.Server{}
--- a/pkg/cockroach/cockroach.go
+++ b/pkg/cockroach/cockroach.go
@@ -16,7 +16,7 @@ type DB struct {
 	*sql.DB
 }
 
-func (d *DB) getDB(ctx context.Context) dssql.Queryable {
+func (d *DB) getQueryable(ctx context.Context) dssql.Queryable {
 	dbi := ctx.Value(txKey)
 	if dbi == nil {
 		// Return the default db
@@ -35,15 +35,15 @@ func SetTx(ctx context.Context, tx *sql.Tx) context.Context {
 }
 
 func (d *DB) QueryContext(ctx context.Context, query string, args ...interface{}) (*sql.Rows, error) {
-	return d.getDB(ctx).QueryContext(ctx, query, args...)
+	return d.getQueryable(ctx).QueryContext(ctx, query, args...)
 }
 
 func (d *DB) QueryRowContext(ctx context.Context, query string, args ...interface{}) *sql.Row {
-	return d.getDB(ctx).QueryRowContext(ctx, query, args...)
+	return d.getQueryable(ctx).QueryRowContext(ctx, query, args...)
 }
 
 func (d *DB) ExecContext(ctx context.Context, query string, args ...interface{}) (sql.Result, error) {
-	return d.getDB(ctx).ExecContext(ctx, query, args...)
+	return d.getQueryable(ctx).ExecContext(ctx, query, args...)
 }
 
 // Dial returns a DB instance connected to a cockroach instance available at
--- a/pkg/cockroach/cockroach.go
+++ b/pkg/cockroach/cockroach.go
@@ -9,7 +9,11 @@ import (
 	dssql "github.com/interuss/dss/pkg/sql"
 )
 
-var txKey = "transaction_interface_key"
+type key int
+
+const (
+	txKey key = iota
+)
 
 // DB abstracts common operations on sql.DB and sql.Tx instances.
 type DB struct {
--- a/pkg/cockroach/cockroach.go
+++ b/pkg/cockroach/cockroach.go
@@ -1,14 +1,49 @@
 package cockroach
 
 import (
+	"context"
 	"database/sql"
 	"errors"
 	"fmt"
+
+	dssql "github.com/interuss/dss/pkg/sql"
 )
 
-// DB models a connection to a CRDB instance.
+var txKey = "transaction_interface_key"
+
+// DB abstracts common operations on sql.DB and sql.Tx instances.
 type DB struct {
 	*sql.DB
+}
+
+func (d *DB) getDB(ctx context.Context) dssql.Queryable {
+	dbi := ctx.Value(txKey)
+	if dbi == nil {
+		// Return the default db
+		return d.DB
+	}
+
+	return dbi.(dssql.Queryable)
+}
+
+func InTx(ctx context.Context) bool {
+	return ctx.Value(txKey) != nil
+}
+
+func SetTx(ctx context.Context, tx *sql.Tx) context.Context {
+	return context.WithValue(ctx, txKey, tx)
+}
+
+func (d *DB) QueryContext(ctx context.Context, query string, args ...interface{}) (*sql.Rows, error) {
+	return d.getDB(ctx).QueryContext(ctx, query, args...)
+}
+
+func (d *DB) QueryRowContext(ctx context.Context, query string, args ...interface{}) *sql.Row {
+	return d.getDB(ctx).QueryRowContext(ctx, query, args...)
+}
+
+func (d *DB) ExecContext(ctx context.Context, query string, args ...interface{}) (sql.Result, error) {
+	return d.getDB(ctx).ExecContext(ctx, query, args...)
 }
 
 // Dial returns a DB instance connected to a cockroach instance available at
--- a/pkg/rid/application/application.go
+++ b/pkg/rid/application/application.go
@@ -14,7 +14,7 @@ type app struct {
 	// TODO: don't fully embed the repos once we reduce the complexity in the store.
 	// Right now it's "coincidence" that the repo has the same signatures as the App interface
 	// but we will want to simplify the repos and add the complexity here.
-	repos.Transactor
+	repos.Repository
 	clock  clockwork.Clock
 	logger *zap.Logger
 }
@@ -24,11 +24,11 @@ type App interface {
 	SubscriptionApp
 }
 
-// NewFromTransactor is a convenience function for creating an App
+// NewFromRepo is a convenience function for creating an App
 // with the given store.
-func NewFromTransactor(transactor repos.Transactor, logger *zap.Logger) App {
+func NewFromRepo(repo repos.Repository, logger *zap.Logger) App {
 	return &app{
-		Transactor: transactor,
+		Repository: repo,
 		clock:      DefaultClock,
 		logger:     logger,
 	}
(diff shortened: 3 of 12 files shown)


steeling:
> I think I probably agree with [this article](https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39) with regard to Context usage based on "informing" versus "controlling". One guideline it has is that, "if your function can’t behave correctly because of a value that may or may not be inside context.Value, then your API is obscuring required inputs too heavily." I agree with that and think it would be best to explicitly list the required arguments to a function in the signature rather than embedding them in Context.

I don't think that quite applies since the function is actually passed in by the user themself, so it's not exactly this obscured API since the entire intent is to create a transaction. The implementation details of how that done is obscured, which I think is exactly what we want.

steeling:
@vosst

vosst:
> @vosst

vosst:
> > I think I probably agree with [this article](https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39) with regard to Context usage based on "informing" versus "controlling". One guideline it has is that, "if your function can’t behave correctly because of a value that may or may not be inside context.Value, then your API is obscuring required inputs too heavily." I agree with that and think it would be best to explicitly list the required arguments to a function in the signature rather than embedding them in Context.
> 
> I don't think that quite applies since the function is actually passed in by the user themself, so it's not exactly this obscured API since the entire intent is to create a transaction. The implementation details of how that done is obscured, which I think is exactly what we want.

Hmm, I'm not exactly sure that we want to obscure things here. I feel like we would also introduce a failure mode that is potentially difficult to debug, i.e.: For the calls to the repo to work properly, the correct `context.Context` instance needs to be used. In particular, the `context.Context` instance must not be used.

@steeling: Mind elaborating why you prefer the ´Context`-based approach?

steeling:
> > > I think I probably agree with [this article](https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39) with regard to Context usage based on "informing" versus "controlling". One guideline it has is that, "if your function can’t behave correctly because of a value that may or may not be inside context.Value, then your API is obscuring required inputs too heavily." I agree with that and think it would be best to explicitly list the required arguments to a function in the signature rather than embedding them in Context.
> > 
> > 
> > I don't think that quite applies since the function is actually passed in by the user themself, so it's not exactly this obscured API since the entire intent is to create a transaction. The implementation details of how that done is obscured, which I think is exactly what we want.
> 
> Hmm, I'm not exactly sure that we want to obscure things here. I feel like we would also introduce a failure mode that is potentially difficult to debug, i.e.: For the calls to the repo to work properly, the correct `context.Context` instance needs to be used. In particular, the `context.Context` instance must not be used.
> 
> @steeling: Mind elaborating why you prefer the ´Context`-based approach?

Obscured is a bad word, I think the implementation details are hidden, but I don't the code is magical, and its relatively simple to dig around to figure out whats going on.

I'd say in any function if you pass in a context and don't use it, bad things can happen. You can miss timeouts, values, or not know when somebody has tried to cancel your function. I'd even go as far to say that not using a context passed in as a param is a bug in and of itself. I'd also add that it makes sense to consider that a transaction is associated with the spirit of what a context is

If we continue on that assumption then this hides all of the details of the implementation details, without obfuscating what is happening, and leaving the API really clean. Having transactions fully contained within a block scope, and handling retries automatically is another nice property, and having 1 object to act on is nice.

vosst:
> > > > I think I probably agree with [this article](https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39) with regard to Context usage based on "informing" versus "controlling". One guideline it has is that, "if your function can’t behave correctly because of a value that may or may not be inside context.Value, then your API is obscuring required inputs too heavily." I agree with that and think it would be best to explicitly list the required arguments to a function in the signature rather than embedding them in Context.
> > > 
> > > 
> > > I don't think that quite applies since the function is actually passed in by the user themself, so it's not exactly this obscured API since the entire intent is to create a transaction. The implementation details of how that done is obscured, which I think is exactly what we want.
> > 
> > 
> > Hmm, I'm not exactly sure that we want to obscure things here. I feel like we would also introduce a failure mode that is potentially difficult to debug, i.e.: For the calls to the repo to work properly, the correct `context.Context` instance needs to be used. In particular, the `context.Context` instance must not be used.
> > @steeling: Mind elaborating why you prefer the ´Context`-based approach?
> 
> Obscured is a bad word, I think the implementation details are hidden, but I don't the code is magical, and its relatively simple to dig around to figure out whats going on.
> 

Agreed. In my mind, this is a question whether we *explicitly* pass in an interface implementation or whether we implicitly pass in a dependency by augmenting the context. I would rather fall on the side of explicit, though. Right now, you cannot implement or test a functor passed to `Transact` without capturing a `Repository` instance in the functor. In thinking about this, this seems to lead to tight coupling.

> I'd say in any function if you pass in a context and don't use it, bad things can happen. You can miss timeouts, values, or not know when somebody has tried to cancel your function. I'd even go as far to say that not using a context passed in as a param is a bug in and of itself. I'd also add that it makes sense to consider that a transaction is associated with the spirit of what a context is
> 

I see where you are coming from, but wouldn't this argument extend to any implementation of an interface then? So instead of passing in an instance explicitly to a function, just wrap it in a context and have a dispatcher object that uses the interface implementation contained in the context?

> If we continue on that assumption then this hides all of the details of the implementation details, without obfuscating what is happening, and leaving the API really clean. Having transactions fully contained within a block scope, and handling retries automatically is another nice property, and having 1 object to act on is nice.

Thanks, that's helpful. I would argue, though, that contained block scopes and automatic retries can be achieved with an explicit `Repository` argument, too.

steeling:
> > > > > I think I probably agree with [this article](https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39) with regard to Context usage based on "informing" versus "controlling". One guideline it has is that, "if your function can’t behave correctly because of a value that may or may not be inside context.Value, then your API is obscuring required inputs too heavily." I agree with that and think it would be best to explicitly list the required arguments to a function in the signature rather than embedding them in Context.
> > > > 
> > > > 
> > > > I don't think that quite applies since the function is actually passed in by the user themself, so it's not exactly this obscured API since the entire intent is to create a transaction. The implementation details of how that done is obscured, which I think is exactly what we want.
> > > 
> > > 
> > > Hmm, I'm not exactly sure that we want to obscure things here. I feel like we would also introduce a failure mode that is potentially difficult to debug, i.e.: For the calls to the repo to work properly, the correct `context.Context` instance needs to be used. In particular, the `context.Context` instance must not be used.
> > > @steeling: Mind elaborating why you prefer the ´Context`-based approach?
> > 
> > 
> > Obscured is a bad word, I think the implementation details are hidden, but I don't the code is magical, and its relatively simple to dig around to figure out whats going on.
> 
> Agreed. In my mind, this is a question whether we _explicitly_ pass in an interface implementation or whether we implicitly pass in a dependency by augmenting the context. I would rather fall on the side of explicit, though. Right now, you cannot implement or test a functor passed to `Transact` without capturing a `Repository` instance in the functor. In thinking about this, this seems to lead to tight coupling.

I think we could easily test this, seeing as it's the DB interface that implements this, and we could create a mock one easily.
> 
> > I'd say in any function if you pass in a context and don't use it, bad things can happen. You can miss timeouts, values, or not know when somebody has tried to cancel your function. I'd even go as far to say that not using a context passed in as a param is a bug in and of itself. I'd also add that it makes sense to consider that a transaction is associated with the spirit of what a context is
> 
> I see where you are coming from, but wouldn't this argument extend to any implementation of an interface then? So instead of passing in an instance explicitly to a function, just wrap it in a context and have a dispatcher object that uses the interface implementation contained in the context?

Not quite; I'd liken what I'm doing in this PR as more similar to using go's mutex to grab a lock, and run a function inside of that lock. The difference is that the implementation details hide how the lock is being grabbed (ie: leveraging contexts, since we need a database lock). It's less about what the interface is; to the user they are always acting on the Repository object, they only know they have a lock on it inside the functor. 

> 
> > If we continue on that assumption then this hides all of the details of the implementation details, without obfuscating what is happening, and leaving the API really clean. Having transactions fully contained within a block scope, and handling retries automatically is another nice property, and having 1 object to act on is nice.
> 
> Thanks, that's helpful. I would argue, though, that contained block scopes and automatic retries can be achieved with an explicit `Repository` argument, too.

I think at this point the argument more or less breaks down to what we want the API to look like for end user, since I believe anything that comes down to testing or "explicitness" or obscurity can be argued in both directions and inches into subjective territory. So let's go full subjective and see which one we prefer ;) (if that sounds reasonable to you!)
a)
```
sub, err := a.Repository.GetSubscription(ctx, id) // Not in a tx

// In a tx
err = a.Repository.InTxnRetrier(ctx, func(ctx context.Context) error {
    sub, err := a.Repository.GetSubscription(ctx, s.ID)
})
```

or b)
```
// Not in a tx
repo, err := a.Store.Interact(ctx)
if err != nil {
	return nil, err
}
return repo.GetISA(ctx, id)

// In a tx
repo, err := a.Store.Transact(ctx)
if err != nil {
	return nil, err
}
return repo.GetISA(ctx, id)
```

Personally I prefer the former, but I'm not super opinionated so I'll defer to you

BenjaminPelletier:
> It's less about what the interface is; to the user they are always acting on the Repository object, they only know they have a lock on it inside the functor.

I think this is my key objection: the user is not acting on the Repository object, they are acting on the Repository object and a special Context -- specifically, a Context that may contain a particular value type at a particular key in its values map.  My main objection is that this is an actual requirement on the interface, but it is not specified or documented in the interface anywhere.  One concrete way we could fix this is to declare an augmented Context composed from Context and require AugmentedContext instead of the plain Context type whenever the consuming function relied on that augmented capability.  Another way would be to accept a normal Context + the additional thing the AugmentedContext might carry.

I would like to clarify the comparison as the second implementation above [does not reflect](https://github.com/interuss/dss/blob/526fe7ae42f67e5ad3f066025296716d112f40b0/pkg/rid/store/store.go#L31) the approach in #333; the example in b should be:

b)
```go
// Not in a tx
repo, err := a.Store.Interact(ctx)
if err != nil {
	return nil, err
}
return repo.GetISA(ctx, id)

// In a tx
err := a.Store.Transact(ctx, func(ctx context.Context, repo repos.Repository) error {
  sub, err := repo.GetSubscription(ctx, s.ID)
})
```

You'll note that b's approach to a transaction is nearly identical to a's with the difference that a different object is provided to perform operations on in b's approach.

The difference in non-transaction interactions between a and b is solely due to the question of whether the Repository/Store should also implement all the data retrieval methods.  The answer in #333 is no, the thing that is capable of performing transactions should be different from the thing that provides all the data retrieval methods.  In answering yes, this PR is adding additional implicit logic to the interface that is not documented.  Specifically:

```go
err = a.Repository.InTxnRetrier(ctx, func(ctx context.Context) error {
    // It is now invalid to call a.Repository.InTxnRetrier,  but it is valid
    // and necessary to call other methods on a.Repository
})
```

I don't have as strong of an opinion on this tradeoff, but do think it is generally better to minimize presentation of invalid or inappropriate capabilities, and a few additional lines of code is worthwhile when it increases clarity regarding the logic and flow of the program.

BenjaminPelletier:
I believe #333 was the approach that we ended up selecting so I'll close this PR as obsolete.

PR #328, opened 2020-06-17
Feature/db schema manager

charlie-pisuraj:


steeling:
I had previously cobbled together a rough outline of schema version system, to fail early if it detected an incompatible schema version. 

Can we tie those two together in this PR as well.

Adding @vosst to review as well

charlie-pisuraj:
> I had previously cobbled together a rough outline of schema version system, to fail early if it detected an incompatible schema version.
> 
> Can we tie those two together in this PR as well.
> 
> Adding @vosst to review as well

Yeah its already in there, I've added a table called schema_versions and advances the version string as it reaches the next step

cmds/db-manager/main.go
```diff
@@ -0,0 +1,218 @@
...
+		log.Fatal("Failed to get Current DB version for confirmation")
+	}
+	migrationStep, dirty, err := myMigrater.Version()
+	if err != nil {
+		log.Fatal("Failed to get Migration Step for confirmation")
+	}
+	log.Printf("DB Version: %s, Migration Step # %d, Dirty: %v", currentDBVersion, migrationStep, dirty)
+}
+
+func directionText(direction int) string {
```
vosst, line 124: Seems like:
```
type direction int

func (d direction) String() string {
  ...
}
```
would be a better fit here.

cmds/db-manager/main.go
```diff
@@ -0,0 +1,218 @@
...
+	} else if direction < 0 {
+		return "Down"
+	}
+	return "No Change"
+}
+
+func getCurrentDBVersion(crdbURI string) (string, error) {
+	crdb, err := dssCockroach.Dial(crdbURI)
+	if err != nil {
+		log.Panic("Failed to dial CockroachDB")
```
vosst, line 136: I think you rather want to return the error here.

cmds/db-manager/main.go
```diff
@@ -0,0 +1,218 @@
...
+	// check if schema_versions table exists
+	const checkTableQuery = `
+		SELECT EXISTS (
+  		SELECT *
+		  FROM information_schema.tables 
+   		WHERE table_name = 'schema_versions'
+  )`
+	row := crdb.QueryRow(checkTableQuery)
+	var ret bool
+	scanErr := row.Scan(&ret)
```
vosst, line 147: You can happily re-use error instances in golang. Usually leads to more readable code, too.

cmds/db-manager/main.go
```diff
@@ -0,0 +1,218 @@
...
+		log.Panic("Failed to dial CockroachDB")
+	}
+	// check if schema_versions table exists
+	const checkTableQuery = `
+		SELECT EXISTS (
+  		SELECT *
+		  FROM information_schema.tables 
+   		WHERE table_name = 'schema_versions'
+  )`
+	row := crdb.QueryRow(checkTableQuery)
```
vosst, line 145: I would probably re-shuffle this to:
```
var (
  version = "v0.0.0"
  exists bool
)

if err := crdb.QueryRow(checkTableQuery).Scan(&exists); err != nil {
  return "", err
}

if !exists {
  return version, nil
}
```

cmds/db-manager/main.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	return "No Change"
+}
+
+func getCurrentDBVersion(crdbURI string) (string, error) {
+	crdb, err := dssCockroach.Dial(crdbURI)
+	if err != nil {
+		log.Panic("Failed to dial CockroachDB")
+	}
+	// check if schema_versions table exists
```
vosst, line 138: Please add `defer crdb.Close()` here, and remove the manual close by the end of the function.

cmds/db-manager/main.go
```diff
@@ -0,0 +1,218 @@
...
+		if version == "" {
+			version = "v0.0.0"
+		}
+	}
+	crdb.Close()
+	return version, nil
+}
+
+func parseVersion(version string) [3]int {
+	splitVersion := strings.Split(version, ".")
```
vosst, line 171: Seems like:
```
var (
  major, minor, patch int
)
if n, err := fmt.Sscanf(version, "v%d.%d.%d", &major, &minor, &patch); err != nil || n != 3 {
  ...
}
```
would be more readable and more robust.

cmds/db-manager/main.go
```diff
@@ -0,0 +1,218 @@
...
+	"go.uber.org/zap"
+)
+
+// MyMigrate is an alias for exstending migrate.Migrate
+type MyMigrate struct {
+	*migrate.Migrate
+}
+
+var (
+	path      = flag.String("migration_files", "", "path to db migration files directory")
```
vosst, line 30: Seems like `migrations_dir` is a better fit here.

charlie-pisuraj:
> Great start, I left a couple of comments inline. For organizing the migration files: Did you consider grouping `up.sql` and `down.sql` into subdirectories?

The way I wanted this tool to work is for the Us to specify the desired db version, and have the tool determine it by itself if it needs to do an "Up" or "Down", because of that the framework needed to be able to see all the files in 1 directory and I didn't see a place where they implemented the ability to recursively read subdirectories.

vosst:
> > Great start, I left a couple of comments inline. For organizing the migration files: Did you consider grouping `up.sql` and `down.sql` into subdirectories?
> 
> The way I wanted this tool to work is for the Us to specify the desired db version, and have the tool determine it by itself if it needs to do an "Up" or "Down", because of that the framework needed to be able to see all the files in 1 directory and I didn't see a place where they implemented the ability to recursively read subdirectories.

You could use https://godoc.org/github.com/golang-migrate/migrate#NewWithSourceInstance to supply a custom `source.Driver` instance for aggregating across a directory subtree. Probably not worth doing so in the scope of this PR, though.

vosst:
One other findings: It seems like you accidentally bumped the revision of a lot of dependencies. Did you build locally with a more recent Golang version?

charlie-pisuraj:
> One other findings: It seems like you accidentally bumped the revision of a lot of dependencies. Did you build locally with a more recent Golang version?

Yeah I did run go build once, what can i do to fix this issue

steeling:
> > I had previously cobbled together a rough outline of schema version system, to fail early if it detected an incompatible schema version.
> > Can we tie those two together in this PR as well.
> > Adding @vosst to review as well
> 
> Yeah its already in there, I've added a table called schema_versions and advances the version string as it reaches the next step

Can you change the GetVersion function in rid/cockroach/store.go?

@vosst what're your thoughts on splitting this between RID/SCD? If we start merging some of the concepts (or at least flattening), it makes sense to have 1 schema versioning system, but if we want to keep them separate we should split this up.

charlie-pisuraj:
> 
> Can you change the GetVersion function in rid/cockroach/store.go?
> 

I was going to do this in a separate PR where I remove the bootstrapping code out of GRPC Backend

BenjaminPelletier:
> > I had previously cobbled together a rough outline of schema version system, to fail early if it detected an incompatible schema version.
> > Can we tie those two together in this PR as well.
> 
> Yeah its already in there, I've added a table called schema_versions and advances the version string as it reaches the next step

We definitely want to update [GetVersion](https://github.com/interuss/dss/blob/f707d47e332936eb25e2bcd75de54b0d6679ed0f/pkg/rid/cockroach/store.go#L165) in another PR as quickly after this PR as possible.  But in the spirit of small PRs, I think this PR has a good, self-contained scope as-is and we can separate the update of that function (and associated logic) into a follow-up PR.

BenjaminPelletier:
> > Can you change the GetVersion function in rid/cockroach/store.go?
> 
> I was going to do this in a separate PR where I remove the bootstrapping code out of GRPC Backend

+1 to small PRs; I think the current scope of the PR is appropriate, and a follow-up PR to change the gRPC backend, etc will be great.

BenjaminPelletier:
> @vosst what're your thoughts on splitting this between RID/SCD? If we start merging some of the concepts (or at least flattening), it makes sense to have 1 schema versioning system, but if we want to keep them separate we should split this up.

I recommend versioning the database schema separately for RID and SCD because those two sets of functionality are driven by two different standards that will evolve on different timelines.  This also allows, for instance, a deployed RID instance to completely ignore database schema changes for SCD.  As a consequence of this, @charlie-pisuraj I recommend:
1) Changing the `db-schemas` folder name to `rid-schemas`
2) Renaming the `schema_versions` table to `rid_schema_versions`, or else changing it to a key/value table where `rid` is the key for a single row (I prefer the latter)

cmds/db-manager/main.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	totalMove := 0
+	migrateDirection := myMigrater.MigrationDirection(postgresURI, *dbVersion, *step)
+	for migrateDirection != 0 {
+		migraterErr = myMigrater.Steps(int(migrateDirection))
+		if migraterErr != nil {
+			log.Fatal(migraterErr)
+		}
+		totalMove += int(migrateDirection)
+		log.Printf("Migrated %s by %d step", migrateDirection.String(), int(math.Abs(float64(int(migrateDirection)))))
```
BenjaminPelletier, line 116: Let's record myMigrater.Version() before calling .Steps and after calling .Steps and report that version transition in this log statement.

cmds/db-manager/main.go
```diff
@@ -0,0 +1,219 @@
...
+		migrateDirection = myMigrater.MigrationDirection(postgresURI, *dbVersion, *step)
+	}
+	if totalMove == 0 {
+		log.Println("No Changes")
+	} else {
+		log.Printf("Moved %d steps %s in total", int(math.Abs(float64(totalMove))), migrateDirection.String())
```
BenjaminPelletier, line 122: This `int(math.Abs(float64(` cast is unnecessary; let's just declare totalMove as an int to start with.

cmds/db-manager/main.go
```diff
@@ -0,0 +1,219 @@
...
+		log.Fatal("Failed to get Current DB version for confirmation")
+	}
+	migrationStep, dirty, err := myMigrater.Version()
+	if err != nil {
+		log.Fatal("Failed to get Migration Step for confirmation")
+	}
+	log.Printf("DB Version: %s, Migration Step # %d, Dirty: %v", currentDBVersion, migrationStep, dirty)
+}
+
+func getCurrentDBVersion(crdbURI string) (string, error) {
```
BenjaminPelletier, line 136: Let's mirror [the migrater's approach](https://github.com/golang-migrate/migrate/blob/8b6e3f1a2ea1fd3016ad479965c573d52469fe51/database/cockroachdb/cockroachdb.go#L256) here since they probably have best practices.  After making the change mentioned above to [construct a migration driver](https://github.com/golang-migrate/migrate/blob/8b6e3f1a2ea1fd3016ad479965c573d52469fe51/database/cockroachdb/cockroachdb.go#L55), we'll have a DB ready to use without redialing.

build/db-schemas/000004_drop_cells_table.up.sql
```diff
@@ -0,0 +1,3 @@
+DROP TABLE IF EXISTS cells_identification_service_areas;
+DROP TABLE IF EXISTS cells_subscriptions;
+UPDATE schema_versions set schema_version = 'v2.0.0' WHERE onerow_enforcer = TRUE
```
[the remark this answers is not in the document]
charlie-pisuraj: We have made changes that rely on versioning, this tool allows for backwards compatibility and it will safely migrate all existing data to the "v2.0.0", or down to "v1.0.0" if necessary. This means that for those who already ran the old binary can freely upgrade without much work, the new deployment jsonnet configs that I'm working on would make it easier as well.

build/db-schemas/000002_inverted_index.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE identification_service_areas ADD COLUMN IF NOT EXISTS cells INT64[];
```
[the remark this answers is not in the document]
charlie-pisuraj: The advantage to this is that we will not risk dropping any data if a single step is corrupted. Each step is idempotent can can be run multiple times if needed.

cmds/db-manager/main.go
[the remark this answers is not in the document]
charlie-pisuraj: This shows how many steps we have taken, how many jump we did successfully and in which direction. I think it is a good information for the person who ran this to know as a summary.

cmds/db-manager/main.go
charlie-pisuraj: Unfortunately Go Doesn't do have a built in absolute method for int. I find that absurd but thats the reason for this weird thing.

cmds/db-manager/main.go
```diff
@@ -0,0 +1,219 @@
...
+		"application_name": "SchemaManager",
+	}
+	postgresURI, err := dssCockroach.BuildURI(uriParams)
+	if err != nil {
+		log.Panic("Failed to build URI", zap.Error(err))
+	}
+	crdbURI := strings.Replace(postgresURI, "postgresql", "cockroachdb", 1)
+	*path = fmt.Sprintf("file://%v", *path)
+	migrater, migraterErr := migrate.New(*path, crdbURI)
+	myMigrater := &MyMigrate{migrater}
```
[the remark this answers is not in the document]
charlie-pisuraj: there is a check for migrateErr != nil a few lines below. Creating this "MyMigrate" object makes the call more uniform with the deferred function

steeling:
> > @vosst what're your thoughts on splitting this between RID/SCD? If we start merging some of the concepts (or at least flattening), it makes sense to have 1 schema versioning system, but if we want to keep them separate we should split this up.
> 
> I recommend versioning the database schema separately for RID and SCD because those two sets of functionality are driven by two different standards that will evolve on different timelines. This also allows, for instance, a deployed RID instance to completely ignore database schema changes for SCD. As a consequence of this, @charlie-pisuraj I recommend:
> 
> 1. Changing the `db-schemas` folder name to `rid-schemas`
> 2. Changing the `schema_versions` table to a key/value table where `rid` is the key for a single row
> 
> More details on what we'll need to do to support migrations per component in review comments.

If we separate these we should use separate databases. It's a little strange to do partial migrations that you need to make sure don't conflict and maintaining separate migrations files for the same database. Much easier to do a full split.

If that's the case we'll need to refactor the code, which I believe currently shares a database connection (probably one line that adds an extra dial. I'd recommend we keep rid in the current db, and move scd to a new one

charlie-pisuraj:
I agree with what Sean said above, since SCD is still in development, I think a split now is much easier. I can put that on my todo list and get that rolling too. For this PR the changes required will be having the DB name added in the URI before dialing.

cmds/db-manager/main.go
BenjaminPelletier: Ah, that is ridiculous.  nit: could we write a little utility function to improve developer readability?

cmds/db-manager/main.go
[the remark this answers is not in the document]
BenjaminPelletier: Ack; same nit as below with regard to an abs utility function for readability.

build/db-schemas/000004_drop_cells_table.up.sql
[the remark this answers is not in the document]
BenjaminPelletier: It seems weird that we're going to have two database schemas that are both considered "v1.0.0" and two database schemas that are both considered "v2.0.0" (each one with a schema_versions table + schema_migrations table, and each one without both), but actually it does look like an original-v2.0.0 configuration can be migrated to to a migrations-v2.0.0 configuration by applying all the up steps from the beginning (what will happen when the migration version isn't detected correctly, IIUC), and likewise for original-v1.0.0.  I think it would help a lot to document the intended usage and expected behavior of this tool under normal circumstances and under anticipated legacy cases like these in a README, but perhaps that is best suited for a separate PR to limit PR scope.  And since I was wrong about backwards compatibility, I'll resolve this comment.

cmds/db-manager/main.go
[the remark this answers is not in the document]
BenjaminPelletier: Yep; it seems like this could be much simpler with:
```go
migrater, err := migrate.New(*path, crdbURI)
if err != nil {
  log.Panic(err)
}
myMigrater := &MyMigrate{migrater}
defer func() {
	if _, err := myMigrater.Close(); err != nil {
		log.Println(err)
	}
}()
```
(then no consideration of migrateErr anywhere else)

As written currently, it seems like an error on line 84 will result in an invalid migrater, yet line 109 and beyond will still run resulting in undefined behavior, so basically I'm asking here whether it is valid to do anything else in the routine if migrateErr is not nil (therefore implying that migrater is not valid).

build/db-schemas/000002_inverted_index.up.sql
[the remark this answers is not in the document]
BenjaminPelletier: If 000002, 000003, and 000004 steps were combined into a single step, would that combined step have a risk of dropping data if the combined step were corrupted?  Would the combined step be idempotent?

cmds/db-manager/main.go
```diff
@@ -0,0 +1,219 @@
...
+		"ssl_dir":          *cockroachParams.sslDir,
+		"application_name": "SchemaManager",
+	}
+	postgresURI, err := dssCockroach.BuildURI(uriParams)
+	if err != nil {
+		log.Panic("Failed to build URI", zap.Error(err))
+	}
+	crdbURI := strings.Replace(postgresURI, "postgresql", "cockroachdb", 1)
+	*path = fmt.Sprintf("file://%v", *path)
+	migrater, migraterErr := migrate.New(*path, crdbURI)
```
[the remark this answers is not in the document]
BenjaminPelletier: The separate-databases approach is an alternative to this suggestion, so assuming we adopt that, this comment is obsolete.

cmds/db-manager/main.go
```diff
@@ -0,0 +1,219 @@
...
+	if d > 0 {
+		return "Up"
+	} else if d < 0 {
+		return "Down"
+	}
+	return "No Change"
+}
+
+var (
+	path      = flag.String("migration_dir", "", "path to db migration files directory")
```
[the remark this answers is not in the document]
BenjaminPelletier: Assuming we adopt the separate-databases approach, I now suggest that we reduce the possibility of inconsistency (and reduce complexity) by naming migrations folders according to database name, and just pass in the database name as a single argument to this tool.

BenjaminPelletier:
> If we separate these we should use separate databases. It's a little strange to do partial migrations that you need to make sure don't conflict and maintaining separate migrations files for the same database. Much easier to do a full split.

This seems like a good solution to me.  @vosst do you have any objections?

> I'd recommend we keep rid in the current db, and move scd to a new one

I think the best path will be to have good names for both databases.  It looks like we're just using the default database currently, and if that's the case, I think we'll want to switch to one that is explicitly RID on the RID side.

> However, if there is be a need to join data between SCD and RID tables to serve something back to the requester, then this proposal might make that more difficult

I don't think there is any foreseeable need to join SCD and RID tables.

build/db-schemas/000002_inverted_index.up.sql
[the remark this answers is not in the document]
charlie-pisuraj: we can make it idempotent but it would be harder to debug if anything goes wrong, breaking down in small steps where 1 step adds columns, 1 step migrates the data, and 1 final step dropping the old data makes it really clean.

cmds/db-manager/main.go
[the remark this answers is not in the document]
charlie-pisuraj: We could do that, but would it be clearer to to have "rid-schema" as the dir name instead of "defaultdb" ? I don't have a strong opinion on this, just wondering if it would be confusing or not. What I'd do to to keep your suggestion would be to pass ./db-schemas/defaultdb to --schema_dir and take the leaf directory name as the db name to upgrade

cmds/db-manager/main.go
```diff
@@ -0,0 +1,228 @@
...
+		"application_name": "SchemaManager",
+		"db_name":          filepath.Base(*path),
+	}
+	postgresURI, err := dssCockroach.BuildURI(uriParams)
+	if err != nil {
+		log.Panic("Failed to build URI", zap.Error(err))
+	}
+	crdbURI := strings.Replace(postgresURI, "postgresql", "cockroachdb", 1)
+	*path = fmt.Sprintf("file://%v", *path)
+	migrater, migraterErr := migrate.New(*path, crdbURI)
```
BenjaminPelletier, line 85: In general, I think we [want to reuse the `err` name for errors](https://github.com/interuss/dss/pull/328#discussion_r442495313); no need for each error to be named differently.

cmds/db-manager/main.go
```diff
@@ -0,0 +1,228 @@
...
+	if migraterErr != nil {
+		log.Panic(migraterErr)
+	}
+	myMigrater := &MyMigrate{migrater}
+	defer func() {
+		if migraterErr == nil {
```
BenjaminPelletier, line 91: No need for this check any more; this code block can only be reached if migrateErr == nil (because of line 86).

cmds/db-manager/main.go
```diff
@@ -0,0 +1,228 @@
...
+			if _, err := myMigrater.Close(); err != nil {
+				log.Println(err)
+			}
+		}
+	}()
+	if migraterErr == nil {
```
BenjaminPelletier, line 97: No need for this check any more; this code block can only be reached if migrateErr == nil (because of line 86).

cmds/db-manager/main.go
```diff
@@ -0,0 +1,228 @@
...
+			for range signals {
+				log.Println("Stopping after this running migration ...")
+				myMigrater.GracefulStop <- true
+				return
+			}
+		}()
+	}
+	totalMove := 0
+	migrateDirection := myMigrater.MigrationDirection(postgresURI, *dbVersion, *step)
+	preMigrationStep, _, _ := myMigrater.Version()
```
BenjaminPelletier, line 114: This could return an error in the last parameter; something bad should happen if that unexpected situation occurs (probably log.Panic)

cmds/db-manager/main.go
```diff
@@ -0,0 +1,228 @@
...
+	for migrateDirection != 0 {
+		migraterErr = myMigrater.Steps(int(migrateDirection))
+		if migraterErr != nil {
+			log.Fatal(migraterErr)
```
BenjaminPelletier, line 118: I think Fatal calls os.Exit without running defers, so do we want log.Panic here to ensure myMigrater is Closed in the above defer?

cmds/db-manager/main.go
```diff
@@ -0,0 +1,228 @@
...
+
+	currentDBVersion, err := getCurrentDBVersion(postgresURI)
+	if err != nil {
+		log.Fatal("Failed to get Current DB version for confirmation")
+	}
+	log.Printf("DB Version: %s, Migration Step # %d, Dirty: %v", currentDBVersion, postMigrationStep, dirty)
+}
+
+func intAbs(x int) int {
+	return int(math.Abs(float64(x)))
```
BenjaminPelletier, line 142: nit (don't need to change if not convenient):
```go
if x < 0 {
  return -x
} else {
  return x
}
```

cmds/db-manager/main.go
BenjaminPelletier: Per the migrater's approach, we can make just one query and check for 42P01 to detect a missing table; can we skip the entire first half of this method with the following?
```go
  const defaultVersion = "v0.0.0"
	err = c.db.QueryRow(getVerisonQuery).Scan(&version)

	switch {
	case err == sql.ErrNoRows:
		return defaultVersion, nil

	case err != nil:
		if e, ok := err.(*pq.Error); ok {
			// 42P01 is "UndefinedTableError" in CockroachDB
			// https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/pgwire/pgerror/codes.go
			if e.Code == "42P01" {
				return defaultVersion, nil
			}
		}
		return "", err

	default:
		return version, nil
	}
```

cmds/db-manager/main.go
```diff
@@ -0,0 +1,228 @@
...
+	if n, err := fmt.Sscanf(version, "v%d.%d.%d", &major, &minor, &patch); err != nil || n != 3 {
+		log.Panic(err)
+	}
+	result := [3]int{major, minor, patch}
+	return result
+}
+
+// MigrationDirection reads our custom DB version string as well as the Migration Steps from the framework
+// and returns a signed integer value of the Direction and count to migrate the db
+func (m *MyMigrate) MigrationDirection(dbURI string, desiredVersion string, desireStep int) Direction {
```
BenjaminPelletier, line 200: nit: I think "desire**d**Step" would be cleaner

cmds/db-manager/main.go
```diff
@@ -0,0 +1,228 @@
...
+	if desireStep != 0 {
+		migrateStep, dirty, err := m.Version()
+		if err != migrate.ErrNilVersion && err != nil {
+			log.Panic(err)
```
BenjaminPelletier, line 204: I think we want to [return the error here](https://github.com/interuss/dss/pull/328#discussion_r442494747) (and below in this function)

cmds/db-manager/main.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	result := [3]int{major, minor, patch}
+	return result
+}
+
+// MigrationDirection reads our custom DB version string as well as the Migration Steps from the framework
+// and returns a signed integer value of the Direction and count to migrate the db
+func (m *MyMigrate) MigrationDirection(dbURI string, desiredVersion string, desireStep int) Direction {
+	if desireStep != 0 {
+		migrateStep, dirty, err := m.Version()
```
BenjaminPelletier, line 202: nit: IMO, currentStep would be clearer than migrateStep

cmds/db-manager/main.go
```diff
@@ -0,0 +1,228 @@
...
+	if d > 0 {
+		return "Up"
+	} else if d < 0 {
+		return "Down"
+	}
+	return "No Change"
+}
+
+var (
+	path      = flag.String("schemas_dir", "/Users/pcharlie/charlie_dss/dss/build/db-schemas/scd", "path to db migration files directory")
```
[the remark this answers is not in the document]
charlie-pisuraj: oops, that was a left over from testing, VS code was annoying me so this is my work around, good catch

cmds/db-manager/main.go
charlie-pisuraj: There is a reassignment in Line 116 when `.Steps()` is called so this deferred function will check it once before trying to close the connection

cmds/db-manager/main.go
charlie-pisuraj: Since we are reassigning the migraterErr everytime we call .Steps() I think its best not to muddle the other err here so it could be used in the deferred .Close() call

cmds/db-manager/main.go
charlie-pisuraj: I'll have to handle the error and log.Panic outside anyway but I guess its good convention

cmds/db-manager/main.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	myMigrater := &MyMigrate{migrater}
+	defer func() {
+		if migraterErr == nil {
+			if _, err := myMigrater.Close(); err != nil {
+				log.Println(err)
+			}
+		}
+	}()
+	myMigrater.PrefetchMigrations = 10
```
steeling, line 97: do we need to set this? it's reading locally so looks like it doesn't matter too much

cmds/db-manager/main.go
```diff
@@ -0,0 +1,235 @@
...
+	myMigrater.LockTimeout = time.Duration(15) * time.Second
```
steeling, line 98: Also looks like we're setting this as the default, might as well remove?

cmds/db-manager/main.go
steeling: also looks like this is the default from the docs

cmds/db-manager/main.go
```diff
@@ -0,0 +1,235 @@
...
+
+	// handle Ctrl+c
+	signals := make(chan os.Signal, 1)
```
steeling, line 101: Can you put this on MyMigrater.New(), and instead of instantiating inline call new?

cmds/db-manager/main.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		log.Panic("Failed to build URI", zap.Error(err))
+	}
+	crdbURI := strings.Replace(postgresURI, "postgresql", "cockroachdb", 1)
+	*path = fmt.Sprintf("file://%v", *path)
+	migrater, migraterErr := migrate.New(*path, crdbURI)
+	if migraterErr != nil {
+		log.Panic(migraterErr)
+	}
+	myMigrater := &MyMigrate{migrater}
```
steeling, line 89: add a mymigrate.New which calls migrate.New(*path, crdbURI), and does other initialization (see comments below)

cmds/db-manager/main.go
```diff
@@ -0,0 +1,235 @@
...
+		return "", err
+	}
+	// if for some reason the string returned is empty
+	if version == "" {
+		version = "v0.0.0"
+	}
+	return version, nil
+}
+
+func parseVersion(version string) [3]int {
```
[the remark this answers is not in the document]
charlie-pisuraj: dang should have known about this, would have made my life easier

cmds/db-manager/main.go
```diff
@@ -0,0 +1,229 @@
...
+)
+
+// MyMigrate is an alias for extending migrate.Migrate
+type MyMigrate struct {
+	*migrate.Migrate
+	postgresURI string
+}
+
+// Direction is an alias for int indicating the direction and steps of migration
+type Direction int
```
steeling, line 32: nit:

add a 
```
const (
  Up Direction = 1
  Down Direction = 2
)
```

charlie-pisuraj: discussed offline, I don't foresee where we are going to use this Enum

cmds/db-manager/main.go
```diff
@@ -0,0 +1,229 @@
...
+		log.Printf("Migrated %s by %d step", migrateDirection.String(), intAbs(int(migrateDirection)))
+		migrateDirection, err = m.MigrationDirection(desiredDBVersion, *step)
+		if err != nil {
+			return totalMoves, err
+		}
+	}
+	return totalMoves, nil
+}
+
+// New instanciates a new migrate object
```
BenjaminPelletier, line 143: nit: instantiates

cmds/db-manager/main.go
```diff
@@ -0,0 +1,229 @@
...
+func New(path string, dbURI string) (*MyMigrate, error) {
+	crdbURI := strings.Replace(dbURI, "postgresql", "cockroachdb", 1)
+	path = fmt.Sprintf("file://%v", path)
+	migrater, err := migrate.New(path, crdbURI)
```
BenjaminPelletier, line 147: Either migrater should be valid or err should be != nil; both of those conditions shouldn't happen at the same time.  So, we should immediately return nil, err if err != nil here and we should return myMigrater, nil at the end of the function.

cmds/db-manager/main.go
```diff
@@ -0,0 +1,229 @@
...
+	if d > 0 {
+		return "Up"
+	} else if d < 0 {
+		return "Down"
+	}
+	return "No Change"
+}
+
+var (
+	path      = flag.String("schemas_dir", "", "path to db migration files directory")
```
BenjaminPelletier, line 44: Let's be clear about the behavior; since we don't yet have a README for this tool and process, how about: "path to db migration files directory.  the migrations found there will be applied to the database whose name matches the folder name."

charlie-pisuraj:
> Thanks for all this work; I feel like all the iteration here has produced a pretty good result, but if you feel like the process was too burdensome or could be improved, let's definitely discuss that. This looks good to me with just a couple small comments. If possible, let's give @vosst one more chance to take a look, but then we can merge by around noon Pacific on Monday. I'm looking forward to using this PR and the followup where we check schema version in code in #334 so this is going to help me immediately, as well as being a big process/architecture improvement in general.

I'm ok with the process, however I do think that it does take a lot of your time to review it over and over again. Also the fact that I'm not super fluent with Go doesn't help.

charlie-pisuraj:
@vosst can we get a review?

@BenjaminPelletier I'm not able to merge it becaused there is a requested change, however I think I've already addressed them all + marked them as resolved

vosst:
> @vosst can we get a review?
> 
> @BenjaminPelletier I'm not able to merge it becaused there is a requested change, however I think I've already addressed them all + marked them as resolved

This looks good to me. Going forward (and certainly in follow-up PRs :-)) we should probably think about:
  * introducing a custom driver that allows us to specify the order of migrations without relying on the ordering of filenames (thinking about a central manifest that describes all of the migrations and their ordering).
  * introducing tests that exercise individual migrations/chains of migrations for idempotency.

PR #329, opened 2020-06-18
[scd] Introduce repo interfaces and refactor store interfaces

vosst:


pkg/scd/store/cockroach/store.go
```diff
@@ -15,81 +17,61 @@ var (
...
-func (t *Transactor) Transact() (scdstore.Transaction, error) {
-	tx, err := t.db.Begin()
-	if err != nil {
-		return nil, err
-	}
-	return &Transaction{
-		tx:     tx,
-		logger: t.logger,
-		clock:  t.clock,
+// Interact returns a new repo instance.
```
BenjaminPelletier, line 45: nit: I think it would be best to comment interface implementations by referring to the interface rather than describing what the function does to prevent comment inconsistency and to make it clear that the function is implementing an interface.  So, here: "Interact implements store.Interactor interface."

pkg/scd/store/store.go
```diff
@@ -2,128 +2,31 @@ package store
...
-	// subscription.
-	UpsertSubscription(ctx context.Context, sub *scdmodels.Subscription) (*scdmodels.Subscription, []*scdmodels.Operation, error)
-
-	// DeleteSubscription deletes a Subscription from the store and returns the
-	// deleted subscription.  Returns nil and an error if the Subscription does
-	// not exist, or is owned by someone other than the specified owner.
-	DeleteSubscription(ctx context.Context, id scdmodels.ID, owner dssmodels.Owner, version scdmodels.Version) (*scdmodels.Subscription, error)
-}
-
 // Store abstracts strategic conflict detection interactions with the backing
```
BenjaminPelletier, line 9: I'd recommend more documentation regarding the roles of Store versus Repo; perhaps something like "Store provides the means by which to obtain Repos with which to interact with the strategic conflict detection backing store."

pkg/scd/repos/repos.go
```diff
@@ -0,0 +1,52 @@
...
+	// subscription.
+	UpsertSubscription(ctx context.Context, sub *scdmodels.Subscription) (*scdmodels.Subscription, []*scdmodels.Operation, error)
+
+	// DeleteSubscription deletes a Subscription from the store and returns the
+	// deleted subscription.  Returns nil and an error if the Subscription does
+	// not exist, or is owned by someone other than the specified owner.
+	DeleteSubscription(ctx context.Context, id scdmodels.ID, owner dssmodels.Owner, version scdmodels.Version) (*scdmodels.Subscription, error)
+}
+
+// Repository aggregates all entity-specific repo interfaces.
```
BenjaminPelletier, line 48: nit: "all SCD-specific" (Subscriptions aren't really entities, and this is just entities for SCD)

PR #330, opened 2020-06-19
Add easy scope specification to integration test suite

BenjaminPelletier:
Previously, our integration test suite almost always depended on the assumption that every token would be a super token, containing all scopes necessary to perform any action on the system.  This PR begins the process of making intended scopes more explicit and more customizable in tests.

The primary means by which this is accomplished are that 1) a new Session type is used which attaches an `auth` parameter to requests automatically according to an additional `scope` or `scopes` argument in each verb (`get`, `put`, etc) and 2) a test function decorator `default_scope` modifies each of the session inputs to that test function to default to the specified scope for the duration of that test function.

This PR also updates test_operation_simple.py to always explicitly specify scopes (using both the argument and decorator as appropriate), and to include additional tests that verify incorrect scopes do not permit access to the Operation reference endpoints.

Adapting all other existing tests to use explicitly-scoped requests will occur in a future PR, after which the default set of scopes can be removed and an error thrown instead (so that all future tests will be required to be explicit about the scope(s) they are using to perform their test actions).

### DIFF
--- a/monitoring/prober/rid/test_token_validation.py
+++ b/monitoring/prober/rid/test_token_validation.py
@@ -25,14 +25,11 @@ def test_validate_token_bad_user(session):
   resp = session.get('/validate_oauth?owner=bad_user')
   assert resp.status_code == 403
 
-def test_put_isa_with_read_only_scope_token(rogue_session, session, isa2_uuid):
-  read_only_token = session.issue_token(['dss.read.identification_service_areas'])
-  rogue_session.headers['Authorization'] = f'Bearer {read_only_token}'
-
+def test_put_isa_with_read_only_scope_token(session, isa2_uuid):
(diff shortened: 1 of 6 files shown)


monitoring/prober/rid/test_token_validation.py
```diff
@@ -47,7 +44,7 @@ def test_put_isa_with_read_only_scope_token(rogue_session, session, isa2_uuid):
               'time_end': time_end.strftime(common.DATE_FORMAT),
           },
           'flights_url': 'https://example.com/dss',
-      })
+      }, scope='dss.read.identification_service_areas')
```
vosst, line 47: nit: Seems like this could use a constant, too.

BenjaminPelletier: Done.

PR #332, opened 2020-06-23
Simplify SCD repo APIs (Subscription + SearchOperations)

BenjaminPelletier:
This PR extracts most application logic embedded in the SCD store/repo layer for Subscription repo actions as well as SearchOperations.  This allows us to move toward a simple CRUD store which allows simpler mocks that more realistically test the application logic checks and behavior currently embedded in the store implementation.

One functional change in this PR is that the Subscription count is no longer checked in SCD because there is no maximum Subscription count requirement in the SCD draft standard, and there is unlikely to be a maximum requirement because of implicit Subscriptions for Operations.

### DIFF
--- a/pkg/scd/subscriptions_handler.go
+++ b/pkg/scd/subscriptions_handler.go
@@ -142,9 +142,17 @@ func (a *Server) GetSubscription(ctx context.Context, req *scdpb.GetSubscription
 	var response *scdpb.GetSubscriptionResponse
 	action := func(ctx context.Context, r repos.Repository) (err error) {
 		// Get Subscription from Store
-		sub, err := r.GetSubscription(ctx, id, owner)
+		sub, err := r.GetSubscription(ctx, id)
 		if err != nil {
 			return err
+		}
+		if sub == nil {
+			return dsserr.NotFound(id.String())
+		}
+
+		// Check if the client is authorized to view this Subscription
+		if owner != sub.Owner {
+			return dsserr.PermissionDenied("Subscription owned by a different client")
 		}
 
 		// Convert Subscription to proto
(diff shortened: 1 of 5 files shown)


pkg/scd/subscriptions_handler.go
```diff
@@ -67,32 +68,73 @@ func (a *Server) PutSubscription(ctx context.Context, req *scdpb.PutSubscription
 		NotifyForConstraints: params.NotifyForConstraints,
 	}
 
-	// Validate requested Subscription
-	if !sub.NotifyForOperations && !sub.NotifyForConstraints {
-		return nil, dsserr.BadRequest("no notification triggers requested for Subscription")
+	// Validate subrequested Subscription
+	if !subreq.NotifyForOperations && !subreq.NotifyForConstraints {
+		return nil, dsserr.BadRequest("no notification triggers subrequested for Subscription")
```
vosst, line 73: Looks like a search'n'replace artifact.

pkg/scd/subscriptions_handler.go
```diff
@@ -142,10 +184,18 @@ func (a *Server) GetSubscription(ctx context.Context, req *scdpb.GetSubscription
 	var response *scdpb.GetSubscriptionResponse
 	action := func(ctx context.Context, r repos.Repository) (err error) {
 		// Get Subscription from Store
-		sub, err := r.GetSubscription(ctx, id, owner)
+		sub, err := r.GetSubscription(ctx, id)
 		if err != nil {
 			return err
 		}
+		if sub == nil {
```
vosst, line 191: I'm not in favor of returning `nil, nil` instead of `nil, NotFound`. Seems like we rather want the more explicit error code returned from the repo instead of interpreting the implicit `nil, nil` pair. I have local changes to clean up error handling, though, and I can happily rebase and fix once your PR is in.

pkg/scd/subscriptions_handler.go
BenjaminPelletier: Thanks; fixed.

pkg/scd/subscriptions_handler.go
steeling: I made this change originally to try to simplify some of our error handling, and since it's more in line with how the raw sql package handles this (we do an explicit check for nil, and return the NotFound error manually).

Open to moving it back though

BenjaminPelletier: If you don't mind rebasing and fixing, then I'll leave it to you to adjust.  However, I would like to understand more about why nil, nil wouldn't be a reasonable return value.  It seems like looking for a Subscription and not finding it would be a normal (non-exceptional) case -- for instance, when deciding what action to take when upserting.  And that approach has the advantage of reducing the number of specific errors the interface must specify by 1.

vosst: -ish :) I think it's easier to reason along the lines of the following invariant: if the error is nil, the actual result is not, thus supporting the following (very common) pattern:
```
result, err := someOperation(...)
if err != nil {
  ...
}
// Code from this point on assumes that result refers to some reasonable object.
```

BenjaminPelletier: The error returned upon the nominal case of the object not existing will have to be part of the interface definition (in a comment), so to avoid changing it once in this PR and again when the error code harmonization PR comes out, I'll just acknowledge the comment and use a different PR to address.

PR #333, opened 2020-06-23
[rid] Unify store/repo layout

vosst:
This PR unifies the store/repo layout between `rid` and `scd`.


pkg/rid/store/cockroach/identification_service_area.go
```diff
@@ -21,14 +21,14 @@ const (
 	updateISAFields = "id, url, cells, starts_at, ends_at, updated_at"
 )
 
-// ISAStore is an implementation of the ISARepo for CRDB.
-type ISAStore struct {
+// isa is an implementation of the ISARepo for CRDB.
+type isa struct {
```
BenjaminPelletier, line 25: Could we keep isaRepo here?  That seems like a more descriptive data type than "ISA" which is an entirely different data type.

pkg/rid/store/cockroach/store.go
```diff
@@ -25,23 +25,45 @@ var (
 	DefaultTimeout = 10 * time.Second
 )
 
+type repo struct {
+	*isa
+	*subscription
+}
+
 // Store is an implementation of dss.Store using
```
BenjaminPelletier, line 33: Let's update the package reference for completeness

pkg/rid/store/cockroach/store.go
```diff
@@ -25,23 +25,45 @@ var (
...
+			clock:     s.clock,
+		},
+	}, nil
 }
 
-// InTxnRetrier supplies a new repo, that will perform all of the DB accesses
+// Transact supplies a new repo, that will perform all of the DB accesses
 // in a Txn, and will retry any Txn's that fail due to retry-able errors
 // (typically contention).
 // Note: Currently the Newly supplied Repo *does not* support nested calls
```
BenjaminPelletier, line 63: This note is no longer applicable

pkg/rid/store/cockroach/subscriptions.go
```diff
@@ -21,16 +21,16 @@ const (
 	updateSubscriptionFields = "id, url, notification_index, cells, starts_at, ends_at, updated_at"
 )
 
-// SubscriptionStore is an implementation of the SubscriptionRepo for CRDB.
-type SubscriptionStore struct {
+// subscriptions is an implementation of the SubscriptionRepo for CRDB.
+type subscription struct {
```
BenjaminPelletier, line 25: Here too I think subscriptionRepo is a more appropriate data type name

steeling:
I think we have 2 competing PR's at this point between this one and https://github.com/interuss/dss/pull/327. Thoughts on how to move forward, or weight the pros/cons of each approach?

vosst:
> Marking a comment as unresolved to discuss the two approaches

Nice, thank you. I'll only find time to get back to this tomorrow, though.

pkg/rid/store/cockroach/store.go
```diff
@@ -25,23 +25,44 @@ var (
...
 }
 
-// InTxnRetrier supplies a new repo, that will perform all of the DB accesses
+// Transact supplies a new repo, that will perform all of the DB accesses
 // in a Txn, and will retry any Txn's that fail due to retry-able errors
 // (typically contention).
-// Note: Currently the Newly supplied Repo *does not* support nested calls
-// to InTxnRetrier.
-func (s *Store) InTxnRetrier(ctx context.Context, f func(repo repos.Repository) error) error {
+func (s *Store) Transact(ctx context.Context, f func(repo repos.Repository) error) error {
```
BenjaminPelletier, line 64: I believe we want this signature to be
```go
func (s *Store) Transact(ctx context.Context, f func(ctx context.Context, repo repos.Repository) error) error {
```
...because the transactor may want to adjust the context used by each transaction attempt.

vosst: Yup, that makes sense. I would propose to postpone until after this PR is merged.

vosst:
@steeling given our discussion on #327, are you fine with proceeding in merging this one?

pkg/rid/store/cockroach/store.go
```diff
@@ -25,23 +25,44 @@ var (
...
-	*ISAStore
-	*SubscriptionStore
+	db     *cockroach.DB
+	logger *zap.Logger
+	clock  clockwork.Clock
+}
 
-	db *cockroach.DB
+// Interact implements store.Interactor interface.
+func (s *Store) Interact(ctx context.Context) (repos.Repository, error) {
```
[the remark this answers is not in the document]
BenjaminPelletier: In general, there might be other data backing implementations that could return an error here.  For instance, a file-based data backing that needs to open a file.

PR #334, opened 2020-06-24
Add Constraint functionality to SCD

BenjaminPelletier:
This PR completes required SCD functionality by adding support for Constraints.  It includes a minimal set of tests to verify Constraint functionality, but future PRs should build additional tests for more complicated interactions and edge cases.

This PR also makes a tiny change to docker_e2e.sh to allow targeting of specific tests without affecting current behavior (when no additional arguments are passed to docker_e2e.sh), and documents this new capability in the testing README.


BenjaminPelletier:
Constraints require two new database tables, making this a different schema than we have currently.  Although this PR uses the current hack to verify database schema compatibility and fail fast with old database schemas, it would be preferable to first merge #328 or equivalent to improve our database schema detection and migration capabilities, and then update this PR with the new approach.

BenjaminPelletier:
@vosst I'm expecting to see and merge your upcoming PR for error handling before this PR is merged.  When that PR is merged, I intend to update this PR to reflect that approach.

pkg/scd/constraints_handler.go
```diff
@@ -2,29 +2,350 @@ package scd
...
 // DeleteConstraintReference deletes a single constraint ref for a given ID at
 // the specified version.
 func (a *Server) DeleteConstraintReference(ctx context.Context, req *scdpb.DeleteConstraintReferenceRequest) (*scdpb.ChangeConstraintReferenceResponse, error) {
-	return nil, dsserr.BadRequest("not yet implemented")
+	// Retrieve Constraint ID
+	idString := req.GetEntityuuid()
+	if idString == "" {
+		return nil, dsserr.BadRequest("missing Constraint ID")
+	}
+	id := scdmodels.ID(idString)
```
vosst, line 25: nit: You could fold this to:
```
id := scdmodels.ID(req.GetEntityUuid())
if len(id) == 0 {
  return ...
}
```

pkg/scd/constraints_handler.go
```diff
@@ -2,29 +2,350 @@ package scd
...
+	}
+
+	return response, nil
 }
 
 // GetConstraintReference returns a single constraint ref for the given ID.
 func (a *Server) GetConstraintReference(ctx context.Context, req *scdpb.GetConstraintReferenceRequest) (*scdpb.GetConstraintReferenceResponse, error) {
-	return nil, dsserr.BadRequest("not yet implemented")
+	id := scdmodels.ID(req.GetEntityuuid())
+	if id.Empty() {
```
vosst, line 107: Same nit for folding applies here.

pkg/scd/store/cockroach/constraints.go
```diff
@@ -0,0 +1,291 @@
...
+
+	constraintFieldsWithPrefix = strings.Join(
+		withPrefix[:], ",",
+	)
+}
+
+func (c *repo) fetchConstraints(ctx context.Context, q dsssql.Queryable, query string, args ...interface{}) ([]*scdmodels.Constraint, error) {
+	rows, err := q.QueryContext(ctx, query, args...)
+	if err != nil {
+		logging.Logger.Warn(fmt.Sprintf("fetchConstraints: error in QueryContext with %s", query))
```
vosst, line 54: I would prefer contextual logging here, i.e.:
```
logger := logging.WithValuesFromContext(ctx, logging.Logger)
logger.Warn("error in issueing db query", zap.String("query", query), zap.Error(err)
```
However, in general, we should avoid logging in library code and leave it to outer layers to log rich error messages.

pkg/scd/constraints_handler.go
BenjaminPelletier: Thanks, done.  Also, this was copy-pasted from somewhere else, so I made the change in the other places as well so the 9 places we do this are consistent.  I used id.Empty() to conform with the plurality of the other places and so that hypothetically there might be a slight performance advantage.

pkg/scd/constraints_handler.go
BenjaminPelletier: I'm not sure I follow on this one; seems like this is already folded per suggestion, with the only difference being `id.Empty()` instead of `len(id) == 0`

pkg/scd/store/cockroach/constraints.go
BenjaminPelletier: Ah yes, this was an oversight; that logging was for debugging and I had intended to remove it.  Now done.

cmds/grpc-backend/main.go
```diff
@@ -35,8 +35,8 @@ import (
 )
 
 const (
-	// The code at this version requires a major schema version equal to 2.
-	RequiredMajorSchemaVersion = "v2"
+	// The code at this version requires a major schema version equal to 3.
+	RequiredMajorSchemaVersion = "v3"
```
steeling, line 39: we shouldn't require this version here; it will break rid if we don't run a migration

cmds/grpc-backend/main.go
```diff
@@ -156,6 +154,8 @@ func RunGRPCServer(ctx context.Context, address string) error {
 		scopesValidators = auth.MergeOperationsAndScopesValidators(scopesValidators, scdServer.AuthScopes())
 	}
 
+	MustSupportSchema(ctx, store)
```
steeling, line 157: why move this?

cmds/grpc-backend/main.go
steeling: while scd is in development i'd only increment minor version numbres for an scd specific version, or not increment at all

pkg/scd/constraints_handler.go
steeling: instead of all of that, could we create the model, and the call `err := constraint.Validate()`

pkg/scd/store/cockroach/store.go
```diff
@@ -127,6 +127,29 @@ func (s *Store) Bootstrap(ctx context.Context) error {
...
+		altitude_upper REAL,
+		starts_at TIMESTAMPTZ,
+		ends_at TIMESTAMPTZ,
+		updated_at TIMESTAMPTZ NOT NULL,
+		INDEX owner_idx (owner),
+		INDEX starts_at_idx (starts_at),
+		INDEX ends_at_idx (ends_at),
+		CHECK (starts_at IS NULL OR ends_at IS NULL OR starts_at < ends_at)
+	);
+	CREATE TABLE IF NOT EXISTS scd_cells_constraints (
```
steeling, line 145: i think we should be using the new index/cells format leveraged in rid

pkg/scd/constraints_handler.go
BenjaminPelletier: @steeling could you elaborate?  This is a request to get a constraint, so we don't already have the constraint to `.Validate()` at this point.

cmds/grpc-backend/main.go
BenjaminPelletier: Like RID, SCD is in active use currently and we need database-binary consistency.  This PR is a breaking change for the SCD schema, so we need to make sure users who run this code against an older database are clearly alerted to the problem.  This approach mirrors what was done in b16976a to fix the similar problem introduced by 8f2687c and I agree that is not a particularly good solution which is why I [noted](https://github.com/interuss/dss/pull/334#issuecomment-648621753) that I intend to adopt the approach introduced in #328 and its successors before merging this PR.  However, we need this constraint functionality fairly soon, so we may have to use the b16976a approach in the interim while #328 and its successors are finished up.

cmds/grpc-backend/main.go
BenjaminPelletier: If SCD is run at this commit, it needs to fail if the database is at a lower schema version than the schema altered in this PR.  We previously accomplished this task with 8f2687c and this PR follows that pattern.  This statement needed to be moved because [the check](https://github.com/interuss/dss/blob/c90bee80069b1376bb45034eb0da16c17ecf0572/pkg/rid/cockroach/store.go#L192) must follow the SCD bootstrapping.  Again, I would prefer to wait for #328 and successors, but if that can't be finished in time, we'll need to take the existing approach to database schema validation.

pkg/rid/cockroach/store.go
```diff
@@ -182,6 +182,24 @@ func (s *Store) GetVersion(ctx context.Context) (string, error) {
 		return "v1.0.0", nil
 	}
 	// Version without cells joins table.
+
+	// We treat the existence of scd_constraints as running on the
+	// newest version, 3.0.0
+	const query3 = `
```
steeling, line 188: This is really polluting the api.. i know you're hoping to wait for the schema management PR, but we could accomplish this with 2 separate versions, and 2 separate calls to the DB if we decide not to

cmds/grpc-backend/main.go
steeling: if we do have to merge this pr before the schema management, let's have a rid version and an SCD version

pkg/scd/constraints_handler.go
steeling: ya this comment should be on the Put/Update api

pkg/scd/store/cockroach/store.go
BenjaminPelletier: Done, and yes, that makes things a lot better -- thanks!

cmds/grpc-backend/main.go
BenjaminPelletier: This PR has been updated to use the recently-developed migrations (in particular, applying them to SCD in #353).

pkg/rid/cockroach/store.go
BenjaminPelletier: This PR has been updated to use the recently-developed migrations (in particular, applying them to SCD in #353).

PR #336, opened 2020-07-06
Deployment Config for Schema Manager

charlie-pisuraj:
Right now it only has the Schema manager for the DefaultDB database, we can modify this deployment to also bootstrap the SCD database in the future


build/deploy/db-schemas/defaultdb.libsonnet
```diff
@@ -0,0 +1,12 @@
+{
+  data:{
```
steeling, line 2: is there a way to import a whole directory?

charlie-pisuraj: I looked and I can't find it

build/README.md
```diff
@@ -229,3 +229,9 @@ Pick a username and create an account:
 
 Then go to https://localhost:8080. You'll have to ignore the HTTPS certificate
 warning.
+
+## Updgrading Database Schemas
+
+All Schemas related files are in `deploy/db-schemas` directory, any changes you wish to make to the Database schema should be done in their respective Database folders. The files are applied in steps from 1 to N in the direction specified in the file's suffix.
+
+For the first ever run during the CRDB cluster initilization the db-manager will run once to bootstrap and bring up the database up to date. In order to upgrade existing clusters you will need to update the `desired_xyz_db_version` field in `main.jsonnet`, delete the existing db-manager job in your k8s cluster and redeploy the newly configured db-manager. It should automatically up/down grade your database schema to your desired version.
```
BenjaminPelletier, line 237: What would happen if a new participant started up their Kubernetes cluster for the first time when connecting to an existing CockroachDB cluster currently hosted by other participants?

Basically, I have these use cases in mind:
1) An individual user wants to start a brand new CRDB cluster + Kubernetes cluster
2) A first-in-region organization wants to start a brand new CRDB cluster (in a brand new Kubernetes cluster), and the CRDB cluster will be later be joined by other participants
3) An organization wants to start a brand new Kubernetes cluster that joins an existing CRDB cluster
4) An organization wants to migrate the schema of an existing CRDB cluster in an existing Kubernetes cluster

I want to make sure migration will not be run for use case 3 (thereby potentially breaking the system for all other participants), even if the joining organization accidentally had `shouldInit=true` or some other easy-to-miss incorrect configuration.

README.md
```diff
@@ -46,6 +46,10 @@ authentication. This backend talks directly to CockroachDB.
 Individual CockroachDB nodes hosting sharded data of the DAR. More information about CockroachDB
 [here](https://www.cockroachlabs.com/docs/stable/architecture/overview.html).
 
+### DB Manager
+
+Component responsible for bootstrapping the Database on first run when the CRDB Cluster is initialized for the first time. Only the first run is automatic to bootstrap the Database to the desired version. This component is also used to upgrade the Database Schemas, and if the cluster has already been bootstrapped with this, it must be invoked manually.
```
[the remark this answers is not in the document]
charlie-pisuraj: this tool will not "initialize" the database, that is responsibility goes to "cluster-init" job, and "first run" is determined by the fact that it runs the same time as "cluster-init" and that is only deployed once, when CRDB is brought up for the first time.

build/deploy/schema-manager.libsonnet
```diff
@@ -0,0 +1,51 @@
...
+  readOnly: false,
+  mountPath: '/db-schemas/defaultdb',
+};
+
+{
+  all(metadata): {
+    rid_schema: base.ConfigMap(metadata, 'db-rid-schema') {
+      data: defaultdb_schema.data
+    },
+    RIDSchemaManager: if metadata.cockroach.shouldInit then base.Job(metadata, 'rid-schema-manager') {
```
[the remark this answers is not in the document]
charlie-pisuraj: For this config it is a pre requisite because it is meant for the first run that bootstraps the database schema. I did this because we are going to remove the bootstrapping from GRPC backend

build/README.md
charlie-pisuraj: I'll add an assert to make sure that only shouldInit is true or Joining Cluster is populated but never both. This will prevent use case 2,3,4 to not run the db-manager

PR #337, opened 2020-07-06
Remove Bootstrapping Functionality from GRPC

charlie-pisuraj:
GRPC will now try to get the DB Schema Version and fail if the DB was not bootstrapped with the Schema Manager

### DIFF
--- a/pkg/cockroach/cockroach.go
+++ b/pkg/cockroach/cockroach.go
@@ -68,14 +68,14 @@ func BuildURI(params map[string]string) (string, error) {
 
 // GetVersion returns the Schema Version of the requested DB Name
 func GetVersion(ctx context.Context, db *DB, dbName string) (string, error) {
-	query := fmt.Sprintf(`
+	const query = `
 		SELECT EXISTS (
   		SELECT *
 		  FROM information_schema.tables 
 		WHERE table_name = 'schema_versions'
-		AND table_catalog = '%s'
-   )`, dbName)
-	row := db.QueryRowContext(ctx, query)
+		AND table_catalog = $1
+   	)`
+	row := db.QueryRowContext(ctx, query, dbName)
 	var ret bool
 	if err := row.Scan(&ret); err != nil {
 		return "", err
--- a/pkg/cockroach/cockroach.go
+++ b/pkg/cockroach/cockroach.go
@@ -1,6 +1,7 @@
 package cockroach
 
 import (
+	"context"
 	"database/sql"
 	"errors"
 	"fmt"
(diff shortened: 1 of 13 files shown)


pkg/rid/cockroach/store.go
```diff
@@ -161,44 +162,35 @@ func (s *Store) Bootstrap(ctx context.Context) error {
...
-	// We treat the existence of cells_subscriptions as running on the initial
-	// version, 1.0.0
+	// when we have multiple databases we will have to substitute
+	// defaultdb for other database names, this field is current useless
 	const query = `
 		SELECT EXISTS (
   		SELECT *
 		  FROM information_schema.tables 
-   		WHERE table_name = 'cells_subscriptions'
+		WHERE table_name = 'schema_versions'
```
charlie-pisuraj, line 174: I'm debating if I should add a fall back method for determining the DB Version that was bootstrapped before Schema Manager is added. This is not hard, I would simply check the for the existence of the tables that should exists and go on from there.

BenjaminPelletier: I think it's cleaner to just require the table (thereby requiring migrations to be run).  I think it's ok to have a one-time breaking change requiring users to migrate since that will be the process going foward.

pkg/rid/cockroach/store.go
```diff
@@ -161,44 +162,35 @@ func (s *Store) Bootstrap(ctx context.Context) error {
 	return err
 }
 
-// GetVersion returns the current semver of the schema
+// GetVersion returns the Version string for the Database.
+// If the DB was is not bootstrapped using the schema manager we throw and error
 func (s *Store) GetVersion(ctx context.Context) (string, error) {
```
BenjaminPelletier, line 167: Can we use the same function from both db-manager and inside the gRPC backend?

pkg/rid/cockroach/store.go
steeling: if we're doing a breaking change, let's bump the RequiredMajorSchemaVersion, and add a migration that increments that as well.

pkg/rid/cockroach/store.go
charlie-pisuraj: I'm going to have to refactor the db-manager, I can do this in the next PR

pkg/rid/store/cockroach/store.go
```diff
@@ -109,125 +110,45 @@ func NewStore(db *cockroach.DB, logger *zap.Logger) *Store {
...
+	DELETE FROM identification_service_areas WHERE id IS NOT NULL;`
 
 	_, err := s.db.ExecContext(ctx, query)
 	return err
 }
 
-// GetVersion returns the current semver of the schema
+// GetVersion returns the Version string for the Database.
+// If the DB was is not bootstrapped using the schema manager we throw and error
 func (s *Store) GetVersion(ctx context.Context) (string, error) {
```
BenjaminPelletier, line 125: We know we need to support different databases, so let's just accept the database name as a string argument here.

pkg/rid/store/cockroach/store.go
```diff
@@ -109,125 +110,45 @@ func NewStore(db *cockroach.DB, logger *zap.Logger) *Store {
...
-	if err != nil {
+	if err := row.Scan(&ret); err != nil {
 		return "", err
 	}
-	if ret {
-		// Base version
-		return "v1.0.0", nil
+	if !ret {
+		// Database has not been bootstrapped using DB Schema Manager
+		return "", fmt.Errorf("DB has not been bootstrapped with Schema Manager")
```
BenjaminPelletier, line 142: Let's add some additional information here about how the problem can be fixed as the most likely encounter with this error message will be a sysadmin troubleshooting why his installation doesn't work.  Pointing to the appropriate README would be great.  And let's remember that when the database name is passed as a parameter, this message may mean that the initialization simply hasn't occurred for that database specifically.

pkg/cockroach/cockroach.go
```diff
@@ -64,3 +65,33 @@ func BuildURI(params map[string]string) (string, error) {
...
+
+// GetVersion returns the Schema Version of the requested DB Name
+func GetVersion(ctx context.Context, db *DB, dbName string) (string, error) {
+	query := fmt.Sprintf(`
+		SELECT EXISTS (
+  		SELECT *
+		  FROM information_schema.tables 
+		WHERE table_name = 'schema_versions'
+		AND table_catalog = '%s'
+   )`, dbName)
```
BenjaminPelletier, line 77: nit: We should probably pass the database name in using SQL parameters (i.e., [$1](https://github.com/interuss/dss/blob/0f151b572a5711e999f7283e55b4b1f4c3189a65/pkg/rid/store/cockroach/identification_service_area.go#L87))

PR #339, opened 2020-07-08
Add flag to GRPC Backend for self identification

Dos-Ph:
Currently there are is no flag in the GRPC for self identification, we can reuse the "Locality" variable that we use for deploying CockroachDB to be the value for this self identification flag


charlie-pisuraj:
it looks good to me. I think @Dos-Ph can rebase the branch and merge it

PR #340, opened 2020-07-09
Add response time test for a multi-volume operation creation request

BenjaminPelletier:


BenjaminPelletier:
As this test does not currently reproduce an issue, I'll close this PR without merging.  In the future, we will probably want to add a load test with a large volume of requests, possibly in parallel.

PR #343, opened 2020-07-13
Improve deployment documentation

BenjaminPelletier:
This PR adds a considerable level of detail to our deployment documentation based on lessons learned from following the deployment procedure, including documenting variable values for example main.jsonnet & spec.json files in the deployment README.  It also exposes some configuration parameters (JWKS and public key files) in the example deployment configuration.

I welcome suggestions on how to satisfy the current `tk show --dangerous-allow-redirect build/deploy/examples/minimum` presubmit check.  It seems like minimum.jsonnet can't actually have a minimum set of values to generate a reasonable deployment configuration, so my baseline suggestion is that we should remove that presubmit check.


build/README.md
```diff
@@ -14,188 +16,380 @@ To run a local DSS instance for testing, evaluation, or development, see [dev/st
...
+      - Under Frontend for Details, the Certificate column for HTTPS protocol
+        will have an icon next to it which will change to a green checkmark when
+        provisioning is complete.
+      - Click on the certificate link to see provisioning progress.
+      - If everything indicates OK and you still receive a cipher mismatch error
+        message when attempting to visit /healthy, wait an additional 5 minutes
+        before attempting to troubleshoot further.
+
+1.  If joining an existing cluster, share your CRDB node addresses with the
+    operator of the existing cluster.  They will add these node addresses to
```
BenjaminPelletier, line 341: @steeling or @charlie-pisuraj please verify this is what should happen -- basically, when a new DSS instance joins a DSS Region, the existing instances should be updated to point to the new instance's CRDB nodes, right?  Assuming so, is this the right way to do that?

steeling: they'll also need to update the certificates provided to their cockroachdb backends to contain both certs

build/README.md
```diff
@@ -14,188 +16,380 @@ To run a local DSS instance for testing, evaluation, or development, see [dev/st
...
-1.  Use the `make-certs.py` script in this directory to create certificates for
+your chosen provider.
+
+If you can augment this documentation with specifics for another cloud provider,
+a PR to that effect would be greatly appreciated.
+
+1.  Create a new Kubernetes cluster. We recommend a new cluster for each DSS
+    instance.  A reasonable cluster name might be `dss-us-prod`,
+    `dss-ca-staging`, `dss-mx-integration`, etc.  The name of this cluster
+    will be referred to as CLUSTER_CONTEXT for the remainder of this document.
```
steeling, line 109: The cluster context is likely not the name of the cluster, but some longer key generated by kubectl

build/README.md
```diff
@@ -14,188 +16,380 @@ To run a local DSS instance for testing, evaluation, or development, see [dev/st
...
-      DNS entries with a common hostname suffix, ie: `*.db.interuss.com`
-
-1.  Use the `make-certs.py` script in this directory to create certificates for
+your chosen provider.
+
+If you can augment this documentation with specifics for another cloud provider,
+a PR to that effect would be greatly appreciated.
+
+1.  Create a new Kubernetes cluster. We recommend a new cluster for each DSS
+    instance.  A reasonable cluster name might be `dss-us-prod`,
```
steeling, line 107: nit: why are we providing naming suggestions? If we are we might want to put more thought into the suggestions themselves. I would likely recommend including the availability zone in the name as well

build/README.md
BenjaminPelletier: Yep, as described on line 196 above?

build/README.md
BenjaminPelletier: To illustrate good practice and imply a set of rules a reasonable name should follow without actually listing out those rules -- we want someone to be able to follow these instructions with minimal interpretation or additional knowledge necessary and still end up with a good DSS instance.

Including zone sounds good; would you recommend `dss-us-us-east4-a-prod` or `dss-us-east4-a-prod` or `dss-us-prod-us-east-4a` or something else?

build/README.md
BenjaminPelletier: Done; thanks.

build/README.md
steeling: yup, missed that

build/README.md
steeling: anything unique to the AZ sounds good to me. I'm partial to short names like `us-e4a`, resulting in `dss-prod-us-e4a`

BenjaminPelletier: Sounds good; done, thanks.

PR #348, opened 2020-07-15
Feature/create db if not exists

charlie-pisuraj:


cmds/db-manager/main.go
```diff
@@ -166,44 +173,51 @@ func intAbs(x int) int {
...
+func getCurrentDBVersion(crdbURI string, database string) (string, error) {
+	crdb, err := dssCockroach.Dial(crdbURI)
+	defer func() {
+		crdb.Close()
+	}()
+	if err != nil {
+		return "", fmt.Errorf("Failed to dial CRDB while getting DB version: %v", err)
+	}
+	version, err := dssCockroach.GetVersion(context.Background(), crdb, database)
+	if err != nil && version == "" {
```
BenjaminPelletier, line 218: If `err != nil`, it shouldn't matter what version is; we can remove `&& version == ""`.

charlie-pisuraj: so I had to modify the GetVersion method itself, so that if the schema_versions table doesn't exists it should return v0.0.0 by default unless the it is a different error

pkg/cockroach/cockroach.go
```diff
@@ -70,24 +70,25 @@ func BuildURI(params map[string]string) (string, error) {
...
+	`
 	row := db.QueryRowContext(ctx, query, dbName)
 	var ret bool
 	if err := row.Scan(&ret); err != nil {
 		return "", err
 	}
 	if !ret {
 		// Database has not been bootstrapped using DB Schema Manager
-		return "", fmt.Errorf("RID DB has not been bootstrapped with Schema Manager, Please check https://github.com/interuss/dss/tree/master/build#updgrading-database-schemas")
+		return "v0.0.0", fmt.Errorf("%s has not been bootstrapped with Schema Manager, Please check https://github.com/interuss/dss/tree/master/build#updgrading-database-schemas", dbName)
```
charlie-pisuraj, line 86: Right here

cmds/db-manager/main.go
BenjaminPelletier: @vosst could you take a look at this?  IIUC, idiomatic Go returns either err or value, but never both at the same time.

PR #349, opened 2020-07-15
Modify CRDB tables to have a "writer" column

Dos-Ph:
Due to the nature of the how the DSS is setup with multiple companies owning a piece of the infrastructure and therefore multiple "GRPC Backend" server running, we've decided to make each instance be responsible for cleaning up its own garbage, therefore with every write to the the DB we need it to also put its name down as the writer so that it can know which row its responsible for GC.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,7 +99,7 @@ test:
 .PHONY: test-cockroach
 test-cockroach: cleanup-test-cockroach
 	@docker run -d --name dss-crdb-for-testing -p 26257:26257 -p 8080:8080  cockroachdb/cockroach:v20.1.1 start --insecure > /dev/null
-	go run ./cmds/db-manager/main.go --schemas_dir ./build/deploy/db-schemas/defaultdb --db_version v3.0.0 --cockroach_host localhost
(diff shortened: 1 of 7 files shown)


steeling:
> These looks good to me, my 1 debate is if we should force the writer column to be NOT NULL, which would make it a breaking change which means it will be a major version. However, we could always keep it as a minor change and on the next major change to V4 we can enforce it to NOT NULL

+1 to the latter; we should get into the habit of making changes this way when convenient.

Makefile
```diff
@@ -99,7 +99,7 @@ test:
 .PHONY: test-cockroach
 test-cockroach: cleanup-test-cockroach
 	@docker run -d --name dss-crdb-for-testing -p 26257:26257 -p 8080:8080  cockroachdb/cockroach:v20.1.1 start --insecure > /dev/null
-	go run ./cmds/db-manager/main.go --schemas_dir ./build/deploy/db-schemas/defaultdb --db_version v3.0.0 --cockroach_host localhost
+	go run ./cmds/db-manager/main.go --schemas_dir ./build/deploy/db-schemas/defaultdb --db_version v3.1.0 --cockroach_host localhost
```
vosst, line 102: Not for this PR, but we should probably keep track of the db version in a central file.

BenjaminPelletier: Added #351

PR #353, opened 2020-07-15
Split SCD data into dedicated `scd` database

BenjaminPelletier:
This PR moves SCD data into its own database (named "scd") and introduces migrations to that database.

One consequence of the approach here is that a single CRDB connection is no longer shared between RID & SCD.  We could share a single connection, but all the SQL statements would need to be rewritten to prefix table names with the correct database.  This probably seems worthwhile, but the RID database should be changed from "defaultdb" to "rid" before adding all those database name prefixes, and this is better-suited to a PR separate from this one.

The PR also pipes the enableScd grpc-backend option through deployment configurations.

### DIFF
--- a/build/deploy/schema-manager.libsonnet
+++ b/build/deploy/schema-manager.libsonnet
@@ -1,6 +1,7 @@
 local base = import 'base.libsonnet';
 local volumes = import 'volumes.libsonnet';
 local defaultdb_schema = import "db-schemas/defaultdb.libsonnet";
+local scd_schema = import "db-schemas/scd.libsonnet";
 
 local rid_schema_vol = {
   name: 'db-rid-schema',
@@ -15,12 +16,28 @@ local rid_schema_mount = {
   mountPath: '/db-schemas/defaultdb',
 };
 
+local scd_schema_vol = {
+  name: 'db-scd-schema',
+  configMap: {
+    defaultMode: 420,
+    name: 'db-scd-schema',
+  },
+};
+local scd_schema_mount = {
+  name: 'db-scd-schema',
+  readOnly: false,
+  mountPath: '/db-schemas/scd',
+};
+
 {
   all(metadata): {
     assert metadata.cockroach.shouldInit == true && metadata.cockroach.JoinExisting == [] : "If shouldInit is True, JoinExisiting should be empty",
     rid_schema: base.ConfigMap(metadata, 'db-rid-schema') {
       data: defaultdb_schema.data
     },
+    scd_schema: if metadata.enableScd then base.ConfigMap(metadata, 'db-scd-schema') {
+      data: scd_schema.data
+    } else null,
     RIDSchemaManager: if metadata.cockroach.shouldInit then base.Job(metadata, 'rid-schema-manager') {
       spec+: {
         template+: {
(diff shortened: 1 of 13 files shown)


build/deploy/schema-manager.libsonnet
```diff
@@ -48,5 +65,32 @@ local rid_schema_mount = {
...
+              image: metadata.schema_manager.image,
+              args_:: {
+                cockroach_host: 'cockroachdb-balanced.' + metadata.namespace,
+                cockroach_port: metadata.cockroach.grpc_port,
+                cockroach_ssl_mode: 'verify-full',
+                cockroach_user: 'root',
+                cockroach_ssl_dir: '/cockroach/cockroach-certs',
+                db_version: metadata.schema_manager.desired_scd_db_version,
+                schemas_dir: scd_schema_mount.mountPath,
+
```
charlie-pisuraj, line 87: nit: extra blank line

pkg/cockroach/cockroach.go
```diff
@@ -83,10 +83,10 @@ func GetVersion(ctx context.Context, db *DB, dbName string) (string, error) {
 	}
 	if !ret {
 		// Database has not been bootstrapped using DB Schema Manager
-		return "v0.0.0", fmt.Errorf("%s has not been bootstrapped with Schema Manager, Please check https://github.com/interuss/dss/tree/master/build#updgrading-database-schemas", dbName)
+		return "v0.0.0", nil
```
charlie-pisuraj, line 86: I was worried about this above, but with this it looks good. thanks

pkg/scd/store/cockroach/store.go
```diff
@@ -67,67 +67,8 @@ func (s *Store) Close() error {
...
-		INDEX cell_id_idx (cell_id),
-		INDEX operation_id_idx (operation_id)
-	);
-	`
-	_, err := s.db.ExecContext(ctx, query)
-	return err
+// GetVersion returns the Version string for the Database.
+// If the DB was is not bootstrapped using the schema manager we throw and error
+func (s *Store) GetVersion(ctx context.Context) (string, error) {
+	return cockroach.GetVersion(ctx, s.db, "scd")
```
charlie-pisuraj, line 73: I think it might be good to add the DB name into the struct so that we won't have to have it hardcoded everywhere, when we do change them like defaultdb to rid we won't easily forget it. but this is extremely minor

test/docker_e2e.sh
```diff
@@ -78,6 +78,16 @@ docker run --rm --name rid-db-manager \
...
 	--cockroach_host crdb
 
+sleep 1
+echo "Bootstrapping SCD Database tables"
+docker run --rm --name scd-db-manager \
+	--link dss-crdb-for-debugging:crdb \
+	-v $(pwd)/build/deploy/db-schemas/scd:/db-schemas/scd \
+	local-db-manager \
+	--schemas_dir db-schemas/scd \
+	--db_version v1.0.0 \
```
charlie-pisuraj, line 88: Keeping track of these things are annoying for our test infra. I'll add an issue to add flag to always upgrade to latest in test

build/deploy/schema-manager.libsonnet
BenjaminPelletier: I'm following [general best practices](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file) I think.

pkg/scd/store/cockroach/store.go
BenjaminPelletier: Named constants are better than magic values -- done; thanks.

test/docker_e2e.sh
BenjaminPelletier: Added #354

PR #367, opened 2020-07-21
Update business object conversions

### LINKED ISSUE #360: SCD Subscription with no `time_start` crashes backend
When a Subscription with no `time_start` is submitted, it crashes the gRPC backend:

```
panic: runtime error: invalid memory address or nil pointer dereference 
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa79019] 
goroutine 613 [running]: 
github.com/interuss/dss/pkg/scd/models.(*Subscription).AdjustTimeRange(0xc0002a17c0, 0xbfbc97f482c5293d, 0x579a6f64475, 0x12cd920, 0xc0002a17c0, 0x0, 0x0) 
	/app/pkg/scd/models/subscriptions.go:113 +0x209 
github.com/interuss/dss/pkg/scd.(*Server).PutSubscription(0xc0000b4240, 0xd751e0, 0xc0002e2870, 0xc0005a4780, 0xc0000b4240, 0x0, 0xc00094c9a0) 
	/app/pkg/scd/subscriptions_handler.go:77 +0x3f3 
github.com/interuss/dss/pkg/api/v1/scdpb._UTMAPIUSSDSSAndUSSUSSService_PutSubscription_Handler.func1(0xd751e0, 0xc0002e2870, 0xbc8640, 0xc0005a4780, 0xc00045cbf0, 0x2, 0x2, 0x0) 
	/app/pkg/api/v1/scdpb/scd.pb.go:5681 +0x86 
```

BenjaminPelletier:
Currently, all geo business object conversions to and from subsystem-specific API objects are glommed into pkg/models/geo.go.  This PR splits these conversions into their own rid_conversions.go and scd_conversions.go files.  It also increases protection against nil pointer dereference for unspecified options during conversion, and hopefully resolves #360.


vosst:
Seems like we could as well factor the business object conversion into `pkg/rid/modes` and `pkg/scd/models`, respectively, thus slimming down the common (geo-)object package. WDYT?

BenjaminPelletier:
> Seems like we could as well factor the business object conversion into `pkg/rid/modes` and `pkg/scd/models`, respectively, thus slimming down the common (geo-)object package. WDYT?

Yeah, the history here is that conversions were initially methods on the API objects, defined in pkg/api/... along side the generated code, but this caused a problem with a build system that regenerated the code itself (rather than using the version in the repo).  Sean fixed this by moving all the conversion code into pkg/models/geo.go.  This PR splits the conversion code out into parallel files along side geo.go.  It may make sense to move those files somewhere in a future PR, but I think it probably wouldn't be pkg/rid/models.  The conversion code converts between objects in, e.g., pkg/api/v1/ridpb and pkg/models.  None of the models in pkg/rid/models are involved, so it seems like we wouldn't want the conversion code to live there.  It may make sense to have pkg/conversion or pkg/models/conversion, but it seems like pkg/models should be ok for now.

PR #368, opened 2020-07-21
Fix deployment and deployment-prober issues

BenjaminPelletier:
This PR fixes a few more bugs found while redeploying a Kubernetes-driven DSS instance.  Along with that, running the prober on a fully-deployed instance currently has some issues (especially around token acquisition and usage) and this PR addresses those as well, mainly by adding explicit scopes to all requests within the prober test suite.

Also, it removes a check that causes an Operation emplacement to fail when invalid OVNs are provided -- invalid OVNs should be ignored because it doesn't hurt for a USS to think there are phantom Operations that don't exist when deconflicting, and the planning area of the USS might not perfectly line up with the cell-gridded area of the Operation they are emplacing (so they may have additional OVNs for Operations not relevant from a DSS perspective).


pkg/scd/store/cockroach/operations.go
```diff
@@ -358,9 +358,6 @@ func (s *repo) UpsertOperation(ctx context.Context, operation *scdmodels.Operati
 			}
 			delete(keyIdx, op.OVN)
 		}
-		if len(keyIdx) != 0 {
```
vosst, line 361: Mind elaborating a little why we can drop this check?

BenjaminPelletier: Sure -- second part of the PR description:

Also, it removes a check that causes an Operation emplacement to fail when invalid OVNs are provided -- invalid OVNs should be ignored because it doesn't hurt for a USS to think there are phantom Operations that don't exist when deconflicting, and the planning area of the USS might not perfectly line up with the cell-gridded area of the Operation they are emplacing (so they may have additional OVNs for Operations not relevant from a DSS perspective).

Basically, requiring an exact match on OVNs (specifically, no invalid OVNs) may break some workflows (specifically, a persistent planning area that is a superset of an individual operation area) and isn't required for safety.

vosst: ACK and thx.

PR #369, opened 2020-07-22
[build] Generate code using docker container

### LINKED ISSUE #352: Add docker container to produce auto-generated code
We auto-generate code in a few different places and store that generated code in the repo.  Currently, the autogeneration commands use the system configuration of the user running the autogeneration.  Instead, we should create a docker image to autogenerate the code in a reproducible way.

_Originally posted by @vosst in https://github.com/interuss/dss/pull/350/files_

vosst:
Fixes #352


pkg/api/v1/auxpb/aux_service.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.22.0
-// 	protoc        v3.11.4
```
vosst, line 4: This should be fine, and upgrading to `3.11.2` in alpine would require using `edge`.

Makefile
```diff
@@ -41,42 +43,72 @@ lint:
 	docker run --rm -v $(CURDIR):/dss -w /dss golangci/golangci-lint:v1.26.0 golangci-lint run --timeout 5m -v --disable-all  -E staticcheck --skip-dirs '^cmds/http-gateway,^pkg/logging'
 
 pkg/api/v1/ridpb/rid.pb.go: pkg/api/v1/ridpb/rid.proto
-	protoc -I/usr/local/include -I.   -I$(GOPATH)/src   -I$(GOPATH)/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@v1.14.3/third_party/googleapis   --go_out=plugins=grpc:. $<
+	docker run -v$(CURDIR):/src:delegated -w /src $(GENERATOR_TAG) protoc \
```
BenjaminPelletier, line 46: I think it would probably be good to specify an explicit image version for protoc in case there's a breaking change (same thing in general for other docker runs that don't involve images we built)

docker/generator/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.14.3-alpine
```
BenjaminPelletier, line 1: nit: We have a bunch of docker artifacts that aren't in this docker subfolder; could we put this `generator` folder under `build`?

Makefile
```diff
@@ -41,42 +43,72 @@ lint:
...
-		-out $@ \
 		-tag dss \
 		-indent 2 \
-		-package scdpb
+		-package scdpb > $@
+
+generator:
+	docker build --rm -t $(GENERATOR_TAG) docker/generator
 
 .PHONY: install-proto-generation
```
BenjaminPelletier, line 113: Should we remove this target now?

Makefile
vosst: The image tag already is `$(GENERATOR_TAG)`, just invoking the `protoc` executable in that image (that we have built).

Issue #314, opened 2020-06-02
500 (Internal Server Error) on second Subscription update 


arrcher:
**Context**
master branch, images are built locally

**Scenario**
1. Clean database
2. Send the subscription the first time
```
PUT http://localhost:8082/dss/v1/subscriptions/b61a6450-db42-4d0d-91f2-7c1334eda399
{
  "extents": {
    "volume": {
      "outline_polygon": {
        "vertices": [
          { "lng": -91.49723052978516, "lat": 41.70085834502109 },
          { "lng": -91.50341033935547, "lat": 41.6770148220322 },
          { "lng": -91.47989273071289, "lat": 41.67509157220958 },
          { "lng": -91.4663314819336, "lat": 41.69329603398001 },
          { "lng": -91.49723052978516, "lat": 41.70085834502109 }
          ]
      },
      "altitude_upper": {"units": "M", "reference": "W84", "value": 764.79037},
      "altitude_lower": {"units": "M", "reference": "W84", "value": 23.24352}
    },
    "time_start": {"value": "2020-06-03T01:30:00.000Z", "format": "RFC3339"},
    "time_end": {"value": "2020-06-03T22:15:00.000Z", "format": "RFC3339"}
  },
  "old_version": 0,
  "uss_base_url": "http://localhost:12012/services/uss/public/uss/v1/",
  "notify_for_constraints": true
}
```
The server responds with 
```
{
  "constraints": [],
  "operations": [],
  "subscription": {
    "dependent_operations": [],
    "id": "b61a6450-db42-4d0d-91f2-7c1334eda399",
    "implicit_subscription": false,
    "notification_index": 0,
    "notify_for_constraints": true,
    "notify_for_operations": false,
    "time_end": {"format": "RFC3339", "value": "2020-06-03T22:15:00Z"},
    "time_start": {"format": "RFC3339", "value": "2020-06-03T01:30:00Z"},
    "uss_base_url": "http://localhost:12012/services/uss/public/uss/v1/",
    "version": 1
  }
}
```
3. Now we send the update request with a modified time_start (+1h)
```
{
  ...
    "time_start": {"value": "2020-06-03T02:30:00.000Z", "format": "RFC3339"},
  ...
  "old_version": 1,
  ...
}
```
The server responses with `500 Internal Server Error`

Here is the log from dss_sandbox_grpc-backend-for-debugging_1
```
[...]

PR #375, opened 2020-07-23
Improve prober tests

BenjaminPelletier:
Currently, our prober tests may negatively affect long-lived production instances on which they are run.  They use random IDs for all resources, so test failures may orphan those resources.  The IDs used cannot easily be distinguished from real IDs, so it may be confusing to debug a real-world issue affected by a resource deposited by a past test.  Tests can sometimes fail if a resource was left behind from a previous (perhaps incomplete) test run.

This PR addresses all of the above issues and adds documentation on guidelines to follow to avoid these problems in the future.  After this PR, all tests should pass or skip for any current DSS deployment regardless of past possibly-incomplete test runs.


monitoring/prober/rid/test_isa_expiry.py
```diff
@@ -7,14 +7,29 @@
 from . import common
 from .common import SCOPE_READ, SCOPE_WRITE
 
+ISA_ID = '00000098-ba6d-4c20-a575-6e412e000000'
```
vosst, line 10: I would prefer if we introduced a pool of well-known test IDs here, and pick an ID at random at the beginning of each test run (for all parts that need an ID).

BenjaminPelletier: What I'm thinking we get from defining them individual in files is that if someone sees, e.g., "00000098-ba6d-4c20-a575-6e412e000000" somewhere in their database, they can quickly search the codebase and find out what test caused the resource with that test ID to be injected.  If we choose randomly from a pool of well-known IDs, it would be harder to track down the reason an artifact was left behind in the database.  I agree that always using the same IDs doesn't exercise the system quite as much (we would never have discovered #314, for instance), but it seems like the tradeoff is worth it.  What were the benefits you wanted to capture from having a pool of well-known test IDs?

vosst: I think maintaining a pool of random IDs would allow for running multiple prober test instances in parallel. With only one ID, we would run the risk of clashing tests (not the end of the world). More to this argument, some sort of `fuzzing` usually helps a lot in uncovering more elaborate issues faster (as you pointed out already).

I feel like maintaining a set of IDs (in a file, one ID per line) would enable both the tests and GC-infrastructure to use the IDs and semi-atuomatically clean up, too, even in the case of failing tests.

BenjaminPelletier: Could you elaborate on how the one-per-line IDs in a file would make multiple prober test instances in parallel possible?  I could imagine a function that generates a random UUID and then zeroes the first and last 6 digits, but that wouldn't need a file and would be prone to permanently leaving artifacts behind (even upon a subsequent successful test run).  If a file were used, we would need a lot of IDs in it to avoid the [birthday problem](https://en.wikipedia.org/wiki/Birthday_problem) unless IDs were drawn sequentially, which would disable multiple parallel test instances.  With that many IDs, I'm not sure it would be reasonable to attempt to clean them all up on every test run.  Also, the method of cleanup is dependent on the object type represented by the ID, so cleaning up everything at the very end would necessitate keeping a type registry which I'm not sure how we'd do.

It seems like putting test sequences in test classes would be the best intermediate suggestion, as test classes can have a destructor that cleans up after themselves.  But I'm still not super excited about that approach with random UUIDs because that will leave behind artifacts that won't be cleaned up next test run if something (like the cleanup phase) fails.

Here are benefits I think we may be interested in:
1. Make sure a successful test suite run cleans any artifacts from previously-failed test suite runs
1. Be able to identify which test(s) did not clean up an artifact
1. Be able to differentiate test resource IDs from real resource IDs at a glance
1. Run multiple prober instances in parallel
1. Explore a wide space of UUID values
1. Clean up artifacts by the end of the test suite run

I think the current PR approach achieves 1, 2, & 3, does not achieve 4 & 5, and achieves 6 as well as is reasonably possible.  I'm not sure how drawing IDs from a fixed pool (like a file containing a list of IDs) achieves any of the above except 3, and maybe 5 slightly (but not much) better than the current PR approach.  Drawing IDs from a function (like one that generates a random ID and then zeros the first and last 6 digits) achieves 3, 4, and 5, and 6 in the same way as the current PR, but does not achieve 1 or 2.  I think 1 & 2 are currently more valuable than 4 & 5 which is why I have the PR as-is rather than drawing from a function.  I think I don't yet fully understand the pool-of-IDs approach as my visualization of it doesn't seem to check as many boxes as either alternative.

Perhaps it might be best to consider a different method of supplying IDs in a follow-up PR?

vosst: Works for me.

PR #380, opened 2020-07-31
Fix SQL statements

charlie-pisuraj:
Constraints name must be unique and applying the constraints to upgrade from old V2 where bootstrap scripts was in GRPC we needed to rename the constraints the make the run valid

### DIFF
--- a/build/deploy/db_schemas/defaultdb/000003_migrate_data_to_index.up.sql
+++ b/build/deploy/db_schemas/defaultdb/000003_migrate_data_to_index.up.sql
@@ -27,3 +27,10 @@ WHERE subs.id = compact_sub_cells.subscription_id
     AND cells IS NULL;
 
 COMMIT;
+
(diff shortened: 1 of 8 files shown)


build/deploy/db_schemas/defaultdb/000003_migrate_data_to_index.up.sql
```diff
@@ -26,4 +26,11 @@ FROM compact_sub_cells
 WHERE subs.id = compact_sub_cells.subscription_id
     AND cells IS NULL;
 
-COMMIT;
\ No newline at end of file
+COMMIT;
+
+ALTER TABLE identification_service_areas ALTER COLUMN cells SET NOT NULL;
```
BenjaminPelletier, line 31: Do these statements not need a trailing COMMIT?

charlie-pisuraj: No, they will not need a COMMIT statement as they are not moving any data around and therefore can be done outside of a transaction.

Issue #379, opened 2020-07-30
Ensure detailed error message availability for 5xx errors


BenjaminPelletier:
Currently, usually only the lowest-level error message is present in the logs (and the response, when `DSS_ERRORS_OBFUSCATE_INTERNAL_ERRORS` is false).  When internal server errors occur, the detailed reason for those errors should be accessible to administrators, preferably with a full stack trace or similar, so that the problem can be identified and resolved more quickly and effectively.

BenjaminPelletier:
It seems like a good approach here would be to use [Palantir's stacktrace](https://github.com/palantir/stacktrace), change every `return ..., err` (and similar) statement to `return ..., stacktrace.Propagate(err, "Couldn't do the thing we're attempting to do in this context")`, and make sure the stack trace is printed in the gRPC backend logs.

PR #382, opened 2020-08-03
Add detailed error message infrastructure

BenjaminPelletier:
This PR begins to address #379 by:

1. Replacing all of our explicit Internal errors with `stacktrace` messages and propagations
1. Adding an error ID so a particular client-facing error message can be connected to backend logs quickly and easily, even if error message details are obfuscated
1. Fixing a number of issues with input validation and implicit Subscriptions exposed by the work above

In future PRs, we can add more `stacktrace.Propagate` statements to improve visibility into error causes, but that will be a large task and this PR aims to be a targeted introduction to the work.

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -36,13 +36,12 @@ func init() {
 	}
 }
 
-func makeErrId() string {
-	errUuid, err := uuid.NewRandom()
+func MakeErrID() string {
+	errUUID, err := uuid.NewRandom()
 	if err == nil {
-		return errUuid.String()
-	} else {
-		return fmt.Sprintf("<error ID could not be constructed: %s>", err)
+		return errUUID.String()
 	}
+	return fmt.Sprintf("<error ID could not be constructed: %s>", err)
 }
 
 // Interceptor returns a grpc.UnaryServerInterceptor that inspects outgoing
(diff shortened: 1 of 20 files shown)


pkg/errors/errors.go
```diff
@@ -35,27 +36,53 @@ func init() {
...
-			logger.Error("encountered internal Status error during unary server call",
-				zap.String("method", info.FullMethod),
-				zap.Stringer("code", statusErr.Code()),
-				zap.String("message", statusErr.Message()),
-				zap.Any("details", statusErr.Details()),
-				zap.Error(err))
-			if statusErr.Code() == codes.Internal && obfuscateInternalErrors {
-				err = errInternal
+
+		if err != nil {
```
vosst, line 54: Please return early, helps with readability.

pkg/errors/errors.go
```diff
@@ -35,27 +36,53 @@ func init() {
...
+			// Separate the root cause from the stacktrace wrapping.
+			trace := err.Error()
+			rootErr := stacktrace.RootCause(err)
+
+			statusErr, ok := status.FromError(rootErr)
+			switch {
+			case !ok:
+				errID := MakeErrID()
+				logger.Error(
+					fmt.Sprintf("encountered non-Status error %s during unary server call", errID),
```
vosst, line 64: `logger.Errorf` should give you the same functionality.

pkg/errors/errors.go
```diff
@@ -35,27 +36,53 @@ func init() {
...
+					zap.String("stacktrace", trace),
+					zap.Error(rootErr))
+				if obfuscateInternalErrors {
+					err = status.Error(codes.Internal, fmt.Sprintf("Internal server error %s", errID))
+				} else {
+					err = status.Error(codes.Internal, err.Error())
+				}
+			case statusErr.Code() == codes.Internal, statusErr.Code() == codes.Unknown:
+				errID := MakeErrID()
+				logger.Error(
```
vosst, line 75: `logger.Errorf`.

pkg/errors/errors.go
```diff
@@ -35,27 +36,53 @@ func init() {
 	}
 }
 
+func MakeErrID() string {
```
vosst, line 39: Please rename to `makeErrID()`. No need to export the function.

BenjaminPelletier: Done.

pkg/errors/errors.go
BenjaminPelletier: Done.

pkg/errors/errors.go
BenjaminPelletier: `type *zap.Logger has no field or method Errorf`; what did you have in mind?

BenjaminPelletier: More specifically: https://godoc.org/go.uber.org/zap#Logger

vosst: I was thinking about `zap.SugaredLogger` but it turns out that it does not support additional, unformatted fields. With that, comment resolved.

PR #388, opened 2020-08-10
[build] Introduce release target

### LINKED ISSUE #359: Add release tooling/scripts
Provide tooling and a `Makefile` target to help with releasing a new version of DSS, both in terms of code and artifacts.

vosst:
Fixes #359

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,3 +143,8 @@ cleanup-test-cockroach:
 .PHONY: test-e2e
 test-e2e:
 	test/docker_e2e.sh
+
+release: VERSION = v$(MAJOR).$(MINOR).$(PATCH)
+
+release:
+	scripts/release.sh $(VERSION)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -143,3 +143,8 @@ cleanup-test-cockroach:
 .PHONY: test-e2e
 test-e2e:
 	test/docker_e2e.sh
+
+release: VERSION = v$(MAJOR).$(MINOR).$(PATCH)
```
BenjaminPelletier, line 147: Let's throw an error if any of these are not specified (e.g., `"${MAJOR:?Need to set MAJOR version parameter}"`).  Or, does it really need to be split up like this?  Why not just `make release VERSION=v0.0.1`?

vosst: We are error'ing out later in the script if the format is not correct. I actually prefer the stricter `MAJOR`, `MINOR` and `PATCH` variables. Let's run with this and see if we find the strict approach to be burdensome.

PR #387, opened 2020-08-07
Update update sql script in store.go to support writer field

Dos-Ph:
#349 

1. Update store.go to have support both rid db v3 and v3.1 (Writer field table is added to identification_service_areas table in rid db v3.1)
2. Add prober test file docker_e2e_rid_v3.sh to support pervious version
3. Add public 'GetVersion' function to Store obj

Next: update InsertISA function to support rid db v3.1

PR #389, opened 2020-08-10
Update update sql script in store.go to support writer field

Dos-Ph:
#349 

Sorry, I accidentally deleted the branch so the previous PR was closed. #387 

Changes:
Create isaRepoV3 to support rid db v3.0.0 or below
Create IISARepo interface and use it in repo struct(cockroach/store.go). Then, uses rid dbVersion to decide between isaRepo and isaRepoV3
Set writer = store.Locality when create isa obj.


pkg/rid/store/store.go
```diff
@@ -14,6 +14,9 @@ type Store interface {
 
 	// Close closes the store and releases all of its resources.
 	Close() error
+
+	// Get store version
+	GetVersion(ctx context.Context) (string, error)
```
Dos-Ph, line 19: I will update this to return semver.Version

pkg/rid/store/cockroach/identification_service_area_v3.go
```diff
@@ -0,0 +1,204 @@
+package cockroach
```
BenjaminPelletier, line 1: Let's add some documentation to this file saying that its purpose is solely to support backwards compatibility and that it will be deleted from the codebase when all existing production deployments have been upgraded to 3.1.0+.

pkg/rid/store/cockroach/store.go
```diff
@@ -47,12 +47,13 @@ type Store struct {
 // Interact implements store.Interactor interface.
 func (s *Store) Interact(ctx context.Context) (repos.Repository, error) {
 	logger := logging.WithValuesFromContext(ctx, s.logger)
+	storeVersion, err := s.GetVersion(ctx)
+	if err != nil {
+		return nil, err
```
BenjaminPelletier, line 52: Because Go doesn't natively support stacktrace production, please generally add context when propagating errors by, e.g.: `return nil, stacktrace.Propagate(err, "Error determining database RID schema version")`.  This command is supported by adding `"github.com/palantir/stacktrace"` to the imports.  (This is a new general request and hasn't yet been documented in developer guidelines)

pkg/rid/store/cockroach/store.go
```diff
@@ -47,12 +47,13 @@ type Store struct {
 // Interact implements store.Interactor interface.
 func (s *Store) Interact(ctx context.Context) (repos.Repository, error) {
 	logger := logging.WithValuesFromContext(ctx, s.logger)
+	storeVersion, err := s.GetVersion(ctx)
```
BenjaminPelletier, line 50: Querying the database version every time an interaction is performed seems like an optimization opportunity -- could we check the version just once upon creation of the Store?

pkg/rid/store/cockroach/store.go
```diff
@@ -71,14 +72,16 @@ func (s *Store) Transact(ctx context.Context, f func(repo repos.Repository) erro
 	// "store" for everything
 	ctx, cancel := context.WithTimeout(ctx, DefaultTimeout)
 	defer cancel()
+
+	storeVersion, err := s.GetVersion(ctx)
```
BenjaminPelletier, line 76: Querying the version every transaction seems like the same optimization opportunity as for interactions.

Makefile
```diff
@@ -142,4 +142,4 @@ cleanup-test-cockroach:
 
 .PHONY: test-e2e
 test-e2e:
-	test/docker_e2e.sh
+	test/docker_e2e.sh
```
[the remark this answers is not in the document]
Dos-Ph: Fixed

pkg/rid/store/cockroach/identification_service_area.go
```diff
@@ -15,13 +15,35 @@ import (
...
 )
 
 const (
-	isaFields       = "id, owner, url, cells, starts_at, ends_at, updated_at"
-	updateISAFields = "id, url, cells, starts_at, ends_at, updated_at"
+	isaFields       = "id, owner, url, cells, starts_at, ends_at, writer, updated_at"
+	updateISAFields = "id, url, cells, starts_at, ends_at, writer, updated_at"
 )
 
+type IISARepo interface {
```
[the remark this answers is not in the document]
Dos-Ph: Removed and used repos.ISA instead

pkg/rid/store/cockroach/identification_service_area_v3.go
Dos-Ph: Added

pkg/rid/store/cockroach/store.go
Dos-Ph: Done

pkg/rid/store/cockroach/store.go
BenjaminPelletier: Never mind; we would need to do this to determine when the database was upgraded to 3.1.0.

pkg/rid/store/cockroach/store.go
BenjaminPelletier: Never mind; we would need to do this to determine when the database was upgraded to 3.1.0.

pkg/rid/store/store.go
BenjaminPelletier: Sounds good, but that can be a follow-up PR.

PR #393, opened 2020-08-11
Added prober test case for heavy traffic scenario

aegisbing8611:
The test cases cover:
1.Single USS creates hundred of nonoverlapping operations that are close to each other sequentially
2.Get & Query to verfiy creation
3.Mutate created opertions
4.Delete
5.Verify deletion

Tested through both 'pytest' and 'docker_e2e' cmds.


monitoring/prober/scd/test_operation_simple_heavy_traffic.py
```diff
@@ -0,0 +1,302 @@
...
+  end_time = time_now + datetime.timedelta(hours=1)
+
+  resp = scd_session.post('/operation_references/query', json={
+    'area_of_interest': common.make_vol4(time_now, end_time, 0, 5000, common.make_circle(-56, 178, 12000))
+  }, scope=SCOPE_SC)
+  assert resp.status_code == 200, resp.content
+  found_ids = [op['id'] for op in resp.json().get('operation_references', [])]
+  assert not _intersection(OP_IDS, found_ids)
+
+  resp = scd_session.post('/operation_references/query', json={
```
BenjaminPelletier, line 67: nit: I think we can remove this and the SCOPE_CM check -- they are there in the original file to make sure queries in general don't succeed with the wrong scope, but that check is already covered in the other file and doesn't need to be repeated here.  Ditto on other SCOPE_CI & SCOPE_CM 403 checks in other sections below.

monitoring/prober/scd/test_operation_simple_heavy_traffic.py
```diff
@@ -0,0 +1,302 @@
...
+# Preconditions: Operations with ids in OP_IDS created by scd_session user
+# Mutations: None
+@default_scope(SCOPE_SC)
+def test_get_ops_by_search(scd_session):
+  resp = scd_session.post('/operation_references/query', json={
+    'area_of_interest': common.make_vol4(None, None, 0, 5000, common.make_circle(-56, 178, 12000))
+  })
+  assert resp.status_code == 200, resp.content
+  found_ids = [op['id'] for op in resp.json().get('operation_references', [])]
+  assert len(_intersection(OP_IDS, found_ids)) == 100
```
BenjaminPelletier, line 142: nit: Maybe `len(OP_IDS)` instead of `100`?

monitoring/prober/scd/test_operation_simple_heavy_traffic.py
```diff
@@ -0,0 +1,302 @@
...
+    'area_of_interest': common.make_vol4(None, None, 0, 5000, common.make_circle(-56, 178, 12000))
+  })
+  assert resp.status_code == 200, resp.content
+  found_ids = [op['id'] for op in resp.json().get('operation_references', [])]
+  assert not _intersection(OP_IDS, found_ids)
+
+
+# Generate request with volumes that cover a circle area that initially centered at (-56, 178)
+# The circle's center lat shifts 0.001 degree (111 meters) per sequential idx change
+def _make_op_request(idx):
```
BenjaminPelletier, line 285: nit: Let's put utility functions at the top of the file so a reader has a chance to see them before seeing their use.

monitoring/prober/scd/resources/op_ids_100.json
```diff
@@ -0,0 +1,102 @@
+[
+  "283098e3-4b43-46e2-b5db-e35e1ec4c439",
```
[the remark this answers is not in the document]
aegisbing8611: Good catch! Sorry I missed some portions of the prober ReadMe. Now all static test IDs are prefixed and suffixed by 6 0s.

monitoring/prober/scd/test_operation_simple_heavy_traffic.py
aegisbing8611: Makes sense, removed.

monitoring/prober/scd/test_operation_simple_heavy_traffic.py
aegisbing8611: Good call, changed to `len(OP_IDS)`

monitoring/prober/scd/test_operation_simple_heavy_traffic.py
aegisbing8611: Done, moved to the top of the file.

Issue #391, opened 2020-08-10
Missing Constraint OVNs are not be included in PUT Operation 409


BenjaminPelletier:
Currently, scd.errors.MissingOVNsErrorResponse only accepts missing Operations, not missing Constraints.  Missing Constraint OVNs should also be returned if the Operation's Subscription notifies for Constraints.

Issue #386, opened 2020-08-06
SCD Subscriptions do not track DependentOperations


BenjaminPelletier:
Each Operation is required to be covered by a Subscription.  This requirement can be fulfilled by either providing the necessary information to simultaneously create a new (implicit) Subscription along with the Operation, or by providing the ID of an existing Subscription that covers the Operation.  Either way, the Operation becomes dependent on the Subscription, so it becomes invalid to shrink the Subscription spacetime area in a way that would exclude any part of the Operation, or to delete the Subscription entirely.  We should track these dependences with a Subscription's DependentOperations.  However, currently, we do not validate against DependentOperations when mutating or deleting a Subscription -- those checks must be added.

BenjaminPelletier:
More importantly, the SCD database schema does not even include dependent Operations for Subscriptions; this column needs to be added.

aegisbing8611:
https://github.com/interuss/dss/pull/452
aims to resolve this and is being reviewed

BenjaminPelletier:
We need to uncomment text in monitoring.prober.scd.test_operations_simple that refers to this bug

pratibhagupta2109:
Test cases updated in #580.

Issue #385, opened 2020-08-06
Subscriptions to notify for Operation changes are not correctly identified


BenjaminPelletier:
When an Operation's area is mutated, the DSS should instruct the client to inform all affected Subscriptions.  Affected Subscripions include both Subscriptions in the new area of the Operation, and Subscriptions in the old area of the Operation.  Currently, the DSS only includes Subscriptions in the new area of the Operation.

The Operations repo API should be reworked to be more simple-CRUD-like (#277), and then the Subscription identification logic should be fixed in the application layer.  See Constraints for a reference implementation.

Issue #277, opened 2020-05-13
Separate application logic from Store implementation


BenjaminPelletier:
Currently, we're performing a lot of application-level actions within the Store implementations, likely because we didn't want to deal with multiple Store actions needing to happen in a single transaction.  This makes it difficult to test the application logic independent of the storage layer by mocking just the storage layer, and it also makes it more difficult to implement different storage methods because the same application logic must be duplicated between them.

One way to address this may be to have the Store interface have a single Transact() method which returns a Transactor object that has all the basic CRUD methods attached to it.

PR #396, opened 2020-08-13
Simplify Operations repo interface and fix some Operation behavior

### LINKED ISSUE #344: DSS returning too many Entities on 409'd PUT
**Describe the bug**
I'm PUTing to the operation_references endpoint.  There are two other ops owned by other USSs that intersect my operation.  I supply a valid OVN for only one of those two operations.  I get 409 response as expected.  But I get BOTH operations returned with the 409 response.

**To Reproduce**
1. Submit two operations as USS1
2. Submit one operation as USS2 that intersects both ops from Step 1.
    a. Make sure the operation intesects the other two submitted in Step 1.
    b. Only submit one OVN and let it be valid for one of the existing ops.
3. Receive 409 with both operations in the entity_conflicts array

**Expected behavior**
Given two operations in DSS by another USS
When I submit an operation that intersects both of those operations
And I submit only one valid OVN for those operations
Then I should get a 409
And the entity_conflicts array should contain the operation for the OVN I did not submit

**Additional context**
We may have scripts that reproduce this.  May be able to add stdout to show the issue.

BenjaminPelletier:
This PR primarily simplifies the internal store/repo interface for Operations per #277, and, in the process, incidentally fixes #344, #385, and #391, and partially addresses #386.  An additional adjustment regarding Operations is that DependentSubscriptions is removed from the Subscription model since it was not consistently populated -- instead, this information can be obtained via a separate repo call when needed.  A few tests are augmented, and the validity of end times in the past is clarified in a few cases (ok to query Operations in the past, not ok to create/update an Operation to have its end time in the past).

### DIFF
--- a/pkg/scd/errors/errors.go
+++ b/pkg/scd/errors/errors.go
@@ -18,7 +18,7 @@ var (
 // MissingOVNsErrorResponse is Used to return sufficient information for an
 // appropriate client error response when a client is missing one or more
 // OVNs for relevant Operations or Constraints.
-func MissingOVNsErrorResponse(missingOps []*dssmodels.Operation) (*spb.Status, error) {
+func MissingOVNsErrorResponse(missingOps []*dssmodels.Operation, missingConstraints []*dssmodels.Constraint) (*spb.Status, error) {
 	detail := &scdpb.AirspaceConflictResponse{
 		Message: errMessageMissingOVNs,
 	}
@@ -29,6 +29,16 @@ func MissingOVNsErrorResponse(missingOps []*dssmodels.Operation) (*spb.Status, e
 		}
 		entityRef := &scdpb.EntityReference{
 			OperationReference: opRef,
+		}
+		detail.EntityConflicts = append(detail.EntityConflicts, entityRef)
+	}
+	for _, missingConstraint := range missingConstraints {
+		constraintRef, err := missingConstraint.ToProto()
+		if err != nil {
+			return nil, stacktrace.Propagate(err, "Error converting missing Constraint to proto")
+		}
+		entityRef := &scdpb.EntityReference{
+			ConstraintReference: constraintRef,
 		}
 		detail.EntityConflicts = append(detail.EntityConflicts, entityRef)
 	}
--- a/pkg/scd/errors/errors.go
+++ b/pkg/scd/errors/errors.go
@@ -37,10 +37,9 @@ func MissingOVNsErrorResponse(missingOps []*dssmodels.Operation, missingConstrai
 		if err != nil {
 			return nil, stacktrace.Propagate(err, "Error converting missing Constraint to proto")
 		}
-		entityRef := &scdpb.EntityReference{
+		detail.EntityConflicts = append(detail.EntityConflicts, &scdpb.EntityReference{
 			ConstraintReference: constraintRef,
-		}
-		detail.EntityConflicts = append(detail.EntityConflicts, entityRef)
+		})
 	}
 
 	p, err := dsserrors.MakeStatusProto(codes.Code(uint16(dsserrors.MissingOVNs)), errMessageMissingOVNs, detail)
--- a/pkg/scd/operations_handler.go
+++ b/pkg/scd/operations_handler.go
@@ -2,10 +2,7 @@ package scd
 
 import (
 	"context"
-	"fmt"
 	"time"
-
-	"github.com/golang/protobuf/ptypes"
 
 	"github.com/golang/geo/s2"
 	"github.com/google/uuid"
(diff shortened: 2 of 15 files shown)


pkg/scd/errors/errors.go
```diff
@@ -32,6 +32,16 @@ func MissingOVNsErrorResponse(missingOps []*dssmodels.Operation) (*spb.Status, e
...
 	}
+	for _, missingConstraint := range missingConstraints {
+		constraintRef, err := missingConstraint.ToProto()
+		if err != nil {
+			return nil, stacktrace.Propagate(err, "Error converting missing Constraint to proto")
+		}
+		entityRef := &scdpb.EntityReference{
+			ConstraintReference: constraintRef,
+		}
+		detail.EntityConflicts = append(detail.EntityConflicts, entityRef)
```
vosst, line 43: nit: You could save a temporary here and do:
```
detail.EntityConflicts = append(detail.EntityConflicts, &scdpb.EntityReference{})
```

pkg/scd/operations_handler.go
```diff
@@ -37,17 +34,88 @@ func (a *Server) DeleteOperationReference(ctx context.Context, req *scdpb.Delete
...
+		var subs []*scdmodels.Subscription
+		for _, s := range allsubs {
+			if s.NotifyForOperations {
+				subs = append(subs, s)
+			}
+		}
+
+		// Increment notification indices for Subscriptions to be notified
+		err = incrementNotificationIndices(ctx, r, subs)
+		if err != nil {
```
vosst, line 99: nit: More compact `if err = incrementNotificationIndex; err != nil { ... }`.

pkg/scd/operations_handler.go
```diff
@@ -37,17 +34,88 @@ func (a *Server) DeleteOperationReference(ctx context.Context, req *scdpb.Delete
...
+			return stacktrace.Propagate(err, "Unable to increment notification indices")
+		}
+
+		// Delete Operation from repo
+		err = r.DeleteOperation(ctx, id)
```
vosst, line 104: Same nit here.

pkg/scd/server.go
```diff
@@ -69,3 +71,18 @@ func (a *Server) AuthScopes() map[auth.Operation]auth.KeyClaimedScopesValidator
 func (a *Server) MakeDssReport(ctx context.Context, req *scdpb.MakeDssReportRequest) (*scdpb.ErrorReport, error) {
 	return nil, stacktrace.NewErrorWithCode(dsserr.BadRequest, "Not yet implemented")
 }
+
+func incrementNotificationIndices(ctx context.Context, r repos.Repository, subs []*scdmodels.Subscription) error {
```
vosst, line 75: nit: It seems like we want to introduce
```
package models

type Subscriptions []*Subscription

func (subs Subscriptions) IncrementNotificationIndices(context.Context, repos.Repository) error {
  ...
}
```

pkg/scd/store/cockroach/operations.go
```diff
@@ -84,6 +82,13 @@ func (s *repo) fetchOperations(ctx context.Context, q dsssql.Queryable, query st
 		return nil, stacktrace.Propagate(err, "Error in rows query result")
 	}
 
+	for _, op := range payload {
+		err = s.populateOperationCells(ctx, q, op)
```
vosst, line 86: Same nit for more compactness.

pkg/scd/errors/errors.go
BenjaminPelletier: Done.

pkg/scd/operations_handler.go
BenjaminPelletier: Done.

pkg/scd/operations_handler.go
BenjaminPelletier: Done.

pkg/scd/server.go
BenjaminPelletier: Nice :) Done.

pkg/scd/store/cockroach/operations.go
BenjaminPelletier: Done.

pkg/scd/server.go
BenjaminPelletier: Actually, only sort of nice.  Can't define that function in models because repos depends on models and that would create a circular dependency.  So, next best is to put it as a utility function in repos rather than models.  But that also requires the Subscriptions type to be defined there, and that seems like a non-ideal place to define it.  And then, we have to change all the explicit []*scdmodels.Subscription type declarations to the more opaque repos.Subscriptions type.  I'm not sure this actually increases overall clarity.

BenjaminPelletier:
I've converted all the mentioned `err = ...` conversions to single-line if statements, but I think there's a larger conversation to be had here.  There are a very large number of remaining instances in the codebase of two-line `err = ...` + `if err != nil` and we should pick a single preferred style.  I suggest that the primary advantages of the two-line style are:
1) It emphasizes that a statement happens, and then a check occurs -- the fact that an important, possibly-expensive operation is invoked seems more easily lost inside a one-line if statement
2) It is more consistent with the required style when another result other than error is returned -- it doesn't seem ideal to change style just because a return value was added or removed
3) It seems more consistent with the spirit of Go -- [the ternary operator is omitted](https://golang.org/doc/faq#Does_Go_have_a_ternary_form) because clarity is more valuable than brevity

pkg/scd/server.go
vosst: Ah, the circular dependency issue is a deal breaker, indeed.

vosst:
> I've converted all the mentioned `err = ...` conversions to single-line if statements, but I think there's a larger conversation to be had here. There are a very large number of remaining instances in the codebase of two-line `err = ...` + `if err != nil` and we should pick a single preferred style. I suggest that the primary advantages of the two-line style are:
> 
> 1. It emphasizes that a statement happens, and then a check occurs -- the fact that an important, possibly-expensive operation is invoked seems more easily lost inside a one-line if statement
> 2. It is more consistent with the required style when another result other than error is returned -- it doesn't seem ideal to change style just because a return value was added or removed
> 3. It seems more consistent with the spirit of Go -- [the ternary operator is omitted](https://golang.org/doc/faq#Does_Go_have_a_ternary_form) because clarity is more valuable than brevity

I think we should scope variables as tightly as possible as a general guidance, but allow for exceptions. I find https://www.calhoun.io/one-liner-if-statements-with-errors/ to be a good overview, also concluding that it's a case-by-case decision.

PR #397, opened 2020-08-14
Switch to explicit semver type

vosst:


cmds/db-manager/main.go
```diff
@@ -239,8 +246,6 @@ func (m *MyMigrate) MigrationDirection(desiredVersion string, desiredStep int) (
 	if err != nil {
 		return 0, fmt.Errorf("Failed to get current DB version to determine migration direction: %v", err)
 	}
-	if !semver.IsValid(currentVersion) {
-		return 0, fmt.Errorf("The current DB Version format is in valid")
-	}
-	return Direction(semver.Compare(desiredVersion, currentVersion)), nil
+
+	return Direction(desiredVersion.Compare(*currentVersion)), nil
```
Dos-Ph, line 250: Do you have to cover the case where desiredVersion == nil?

Dos-Ph:
> We will need to also change the db-schemas/*.sql files where we bump the version numbers as well

I don't think we need. We just make sure that the new version after this doesn't contain 'v'.

cmds/db-manager/main.go
vosst: Nope, `desiredVersion` is not a pointer.

vosst:
> We will need to also change the db-schemas/*.sql files where we bump the version numbers as well

The code handles versions with and without prefix `v`.

vosst:
@charlie-pisuraj ping 😄

PR #398, opened 2020-08-18
Harmonize and document off-nominal error responses

BenjaminPelletier:
This PR simplifies, harmonizes, and better-documents error messages, especially in off-nominal circumstances.


vosst (approved):
LGTM

PR #400, opened 2020-08-19
[build] Fix git commit script

vosst:


monitoring/prober/aux_/test_version.py
```diff
@@ -7,4 +7,4 @@ def test_version(aux_session):
   assert resp.status_code == 200
   version = resp.json()['version']['as_string']
   assert version
-  assert 'undefined' not in version, version
+  assert '0.0.0-undefined' not in version, version
```
BenjaminPelletier, line 10: This relaxes the check a bit; is there a reason the original assertion wouldn't pass in a valid scenario?

vosst: Some hiccup on my ubuntu setup, committed accidentally.

PR #401, opened 2020-08-19
store database version in store

Dos-Ph:
Store db version when create. So, we don't have to query db every time that interact function is triggered.


pkg/rid/store/store.go
```diff
@@ -16,7 +16,7 @@ type Store interface {
 	Transactor
 
 	// Get store version
-	GetVersion(ctx context.Context) (*semver.Version, error)
+	GetVersion() semver.Version
```
vosst, line 19: I don't think we should adjust the interface and instead make the caching optimization an implementation detail. After all, an implementation talking to the backend data store would certainly be reasonable.

Dos-Ph: Sorry, I'm not quite get it. It doesn't hurt to have this optimization. Is there a chance that db version is changed in the fly? Does db migration process happens only when we restart server?

charlie-pisuraj: What I think @vosst might mean is, that we are change the way we check the version when we don't really need to. I think he is suggesting we keep the function signature the same. What we can do is have GetVersion still try to reach the backend and simply store the version when it gets it for the first time, and on subsequent gets check if its already cached the information then return cache if it exists.

Dos-Ph: oh. Got it. Thanks :)

pkg/rid/store/cockroach/store.go
```diff
@@ -158,5 +162,8 @@ func (s *Store) CleanUp(ctx context.Context) error {
 // GetVersion returns the Version string for the Database.
 // If the DB was is not bootstrapped using the schema manager we throw and error
 func (s *Store) GetVersion(ctx context.Context) (*semver.Version, error) {
+	if s.version != nil {
```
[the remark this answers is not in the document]
charlie-pisuraj: I think in light of the issue we see with transactions being aborted due to bubbling up the error too quickly with stack tracing, it might be a good idea to revisit this. as for what happens to the DSS instance when the db is migrated from under it? I think the schema will reject the call if there is a major change to it. Either case the schema should only be migrated after the supported code has been rolled out where the code should be backwards compatible to the older schema. But I can also see that we might not get that all the time in the real world.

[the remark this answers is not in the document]
BenjaminPelletier: What I'm thinking is that if we cache the DB version, then the 3.0.0 interaction methods will be used even after the DB gets migrated to 3.1.0.  So, someone may be running the newest code, but all their endpoints will start failing as soon as the DB is migrated from 3.0.0 to 3.1.0 since their endpoints won't recognize the DB version change, right?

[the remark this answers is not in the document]
Dos-Ph: If we have database v3.1.0 run with the old code, the system will not fail. The system just doesn't update 'writer' field.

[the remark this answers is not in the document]
charlie-pisuraj: Thats true since we are not enforcing that the writer column must not be null the call should be successful with the column written to be null. when we do move to V4.0.0 which we plan to enforce the column then at that point the code will fail. it is after all the next major version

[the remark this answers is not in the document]
BenjaminPelletier: This seems complex enough to warrant a note in build/README.md -- I think we should explain what will work (new code with 3.0.0, new code with 3.1.0), what won't work (old code with 3.1.0), and what works with what caveats (new code with container started when DB was 3.0.0 then updated to 3.1.0; caveat that `writer` is omitted).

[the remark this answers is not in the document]
Dos-Ph: Update readme file

build/README.md
```diff
@@ -520,3 +520,13 @@ existing clusters you will need to:
...
+
+### Database version and server cache
+Current version: 3.1.0
+Minimum support version: 3.0.0
+
+Changes in v3.1.0
+* Added a new writer field to identification_service_areas and subscriptions table. 
+* Read the database version and cache it in memory. 
+
+If you start server with database v3.0.0. and migrate database to v3.1.0 without restart server, the service will work fine, just the writer field gets ignored. However, if you start the server with database v3.1.0 and you downgrade the database to 3.0.0, the service will fail because the server tries to update the writer field.
```
charlie-pisuraj, line 532: nit: without restarting the server.

nit: the service will work, the writer field will be ignored.

nit: the service will fail because the server will try to write/update the writer field.

pkg/scd/store/cockroach/store.go
```diff
@@ -51,15 +51,15 @@ func NewStore(ctx context.Context, db *cockroach.DB, logger *zap.Logger) (*Store
 		clock:  DefaultClock,
 	}
 
-	if err := store.CheckCurrentMajorSchemaVersion(ctx); err != nil {
+	if err := store.checkCurrentMajorSchemaVersion(ctx); err != nil {
```
charlie-pisuraj, line 54: Why are we making these private again?

pkg/rid/store/cockroach/store.go
```diff
@@ -48,33 +48,37 @@ type repo struct {
...
 }
 
 // CheckCurrentMajorSchemaVersion checks that store supports the current major schema version.
-func (s *Store) CheckCurrentMajorSchemaVersion(ctx context.Context) error {
-	vs, err := s.GetVersion(ctx)
-	if err != nil {
-		return stacktrace.Propagate(err, "Failed to get database schema version for remote ID")
-	}
-	if vs == cockroach.UnknownVersion {
+func (s *Store) checkCurrentMajorSchemaVersion(vs semver.Version) error {
```
charlie-pisuraj, line 79: I don't think we need to pass in the version as a parameter right, since its already part of the store object when New Store is called, also I believe we want to keep this as public method as well to preserve the interface, as Thomas mentioned

Dos-Ph: This method is not part of store interface and also no other class call it. Do we have up coming changes that use this method?

pkg/rid/store/cockroach/store.go
```diff
@@ -152,5 +156,8 @@ func (s *Store) CleanUp(ctx context.Context) error {
 // GetVersion returns the Version string for the Database.
 // If the DB was is not bootstrapped using the schema manager we throw and error
 func (s *Store) GetVersion(ctx context.Context) (*semver.Version, error) {
+	if s.version != nil {
+		return s.version, nil
+	}
 	return s.db.GetVersion(ctx, DatabaseName)
```
charlie-pisuraj, line 162: if s.version == nil we should set the value before returning

charlie-pisuraj (changes requested):

build/README.md
```diff
@@ -520,3 +520,13 @@ existing clusters you will need to:
...
+
+### Database version and server cache
+Current version: 3.1.0
+Minimum support version: 3.0.0
+
+Changes in v3.1.0
+* Added a new writer field to identification_service_areas and subscriptions table. 
+* Read the database version and cache it in memory. 
+
+If you start server with database v3.0.0. and migrate database to v3.1.0 without restarting the server, the service will work, the writer field will be ignored. However, if you start the server with database v3.1.0 and you downgrade the database to 3.0.0, the service will fail because the server will try to write/update the writer field
```
BenjaminPelletier, line 532: We want to make sure the consumer of this documentation understands the expected behavior of the system at the level they're interested in.  The audience for this documentation is someone who is deploying a DSS, but may not be intimately familiar with how it works.  So, they likely would not know what a writer field is, what it does, or what the implications of it being ignored are.  If correct, we should say something like "...the service will work normally but undeleted, expired ISAs will not be automatically removed from the database by the system, even when the garbage collection feature is finished and enabled."

go.mod
```diff
@@ -18,8 +18,9 @@ require (
 	github.com/google/uuid v1.1.1
 	github.com/grpc-ecosystem/go-grpc-middleware v1.2.0
 	github.com/grpc-ecosystem/grpc-gateway v1.14.3
-	github.com/lib/pq v1.5.2
 	github.com/interuss/stacktrace v0.0.0-20200827180054-b2e58cf48818
+	github.com/lib/pq v1.5.2
+	github.com/palantir/stacktrace v0.0.0-20161112013806-78658fd2d177 // indirect
```
BenjaminPelletier, line 23: The palantir reference has now been removed from the codebase; this addition is no longer necessary.

go.sum
```diff
@@ -243,6 +243,8 @@ github.com/opencontainers/go-digest v1.0.0-rc1/go.mod h1:cMLVZDEM3+U2I4VmLI6N8jQ
 github.com/opencontainers/image-spec v1.0.1/go.mod h1:BtxoFyWECRxE4U/7sNtV5W15zMzWCbyJoFRP3s7yZA0=
 github.com/opentracing/opentracing-go v1.1.0/go.mod h1:UkNAQd3GIcIGf0SeVgPpRdFStlNbqXla1AfSYxPUl2o=
 github.com/openzipkin/zipkin-go v0.1.6/go.mod h1:QgAqvLzwWbR/WpD4A3cGpPtJrZXNIiJc5AZX7/PBEpw=
+github.com/palantir/stacktrace v0.0.0-20161112013806-78658fd2d177 h1:nRlQD0u1871kaznCnn1EvYiMbum36v7hw1DLPEjds4o=
```
BenjaminPelletier, line 246: As above, these references can now be removed.

pkg/rid/store/cockroach/store.go
```diff
@@ -48,33 +48,37 @@ type repo struct {
...
 }
 
 // CheckCurrentMajorSchemaVersion checks that store supports the current major schema version.
-func (s *Store) CheckCurrentMajorSchemaVersion(ctx context.Context) error {
-	vs, err := s.GetVersion(ctx)
-	if err != nil {
-		return stacktrace.Propagate(err, "Failed to get database schema version for remote ID")
-	}
-	if vs == cockroach.UnknownVersion {
+func (s *Store) CheckCurrentMajorSchemaVersion(vs semver.Version) error {
```
BenjaminPelletier, line 79: [Charlie's comment](https://github.com/interuss/dss/pull/401#discussion_r481489570) still applies here: there is no need to pass `vs` in because `version` is a pre-existing field in `Store`.  If we want to avoid dealing with nils, probably grabbing the version with `GetVersion` would be a good way to do things.  Or, we could make this a stand-alone function rather than a Store method.

BenjaminPelletier (changes requested):

pkg/rid/store/cockroach/store.go
Dos-Ph: I changed the function back. The function now receives Context instead of version.

charlie-pisuraj (approved):
Fix go.sum and go.mod with go build and we should be good to go, Thank You @Dos-Ph for your work. Lets finish this and move on to the Garbage collection code

go.sum
```diff
@@ -243,6 +243,7 @@ github.com/opencontainers/go-digest v1.0.0-rc1/go.mod h1:cMLVZDEM3+U2I4VmLI6N8jQ
 github.com/opencontainers/image-spec v1.0.1/go.mod h1:BtxoFyWECRxE4U/7sNtV5W15zMzWCbyJoFRP3s7yZA0=
 github.com/opentracing/opentracing-go v1.1.0/go.mod h1:UkNAQd3GIcIGf0SeVgPpRdFStlNbqXla1AfSYxPUl2o=
 github.com/openzipkin/zipkin-go v0.1.6/go.mod h1:QgAqvLzwWbR/WpD4A3cGpPtJrZXNIiJc5AZX7/PBEpw=
+github.com/palantir/stacktrace v0.0.0-20161112013806-78658fd2d177/go.mod h1:ao5zGxj8Z4x60IOVYZUbDSmt3R8Ddo080vEgPosHpak=
```
BenjaminPelletier, line 246: We still want to remove this added reference (that was intended to be part of [this comment](https://github.com/interuss/dss/pull/401#discussion_r487341999)).

BenjaminPelletier (changes requested):
It looks like [this comment](https://github.com/interuss/dss/pull/401#discussion_r487341871) was marked resolved, but I don't see any changes based on it.

go.sum
Dos-Ph: Sorry, there are 2 palantir ref. I missed this one TwT. Updated.

go.sum
Dos-Ph: Done

charlie-pisuraj (approved):

BenjaminPelletier (approved):
Added #442 to capture remaining documentation comment so we can merge.

PR #402, opened 2020-08-19
fix bug GetISA fail when writer=NULL

Dos-Ph:
Cause: migrating from db v3 to db v3.1 added a 'writer' field but didn't set the default value. So, all writer field values are NULL. When we try to read from the table, we get error.

Solution: if writer == NULL, use empty string instead.

### DIFF
--- a/pkg/rid/store/cockroach/identification_service_area.go
+++ b/pkg/rid/store/cockroach/identification_service_area.go
@@ -22,6 +22,7 @@ import (
 
 const (
 	isaFields       = "id, owner, url, cells, starts_at, ends_at, writer, updated_at"
+	getISAFields    = "id, owner, url, cells, starts_at, ends_at, IFNULL(writer, ''), updated_at"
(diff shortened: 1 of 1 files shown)


Dos-Ph:
> I think you can just `Scan` into a `sql.NullString` on all code paths instead of using a different SQL query.

I'm pretty new to this. What do you mean by `Scan into a sql.NullString on all code paths`? When you said `scan`, do you mean row.scan function?

vosst:
> > I think you can just `Scan` into a `sql.NullString` on all code paths instead of using a different SQL query.
> 
> I'm pretty new to this. What do you mean by `Scan into a sql.NullString on all code paths`? When you said `scan`, do you mean row.scan function?

Ah, my apologies, I should have provided more context. The `sql` package provides a couple of types that allow for scanning `NULL` values, see https://golang.org/pkg/database/sql/#NullString for documentation and an example.

PR #407, opened 2020-08-21
Add additional context for some error messages

BenjaminPelletier:
This PR provides some additional details about certain error messages in the server logs to make root cause identification a little easier.


vosst (commented):
Are we running the risk of leaking information here? In particular, when communicating out ownership information for subscriptions or operations?

BenjaminPelletier:
> Are we running the risk of leaking information here? In particular, when communicating out ownership information for subscriptions or operations?

Subscriptions yes, as the client cannot obtain Subscription ownership information otherwise.  Operations, ISAs, and Constraints, no (regarding ownership) -- the client can obtain that information via other means.  I'll change those to just give the information directly.

vosst (approved):
LGTM

PR #410, opened 2020-08-23
Update update sql script to write writer field for Subscriptions table

Dos-Ph:
#349

Changes
1. Add writer file to subscription object
2. Create function returning SubscriptionRepo based on store version
3. Update Subscription handler to pass locality value
4. Move previous /cockroach/subscription.go to /cockroach/subscription_v3.go. This file will be delete after rid db is migrated to v3.1.

### DIFF
--- a/pkg/rid/store/cockroach/subscriptions.go
+++ b/pkg/rid/store/cockroach/subscriptions.go
@@ -77,11 +77,7 @@ func (c *subscriptionRepo) process(ctx context.Context, query string, args ...in
 		if err != nil {
 			return nil, stacktrace.Propagate(err, "Error scanning Subscription row")
 		}
-		if writer.Valid {
-			s.Writer = writer.String
-		} else {
-			s.Writer = ""
-		}
+		s.Writer = writer.String
 		s.SetCells(cids)
 		payload = append(payload, s)
 	}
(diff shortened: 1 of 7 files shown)


pkg/rid/store/cockroach/subscriptions.go
```diff
@@ -52,11 +71,17 @@ func (c *subscriptionRepo) process(ctx context.Context, query string, args ...in
 			&cids,
 			&s.StartTime,
 			&s.EndTime,
+			&writer,
 			&s.Version,
 		)
 		if err != nil {
 			return nil, stacktrace.Propagate(err, "Error scanning Subscription row")
 		}
+		if writer.Valid {
```
vosst, line 80: Ah, no need to check `writer.Valid` in this case. Instead, `s.Writer = writer.String` does the right thing as `writer.String` will be the empty string if `!writer.Valid`.

vosst (approved):
LGTM with one niggle inline.

pkg/rid/store/cockroach/subscriptions.go
Dos-Ph: Done. Fixed it in both cookroach/subscriptions.go and cookroach/identification_service_area.go

PR #414, opened 2020-08-26
Fix deployment and debugging issues

BenjaminPelletier:
This PR was primarily prompted by run_locally.sh being broken by #412, but it addresses a few different small issues encountered when debugging an issue encountered in a particular deployment:
* Fix run_locally.sh (unquote environment variable where empty quotes cause incorrect behavior)
* Break out run_locally's startup sequencing logic into shell scripts rather than trying to put everything on one line
* Explicitly indicate whether SCD is enabled or not (specifying `-trace-requests true -scd_enabled` causes SCD not to be enabled)
* Add log statements to HTTP gateway to fill a gap where errors generated by the gRPC client (but not the gRPC backend), such as JSON deserialization, will be printed to the HTTP gateway logs -- this is in an effort to make sure that every error ID has a log in the server logs
* Pipe `trace-requests` for the HTTP gateway through deployment configurations (but leave it as its current default of false)
* Clarify pooling procedures


vosst:
> This PR was primarily prompted by run_locally.sh being broken by #412, but it addresses a few different small issues encountered when debugging an issue encountered in a particular deployment:
> 

Yup, my clean up PR has fixes for that, too. However, I was wondering if we could test `run_locally.sh` in our CI setup. I think we are running in VMs, and with that, we should be able to run `docker compose`.

> * Fix run_locally.sh (unquote environment variable where empty quotes cause incorrect behavior)
> * Break out run_locally's startup sequencing logic into shell scripts rather than trying to put everything on one line
> * Explicitly indicate whether SCD is enabled or not (specifying `-trace-requests true -scd_enabled` causes SCD not to be enabled)
> * Add log statements to HTTP gateway to fill a gap where errors generated by the gRPC client (but not the gRPC backend), such as JSON deserialization, will be printed to the HTTP gateway logs -- this is in an effort to make sure that every error ID has a log in the server logs
> * Pipe `trace-requests` for the HTTP gateway through deployment configurations (but leave it as its current default of false)
> * Clarify pooling procedures

BenjaminPelletier:
> Yup, my clean up PR has fixes for that, too. However, I was wondering if we could test `run_locally.sh` in our CI setup. I think we are running in VMs, and with that, we should be able to run `docker compose`.

Added #425

PR #415, opened 2020-08-26
Refactor interoperability tests to reuse prober monitoring authorization library

BenjaminPelletier:
The prober now supports two additional authorization methods and has a streamlined interface to specify those methods, but the interoperability test suite was previously re-implementing those methods.  This PR restructures the monitoring folder to reuse a common monitorlib package in both the prober and the interoperability test (which is moved to monitoring for consistency).


vosst (approved):
SGTM

Issue #421, opened 2020-08-26
SCD - Transient Http 500 Error when concurrently create/mutate/delete operations that have volume cover different 2d areas


aegisbing8611:
Currently when we concurrently create/mutate/delete operations that have non-overlapping geospatial areas (> 11 km away from each other), once in a while (1 out of 10 test runs) we observed the transient http 500 error.

```
>       assert resp.status_code == 200, resp.content
E       AssertionError: b'{
E           "error": "Internal server error E:851df5e0-dee4-46c6-93d4-6be66ed7efc4",
E           "code": 13,
E           "message": "Internal server error E:851df5e0-dee4-46c6-93d4-6be66ed7efc4",
E           "error_id": "E:ba9cd948-d307-485c-9a90-f78bd684b801"
E         }'
E       assert 500 == 200
```

Detailed trace:

[...]

Issue #419, opened 2020-08-26
SCD - Concurrently create operations that have different time range returns http 400


aegisbing8611:
**Concurrent Create**

If we submit tens/hundreds of operations with each has new subscription, the same 2d area, but different TIME ranges CONCURRENTLY to DSS, a portion of them will fail with http 400 with error 'Specified Subscription does not exist', while the rest of them will succeed.

```
>       assert resp.status_code == 200, resp.content
E       AssertionError: b'{
E           "error": "Specified Subscription does not exist",
E           "code": 3,
E           "message": "Specified Subscription does not exist",
E           "error_id": "E:da2454fd-a0b1-4af7-b55a-8021f8980b2b"
E         }'
E       assert 400 == 200
E        +  where 400 = <Response [400]>.status_code
```

Detailed trace:
2020-08-26T23:18:33.658Z	error	errors/errors.go:122	Error E:da2454fd-a0b1-4af7-b55a-8021f8980b2b during unary server call	{"method": "/scdpb.UTMAPIUSSDSSAndUSSUSSService/PutOperationReference", "stacktrace": "Specified Subscription does not exist\n --- at /app/pkg/scd/operations_handler.go:367 (Server.PutOperationReference.func1) ---", "grpc_code": "InvalidArgument", "code": 3, "error": "Specified Subscription does not exist"}

pratibhagupta2109:
This issue is addressed via #641 and #610

Issue #417, opened 2020-08-26
SCD - Concurrently mutate operations that have different altitudes returns http 500


aegisbing8611:
**Concurrent Mutate**

If we mutate tens/hundreds of operations with each has he same 2d area, but different ALTITUDE ranges CONCURRENTLY to DSS, a portion of them will fail with http 500:

```
>       assert resp.status_code == 200, resp.content
E       AssertionError: b'{
E           "error": "Internal server error E:1027f1fa-82b4-4545-92aa-ea69336f8442",
E           "code": 13,
E           "message": "Internal server error E:1027f1fa-82b4-4545-92aa-ea69336f8442",
E           "error_id": "E:6f3a07af-deff-40f5-aef5-4891746e9c72"
E         }'
E       assert 500 == 200
E        +  where 500 = <Response [500]>.status_code
```

with below error trace.

[...]

PR #422, opened 2020-08-27
Added SCD prober test cases for heavy CONCURRENT traffic

aegisbing8611:
The test cases cover:

1.Single USS creates hundreds of non-overlapping operations in parallel
2.The operations are designed to be non-overlapping in 2d, altitude ranges and time windows respectively
3.Get & Query operations concurrently
4.Mutate created operations concurrently
5.Get & Query to verify concurrently
5.Delete concurrently
6.Verify deletions

Given known DSS concurrent issues, the thread pool size is set to 1 to enforce sequtial execution to pass the test. Once the issues are fixed, we can just update the thread pool size without touching anything else.
 - https://github.com/interuss/dss/issues/417
 - https://github.com/interuss/dss/issues/418
 - https://github.com/interuss/dss/issues/419
 - https://github.com/interuss/dss/issues/420
 - https://github.com/interuss/dss/issues/421

Tested through both 'pytest' and 'docker_e2e' cmds.


monitoring/prober/scd/test_operation_simple_heavy_traffic_concurrent.py
```diff
@@ -0,0 +1,365 @@
...
+    elif resp.status_code == 404:
+      # As expected.
+      pass
+    else:
+      assert False, resp.content
+
+
+# Preconditions: None
+# Mutations: None
+@default_scope(SCOPE_SC)
```
BenjaminPelletier, line 182: I think we can remove this test and the following one unless they are converted to concurrent requests -- we already have test coverage that verifies non-existent ops are inaccessible via get & query.

monitoring/prober/scd/test_operation_simple_heavy_traffic_concurrent.py
```diff
@@ -0,0 +1,365 @@
...
+def test_ops_do_not_exist_get(scd_session):
+  for op_id in OP_IDS:
+    resp = scd_session.get('/operation_references/{}'.format(op_id))
+    assert resp.status_code == 404, resp.content
+
+
+# Preconditions: None
+# Mutations: None
+@default_scope(SCOPE_SC)
```
BenjaminPelletier, line 355: I think we can also remove this test -- the test above verifies that the operations no longer exist generally, and test coverage elsewhere ensures that requesting by get and by query return consistent results.

monitoring/prober/scd/test_operation_simple_heavy_traffic.py
```diff
@@ -18,7 +18,7 @@
 
 
 def _load_op_ids():
-  with open('./scd/resources/op_ids_100.json', 'r') as f:
+  with open('./scd/resources/op_ids_100_1.json', 'r') as f:
```
BenjaminPelletier, line 21: nit: Let's name the files `op_ids_heavy_traffic_sequential.json` and `op_ids_heavy_traffic_concurrent.json` instead of `op_ids_100_1.json` and `op_ids_100_2.json`.  This will allow developers browsing the code to link the files to their purpose more easily, and allow debuggers to identify which test a particular ID belongs to more easily.

aegisbing8611: Good call, resolved.

monitoring/prober/scd/test_operation_simple_heavy_traffic_concurrent.py
aegisbing8611: Make sense, deleted.

monitoring/prober/scd/test_operation_simple_heavy_traffic_concurrent.py
aegisbing8611: Gotcha, removed the test cases to verify the deletion since as you mentioned they are already covered in other places.

In terms of the concurrent get/query,  we have them right after the concurrent creation to verify the creation.

PR #424, opened 2020-08-28
Feature/load test suite

charlie-pisuraj:
This is a load test suite using Locust. I went with python since it worked well with the rest of our test suite.


monitoring/loadtest/locust_files/locustfile.py
```diff
@@ -0,0 +1,256 @@
...
+    abstract = True
+    isa_dict = {}
+    sub_dict = {}
+
+    def __init__(self, *args, **kwargs):
+        super(USS, self).__init__(*args, **kwargs)
+        oauth_adapter = auth.DummyOAuth("http://localhost:8085/token", "fake_uss")
+        self.client = DSSClient(self.host, oauth_adapter)
+        self.client._locust_environment = self.environment
+        self.client.default_scopes = [common.SCOPE_WRITE, common.SCOPE_READ]
```
BenjaminPelletier, line 66: This results in requesting a token with more permissions than necessary which doesn't follow principle of least privilege best practices; let's leave this None and instead specify `scope=` on each request call ([example](https://github.com/interuss/dss/blob/ca889d11aac4d548a04366ccc58de1753d616ab7/monitoring/prober/rid/test_isa_simple.py#L21)).

monitoring/loadtest/locust_files/locustfile.py
```diff
@@ -0,0 +1,256 @@
...
+        # insert atleast 1 ISA for update to not fail
+        self.create_isa()
+
+    def on_stop(self):
+        # clean up after itself
+        for _ in self.isa_dict:
+            self.delete_isa()
+
+
+class SUB(USS):
```
BenjaminPelletier, line 163: nit: Sub rather than SUB I think (or Subscription)

monitoring/loadtest/requirements.txt
```diff
@@ -0,0 +1,7 @@
+google-auth==1.6.3
```
BenjaminPelletier, line 1: The auth lib requirements are in monitorlib/requirements.txt included below; they don't need to be additionally included here (and we don't want to because that increases the likelihood of version mismatch).  I think we only need to specify `locust` here.

monitoring/loadtest/locust_files/locustfile.py
```diff
@@ -0,0 +1,256 @@
...
+        if (method == "PUT" and len(url.split("/")) > 3) or method == "PATCH":
+            real_method = "UPDATE"
+        else:
+            real_method = method
+        name = url.split("/")[1]
+        start_time = time.time()
+        result = None
+        try:
+            result = super(DSSClient, self).request(method, url, **kwargs)
+        except Exception as e:
```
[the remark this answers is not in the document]
charlie-pisuraj: Locust was able to handle Interrupts and kills the process without attributing it as a failure. This should only catch the failure from sending the request. I think this wide net is ok.

monitoring/loadtest/locust_files/locustfile.py
```diff
@@ -0,0 +1,256 @@
...
+            request_type=real_method,
+            name=name,
+            response_time=total_time,
+            exception=e,
+            response_length=0,
+        )
+
+
+class USS(User):
+    abstract = True
```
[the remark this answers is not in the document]
charlie-pisuraj: I'm not sure. This was suggested by the Locust API docs. It might be a variable that the frame work reads.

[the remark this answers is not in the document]
BenjaminPelletier: Ah, ok -- never mind then.  In that case, let's add a comment like "Suggested by Locust API docs"

monitoring/loadtest/README.md
```diff
@@ -0,0 +1,35 @@
...
+# Adjusting workload ratio
+In each files every action has a weight declared in the `@task(n)` decorator. You can adjust the value of `n` to suite your needs
+
+# Run Locally without Docker
+1. Go to the repository's root directory. We have to execute from root directory due to our directory structure choice.
+1. Create vitrual environment `virtualenv --python python3 ./monitoring/loadtest/env`
+1. Install dependencies `pip install -r ./monitoring/loadtest/requirements.txt`
+1. Configure your OAuth Server Endpoint in `./monitoring/loadtest/locust_files/conf.py`
+
+    a. Currently it only supports Dummy OAuth Server, if there is a request to support other Auth mechaisms, we can implement them.
```
BenjaminPelletier, line 14: I think one of the immediately useful things to do with this tool will be to run it on real deployments, so I think it's likely worth supporting the general auth mechanism in monitorlib.  This should be pretty easy; see my comment below.

monitoring/loadtest/README.md
```diff
@@ -0,0 +1,35 @@
+# LoadTest tool
+
+The LoadTest tool is based on [Locust](https://docs.locust.io/en/stable/index.html) which provides a UI for controlling the number of Users to spawn and make random requests. Currently its configured to make the request in the ratio 10 x Create ISA : 5 x Update ISA : 100 x Get ISA : 1 x Delete ISA. This means the User is 10 times likely to Create an ISA vs Deleting an ISA, and 10 times more likely to Get ISA vs Creating an ISA and so on. Subscription workflow is heavier on the Write side with the ratio of 100 x Create Sub : 50 x Update Sub : 20 x Get Sub : 5 x Delete Sub.
+
+# Adjusting workload ratio
+In each files every action has a weight declared in the `@task(n)` decorator. You can adjust the value of `n` to suite your needs
+
+# Run Locally without Docker
+1. Go to the repository's root directory. We have to execute from root directory due to our directory structure choice.
+1. Create vitrual environment `virtualenv --python python3 ./monitoring/loadtest/env`
```
BenjaminPelletier, line 10: nit: virtual

monitoring/loadtest/README.md
BenjaminPelletier, line 14: nit: mechanisms

monitoring/loadtest/locust_files/client.py
```diff
@@ -0,0 +1,65 @@
...
+
+class USS(User):
+    # Suggested by Locust 1.2.2 API Docs https://docs.locust.io/en/stable/api.html#locust.User.abstract
+    abstract = True
+    isa_dict: typing.Dict[str, str] = {}
+    sub_dict: typing.Dict[str, str] = {}
+
+    def __init__(self, *args, **kwargs):
+        super(USS, self).__init__(*args, **kwargs)
+        oauth_adapter = auth.DummyOAuth(conf.OAUTH_HOST, "fake_uss")
```
BenjaminPelletier, line 60: I think we can support all auth types with just:
```python
oauth_adapter = auth.make_auth_adapter(conf.AUTH_SPEC)
```

...and in conf.py:
```python
AUTH_SPEC = "DummyOAuth(http://localhost:8085/token,fake_uss)"
```

I assume conf.py is the best place to put that argument, otherwise I would have done:
```python
oauth_adapter = auth.make_auth_adapter(os.environ.get("AUTH_SPEC", "DummyOAuth(http://localhost:8085/token,fake_uss)"))
```

monitoring/prober/rid/test_isa_simple.py
```diff
@@ -65,6 +65,41 @@ def test_create_isa(session):
...
+      json={
+          'extents': {
+              'spatial_volume': {
+                  'footprint': {
+                      'vertices': common.VERTICES,
+                  },
+                  'altitude_lo': 20,
+                  'altitude_hi': 400,
+              },
+              'time_start': time_start.strftime(common.DATE_FORMAT),
```
BenjaminPelletier, line 86: You probably want `rid.DATE_FORMAT` (this is why the integration test is failing).  Also on other lines in this test.

monitoring/loadtest/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.7
+# Not -alpine because: https://stackoverflow.com/a/58028091/651139
+# The context for this image should be monitoring
```
BenjaminPelletier, line 3: This comment is wrong in a couple other Dockerfiles which is probably why it's here -- the context should be the root folder of the repo (per README and where the folders mentioned below are located).

BenjaminPelletier (changes requested):

monitoring/loadtest/README.md
charlie-pisuraj: What I will do what the prober tests does. Allow the user to use whatever class we have in infrastructure.auth

monitoring/loadtest/locust_files/client.py
charlie-pisuraj: Let me give this a go with reading from env vars. because locust doesn't have a way to pass flags to modules.

monitoring/prober/rid/test_isa_simple.py
```diff
@@ -83,34 +92,18 @@ def test_update_isa(session):
...
-      rid.DATE_FORMAT)
-  assert data['service_area']['time_end'] == time_end.strftime(
-      rid.DATE_FORMAT)
+  assert data['service_area']['flights_url'] == 'https://example.com/dss/v2'
   assert re.match(r'[a-z0-9]{10,}$', data['service_area']['version'])
   assert 'subscribers' in data
 
 
-@default_scope(SCOPE_READ)
-def test_get_isa_by_id(session):
```
BenjaminPelletier: I think it's a good idea to get the ISA above, but I think we want to keep this test here also -- it would be great to validate the actual change (flights_url -> https://example.com/dss/v2)

BenjaminPelletier (approved):

PR #428, opened 2020-09-01
[build] Express e2e tests with docker compose

vosst:


Makefile
```diff
@@ -142,17 +142,26 @@ cleanup-test-cockroach:
 	@docker rm dss-crdb-for-testing > /dev/null 2>&1 || true
 
 .PHONY: test-e2e
-test-e2e:
-	test/docker_e2e.sh
```
BenjaminPelletier, line 146: One regression in this PR is that docker-compose invokes a build for each of `local-dss-rid-bootstrapper`, `local-dss-scd-bootstrapper`, `local-dss-grpc-backend`, and `local-dss-http-gateway`.  It does (already) manage rebuilding correctly, using the cache for everything if nothing has changed.  However, just invoking the build command (even when using the cache for everything) takes 14-18 seconds on my machine which means doubling the bootstrapper build and doubling the main image build adds 30+ seconds to the e2e invocation cycle versus the current docker_e2e.sh.  Since the e2e test cycle is a primary development tool, I think it's worth trying to just have each of those images build once.  I've added a commit that fixes this (no more action required).

BenjaminPelletier (changes requested):
I'm not sure exactly why, but even with my docker-compose.yaml change to avoid building two images twice each, `make test-e2e` still takes around 25 seconds (31%) longer(1) on my machine to begin the tests than the current `docker-e2e.sh`.  As this is a core development tool, I'm hesitant to add 25 seconds to each e2e iteration -- do you see anything else we could do to reduce this extra time?

(1) Test procedure:
* Check out master
* `test/docker_e2e.sh`
* Start timing
* `test/docker_e2e.sh`
* Stop timing when the first test is reached
* Check out this branch
* `make test-e2e`
* `make stop-locally`
* Start timing
* `make test-e2e`
* Stop timing when the first test is reached

BenjaminPelletier:
Closing this PR due to inactivity, but it can be reopened in the future.

PR #435, opened 2020-09-13
Added OVN remain the same test on PUT / DELETE operation

rpai1:


monitoring/prober/scd/test_operation_references_error_cases.py
```diff
@@ -181,3 +181,29 @@ def test_op_invalid_id(scd_session):
...
+  with open('./scd/resources/op_request_1.json', 'r') as f:
+      req = json.load(f)
+  resp1 = scd_session.put('/operation_references/{}'.format(OP_ID), json=req)
+  assert resp1.status_code == 200, resp1.content
+  ovn1 = resp1.json()['operation_reference']['ovn']
+  existing_op = resp1.json().get('operation_reference', None)
+
+  with open('./scd/resources/op_request_1.json', 'r') as f:
+      req = json.load(f)
+      req['state'] = 'Activated'
```
BenjaminPelletier, line 196: nit: No need for these three lines to be indented; `req = json.load(f)` is the only line that needs the `f` context.  And actually, no need to reload `req` at all.

monitoring/prober/scd/test_operation_references_error_cases.py
```diff
@@ -181,3 +181,29 @@ def test_op_invalid_id(scd_session):
...
+      req['old_version'] = 1
```
BenjaminPelletier, line 197: nit: Better to assign this to `existing_op['version']` rather than assuming 1 (will better support future changes).

monitoring/prober/scd/test_operation_references_error_cases.py
```diff
@@ -181,3 +181,29 @@ def test_op_invalid_id(scd_session):
...
+      req['key'] = [existing_op["ovn"]]
+  resp2 = scd_session.put('/operation_references/{}'.format(OP_ID), json=req)
+  assert resp2.status_code == 200, resp2.content
+  ovn2 = resp2.json()['operation_reference']['ovn']
```
BenjaminPelletier, line 201: Let's add a commented-out `assert ovn2 != ovn1` here with a TODO mentioning an Issue (we can open one if one doesn't currently exist, which I don't think it does).

monitoring/prober/scd/test_operation_references_error_cases.py
```diff
@@ -181,3 +181,29 @@ def test_op_invalid_id(scd_session):
...
+
+  resp3 = scd_session.delete('/operation_references/{}'.format(OP_ID), json=req)
+  assert resp3.status_code == 200, resp3.content
+  ovn3 = resp3.json()['operation_reference']['ovn']
+  assert ovn1 == ovn3 == ovn2
```
BenjaminPelletier, line 206: The expected behavior here is `ovn2 == ovn3`; let's address the ovn1 expectations as mentioned above.

BenjaminPelletier:
Closing this PR due to inactivity, but it can be reopened in the future.

Issue #423, opened 2020-08-28
Garbage collector functionality remove expired data after x period.


Dos-Ph:
Set up scheduler using Go Job Scheduler -> https://golangexample.com/go-job-scheduler-for-humans/

Changes:
1. Add 'deleteExpiredRecords(ctx, duration, writer)' function to ISA interface (dss/pkg/rid/repos/isa.go) and Subscription interface (dss/pkg/rid/repos/subscription.go). Talked to Charlie, he said 30 mins should be enough for an expired duration.

2. Implement the 'deleteExpiredRecords' function in dss/pkg/rid/store/cockroach/identification_service_area.go and dss/pkg/rid/store/cockroach/subscription.go
- sql: DELETE FROM identification_service_areas WHERE ends_at < dateadd(minute, -30, GetDate()) and writer = '<writer>'.
- sql: DELETE FROM subscriptions WHERE ends_at < dateadd(minute, -30, GetDate()) and writer = '<writer>'.
- log only when error occur

3. Create Garbage collector class
- File path: dss/pkg/rid/store/garbage_collector.go
- deleteExpiredRecords(ctx, duration, writer) function: delete expired records from the table after they expire for x minutes. The function call deleteExpiredRecords function in identification_service_area.go and subscription.go.

4. Setup scheduler in main.go when we create server by creating Garbage collector class and setting up a schedule to call it every 1 hour(?).
https://github.com/interuss/dss/blob/23465ba1c47b51f74ecf67b7624dccc64f65460a/cmds/grpc-backend/main.go#L245

Alternative1:
Instead of having a 'deleteExpiredRecords' function, we have 'listExpiredRecords' function. The function returns record obj and call DeleteSubscription and [DeleteISA](https://github.com/interuss/dss/blob/23465ba1c47b51f74ecf67b7624dccc64f65460a/pkg/rid/store/cockroach/identification_service_area.go#L184) function. 

Alternative2:
Instead of having separate 'deleteExpiredRecords' functions for ISA and Subscription, we have only one private function in Garbage collector class which gets table name as an input. 
- sql: DELETE FROM<tablename> WHERE ends_at < dateadd(minute, -30, GetDate()) and writer = '<writer>'.
For this one, we don't need to make a change on the ISA/Subscription interface. The function will be implemented as a private class in the Garbage collector class.

charlie-pisuraj:
@vosst @BenjaminPelletier , thoughts?

BenjaminPelletier:
[...]

PR #436, opened 2020-09-14
Create listExpiredISAs and listExpiredSubscriptions

Dos-Ph:
#423 
Next:
1. set scheduler to call these functions
2. add Garbage collector detail to 'build/README.md' file


pkg/rid/store/cockroach/identification_service_area.go
```diff
@@ -222,3 +222,25 @@ func (c *isaRepo) SearchISAs(ctx context.Context, cells s2.CellUnion, earliest *
...
+		ends_at > $1
+	AND
+		writer = $2`, isaFields)
+	)
+
+	if len(writer) == 0 {
+		return nil, stacktrace.NewError("Writer is required")
+	}
+
+	return c.process(ctx, isasInCellsQuery, expiredTime, writer)
```
charlie-pisuraj, line 245: ok I think I might see a problem here, if you passed in `expiredTime` we run the risk of using in consistent time from different servers. I understand normally we all use unix epoch but we probably don't want to risk it. We should probably just query it with the SQL statement for getting the database time like `CURRENT_TIME` and do offset with it.

charlie-pisuraj (changes requested):

charlie-pisuraj:
This looks good to me, maybe make the 30 Minute a configurable value? but other than that I'm fine with it

@vosst can you take a look?

charlie-pisuraj (approved):

pkg/rid/store/cockroach/identification_service_area.go
```diff
@@ -222,3 +222,25 @@ func (c *isaRepo) SearchISAs(ctx context.Context, cells s2.CellUnion, earliest *
 
 	return c.process(ctx, isasInCellsQuery, earliest, latest, pq.Int64Array(cids))
 }
+
+// ListExpiredISAs lists all expired ISAs based and writer.
```
BenjaminPelletier, line 226: "...all expired ISAs based *on* writer."

pkg/rid/store/cockroach/identification_service_area.go
```diff
@@ -222,3 +222,25 @@ func (c *isaRepo) SearchISAs(ctx context.Context, cells s2.CellUnion, earliest *
...
+	FROM
+		identification_service_areas
+	WHERE
+		ends_at + INTERVAL '%d' MINUTE <= CURRENT_TIMESTAMP
+	AND
+		writer = $1`, isaFields, expiredDurationInMin)
+	)
+
+	if len(writer) == 0 {
+		return nil, stacktrace.NewError("Writer is required")
```
BenjaminPelletier, line 242: How will ISAs with no writer be removed if we can't specify an empty string here?

pkg/rid/store/cockroach/subscriptions.go
```diff
@@ -301,3 +301,25 @@ func (c *subscriptionRepo) SearchSubscriptionsByOwner(ctx context.Context, cells
 
 	return c.process(ctx, query, pq.Int64Array(cids), owner, c.clock.Now())
 }
+
+// ListExpiredSubscriptions lists all expired Subscriptions based and writer.
```
BenjaminPelletier, line 305: "...based *on* writer."

BenjaminPelletier (commented):
This looks good to me with the exception that I think we need to have a way to deal with empty-writer records.  Right now, we can't even query for them, so we definitely won't be able to delete them.  I think an easy fix would be to allow the writer argument to be empty when listing expired resources, and retrieving no-writer entries in that case.  That will allow us to target them for deletion under appropriate circumstances.

pkg/rid/store/cockroach/identification_service_area.go
Dos-Ph: removed condition and add ListExpiredSubscriptionsWithNullWriter function to remove expired record with writer = NULL (writer can be NULL when db is migrated from v3.0 to v3.1)

BenjaminPelletier:
Is there a case where we would want to retrieve entities with writer='' separately from entities with writer=null?  If not, I suggest we retrieve any entities with writer='' or writer=null when a blank writer is passed to ListExpired*.  This simplifies the API by letting us eliminate the separate ListExpired*WithNullWriter methods.

pkg/rid/store/cockroach/identification_service_area.go
```diff
@@ -222,3 +222,26 @@ func (c *isaRepo) SearchISAs(ctx context.Context, cells s2.CellUnion, earliest *
 
 	return c.process(ctx, isasInCellsQuery, earliest, latest, pq.Int64Array(cids))
 }
+
+// ListExpiredISAs lists all expired ISAs based on writer.
+// Records expire if current time is <expiredDurationInMin> minutes more than records' endTime.
+func (c *isaRepo) ListExpiredISAs(ctx context.Context, writer *string) ([]*ridmodels.IdentificationServiceArea, error) {
+	writerQuery := "NULL"
+	if writer != nil {
```
BenjaminPelletier, line 230: With this implementation, if we want to garbage-collect everything without a writer in a robust way, we'll need to call ListExpiredISAs(ctx, "") and ListExpiredISAs(ctx, nil) -- is there ever a case in which we would want to make those calls separately, or distinguish between entries with "" writer and null writer?  If not, let's accept a standard string argument and query for both "" and null when the empty string is supplied.

BenjaminPelletier (approved):

PR #439, opened 2020-09-15
[monitoring] Factor out tracer fetches for reuse and log raw data directly

BenjaminPelletier:
This PR exports the USS client functionality of the tracer monitoring tool so that it can be reused in other monitoring tools.  It also reorganizes the data collection formats to focus on almost-pure raw data rather than storing results of computations that may be adjusted later.  This version of the tool was tested during a test event shakedown.


monitoring/monitorlib/fetch/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+    return obj
+  else:
+    return desired_type(obj)
+
+
+class RequestDescription(dict):
+  @property
+  def token(self) -> Dict:
+    return infrastructure.get_token_claims(self.get('headers', {}))
+yaml.add_representer(RequestDescription, Representer.represent_dict)
```
charlie-pisuraj, line 24: is this line supposed to be indented?

charlie-pisuraj (approved):
Approved but could you answer the comment. I'm curious

monitoring/monitorlib/fetch/__init__.py
BenjaminPelletier: No; this line is telling the yaml renderer that the type just defined (RequestDescription) should be rendered as a standard dict -- it corresponds with the type definition for RequestDescription (directly prior).  Basically, we define the class, and then we register the render method of that class with the yaml library -- this line is the latter step.

charlie-pisuraj: Cool, I've never done something like this before, good to know. Thanks

PR #441, opened 2020-09-17
Auto DB ping and reconnect

charlie-pisuraj:
We've seen that when CRDB dies or when the connection to it dies it doesn't get automatically reconnected. To fix this we ping the DB ever minute to refresh the connection.

cc: @Dos-Ph as an example for scheduling Garbage collector

### DIFF
--- a/cmds/grpc-backend/main.go
+++ b/cmds/grpc-backend/main.go
@@ -30,6 +30,7 @@ import (
 	scdc "github.com/interuss/dss/pkg/scd/store/cockroach"
 	"github.com/interuss/dss/pkg/validations"
 	"github.com/interuss/stacktrace"
+	"github.com/robfig/cron/v3"
 
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	"go.uber.org/zap"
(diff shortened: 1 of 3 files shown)


cmds/grpc-backend/main.go
```diff
@@ -114,6 +130,11 @@ func createSCDServer(ctx context.Context, logger *zap.Logger) (*scd.Server, erro
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "Failed to connect to strategic conflict detection database; verify your database configuration is current with https://github.com/interuss/dss/tree/master/build#upgrading-database-schemas")
 	}
+	// schedule period tasks for SCD Server
+	ridCron := cron.New()
```
charlie-pisuraj, line 134: LOL, copy paste failure

BenjaminPelletier:
This looks good to me, but I don't have the experience or knowledge to evaluate the periodic task invocation approach (it does seem to pull in a lot of dependencies for that job).  @vosst could you take a look?

charlie-pisuraj:
I think its only adding 1 extra dependency which is the robfig/cron and the others might be because i ran `go mod tidy`

BenjaminPelletier (approved):

Issue #351, opened 2020-07-15
Keep track of the db version in a central file


BenjaminPelletier:
Currently, the database version needs to be incremented in multiple places when a change is made.  We should, at a minimum, document where all these places are.  But, a better solution would be to have a single place from which everything obtains a shared piece of information (the database version) from.

_Originally posted by @vosst in https://github.com/interuss/dss/pull/349_

BenjaminPelletier:
An intermediate step could be to document all the places where the database version should be changed when it is incremented in development.

BenjaminPelletier:
A list of places the database version needs to be updated may be found in [build/deploy/db_schemas/README.md](https://github.com/interuss/dss/blob/master/build/deploy/db_schemas/README.md).

MariemBaccari:
The file has been renamed as [build/db_schemas/README.md](https://github.com/interuss/dss/blob/master/build/db_schemas/README.md).

PR #445, opened 2020-09-18
Add Index on end_at for improved performance

charlie-pisuraj:
Querying number of subscriptions is the bottle neck for inserting Subscriptions. Adding an index here making removes the full scan in the query. 
Ideally we want to index on cells however it is not supported in CRDB yet, They anticipate that v20.2 will support creating forward index on array

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -521,19 +521,4 @@ existing clusters you will need to:
 
 1.  Run `tk apply workspace/$CLUSTER_CONTEXT_schema_manager`
 
-### Database version and server cache
-Current version: 3.1.1
-Minimum support version: 3.0.0
-
-#### Note: It is best practice to keep the Server version ahead of the Database version. Having the Server code behind should be fine but to get all the functionalities both Server and Database version should be kept up to date ####
-
-Changes in v3.1.0
-* Added a new `writer` field to identification_service_areas and subscriptions table for the purposes of automatic Garbage Collection 
-* Read the database version and cache it in memory. 
-* Marking data for automatic garbage collection introduced in dss [c789b2b](https://github.com/interuss/dss/commit/c789b2b4a9fa5fb651d202da0a3abc02a03c15d2)  on Aug 25, 2020
-
-Changes in v3.1.1
-* Add index for subscriptions table by end time.
-
-The `writer` column was added to facilitate automatic Garbage Collection, recording the instance that added the record to Database.
-The presence of the `writer` column in the Database will not cause failure in basic functionality of the DSS. When running Server v3.1.x with Database v3.0.0, the `writer` column will simply be ignored and that record will not be automatically Garbage collected. You will need to manually remove expired records either with the existing Delete API ro manual SQL commands.
+Only since commit [c789b2b](https://github.com/interuss/dss/commit/c789b2b4a9fa5fb651d202da0a3abc02a03c15d2) on Aug 25, 2020 will the DSS enable automatic garbage collection of records by tracking which DSS instance is responsible for garbage collection of the record. Expired records added with a DSS deployment running code earlier than this must be manually removed.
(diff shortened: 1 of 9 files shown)


BenjaminPelletier:
(awaiting passing tests for review)

build/README.md
```diff
@@ -522,7 +522,7 @@ existing clusters you will need to:
 1.  Run `tk apply workspace/$CLUSTER_CONTEXT_schema_manager`
 
 ### Database version and server cache
-Current version: 3.1.0
+Current version: 3.1.1
 Minimum support version: 3.0.0
 
 #### Important: Server version should NEVER be behind Database Version ####
```
BenjaminPelletier, line 528: Let's remove this note as the semantic versioning of the database means that code intended to work with v3.0.0 of the database should also work with, e.g., v3.1.1 of the database.

build/README.md
```diff
@@ -531,5 +531,8 @@ Changes in v3.1.0
...
 * Read the database version and cache it in memory. 
 
+Changes in v3.1.1
+* Add index for subscriptions table by end time.
+
 The `writer` column was added to facilitate automatic Garbage Collection, recording the instance that added the record to Database.
-The presence `writer` column in the Database will not cause failure in basic functionality of the DSS, as long as you keep the version of the Server ahead of the Database version. When running Server v3.1.0 with Database v3.0.0, the `writer` column will simply be ignored and that record will not be automatically Garbage collected. You will need to manually remove expired records either with the existing Delete API ro manual SQL commands.
\ No newline at end of file
+The presence `writer` column in the Database will not cause failure in basic functionality of the DSS, as long as you keep the version of the Server ahead of the Database version. When running Server v3.1.x with Database v3.0.0, the `writer` column will simply be ignored and that record will not be automatically Garbage collected. You will need to manually remove expired records either with the existing Delete API ro manual SQL commands.
```
BenjaminPelletier, line 538: How about just, "Only since commit ??? on Aug ??, 2020 will the DSS enable automatic garbage collection of records by tracking which DSS instance is responsible for garbage collection of the record.  Expired records added with a DSS deployment running code earlier than this must be manually removed."

BenjaminPelletier (changes requested):
Until we resolve #351, I think we probably want to change the version [here](https://github.com/interuss/dss/blob/3f141e5b2af432aa30f5695e7d1e124cfa1d793f/pkg/rid/application/isa_test.go#L72), [here](https://github.com/interuss/dss/blob/3f141e5b2af432aa30f5695e7d1e124cfa1d793f/build/deploy/examples/schema_manager/main.jsonnet#L17), [here](https://github.com/interuss/dss/blob/3f141e5b2af432aa30f5695e7d1e124cfa1d793f/test/docker_e2e.sh#L79), and [here](https://github.com/interuss/dss/blob/b1aa7e7aa28174e40d1f90333ddf24786511f987/build/dev/startup/rid_bootstrapper.sh#L16).

build/README.md
```diff
@@ -522,14 +522,18 @@ existing clusters you will need to:
...
+
+Changes in v3.1.1
+* Add index for subscriptions table by end time.
 
 The `writer` column was added to facilitate automatic Garbage Collection, recording the instance that added the record to Database.
-The presence `writer` column in the Database will not cause failure in basic functionality of the DSS, as long as you keep the version of the Server ahead of the Database version. When running Server v3.1.0 with Database v3.0.0, the `writer` column will simply be ignored and that record will not be automatically Garbage collected. You will need to manually remove expired records either with the existing Delete API ro manual SQL commands.
\ No newline at end of file
+The presence of the `writer` column in the Database will not cause failure in basic functionality of the DSS. When running Server v3.1.x with Database v3.0.0, the `writer` column will simply be ignored and that record will not be automatically Garbage collected. You will need to manually remove expired records either with the existing Delete API ro manual SQL commands.
```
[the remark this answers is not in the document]
charlie-pisuraj: I'm also removing the weird change long logs we have as well. This should be enough information for the people deploying the code

BenjaminPelletier (approved):

Issue #438, opened 2020-09-15
DSS crashes with certain geographic inputs


BenjaminPelletier:
When certain inputs are provided to the DSS, it appears to consume a huge amount of memory and then crash.  One such input on SCD Subscriptions may be:

```json
{"extents":{"spatial_volume":{"footprint":{"vertices":[{"lng":180.0,"lat":90.0},{"lng":-180.0,"lat":90.0},{"lng":-180.0,"lat":-90.0},{"lng":180.0,"lat":-90.0}]},"altitude_lo":0.0,"altitude_hi":121.92000000000003},"time_start":"2020-09-15T01:29:51.730Z","time_end":"2020-09-15T01:34:51.730Z"}
```

This may be an issue with [`s2.Loop.Area`](https://github.com/interuss/dss/blob/135c72d76d0370b8bdbe68eec439ab02a0ca8b25/pkg/geo/s2.go#L89); we may want to consider estimating an upper bound of area using a lat,lng bounding rectangle and rejecting areas that have too big of such a bounding rectangle.

PR #446, opened 2020-09-18
[bugfix] Strictly validate polygons

### LINKED ISSUE #444: Invalid polygon geometry sometimes crashes DSS via S2 library
For certain kinds of invalid polygon geometry (especially, perhaps, crossing edges), the s2 library fails to detect the invalid geometry during [Validate](https://github.com/golang/geo/blob/e86565bf3f350d722923af04ce21a09e28d4b133/s2/loop.go#L209), and then locks the process until it dies.  We should implement something similar to the [`findAnyCrossing`](https://github.com/golang/geo/blob/e86565bf3f350d722923af04ce21a09e28d4b133/s2/loop.go#L215) method mentioned in the s2 comment and apply it to any incoming loops.  This will be non-trivial because these are spherical chord intersections rather than planar line segment intersections.

BenjaminPelletier:
This PR fixes #444 and #438 whose root causes were insufficient validation of polygons and a flaw in the s2 library when an invalid polygon was provided.  This PR ensures that no edges of a polygon cross, and that it passes other s2-library validation before attempting to compute a cell covering for it.

This fix revealed a number of problems in existing tests which have been adjusted to conform to the ASTM remote ID standard and draft strategic deconfliction standard.


pkg/geo/s2.go
```diff
@@ -89,26 +89,96 @@ func loopAreaKm2(loop *s2.Loop) float64 {
...
+	} else if l1p.Radians() >= math.Pi && l2p.Radians() >= math.Pi {
+		return true
+	}
+	return false
+}
+
+// validateLoop returns an error if any of the edges formed by the specified
+// points intersect each other.  There is an edge between the last and first
+// vertices.
+func validateLoop(points []s2.Point) error {
```
charlie-pisuraj, line 138: My apologies if this is not the same thing, but do you think we can use this? https://github.com/golang/geo/blob/e86565bf3f350d722923af04ce21a09e28d4b133/s2/loop.go#L225

BenjaminPelletier: We can and do call that method [via loop.Validate](https://github.com/interuss/dss/pull/446/files#diff-0838de1360740955856c6f0c3d004c22R162), however it [does not check for edge intersections](https://github.com/golang/geo/blob/e86565bf3f350d722923af04ce21a09e28d4b133/s2/loop.go#L245).  Once [this TODO](https://github.com/golang/geo/blob/e86565bf3f350d722923af04ce21a09e28d4b133/s2/loop.go#L215) is resolved, we may be able to switch back to using only the s2 library to validate a polygon instead of checking for edge intersections ourselves.

charlie-pisuraj (approved):

PR #447, opened 2020-09-18
[tracer] Handle request timeouts gracefully

BenjaminPelletier:
This PR gracefully handles HTTP request timeouts in tracer and also sets the timeout threshold to a more reasonable 30 seconds (5 seconds connect, 25 seconds read).  It also increases robustness to a few additional possible mistakes.


monitoring/monitorlib/fetch/__init__.py
```diff
@@ -62,7 +65,7 @@ def describe_request(req: requests.PreparedRequest,
 class ResponseDescription(dict):
   @property
   def status_code(self) -> int:
-    return self['code']
+    return self['code'] if self['code'] is not None else 999
```
charlie-pisuraj, line 68: can you get away with self.get('code', 999) ?

monitoring/tracer/uss_receiver/routes.py
```diff
@@ -50,7 +50,7 @@ def rid_isa_notification(id: str) -> Tuple[str, int]:
   label = colored('ISA', 'cyan')
   try:
     json = flask.request.json
-    if 'service_area' in json:
+    if 'service_area' in json and json['service_area']:
```
charlie-pisuraj, line 53: if json.get('service_area') ?

charlie-pisuraj (approved):

monitoring/monitorlib/fetch/__init__.py
BenjaminPelletier: Not in this case because the main condition we're expecting here is that there is an entry at `'code'` containing `None`.  But re: your comment below, I changed this to be resilient to missing `'code'` by using `get`.

monitoring/tracer/uss_receiver/routes.py
BenjaminPelletier: Ah yes...I had been thinking `get` behaved differently.  Changed here and a few other places -- thanks!

Issue #354, opened 2020-07-15
Support "latest" as target version identifier in migrations


BenjaminPelletier:
db-manager should support a value of "latest" for db_version (in addition to, e.g., "v1.0.0") that is automatically replaced with whatever version is most current.  This would make it so the current database version wouldn't need to be manually set in so many places.

Context:
Keeping track of these things are annoying for our test infra. I'll add an issue to add flag to always upgrade to latest in test

_Originally posted by @charlie-pisuraj in https://github.com/interuss/dss/pull/353_

PR #448, opened 2020-09-18
Feature/latest flag db manager

charlie-pisuraj:
Add --latest flag to make testing easier. In production we still want the user to specify the exact version the Schema Version should be for verbosity
#354

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,7 +129,7 @@ test:
 .PHONY: test-cockroach
 test-cockroach: cleanup-test-cockroach
 	@docker run -d --name dss-crdb-for-testing -p 26257:26257 -p 8080:8080  cockroachdb/cockroach:v20.1.1 start --insecure > /dev/null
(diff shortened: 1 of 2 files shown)


BenjaminPelletier (changes requested):
With three flags, there are a lot of opportunities for conflicting options (what happens, for instance, with `--db_version 3.0.0 --latest`?).  Why not have `"latest"` be a special value in the `db_version` flag?

charlie-pisuraj:
> With three flags, there are a lot of opportunities for conflicting options (what happens, for instance, with `--db_version 3.0.0 --latest`?). Why not have `"latest"` be a special value in the `db_version` flag?

Yeah thats a great point actually. right now we have it guarded where only one of the 3 should be specified. I do like the suggestion to use the string "latest" in --db_version. No objections to changing to make it do that.

BenjaminPelletier (approved):

PR #449, opened 2020-09-21
Remove bad conf import

charlie-pisuraj:
This was an artifact that reappeared after rebasing and was not previously caught


BenjaminPelletier:
Looks good, but could we also remove [these two references](https://github.com/interuss/dss/blob/3346c1ed3014f45c4ef1ab9ea5cacbf2abaec0db/go.sum#L266) while we're cleaning up?

BenjaminPelletier (approved):

PR #459, opened 2020-10-14
Fix 'remote error: tls: bad certificate'

charlie-pisuraj:
for prometheus to scrape cockroachdb we need to allow insecure_skip_verify set to true to skip verifying the cert matched the url its scraping. we already enable this in 'k8s-endpoint' job and don't need to double scrape in 'kubernetes-service-endpoint'


BenjaminPelletier:
Closing and reopening to retrigger EasyCLA

BenjaminPelletier (approved):

PR #469, opened 2020-11-30
Upgrade to CRDB 20.2

charlie-pisuraj:
Changes to underlying storage, from RocksDB to a CRL's forked version of RocksDB called Pebble.
There are other supports for geo data.


BenjaminPelletier:
Current status on this PR is that we want some documentation validating that this update functions properly on its own, and that it interoperates correctly with past versions (or else very clear documentation and mitigations regarding what users will need to do to update to the newer version).

BenjaminPelletier (approved):

PR #475, opened 2021-02-16
Add automated testing draft API

BenjaminPelletier:
This PR adds a working API for both the injection side of future automated remote ID testing (where automated testing injects known flight data into a system under test) and the ingestion side of future automated testing (where automated testing senses the state of the system under test from the perspective of a Display Application).


interfaces/automated-testing/rid/injection.yaml
```diff
@@ -0,0 +1,159 @@
+openapi: 3.0.2
+info:
+  title: Remote ID Test Data Injection
+  version: 0.0.1
+  description: >-
+    This interface is implemented by every Service Provider wishing to be tested by the automated testing framework.
+    The automated testing suite calls this interface to inject flight-related test data into the Service Provider system under test.
```
PearGod, line 7: This seems pretty specific.  Is there a design doc that this was covered in?  I would like to review what the overall outline for the approach for Test Data Injection is before we define the schema for it.

interfaces/automated-testing/rid/ingestion.yaml
```diff
@@ -0,0 +1,163 @@
...
+          description: >-
+            Identifier of flight that may be used to obtain details about the flight.
+            This is not necessarily the UTM/flight ID in the remote ID system.
+          type: string
+        most_recent_position:
+          description: Most recent position known for the flight.
+          allOf:
+            - $ref: '#/components/schemas/Position'
+        recent_paths:
+          description: Paths the flight recently traveled, if available.
```
hrishiballal, line 100: Suggestion to specify max /  min number of positions in the path, I know this is optional field.

interfaces/automated-testing/rid/injection.yaml
```diff
@@ -0,0 +1,159 @@
...
+      description: Remove all injected test data associated with this test from the host Service Provider's system under test.
+components:
+  schemas:
+    CreateTestParameters:
+      description: 'A complete set of data to be injected into a Service Provider system under test.'
+      required:
+        - requested_flights
+      type: object
+      properties:
+        requested_flights:
```
hrishiballal, line 85: Are all flight positions of all flights submitted to the endpoint in one `Create Test` request? A concern is that request body data might be too large and we might to set some limit as to how many flights can be created. e.g. one SP can create max two flights (or maybe just one).

interfaces/automated-testing/rid/ingestion.yaml
```diff
@@ -0,0 +1,163 @@
...
+            $ref: '#/components/schemas/Cluster'
+          default: []
+    GetDetailsResponse:
+      description: Response to a request to get details about a flight.
+      type: object
+    Flight:
+      description: ''
+      required:
+        - id
+        - most_recent_position
```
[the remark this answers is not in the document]
BenjaminPelletier: Serial number, registration number, and operation description are all fields in a /flight/details call, not the analog to a /flights query -- this object (`Flight`, a field in `GetDisplayDataResponse`) is the analog to a /flights query.  `GetDetailsResponse` is the response from a query similar to /flight/details and it currently has no fields because none of the automated tests identified in the SUSI document check this data.  Because the data format at the point it is ingested into the automated testing system (roughly the point of the Display Application) is not standardized, we shouldn't assume that all information available at the Display Provider is necessarily available at the Display Application unless that information is strictly required (because a Display Provider may choose not to expose certain information in its Display Application API if it is not required).  In Europe, the serial_number will be required so it may make sense to create an additional test to verify that the Display Application is capable of providing this data to the user, but it should be gated by a flag because not all jurisdictions will require this functionality and it is not required in the standard.  I recommend we implement the core functionality first and then add additional tests after MVP is working, at which point more fields in this API may be warranted.  But, I don't think we need to add them now as fields are easy to add, but hard to move, rename or delete.

interfaces/automated-testing/rid/ingestion.yaml
BenjaminPelletier: This is the ingestion interface, so is providing the data available in the remote ID UTM system to the automated testing framework.  An implementation of this interface is merely relaying information received from other participants in the system to the automated testing framework, so it has no power to ensure that a certain amount of data is populated (because those fields are populated, or not, by other participants).  Specifically, the ingestion interface implementation has no power/ability to ensure that the number of recent positions satisfies a minimum or maximum.  With regard to maximum, and important case we would want to check for is that no more than 60 seconds of recent positions are provided (because that is a requirement in the standard) so we certainly don't want to set a maximum at the ingestion interface as that might cause the ingestion interface implementation to discard an error condition that we want to detect.

I think this raises the question of whether most_recent_position should actually be required at this level, and I think the answer is no -- I will adjust that.

interfaces/automated-testing/rid/injection.yaml
BenjaminPelletier: Yes, the intention here was to include all injected data for all flights for a single Service Provider for a single test in a single call.  The data is carried in a request body rather than URL, so we're not subject to the ~2k limit for URLs.  Is there a reasonable server implementation that would have trouble with request bodies of this size?  If we have a 10 minute test (seems much longer than necessary) with 3 flights per SP (seems high; I think only 1 is needed for currently-identified tests), a standard 1 Hz update rate would result in a request body no larger than about 1 MB if I calculate correctly.  Do you expect a 1 MB (or even 2 MB) request body to be a problem for any web server an SP is likely to use?

Also, just to clarify, the automated testing system would be requesting the creation of these flights (not the SP).  So, with the current set of tests described in the SUSI document, I believe we can limit the number of flights requested of each SP to 1 by simply choosing to only request one flight per SP per test.

interfaces/automated-testing/rid/injection.yaml
BenjaminPelletier: Implied approach added to [README](https://github.com/BenjaminPelletier/dss/tree/automatedtesting/api/interfaces/automated-testing/rid).

interfaces/automated-testing/rid/ingestion.yaml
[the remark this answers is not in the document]
hrishiballal: OK Understood, no problems.

interfaces/automated-testing/rid/injection.yaml
hrishiballal: No problems, having one flight per SP is sounds OK since we may not "simultaneously" test it like the SUSI checkouts. In this case, I dont think it should be a problem.

BenjaminPelletier: Since the automated testing system is the one requesting the creation of flights, it would also be easy to request more than one flight per USS, though I don't think any of the current tests require this as described.  The most important aspect to test "simultaneously" is for each SP to have at least one flight in an area at the same time.  Multiple simultaneous flights for a single SP would probably also be nice to test, but I think the simultaneity with multiple SPs is the most important dimension to exercise.

interfaces/automated-testing/rid/README.md
```diff
@@ -0,0 +1,58 @@
...
+remote ID information available in a requested area, at the level of the Display
+Application, and provide that information to the automated test driver via the
+ingestion interface defined in this folder.
+
+The automated test driver would perform a test by injecting test data into each
+participating Service Provider, polling the resulting information from the
+remote ID system via the display data test receiver, and checking that the
+observed data satisfied a set of success criteria (e.g., an injected flight was
+observed at the expected time, not observed at other times, had the right number
+of recent positions, etc).
```
hrishiballal, line 58: Maybe we produce a compliance report at the end as text file or something similar?

interfaces/automated-testing/rid/ingestion.yaml
hrishiballal: A query: Why does it return `most_recent_positions` and not the whole `aircraft state`? Maybe I am getting confused.

interfaces/automated-testing/rid/injection.yaml
```diff
@@ -0,0 +1,180 @@
...
+          type: string
+          example: 'edb7695f-8737-4b9f-91f8-e2afbb333f41'
+        telemetry:
+          description: >-
+            The set of telemetry data that should be injected into the system for this flight.
+            Each element follows the RIDAircraftState schema from the ASTM remote ID standard.
+          type: array
+          items:
+            $ref: 'https://raw.githubusercontent.com/uastech/standards/astm_rid_1.0/remoteid/canonical.yaml#/components/schemas/RIDAircraftState'
+        details_responses:
```
hrishiballal, line 144: `details_responses` is a bit difficult to understand, recommend to rename it to `test_flight_details_responses` or something like that.

interfaces/automated-testing/rid/injection.yaml
```diff
@@ -0,0 +1,180 @@
...
+          items:
+            $ref: '#/components/schemas/TestFlightDetails'
+    TestFlightDetails:
+      description: The set of data with which the Service Provider system under test should respond when queried for the details of a test flight.
+      required:
+        - effective_after
+        - details
+      type: object
+      properties:
+        effective_after:
```
hrishiballal, line 156: At the moment, are we expecting that the track data will be submitted right after the tracks are generated? This is simpler and will work. The other option is to have a `test_start` and `test_end` parameters so we can say tracks / test data is created today for testing tomorrow (for e..g.)

hrishiballal (approved):

interfaces/automated-testing/rid/README.md
BenjaminPelletier: Yes, I think a report detailing compliance will be a great output.

interfaces/automated-testing/rid/injection.yaml
BenjaminPelletier: The object in which `details_responses` is defined is `TestFlight`; could your suggested field name be inferred by prefacing the field with the object type in which the field is located?  (`<TestFlight>.details_response`)

interfaces/automated-testing/rid/injection.yaml
BenjaminPelletier: Eventually, the tracks should be generated once and then committed to the repository so they can be audited (we'll do this once we get good tracks that have been verified to produce good test results).  When playing those tracks back, I'm imagining the replayer will probably leave a minute or so between the time the test data is delivered to a USS and the times the telemetry points should be applicable.  This allows for additional time on the automated testing side to announce test data to other USSs, and it allows time on the USS side to initialize its system with the injected data.

Currently, all injected data has/should have timestamps associated with it, so hypothetically rid_qualifier could provide it to the injector a day before the test.  Though I don't think that is a likely scenario as I wouldn't expect injected data to necessarily persist through a system restart, so if a USS redeployed their system, they may lose the data intended for the next day.

hrishiballal: OK got it. I think then we can drop the `test_start` and `test_end` parameters. I agree that when we have robust track data and track types we can commit it to the repository and at the moment we will test and develop these starting with what we have.

interfaces/automated-testing/rid/injection.yaml
hrishiballal: Agreed, we have that piece covered, with the multiple tracks that can be assigned to a different SP.

interfaces/automated-testing/rid/injection.yaml
hrishiballal: Yes that would work, thanks!

BenjaminPelletier:
I believe the comments have been resolved; I'll plan to merge this soon.

PR #485, opened 2021-03-08
Update jq version checking logic and fixed some typos.

KaiGuo-gkai:
- We have a check to see if the user has _jq_ installed. But check logic does the opposite. The PR fixed the logic.
- We have a Python script written in Python 3. An error occurs when the default local version is Python 2. Specifying the Python version solves this problem. (Another solution would be to modify the script so that both versions work)
- Fixed some typos.

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -32,7 +32,7 @@ Download & install the following tools to your workstation:
   - `kubectl` can be installed from `gcloud` instead of via the method below.
 - [Install kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) to
   interact with kubernetes
-  - Confirm sucessful installation with `kubectl version --client` (should
+  - Confirm successful installation with `kubectl version --client` (should
     succeed from any working directory).
   - Note that kubectl can alternatively be installed via the Google Cloud SDK
    `gcloud` shell if using Google Cloud.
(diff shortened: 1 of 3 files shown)


build/README.md
```diff
@@ -171,12 +171,12 @@ a PR to that effect would be greatly appreciated.
 1.  Use [`make-certs.py` script](./make-certs.py) to create certificates for
     the new CockroachDB cluster:
 
-        ./make-certs.py --cluster $CLUSTER_CONTEXT --namespace $NAMESPACE
+        python3 ./make-certs.py --cluster $CLUSTCONTEXT --namespace $NAMESPACE
```
BenjaminPelletier, line 174: I think we want to keep `$CLUSTER_CONTEXT` rather than replacing with `$CLUSTCONTEXT`

BenjaminPelletier, line 174: To be consistent with other script usage in this repository, could we instead change the first line of `make-certs.py` to be the following?

```python
#!/usr/bin/env python3
```

BenjaminPelletier (commented):
Looks good -- just 1 change request and 1 suggestion

build/README.md
KaiGuo-gkai: Oops, it is a typo. Thank you for pointing out.

KaiGuo-gkai: Good idea. Will do.

BenjaminPelletier (approved):

Issue #472, opened 2021-01-04
Prober test fails when run by different user


BenjaminPelletier:
When the prober test suite is run by one client ID, it then cannot be run (successfully) by a different client ID; there are two failures apparently due to resources that are not cleaned up.

BenjaminPelletier:
#489 addresses a large part of this bug.  We should also make resource IDs dependent on calling USS so that two USSs aren't trying to use the same resource ID.

pratibhagupta2109:
fixed in #549

PR #489, opened 2021-04-14
Update test cases with delete entries in the end.

pratibhagupta2109:


BenjaminPelletier (approved):
Verified to fix #472 when all test runs complete successfully.  Verified using process that we should probably add to the repo (described in #490).

PR #492, opened 2021-04-20
Script to run prober tests on local DSS instance

pratibhagupta2109:


build/dev/probe_local_instance.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+echo "Re/Create probe_local_instance_test_result file"
+RESULTFILE="$(pwd)/probe_local_instance_test_result"
+touch "${RESULTFILE}"
+cat /dev/null > "${RESULTFILE}"
+declare -a localhost_containers=("dss_sandbox_local-dss-http-gateway_1"
```
BenjaminPelletier, line 9: The issue on the `for` loop starts here -- this command only sets `localhost_containers` to "`dss_sandbox_local-dss-http-gateway_1`" (check with `echo $localhost_containers` just before the `for` loop).  Replacing this set of lines with `localhost_containers="dss_sandbox_local-dss-http-gateway_1 dss_sandbox_local-dss-grpc-backend_1 dss_sandbox_local-dss-crdb_1"` seems to resolve the issue, but I'm not sure how to split strings across multiple lines in bash.

BenjaminPelletier, line 9: I think we may set ourselves up for easier extension of this script in the future if we set variables for all the container names and then just use those variables.  So, something like:

```bash
GATEWAY_CONTAINER="dss_sandbox_local-dss-http-gateway_1"
OAUTH_CONTAINER="dss_sandbox_local-dss-dummy-oauth_1"
localhost_containers="$GATEWAY_CONTAINER $OAUTH_CONTAINER"
```

...and then we can reuse those variables again later:

```bash
docker run --network dss_sandbox_default \
  --link $OAUTH_CONTAINER:oauth \
	--link $GATEWAY_CONTAINER:local-gateway \
```

Note that I think we only have to check for these two containers because they're the only ones we talk to.  If the CRDB or gRPC containers are missing, that still might possibly be ok since we don't talk to them directly.

build/dev/probe_local_instance.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+echo "Re/Create probe_local_instance_test_result file"
+RESULTFILE="$(pwd)/probe_local_instance_test_result"
```
BenjaminPelletier, line 6: I think this file should probably have a .xml extension

build/dev/probe_local_instance.sh
```diff
@@ -0,0 +1,51 @@
...
+touch "${RESULTFILE}"
+cat /dev/null > "${RESULTFILE}"
```
BenjaminPelletier, line 8: I think this line is unnecessary given the `touch` line above

build/dev/probe_local_instance.sh
BenjaminPelletier, line 6: Let's add a [.gitignore](https://git-scm.com/docs/gitignore) file to this folder that includes the name of this result file so someone won't accidentally commit it to the repository

build/dev/startup/grpc_backend.sh
```diff
@@ -16,7 +16,7 @@ if [[ $DEBUG_ON == 1 ]]; then
   -reflect_api \
   -log_format console \
   -dump_requests \
-  -accepted_jwt_audiences localhost,host.docker.internal \
+  -accepted_jwt_audiences localhost,,host.docker.internal,local-gateway,dss_sandbox_local-dss-http-gateway_1 \
```
BenjaminPelletier, line 19: nit: I don't think we want the extra comma here as that will allow an empty audience (other additions are good)

BenjaminPelletier (approved):

PR #501, opened 2021-05-20
Add (mis)behaviors to rid_qualifier mock's virtual USSs

BenjaminPelletier:
This PR adds a dynamic behavior configuration to each virtual Service Provider and Display Provider in rid_qualifier mock so that these USSs can be configured to simulate various failures observed in real USSs, and therefore exercise the corresponding code paths in (the yet-to-be-written sections of) rid_qualifier.  For instance, a client of rid_qualifier mock can PUT `{"use_feet_instead_of_meters_for_altitude": true}` to http://localhost:8070/sp/uss1/behavior to simulate the observations that would occur if the uss1 RID Service Provider reported its telemetry in units of feet rather than units of meters.  Or, a client of rid_qualifier mock can PUT `{"do_not_display_flights_from": ["uss2"]}` to http://localhost:8070/dp/uss2/behavior to simulate the observations that would occur if the uss2 RID Display Provider could see other USSs' flights, but not those from its own Service Provider.

This PR also fixes a bug in `ImplicitDict` where optional fields with default values would not have their default values populated into constructed instances.


BenjaminPelletier:
Note that `routes.py` was split up into multiple files, so it's hard to see the actual/substantive changes in the full-PR diff.  To see the substantive changes to the code that was previously in routes, see just the second commit (after splitting `routes.py` in the first commit): 852aab8

pratibhagupta2109 (approved):

monitoring/rid_qualifier/mock/routes_behavior.py
```diff
@@ -0,0 +1,44 @@
...
+  try:
+    json = flask.request.json
+    if json is None:
+      raise ValueError('Request did not contain a JSON payload')
+    dp_behavior = ImplicitDict.parse(json, behavior.DisplayProviderBehavior)
+  except ValueError as e:
+    msg = 'Change behavior for Display Provider {} unable to parse JSON: {}'.format(dp_id, e)
+    return msg, 400
+
+  dp = db.get_dp(dp_id)
```
hrishiballal, line 23: Recommend a try/catch to check for invalid dp/sp id passed.

monitoring/rid_qualifier/mock/routes_behavior.py
```diff
@@ -0,0 +1,44 @@
...
+  try:
+    json = flask.request.json
+    if json is None:
+      raise ValueError('Request did not contain a JSON payload')
+    sp_behavior = ImplicitDict.parse(json, behavior.ServiceProviderBehavior)
+  except ValueError as e:
+    msg = 'Change behavior for Service Provider {} unable to parse JSON: {}'.format(sp_id, e)
+    return msg, 400
+
+  sp = db.get_sp(sp_id)
```
hrishiballal, line 41: Same comment as above: Recommend a try/catch to check for invalid dp/sp id passed.

hrishiballal (approved):

monitoring/rid_qualifier/mock/routes_behavior.py
BenjaminPelletier: Getting the dp/sp [will always succeed](https://github.com/BenjaminPelletier/dss/blob/852aab84ff11de46c6c17fde9ee4c6c0209a0892/monitoring/rid_qualifier/mock/database.py#L51)

monitoring/rid_qualifier/mock/routes_behavior.py
BenjaminPelletier: Getting the dp/sp [will always succeed](https://github.com/BenjaminPelletier/dss/blob/852aab84ff11de46c6c17fde9ee4c6c0209a0892/monitoring/rid_qualifier/mock/database.py#L46)

PR #505, opened 2021-05-26
rid_qualifier: Add first draft of display data evaluator

BenjaminPelletier:
This PR adds a first draft of the part of rid_qualifier that observes an RID system and evaluates whether the data observed is consistent with the injected flight data.  It is primarily located in `display_data_evaluator.py`, called from `test_executor.py`.  A simple JSON-formatted report is generated as defined in `reports.py` tracking what data was injected into the system, what observations were made, and what issues were discovered.

A few utilities are improved to support these efforts, and the delays and default durations in rid_qualifier are reduced to support quicker development iteration.

### DIFF
--- a/monitoring/monitorlib/fetch/__init__.py
+++ b/monitoring/monitorlib/fetch/__init__.py
@@ -2,6 +2,7 @@ import datetime
 import json
 from typing import Dict, Optional
 
+import arrow
 import flask
 import requests
 import yaml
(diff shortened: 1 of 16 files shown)


hrishiballal (approved):
Some methods need docstrings but I will create a PR for that for your review,

BenjaminPelletier:
> Some methods need docstrings but I will create a PR for that for your review,

Sounds good, but remember that we're not shooting for 100% docstrings -- we want docstrings (and documentation in general) when they will aid in understanding what is or should be going on.  If the purpose of the method is or should be obvious from its name and signature or its implementation is trivial, it may not need a docstring, especially if the method is private (indicated with single underscore prefix).

PR #506, opened 2021-05-27
[rid_qualifier]Accept JSON configuration as input

BenjaminPelletier:
This PR enables rid_qualifier to accept a single JSON-based configuration description (`RIDQualifierTestConfiguration`) for its test(s), which fills in the gaps of allowing the user to specify various parameters that were previously hard-coded.

`run_locally.sh` is updated to demonstrate how this configuration is provided, and also includes a multi-injection-target, multi-observer test, and is also adjusted so that it no longer needs to be run from a specific working directory.

The `auth_spec` is kept separate from the configuration because it often contains sensitive data (e.g., passwords) that would be insecure to store along side other configuration data.


hrishiballal (approved):
`test_executor.py` imports arrow that is not used / needed. Looks good otherwise, thanks!

BenjaminPelletier:
> `test_executor.py` imports arrow that is not used / needed. Looks good otherwise, thanks!

Fixed, thanks.

PR #508, opened 2021-05-28
[mock_ridsp] Add test flight data injection and observation of that data

BenjaminPelletier:
This PR enables mock_ridsp to accept incoming test data via the RID automated testing injection API, and then display that data via the standard RID interfaces.  This PR does not create an ISA in the DSS, but a subsequent PR will add this functionality.

This PR is split into two initial commits.  The first commit just refactors shared code related to automated testing (used by both rid_qualifier and rid_qualifier mock) into a single place such that mock_ridsp can effectively re-use it.  The second commit contains the substantive changes for this PR even though it appears to the smaller of the two commits.


monitoring/mock_ridsp/routes.py
```diff
@@ -29,3 +29,4 @@ def handle_exception(e):
 
 
 from . import routes_ridsp
+from . import routes_injection
```
hrishiballal, line 32: Is there a reason these are imported at the bottom?

hrishiballal: Never mind, I got it, there no way to delete a comment.

monitoring/mock_ridsp/routes_ridsp.py
```diff
@@ -1,30 +1,88 @@
...
 import datetime
+from typing import List, Optional
 
 import flask
+import s2sphere
 
-from monitoring.monitorlib import rid
+from monitoring.monitorlib import geo, rid
+from monitoring.monitorlib.rid_automated_testing.injection_api import TestFlight
+from monitoring.monitorlib.typing import StringBasedDateTime
```
hrishiballal, line 10: Nit: StringBasedDateTime takes both string and datetime, recommend to rename

monitoring/mock_ridsp/routes_ridsp.py
```diff
@@ -1,30 +1,88 @@
...
+  include_recent_positions = flask.request.args.get('include_recent_positions', 'False').lower() == 'true'
+
+  diagonal = view.lo().get_distance(view.hi()).degrees * geo.EARTH_CIRCUMFERENCE_KM / 360
+  if diagonal > rid.NetMaxDisplayAreaDiagonal:
+    msg = 'Requested diagonal of {} km exceeds limit of {} km'.format(diagonal, rid.NetMaxDisplayAreaDiagonal)
+    return flask.jsonify(rid.ErrorResponse(message=msg)), 413
 
+  now = arrow.utcnow().datetime
+  flights = []
+  for test_id, record in db.tests.items():
```
hrishiballal, line 71: What happens if by mistake two tests are submitted (with same flight tracks) for the same view, this will give the data twice. I guess I am asking in this API, there is no check here against a test_id

hrishiballal (approved):

monitoring/mock_ridsp/routes_ridsp.py
BenjaminPelletier: StringBasedDateTime is fundamentally a representation of a datetime, and the way it represents that datetime is as a string.  It can be constructed from multiple input formats, but it is not its the formats it can accept.  With this in mind, what would you suggest as a new name?

monitoring/mock_ridsp/routes_ridsp.py
BenjaminPelletier: When its `flights` endpoint is queried, any RID Service Provider should respond with all flights it manages within the area of interest.  There is no concept of a "test_id" when querying an SP (using the ASTM-standard interface) for its flights, nor should there be -- the data source of flights known to the SP is internal to that SP.  If two tests with the same tracks are injected into a single SP, that SP should indeed respond with two flights at identical positions when its `flights` are queried.

We fundamentally want to create as few parallel code paths as possible that are exercised only during the test -- instead, we're trying to test the entire system end-to-end.  Since a production Display Provider will not provide a test_id when querying a Service Provider, we should not provide that test_id when testing either.

mock_ridsp should not assign the same flight ID to two different flights, however, and that is what would happen currently.  In a future PR, we'll add the ability to adjust the injected test data to fix problems like this (or perhaps to reject the injection attempt).

monitoring/mock_ridsp/routes_ridsp.py
BenjaminPelletier: (will merge for now but address suggestions in a future PR)

hrishiballal: Dont really have a good suggest but perhaps `SemanticDateTime` ?

PR #509, opened 2021-06-03
Initial Commit / Adding Flight Passport OAUTH server to AuthSpec

hrishiballal:
First commit for new Auth Adapter based on the open source Flight Passport project

### DIFF
--- a/monitoring/monitorlib/auth.py
+++ b/monitoring/monitorlib/auth.py
@@ -14,7 +14,7 @@ import jwcrypto.jwk
 import jwcrypto.jws
 import jwcrypto.jwt
 import requests
-import json
+
 from google.auth.transport import requests as google_requests
 from google.oauth2 import service_account
 from monitoring.monitorlib.infrastructure import AuthAdapter
--- a/monitoring/monitorlib/auth.py
+++ b/monitoring/monitorlib/auth.py
@@ -261,7 +261,7 @@ class ClientIdClientSecret(AuthAdapter):
     self._oauth_token_endpoint = token_endpoint
     self._client_id = client_id
     self._client_secret = client_secret
-    self.send_request_as_data = send_request_as_data
+    self._send_request_as_data = send_request_as_data
     
   # Overrides method in AuthAdapter
   def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
@@ -273,7 +273,7 @@ class ClientIdClientSecret(AuthAdapter):
       'scope': ' '.join(scopes),
     }
 
-    if self.send_request_as_data: 
+    if self._send_request_as_data: 
       response = requests.post(self._oauth_token_endpoint, data=payload)
     else:
       response = requests.post(self._oauth_token_endpoint, json=payload)
@@ -287,14 +287,11 @@ class FlightPassport(ClientIdClientSecret):
 
   def __init__(self, token_endpoint: str, client_id: str, client_secret: str, send_request_as_data:str = 'true'):
     
-    send_request_as_data = True if (send_request_as_data.lower() == 'true') else False
+    send_request_as_data = (send_request_as_data.lower() == 'true')
       
     super(FlightPassport, self).__init__(token_endpoint, client_id, client_secret, send_request_as_data)
   
-    self._oauth_token_endpoint = token_endpoint
-    self._client_id = client_id
-    self._client_secret = client_secret
-    self.send_request_as_data = send_request_as_data
+    self._send_request_as_data = send_request_as_data
 
 
 class AccessTokenError(RuntimeError):
--- a/monitoring/monitorlib/auth.py
+++ b/monitoring/monitorlib/auth.py
@@ -14,9 +14,11 @@ import jwcrypto.jwk
 import jwcrypto.jws
 import jwcrypto.jwt
 import requests
+import json
 from google.auth.transport import requests as google_requests
 from google.oauth2 import service_account
-
+from monitoring.monitorlib.typing import ImplicitDict
+from monitoring.monitorlib.auth_utils import FlightPassportClientDetails
 from monitoring.monitorlib.infrastructure import AuthAdapter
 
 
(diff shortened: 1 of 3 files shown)


BenjaminPelletier:
Note that the unsuccessful checks are due to #510 and should be fixed by #511

monitoring/monitorlib/auth.py
```diff
@@ -119,6 +121,36 @@ def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
     return response.json()['access_token']
 
 
+class FlightPassport(AuthAdapter):
+  """ Auth adpater for Flight Passport OAUTH server (https://www.github.com/openskies-sh/flight_passport) """
+  
+  def __init__(self,token_endpoint: str,  service_account_json: FlightPassportClientDetails):
```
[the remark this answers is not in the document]
hrishiballal: OK I have now used the ClientIdClientSecret constructor.

monitoring/monitorlib/auth.py
```diff
@@ -119,6 +121,36 @@ def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
...
+    super().__init__()
+    
+    with open(service_account_json) as flight_passport_credentials: 
+      passport_credentials= json.loads(flight_passport_credentials.read())
```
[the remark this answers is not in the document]
hrishiballal: Thanks I have now removed this file reading part all together

monitoring/monitorlib/auth.py
```diff
@@ -119,6 +121,36 @@ def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
...
+    
+    data = {
+      'client_id':urllib.parse.quote(self.passport_config.client_id),
+      'client_secret': urllib.parse.quote(self.passport_config.client_secret),
+      'grant_type': "client_credentials",
+      'scope': scopes, 
+      'audience':urllib.parse.quote(intended_audience)
+
+    }
+    response = requests.post(self.token_endpoint, data = data)
```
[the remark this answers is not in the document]
hrishiballal: Yes, agreed, this is a quirk with [Django OAUTH toolkit](https://github.com/jazzband/django-oauth-toolkit/issues/807#issuecomment-602065242) that it does [not support JSON](https://github.com/jazzband/django-oauth-toolkit/issues/127#issuecomment-94439287) in the requests, when setting up the server you [can configure it](https://github.com/jazzband/django-oauth-toolkit/issues/127#issuecomment-94439287) . I will work on a flag implementation.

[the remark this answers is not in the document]
hrishiballal: I have now removed the file based input parameters so will fix the flag re JSON or Data

monitoring/monitorlib/auth.py
```diff
@@ -119,6 +121,36 @@ def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
...
+      
+    self.passport_config: FlightPassportClientDetails = ImplicitDict.parse(passport_credentials, FlightPassportClientDetails)      
+    
+    self.token_endpoint = token_endpoint
+
+  # Overrides method in AuthAdapter
+  def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
+    
+    data = {
+      'client_id':urllib.parse.quote(self.passport_config.client_id),
```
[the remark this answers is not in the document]
hrishiballal: Agreed, removing this.

monitoring/monitorlib/auth.py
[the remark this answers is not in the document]
hrishiballal: A first version is implemented now.

monitoring/monitorlib/auth.py
```diff
@@ -119,6 +121,36 @@ def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
...
+      'client_secret': urllib.parse.quote(self.passport_config.client_secret),
+      'grant_type': "client_credentials",
+      'scope': scopes,
```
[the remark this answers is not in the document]
hrishiballal: Thanks since we are now using the ClientIdClientCredentials, this is removed.

monitoring/monitorlib/auth.py
```diff
@@ -254,29 +253,46 @@ def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
...
+    if self.send_request_as_data: 
+      response = requests.post(self._oauth_token_endpoint, data=payload)
+    else:
+      response = requests.post(self._oauth_token_endpoint, json=payload)
     if response.status_code != 200:
       raise AccessTokenError('Unable to retrieve access token:\n' + response.content.decode('utf-8'))
     return response.json()['access_token']
 
 
+class FlightPassport(ClientIdClientSecret, AuthAdapter):
```
BenjaminPelletier, line 285: `ClientIdClientSecret` is already an `AuthAdapter` (no need for multiple inheritance), so I think we want:

```python
class FlightPassport(ClientIdClientSecret):
  def __init__(self, token_endpoint: str, client_id: str, client_secret: str):
    super(FlightPassport, self).__init__(token_endpoint, client_id, client_secret, True)
```

(but be careful of the value of that last argument if my other comment is addressed; `True` is not a `str`)

monitoring/monitorlib/auth.py
```diff
@@ -254,29 +253,46 @@ def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
 
 
 class ClientIdClientSecret(AuthAdapter):
-  """Auth adapter that gets JWTs using a client ID and client secret."""
+  """Auth adapter that gets JWTs using a client ID and client secret. By default, this will send the request as JSON, you can use send_request_as_data flag to send the request as form data. """
 
-  def __init__(self, token_endpoint: str, client_id: str, client_secret: str):
+  def __init__(self, token_endpoint: str, client_id: str, client_secret: str, send_request_as_data :bool=False):
```
BenjaminPelletier, line 258: Constructor arguments [may only be strings](https://github.com/interuss/dss/blame/master/monitoring/monitorlib/README.md#L19) and bool isn't a string.  To accept `send_request_as_data` as a `str`, convert to a bool with something like `send_request_as_data.lower() == 'true'`.

hrishiballal: Thanks, fixed this now.

monitoring/monitorlib/auth.py
hrishiballal: Thanks, yes I have now put in a check and converts it into a bool

monitoring/monitorlib/auth.py
```diff
@@ -254,33 +253,57 @@ def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
...
       raise AccessTokenError('Unable to retrieve access token:\n' + response.content.decode('utf-8'))
     return response.json()['access_token']
 
 
+class FlightPassport(ClientIdClientSecret):
+  """ Auth adpater for Flight Passport OAUTH server (https://www.github.com/openskies-sh/flight_passport) """
+
+  def __init__(self, token_endpoint: str, client_id: str, client_secret: str, send_request_as_data:str = 'true'):
+    
+    send_request_as_data = True if (send_request_as_data.lower() == 'true') else False
```
BenjaminPelletier, line 290: nit: can be simplified to `send_request_as_data = (send_request_as_data.lower() == 'true')` (or without parentheses) because the equality comparison operation returns a bool (True or False)

monitoring/monitorlib/auth.py
```diff
@@ -254,33 +253,57 @@ def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
...
+      
+    super(FlightPassport, self).__init__(token_endpoint, client_id, client_secret, send_request_as_data)
+  
+    self._oauth_token_endpoint = token_endpoint
```
BenjaminPelletier, line 294: When `super`'s `__init__` is called above, that's calling the super class's (ClientIdClientSecret) constructor, which [already sets these variables](https://github.com/openskies-sh/dss/blob/2cc9a91bf1820c282cb061de6a8aefe8c37ed3f7/monitoring/monitorlib/auth.py#L261); they don't need to be set here.

monitoring/monitorlib/auth.py
```diff
@@ -254,33 +253,57 @@ def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
...
 
-  def __init__(self, token_endpoint: str, client_id: str, client_secret: str):
+  def __init__(self, token_endpoint: str, client_id: str, client_secret: str, send_request_as_data :bool=False):
     super().__init__()
 
     self._oauth_token_endpoint = token_endpoint
     self._client_id = client_id
     self._client_secret = client_secret
-
+    self.send_request_as_data = send_request_as_data
```
BenjaminPelletier, line 264: nit: Python convention is for private class fields to be prefixed with a single underscore (see fields above this one).  Tools in a shared/reusable library generally carry a higher bar for documentation, so for fields that consumers of class instances will not need to use, a single underscore often removes the need to document the field by marking it as not for use by class instance consumers.  Otherwise, the field should usually be documented.

hrishiballal: Thanks 🙏🏼 , yes I have now fixed it.

monitoring/monitorlib/auth.py
hrishiballal: Thanks agreed, yes, I have now fixed this.

BenjaminPelletier (approved):

PR #513, opened 2021-06-04
[documentation] Improve pooling and deployment architecture documentation

BenjaminPelletier:


build/deploy/README.md
```diff
@@ -0,0 +1,99 @@
+# Kubernetes deployment via Tanka
```
pratibhagupta2109, line 1: @BenjaminPelletier Following use-case is not always guaranteed to be survived. 
For example, if we have these nodes in our cluster: [(n1, n2), (n3, n4), (n5,n6)]
Out of these n2, n4 and n6 are currently having replicated ranges. If I suspend n2 and n4 simultaneously, we lose some of the ranges.
Screen Shot 2021-06-07 at 1 38 39 pm
Screen Shot 2021-06-07 at 1 46 03 pm

BenjaminPelletier: Good catch -- thanks.  Will fix.

pratibhagupta2109 (approved):

PR #518, opened 2021-06-30
Get flight state records from KML.

pratibhagupta2109:


monitoring/monitorlib/kml.py
```diff
@@ -0,0 +1,101 @@
...
+
+def get_folder_description(folder_elem, to_json=True):
+    """Returns folder description from KML.
+    Args:
+        folder_elem: Folder element from KML.
+        to_json: If True: Returns description details as key: value pair, else a description string.
+    """
+    if to_json:
+        description = folder_elem.description
+        return dict([tuple((i.strip()).split(':')) for i in str(description).split('\n')])
```
BenjaminPelletier, line 83: I think this may lead to value strings having a leading space.  Either way, I'm observing the output JSON containing strings like `" Helicopter"` rather than `"Helicopter"`.  We may want to just `trim()` value strings to get rid of leading and trailing whitespace.

monitoring/rid_qualifier/create_flight_record_from_kml.py
```diff
@@ -0,0 +1,358 @@
...
+            track=198.57516619987473,
+            # TODO: Speed to be fetched relative to speed polygon.
+            speed=speed,
+            timestamp_accuracy=float(flight_description.get('timestamp_accuracy', '0.0')),
+            speed_accuracy=flight_description.get('speed_accuracy', ''),
+            # TODO: vertical_speed ?
+            vertical_speed=0.0)
+        flight_telemetry.append(rid_aircraft_state)
+    rid_details = RIDFlightDetails(
+        id=id,
```
BenjaminPelletier, line 176: `id` is not defined here, so we're getting the built-in Python function `id` rather than a data value that we'd like.  It would be great to take `id` from the folder description if present, or else use a UUID (`import uuid`, `str(uuid.uuid4())`).

monitoring/rid_qualifier/create_flight_record_from_kml.py
```diff
@@ -0,0 +1,358 @@
...
+        serial_number=flight_description.get('serial_number'),
+        operation_description=flight_description.get('operation_description'),
+        operator_location=LatLngPoint(lat=operator_location.get('lat'), lng=operator_location.get('lng')),
```
BenjaminPelletier, line 179: At this point, `operator_location` is a `FloatElement` rather than an actual `float`.  I'm not sure what a `FloatElement` is, but they can apparently be converted to plain floats with, e.g., `float(operator_location.get('lat'))`

monitoring/rid_qualifier/create_flight_record_from_kml.py
```diff
@@ -0,0 +1,358 @@
...
+        # flight_state_speeds.append(
+        #     get_interpolated_value(point, flattened_speed_polygons, all_polygon_speeds, round_value=True))
+    flight_state_vertices_unflatten = [unflatten(
+        s2sphere.LatLng.from_degrees(*reference_point[:2]), v) for v in flight_state_vertices]
+
+    # Position Lat, Lng to Lng, Lat order for KML representation.
+    flight_state_coordinates = []
+    for p, alt in zip(flight_state_vertices_unflatten, flight_state_altitudes):
+        flight_state_coordinates.append((
+            str(p.lng().degrees), str(p.lat().degrees), str(alt)
```
BenjaminPelletier, line 292: We want the JSON values of `lat`, `lng`, and `alt` to be numbers rather than strings, so I think we want to drop the `str` calls on this line

monitoring/rid_qualifier/create_flight_record_from_kml.py
```diff
@@ -0,0 +1,358 @@
...
+        # TODO: RIDHeight values ?
+        aircraft_height = RIDHeight(distance=ALTITUDE_AGL, reference="TakeoffLocation")
+        rid_aircraft_state = RIDAircraftState(
+            timestamp=timestamp_isoformat,
+            # TODO: operational_status ?
+            operational_status="Airborne",
+            position=aircraft_position,
+            height=aircraft_height,
+            # TODO: track ?
+            track=198.57516619987473,
```
BenjaminPelletier, line 167: It's fine to leave this as a TODO for this PR, but let's make sure it's obvious this isn't a real value -- so, let's use, e.g., 99999 instead of this value.

monitoring/rid_qualifier/README.md
```diff
@@ -10,6 +10,15 @@ This directory contains a series of tests for qualifying Network Remote ID compl
...
 
+5. **Create Flight Record from KML**: [create_flight_record_from_kml.py](create_flight_record_from_kml.py) accepts a KML file with one/many flights defined in the KML folders and produce a set of JSON files for each such flight, snapshoting the aircraft's state every sample_rate. Every flight needs exactly one path (LineString) and this is the path the aircraft takes over the ground. Speed and altitude of the flight  are defined by the polygons surrounding the path.
+
+- To run script, Set PYTHONPATH to path to the working directory `dss`
+- mark create_flight_record_from_kml.py executable using
```
BenjaminPelletier, line 16: nit: Note that another way to run this tool (instead of `chmod`ing the .py file to be excutable) is to just run `python ./create_flight_record_from_kml.py <args>` -- that might be simpler than making the Python file executable, and that approach is a bit more cross-platform since Windows doesn't have a concept of an executable attribute.

monitoring/rid_qualifier/create_flight_record_from_kml.py
```diff
@@ -0,0 +1,358 @@
...
+    for p, alt in zip(flight_state_vertices_unflatten, flight_state_altitudes):
+        flight_state_coordinates.append((
+            str(p.lng().degrees), str(p.lat().degrees), str(alt)
+        ))
+    return flight_state_coordinates, flight_state_speeds
+
+
+def write_to_json_file(data, file_name, output_folder):
+    with open(f'{output_folder}/{file_name}.json', 'w') as outfile:
+        outfile.write(jsonpickle.encode(data, unpicklable=False))
```
BenjaminPelletier, line 299: After making sure the `data` doesn't have any non-serializable elements (right now, `id`, `lat`, and `lng` are unserializable; see other comments), we can use `json.dumps` and don't need to use `jsonpickle`.

monitoring/rid_qualifier/create_flight_record_from_kml.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+def create_output_folder(folder_path):
+    directory = os.path.dirname(folder_path)
```
BenjaminPelletier, line 303: This is grabbing the directory name assuming that `folder_path` is describing a file, so `folder_path='monitoring/rid_qualifier/test_data'` is resulting in the creation of `monitoring/rid_qualifier`, but not `monitoring/rid_qualifier/test_data`.  As a result, if `test_data` doesn't already exist, the script will crash when trying to `write_to_json_file`.

monitoring/monitorlib/kml.py
```diff
@@ -0,0 +1,101 @@
...
+def get_folders(root):
+    return root.Document.Folder.Folder
+
+
+def get_polygon_speed(polygon_name):
+    """Returns speed unit within a polygon."""
+    result = re.search(r"\(([0-9.]+)\)", polygon_name)
+    return float(result.group(1)) if result else None
+
+def get_folder_details(folder_elem):
```
BenjaminPelletier, line 23: This function is specific enough to rid_qualifier that it seems unlikely we would reuse it in another general `monitoring` tool.  So, we should probably put it in monitoring/rid_qualifier/kml.py rather than here (monitoring/monitorlib/kml.py presumably contains functions and tools that can be reused over many different monitoring tools).

monitoring/monitorlib/kml.py
```diff
@@ -0,0 +1,101 @@
...
+       return [tuple(float(x.strip()) for x in c.split(',')) for c in str(coordinates[0]).split(' ') if c.strip()]
+
+
+def get_folder_description(folder_elem, to_json=True):
+    """Returns folder description from KML.
+    Args:
+        folder_elem: Folder element from KML.
+        to_json: If True: Returns description details as key: value pair, else a description string.
+    """
+    if to_json:
```
BenjaminPelletier, line 81: Do we ever call this function with `to_json=False`?  If not, we should probably remove that branch for simplicity.

monitoring/monitorlib/kml.py
```diff
@@ -0,0 +1,101 @@
...
+    operator_location = {}
+    coordinates = ''
+    for placemark in folder_elem.xpath('.//kml:Placemark', namespaces=KML_NAMESPACE):
+        placemark_name = str(placemark.name)
+        polygons = placemark.xpath('.//kml:Polygon', namespaces=KML_NAMESPACE)
+
+        if placemark_name == 'operator_location':
+            operator_location = {
+                'lng': folder_elem.xpath(
+                    './/kml:Placemark/kml:LookAt/kml:longitude', namespaces=KML_NAMESPACE)[0],
```
BenjaminPelletier, line 35: [The LookAt tag](https://developers.google.com/kml/documentation/kmlreference#lookat) defines where the camera should be looking when the user double-clicks this Placemark in Google Earth, but I think we want to extract the actual location of the Placemark itself instead (e.g., 39.078915 latitude instead of 39.077887 latitude for test_flight_north)

monitoring/monitorlib/kml.py
```diff
@@ -0,0 +1,101 @@
...
+                    './/kml:Placemark/kml:LookAt/kml:latitude', namespaces=KML_NAMESPACE)[0]
+            }
+        if polygons:
+            if placemark_name.startswith('alt:'):
+                polygon_coords = get_coordinates_from_kml(
+                    polygons[0].outerBoundaryIs.LinearRing.coordinates)
+                alt_polygons.update({placemark_name: polygon_coords})
+            if placemark_name.startswith('speed:'):
+                if not get_polygon_speed(placemark_name):
+                    # TODO: raise error
```
BenjaminPelletier, line 46: To avoid failing silently, I think we can go ahead and actually raise the error here (e.g., `raise ValueError('Could not determine Polygon speed from Placemark "{}"'.format(placemark_name))`)

monitoring/rid_qualifier/test_create_flight_record_from_kml.py
```diff
@@ -0,0 +1,21 @@
+from monitoring.rid_qualifier import create_flight_record_from_kml as frk
+
+def test_get_interpolated_altitude(mocker):
```
BenjaminPelletier, line 3: Having tests is great -- let's write a brief comment in this file (or perhaps in this folder's readme) saying how someone should invoke these tests

monitoring/rid_qualifier/create_flight_record_from_kml.py
```diff
@@ -0,0 +1,358 @@
...
+    if not os.path.exists(directory):
+        os.makedirs(directory)
+
+def main(kml_file, debug_mode=None):
+    # kml_file = 'monitoring/rid_qualifier/test_data/dcdemo.kml'
+    kml_content = kml.get_kml_content(kml_file)
+    flight_state_coordinates = {}
+    output_folder = 'monitoring/rid_qualifier/test_data'
```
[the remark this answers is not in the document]
pratibhagupta2109: I still need to work on passing the input/output folders.

monitoring/rid_qualifier/README.md
```diff
@@ -10,6 +10,15 @@ This directory contains a series of tests for qualifying Network Remote ID compl
 
 4. **Display Data Evaluator**: Once test aircraft state data is sent to each Remote ID Service Provider's injection test harness to be injected into their systems, the state of the RID system must be queried and then the results of those queries compared against expected results. This module performs the latter tasks of querying the RID system through each specified Remote ID Display Provider's observation test harness, and comparing those query results against expectations. [Flight Blender](https://github.com/openskies-sh/flight-blender) is an open-source Remote ID Display Provider that will implement an observation test harness, and therefore can be used to test compliance.
 
+5. **Create Flight Record from KML**: [create_flight_record_from_kml.py](create_flight_record_from_kml.py) accepts a KML file with one/many flights defined in the KML folders and produce a set of JSON files for each such flight, snapshoting the aircraft's state every sample_rate. Every flight needs exactly one path (LineString) and this is the path the aircraft takes over the ground. Speed and altitude of the flight  are defined by the polygons surrounding the path.
```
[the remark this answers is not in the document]
pratibhagupta2109: Updated the document with more details.

monitoring/monitorlib/kml.py
pratibhagupta2109: `i.strip()` was meant to remove the spaces but it required some modifications. done.

monitoring/rid_qualifier/create_flight_record_from_kml.py
pratibhagupta2109: done.

monitoring/rid_qualifier/create_flight_record_from_kml.py
pratibhagupta2109: done.

monitoring/rid_qualifier/create_flight_record_from_kml.py
pratibhagupta2109: done.

monitoring/rid_qualifier/README.md
pratibhagupta2109: Updated.

monitoring/rid_qualifier/create_flight_record_from_kml.py
pratibhagupta2109: done.

monitoring/rid_qualifier/create_flight_record_from_kml.py
pratibhagupta2109: Done.

monitoring/rid_qualifier/test_create_flight_record_from_kml.py
BenjaminPelletier: Let's still write this comment, either in this file (probably preferable) or in the README, even though I think it is as simple as running `pytest` from this working directory.  However, also note that the tests fail currently, I think because the function was renamed from `get_interpolated_altitude` to `get_interpolated_value`.

PR #520, opened 2021-07-08
rid_qualifier web-server initial setup.

pratibhagupta2109:


monitoring/rid_qualifier/host/config.py
```diff
@@ -0,0 +1,32 @@
+import os
+
+from monitoring.monitorlib import auth_validation
+
+
+ENV_KEY_PREFIX = 'MOCK_HOST'
+ENV_KEY_PUBLIC_KEY = '{}_PUBLIC_KEY'.format(ENV_KEY_PREFIX)
```
BenjaminPelletier, line 7: I think it's safe to remove `*PUBLIC_KEY`, `*TOKEN_AUDIENCE`, `*BASE_URL`, and `*DSS` from this configuration as we won't use them, even in the future, for this tool.  We should also remove them from run_locally.sh.

monitoring/rid_qualifier/host/auth.py
```diff
@@ -0,0 +1,8 @@
+from monitoring.monitorlib import auth_validation
+from monitoring.rid_qualifier.host import webapp
+from . import config
+
+
+requires_scope = auth_validation.requires_scope_decorator(
```
BenjaminPelletier, line 6: We won't be validating authorization like this for this tool, so I think this file can be deleted.

monitoring/rid_qualifier/host/mockuss.py
```diff
@@ -0,0 +1,12 @@
+import sys
+
+from monitoring.rid_qualifier.host import webapp
+
+
+def main(argv):
```
BenjaminPelletier, line 6: Let's rename this file `debug_host.py` -- it's basically the debugging entrypoint into the `webapp` defined in this module (instead of using `gunicorn` like run_locally.sh does)

BenjaminPelletier (commented):
This looks great and almost ready to merge -- just a few items, mostly pruning.

monitoring/rid_qualifier/host/config.py
pratibhagupta2109: Done.

monitoring/rid_qualifier/host/mockuss.py
pratibhagupta2109: Done.

BenjaminPelletier (approved):

PR #522, opened 2021-07-27
[rid_qualifier] Run rid_qualifier from rid_qualifier host

pratibhagupta2109:
RID host backend changes.


monitoring/rid_qualifier/test_executor.py
```diff
@@ -40,3 +56,9 @@ def main(test_configuration: RIDQualifierTestConfiguration, auth_spec: str):
         report.findings)
     with open('report.json', 'w') as f:
         json.dump(report, f)
+    if job_id:
```
BenjaminPelletier, line 59: Instead of adding extra behaviors to `main`, let's split out most of the functionality originally in `main` into something like `conduct_test` which only returns `report` (and doesn't write it to file).  Then, `main` can call `conduct_test` and write the result to file while the Redis worker can do something else with the report.  Importantly, `rid_qualifier` by itself should probably not need to do anything with Redis (currently, `main` is trying to get the current Redis job) since it's just the `host` which is employing a Redis worker to run it.

monitoring/rid_qualifier/host/forms.py
```diff
@@ -0,0 +1,20 @@
...
+class UserConfig(FlaskForm):
+    auth_spec = StringField('Auth Spec', validators=[DataRequired()])
+    user_config = TextAreaField('User Config', validators=[DataRequired()])
+    submit = SubmitField('Submit')
+
+    def validate_user_config(form, field):
+        try:
+            user_config = json.loads(field.data)
+        except ValueError as e:
+            raise ValueError(e)
```
BenjaminPelletier, line 16: If the only thing we do here in case of a ValueError during `json.loads` is re-raise the error, we can just drop the try block and allow `json.loads` to raise the original error uncaught.

BenjaminPelletier (commented):
When I try to run rid_qualifier host against rid_qualifier mock, my rq-worker container is producing this error message:

```
FileNotFoundError: [Errno 2] No such file or directory: '/app/monitoring/rid_qualifier/test-data'
```

It doesn't look like host/run_locally.sh mounts a volume in this folder, and there is no rid_qualifier/test-data folder in the repo (though there is a rid_qualifier/test_data folder) -- where should the content for this folder come from?

monitoring/rid_qualifier/host/README.md
```diff
@@ -1,10 +1,10 @@
...
 This PR includes the initial setup for the web-server that can be run locally on your system via the (run_locally.sh)[run_locally.sh] script. 
 
-Upcoming feature PRs will include:
-- A page with "Start test" button to execute the qualifier.
-- Ability to run the job as a background task by adding it to the Redis Queue.
-- Accept configuration object and auth spec as user input that can be passed as arguments to the code running as Redis worker.
-- Invoke test_executor.main throguh the Redis worker.
-- Update test_executor.main to return a Report object.
-- Make the resulting report retrievable from the web interface.
+**Prerequisite**: Local DSS instance should be started by running (monitoring/rid_qualifier/mock/run_locally.sh)[monitoring/rid_qualifier/mock/run_locally.sh] script.
```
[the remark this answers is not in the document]
pratibhagupta2109: Updated README file. Please check.

monitoring/rid_qualifier/aircraft_state_replayer.py
```diff
@@ -12,22 +12,24 @@
...
 
-from typing import List
+from typing import List, Optional
 from monitoring.rid_qualifier.utils import RIDQualifierTestConfiguration
 
 class TestBuilder():
     ''' A class to setup the test data and create the objects ready to be submitted to the test harness '''
 
-    def __init__(self, test_configuration: RIDQualifierTestConfiguration) -> None:
+    def __init__(
```
BenjaminPelletier, line 21: nit: when breaking a line into multiple lines, the following lines should be indented twice to distinguish the arguments from the code that follows

monitoring/rid_qualifier/host/README.md
```diff
@@ -1,9 +1,38 @@
...
+
+* redis Container: A redis server to handle background tasks.
+
+* rq-worker Container: A container to run redis queue worker processes.
+
+* rid-host Container: A flask application which accepts flight states' files, collect auth and config spec from user and executes RID Qualifier task. Once task finishes successfully, report.json file is available to download.
+
+## Run via Docker
+
+You should first create a new Python virtual environment to install the
```
BenjaminPelletier, line 18: Is there a need to set up/activate an appropriate virtual environment if Docker is going to be used?  Generally, the appropriate virtual environment is built inside the Docker images so it doesn't need to be built on the host machine, and the run_locally.sh file below just uses Docker

monitoring/rid_qualifier/host/run_locally.sh
```diff
@@ -11,7 +11,7 @@ fi
 cd "${BASEDIR}/../../.." || exit 1
 
 # This sample assumes that a local DSS instance is available similar to the one
-# produced with /build/dev/run_locally.sh
+# produced with monitoring/rid_qualifier/mock/run_locally.sh
```
BenjaminPelletier, line 14: Running monitoring/rid_qualifier/mock/run_locally.sh produces a mock RID system for use with rid_qualifier (rather than a DSS instance, as the previous line currently says)

monitoring/rid_qualifier/host/run_locally.sh
```diff
@@ -37,6 +37,7 @@ docker run --name rq-worker -d --rm \
   -e MOCK_HOST_DSS_URL="${DSS}" \
   -e REDIS_URL=redis://redis-server:6379/0 \
   -v `pwd`/build/test-certs:/var/test-certs:ro \
+  -v /tmp:/mnt/app/input-files \
```
BenjaminPelletier, line 40: Do we need to mount a volume here?  Mounting a volume enables the container to 1) access files pre-existing on the host system and 2) create or modify files on the host system.  Do we need either of these things here?  If not, we can drop the volume mount at runtime (this argument), and instead add a `RUN mkdir -p /mnt/app/input-files` (or any other directory name) in the Dockerfile to create a folder that exists only within the container.  If 1 above is needed, what files should pre-exist in /tmp and where have we documented that?  If 2 above is needed, what files are produced/modified by rid_qualifier host and where is that documented?

monitoring/rid_qualifier/host/README.md
```diff
@@ -1,10 +1,39 @@
...
+
+Change the root directory to dss/. Run the following command to start new containers:
+
+```bash
+(venv) $ ./monitoring/rid_qualifier/host/run_locally.sh
+```
+
+## Task Description
+
+You can now test the application by uploading flight states json files. Input files are **required** to execute the task. When the flask application container is booted, a volume is
```
BenjaminPelletier, line 34: Let's be a little more detailed here -- specifically, which flight record files should be uploaded, and where do they come from? (presumably, from running fligh_data_generator.py would be the simplest, though create_flight_record_from_kml.py can produce them too).  I think we also need to tell the user they need an RID system to test, and the easiest way to have one of those is to run an instance of rid_qualifier mock.

Basically, we want to make sure we fully describe the simplest process to demonstrate that the tool works (performs its basic function).  To do that, I think we want to have a list of steps to describe the simplest way to actually get a report.json that is the result of a real test. I think those instructions would probably include the above (running host/run_locally.sh, uploading flight record files), but also running flight_data_generator to create the files, uploading the flight record file(s), bringing up an rid_qualifier mock instance, and using NoAuth() as the auth spec and the User config from rid_qualifier/run_locally.sh.

monitoring/rid_qualifier/test_executor.py
```diff
@@ -1,16 +1,17 @@
...
 from monitoring.rid_qualifier import display_data_evaluator, reports
 from monitoring.monitorlib.infrastructure import DSSTestSession
 from monitoring.monitorlib.auth import make_auth_adapter
 
 
-def main(test_configuration: RIDQualifierTestConfiguration, auth_spec: str):
-    # This is the configuration for the test.
-    my_test_builder = TestBuilder(test_configuration = test_configuration)
+def main(
+    test_configuration: RIDQualifierTestConfiguration,
```
BenjaminPelletier, line 12: nit: Same as above: when splitting lines, double-indent line continuations to differentiate the single split line from the following lines of code.  Or, indent to match an open paren -- both of the following are ok:

```python
def example(
        p1: Parameter,
        p2: Parameter):
    pass
````

```python
def example(p1: Parameter,
            p2: Parameter):
    pass
````

We want to avoid:

```python
def example(
    p1: Parameter,
    p2: Parameter):
    pass
````

...because it's hard to see the difference between `p1` & `p2` and `pass`

BenjaminPelletier (commented):
Note also that another PR has been merged into master, so you'll need to merge master into this branch to grab those changes and verify everything still works properly

BenjaminPelletier (approved):

PR #524, opened 2021-07-28
rid_qualifier: Reducing query variation in the display data evaluator

hrishiballal:
Added a new test config reduce query variations, added to keep every third view box to be the same, refactor to have rect generation before observation

### DIFF
--- a/monitoring/rid_qualifier/utils.py
+++ b/monitoring/rid_qualifier/utils.py
@@ -28,6 +28,9 @@ class EvaluationConfiguration(ImplicitDict):
     min_query_diagonal: float = 100
     """Do not make queries with diagonals smaller than this many meters."""
 
+    reduce_query_variation: bool = True
+    """Reduce variations in queries performed during obsverations."""
+
 
 class RIDQualifierTestConfiguration(ImplicitDict):
     locale: str
--- a/monitoring/rid_qualifier/utils.py
+++ b/monitoring/rid_qualifier/utils.py
@@ -30,6 +30,10 @@ class EvaluationConfiguration(ImplicitDict):
 
     reduce_query_variation: bool = True
     """Reduce variations in queries performed during obsverations."""
+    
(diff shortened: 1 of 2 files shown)


monitoring/rid_qualifier/utils.py
```diff
@@ -28,6 +28,9 @@ class EvaluationConfiguration(ImplicitDict):
     min_query_diagonal: float = 100
     """Do not make queries with diagonals smaller than this many meters."""
 
+    reduce_query_variation: bool = True
```
BenjaminPelletier, line 31: Let's make this behavior more self-documenting at the same time as making it more flexible:

```python
repeat_query_rect_period: int = 3
"""If set to a value above zero, reuse the most recent query rectangle/view every this many queries."""
```

and

```python
if (last_rect and
    config.reduce_query_variation > 0 and
    query_counter % config.reduce_query_variation == 0):
```

BenjaminPelletier (commented):
Note that the failing CI check is not the fault of this PR; I will try to address that separately.

hrishiballal:
For some reason EasyCLA is erroring out, I will try again.

BenjaminPelletier:

monitoring/rid_qualifier/utils.py
BenjaminPelletier: `reduce_query_variation` can now be removed -- since `repeat_query_rect_period=0` doesn't make sense, we can use this as the value to disable the feature.  So, the check can be `config.repeat_query_rect_period > 0` rather than `config.reduce_query_variation > 0` which we currently have but doesn't make sense any way because `config.reduce_query_variation` is a `bool` not an `int`.

BenjaminPelletier:
Current status is now that EasyCLA issue has been resolved and all checks pass; just [one more comment](https://github.com/interuss/dss/pull/524#discussion_r681111276) to address.

BenjaminPelletier (approved):

PR #526, opened 2021-07-30
[mock_ridsp] Fix ISA duration by adding trailing buffer

BenjaminPelletier:
mock_ridsp previously created an ISA that tightly wrapped the injected positions, but this was non-compliant with NET0610 which requires that the flight be encompassed in an ISA for 60 seconds after the end of the flight.  This PR buffers the end time of the ISA more than 60 seconds past the last position report to be compliant with NET0610.


BenjaminPelletier:

PR #527, opened 2021-07-31
Add script to run flight_data_generator of rid_qualifier in the docker container

barroco:
This PR adds a script to run flight_data_generator.py in the docker container. It allows a developer to generate the initial test dataset without installing the python dependencies on his local host.


BenjaminPelletier (approved):
This looks good; we seem to be having EasyCLA issues (see 3 of the other most recent PRs also), so I will comment again when that should be resolved.

PR #528, opened 2021-08-01
[rid_qualifier] Resolve TODOs in create_flight_record_from_kml

pratibhagupta2109:


BenjaminPelletier:

monitoring/rid_qualifier/create_flight_record_from_kml.py
```diff
@@ -152,24 +162,20 @@ def generate_flight_record(
...
             accuracy_v=flight_description.get('accuracy_v'),
-            # TODO: need to check value for extrapolated
+            # TODO: extrapolated ?
             extrapolated=False,
             )
-        # TODO: RIDHeight values ?
         aircraft_height = RIDHeight(distance=ALTITUDE_AGL, reference="TakeoffLocation")
         rid_aircraft_state = RIDAircraftState(
             timestamp=timestamp_isoformat,
             # TODO: operational_status ?
```
BenjaminPelletier, line 171: Hardcoding this operation_status is fine.

monitoring/rid_qualifier/create_flight_record_from_kml.py
```diff
@@ -48,23 +42,36 @@ def check_if_vertex_is_correct(point1, point2, point3, flight_distance):
         ) < 0.2, f'Generated vertex is not correct. {point1}, {point2}, {point3}, {flight_distance}'
 
 
+def get_track_angle(point1, point2):
+    x1, y1 = point1
+    x2, y2 = point2
+    if x2 == x1:
+        return 0
+    else:
+        return math.degrees(math.atan((y2 -y1)/(x2 - x1)))
```
BenjaminPelletier, line 51: Looks like this isn't quite what we want.  A north track should be 0, east should be 90, south 180, and west 270 ("degrees east of north").  An increase in `x` here should indicate that longitude is increasing (an eastward movement) and an increase in `y` should indicate that latitude is increasing (a northward movement).  Let's at least test all the cardinal directions (N, E, S, W).

Also, we'll sometimes get no movement in the denominator (regardless of whether it's `x` or `y`) and we still want this to compute correctly.  So, let's use [`atan2`](https://docs.python.org/3/library/math.html#math.atan2).  Since this might be tricky to ensure we get right, we can add a unit test that verifies we obtain known angles with known `point1` and `point2`.  For instance, if `point2.x` = `point1.x` and `point2.y` = `point2.y + <something positive>`, then we know the track should be `0` (north).

monitoring/rid_qualifier/create_flight_record_from_kml.py
```diff
@@ -152,24 +162,20 @@ def generate_flight_record(
             alt=coordinates[2],
             accuracy_h=flight_description.get('accuracy_h'),
             accuracy_v=flight_description.get('accuracy_v'),
-            # TODO: need to check value for extrapolated
+            # TODO: extrapolated ?
             extrapolated=False,
             )
-        # TODO: RIDHeight values ?
         aircraft_height = RIDHeight(distance=ALTITUDE_AGL, reference="TakeoffLocation")
```
BenjaminPelletier, line 168: Let's actually omit the `RIDHeight` since it is optional and it will be tricky to compute reasonably-correct values for it.  That will also make it so we don't need `ALTITUDE_AGL`.

monitoring/rid_qualifier/create_flight_record_from_kml.py
pratibhagupta2109: Updated with supported use-cases.

monitoring/rid_qualifier/test_create_flight_record_from_kml.py
```diff
@@ -25,3 +25,47 @@ def test_get_interpolated_value(mocker):
     result_alt = frk.get_interpolated_value('point', 'polygons', [140,125,116])
     assert abs(result_alt - 135.84) < 0.1
     
+def test_get_track_angle():
```
BenjaminPelletier, line 28: Beautiful :)

BenjaminPelletier (approved):
I missed this before and not worth blocking this PR for, but we should fix some of the `raise` instructions.  Raising a `str` is not valid; instead, an Error/Exception has to be raised.  So, instead of

```python
raise 'Path to output folder not provided.'
```

...it should be

```python
raise ValueError('Path to output folder not provided.')
```

monitoring/rid_qualifier/create_flight_record_from_kml.py
pratibhagupta2109: Getting error on removing RIDHeight value
raise ValueError('Expected to find dictionary data to populate {} object but instead found {} type'.format(parse_type.__name__, type(source).__name__))
ValueError: Expected to find dictionary data to populate RIDHeight object but instead found NoneType type

pratibhagupta2109 (commented):
Value is expected for RIDHeight. Getting following error:

**ValueError: Expected to find dictionary data to populate RIDHeight object but instead found NoneType type**

PR #529, opened 2021-08-03
Added aircraft_state_files parameter per the new requirements

hrishiballal:
Added required variable when calling test_executor.py


BenjaminPelletier:
This argument is intended to be optional, so I think the better fix is to correctly make it optional rather than providing an explicit value here -- I recommend #530 instead of this one.

hrishiballal:
OK, thanks I had synced with upstream and it was giving this issue, I will close off this pull request.

PR #530, opened 2021-08-03
[rid_qualifier] Fix issues from #527 and #522

BenjaminPelletier:


BenjaminPelletier:
CI checks pending resolution of [a dlv issue](https://github.com/go-delve/delve/issues/2639)

pratibhagupta2109 (approved):

hrishiballal (approved):

monitoring/rid_qualifier/aircraft_state_replayer.py
```diff
@@ -20,7 +20,7 @@ class TestBuilder():
 
     def __init__(
             self, test_configuration: RIDQualifierTestConfiguration,
-            aircraft_state_files: Optional[list]) -> None:
+            aircraft_state_files: Optional[list] = None) -> None:
```
pratibhagupta2109, line 23: when a type is wrapped with Optional, e.g. Optional[list], isn't it same as default None?

BenjaminPelletier: No; type hints merely indicate what kind of data may be present in the variable.  The `Optional` type hint means that the data may either be the type argument, or else `None` -- so, for `def f(x: Optional[MyType])`, `x` may either be an instance of `MyType`, or it may be `None`.  If `None` were passed as `x` to `def f(x: MyType)`, the static type checker (lint) would complain that `None` is not an instance of `MyType`.  Here's an example summary:

| Definition | `f([])` | `f(None)` | `f()` | `x` value if omitted |
|-|-|-|-|-|
| `def f(x: List)` | Valid | Invalid (lint) | Invalid (argument error) | N/A |
| `def f(x: Optional[List])` | Valid | Valid | Invalid (argument error) | N/A |
| `def f(x: List = [])` | Valid | Invalid (lint) | Valid | `[]` |
| `def f(x: Optional[List] = None)` | Valid | Valid | Valid | `None` |
| `def f(x: Optional[List] = [])` | Valid | Valid | Valid | `[]` |

PR #531, opened 2021-08-05
[rid_qualifier] Add user identity via SSO in `host`

pratibhagupta2109:
Please suggest where should we keep client_secret file for secure access.


BenjaminPelletier:
Looks like this branch needs to be synced with master (merge `master` into `rid-host`)

monitoring/rid_qualifier/host/routes.py
```diff
@@ -1,69 +1,148 @@
...
+def logout():
+    session.clear()
+    return render_template('logout.html')
+
+
 @webapp.route('/')
+@login_required
 def home_page():
-  return render_template('home.html', title='Home', greetings='Hello RID Host !!')
+    if session.get('google_id'):
```
BenjaminPelletier, line 84: Can we remove this check for `google_id` since this function is decorated with `@login_required`?

BenjaminPelletier (commented):
To provide `client_secret.json`, let's add `client_secret.json` to a new `.gitignore` file in the `host` folder.  Then, let's add [some logic](https://linuxize.com/post/bash-check-if-file-exists/) to `host`'s `run_locally.sh` to raise an error pointing to documentation on how to fix the problem if `client_secret.json` does not exist.  Then, we can mount it to the container by passing `-v $(pwd)/client_secret.json:/app/monitoring/rid_qualifier/host/client_secret.json` when starting the Docker container.

We will also need to add instructions to `host`'s README.md explaining how a GCP project must be configured to act as an SSO provider here, including where to get the `GOOGLE_CLIENT_ID` and `client_secret.json`.

This may be in a follow-up PR, but it would also be nice to condition SSO user validation on whether the user provides `client_secret.json` or not -- if provided, we would validate users as in this PR.  Otherwise, we wouldn't validate users at all (and we would probably want to print a warning message to that effect on startup).  A small change to the `login_required` decorator should let us do this without too much trouble.

BenjaminPelletier:
I believe this PR is obsolete after merging #534; should we close this PR?

PR #534, opened 2021-08-18
[rid_qualifier] Update `Run Tests` page with I/O files

pratibhagupta2109:


BenjaminPelletier (approved):
Overall this looks really great!  It looks like there's a bug where one test run generates a bunch of JSON files:
image
That screenshot is from running the rid_qualifier host against the rid_qualifier mock using the 6 aircraft states produced by the flight_data_generator.

A different time I ran it, I never seemed to get any results.

One small thing that would be nice in a future PR would be to have the flight files' boxes checked by default immediately after they are uploaded.

Another thing we should address in a separate PR is that when I navigate to the /tests page after kicking off a test, I don't see an indication that a test is in progress -- that would be good to have.

I think this is a great base to work from though.

PR #535, opened 2021-08-20
[rid_qualifier host] Bypass identity checks when client_secret is not provided

pratibhagupta2109:


monitoring/rid_qualifier/host/config.py
```diff
@@ -10,22 +10,21 @@
 
 # These keys map to entries in the Config class
 KEY_AUTH_SPEC = 'AUTH_SPEC'
+ENV_KEY_REDIS_URL = os.environ['REDIS_URL']
```
BenjaminPelletier, line 13: These values should be fixed and be located next to `ENV_KEY_PREFIX` above -- these are the names of the environment variables that used to populate each variable.  So, for instance, `ENV_KEY_REDIS_URL = 'REDIS_URL`

monitoring/rid_qualifier/host/config.py
```diff
@@ -10,22 +10,21 @@
...
+ENV_KEY_PORT = os.environ['PORT']
 
 
 workspace_path = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'workspace')
 os.environ["OAUTHLIB_INSECURE_TRANSPORT"] = "1"
 
 
 class Config(object):
-  AUTH_SPEC = os.environ[ENV_KEY_AUTH]
-  REDIS_URL = os.environ['REDIS_URL']
```
BenjaminPelletier, line 21: We still want these variables defined down here in the `Config`, we just want to use, e.g., `ENV_KEY_REDIS_URL` rather than the magic string `'REDIS_URL'`.

monitoring/rid_qualifier/host/routes.py
```diff
@@ -47,8 +46,12 @@ def info():
 def login_required(function):
     @wraps(function)
     def decorated_function(*args, **kwargs):
-        if 'google_id' not in session:
+        if flow and 'google_id' not in session:
             return redirect(url_for('login', next=request.url))
+        elif 'google_id' not in session:
+            session['google_id'] = 'testuser'
```
BenjaminPelletier, line 52: Let's just use `'localuser'` since an individual developer may want to run rid_qualifier host on their own local machine to do real (not just test) development (and likewise, `'Local User'` for `session['name']`; I'm not sure what `session['state']` should be)

BenjaminPelletier (approved):

PR #537, opened 2021-08-28
[prober] Add API-dependence tool for prober tests

BenjaminPelletier:
To prepare for upgrading tests to support SCD DSS API version 0.3.15 (or later), this PR introduces a tool to annotate prober tests as supporting a particular set of API versions.

### DIFF
--- a/monitoring/prober/conftest.py
+++ b/monitoring/prober/conftest.py
@@ -1,7 +1,8 @@
 from typing import Optional
 
 from monitoring.monitorlib import infrastructure
-from monitoring.monitorlib import auth
+from monitoring.monitorlib import auth, scd
+from monitoring.prober.infrastructure import VersionString
 
 import pytest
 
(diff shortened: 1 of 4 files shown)


monitoring/prober/conftest.py
```diff
@@ -53,3 +55,10 @@ def scd_session2(pytestconfig):
 @pytest.fixture(scope='function')
 def no_auth_session(pytestconfig):
   return make_session(pytestconfig, '/v1/dss')
+
+@pytest.fixture(scope='session')
+def scd_api(pytestconfig):
+  api = pytestconfig.getoption('scd_api_version')
+  if api is None:
+    api = scd.API_0_3_5
+  return VersionString(api)
```
pratibhagupta2109, line 64: What is the purpose of wrapping api string to VersionString when class has no properties?

BenjaminPelletier: In `pytest`, test "fixtures" are provided by name according to the signature of the test.  The decorator (`for_api_versions`) needs to look at this `scd_api` test fixture (or a similar test fixture for RID), but it only has access to `*args` and `**kwargs` and doesn't know the signature of the function being wrapped.  To determine which argument is the `scd_api` test fixture, it looks at the class (`isinstance(value, VersionString)`).  This could present some difficulty if both `scd_api` and (the hypothetical future) `rid_api` fixtures were present in a particular test's signature, but I think we can afford to sort that out later.

pratibhagupta2109 (approved):

PR #551, opened 2021-09-14
[scd] Update the adjuster tool to handle version 0.3.16

barroco:


pkg/api/v1/scdpb/scd.proto
```diff
@@ -4,716 +4,3 @@ package scdpb;
 
 import "google/api/annotations.proto";
 import "google/protobuf/timestamp.proto";
```
BenjaminPelletier, line 6: Looks like this didn't get generated properly?

barroco: Thanks for spotting it. I just found the issue. For some reason, `make protos` fails randomly generating this partial file with the following error:
```
error: failed to transpile: failed to encode protocol buffers to text: failed to encode protocol buffers package: failed to encode type definition: failed to encode string: unknown type protobuf.Builtin (string)
make: *** [pkg/api/v1/scdpb/scd.proto] Error 1
```
Running the command directly without `make` generated the correct file. I will try to find the reason and create a new issue if the problem persists.

BenjaminPelletier (approved):

PR #554, opened 2021-09-17
SCD Qualifier nominal interoperability test API

hrishiballal:
First pass of APIs to implement SCD qualifier on the USSP side.


interfaces/automated-testing/scd/observation.yaml
```diff
@@ -0,0 +1,473 @@
...
+  /scd_qualifier_data/{id}:
+    get:
+      parameters:
+        -
+          examples:
+            default:
+              value: '2fa86378-5915-42bf-9e87-99ff5bce543c'
+          name: id
+          description: >-
+            ID of test identified in `/scd_qualifier_data`.
```
BenjaminPelletier, line 451: There should be no reason to specify a test ID here; this endpoint should be asking "what information does the USS observe that is relevant to strategic coordination?" and the fact that there may or may not be tests injected into one or more USSs should not affect the behavior of that observation.

interfaces/automated-testing/scd/observation.yaml
```diff
@@ -0,0 +1,473 @@
...
+    OperationalIntentList:
+      description: A list of operational intent 
+      type: array
+      items: 
+        $ref: '#/components/schemas/OperationalIntent'
+
+paths:
+  /scd_qualifier_data/{id}:
+    get:
+      parameters:
```
BenjaminPelletier, line 444: We need to specify where we want to look for operational intents; I would expect a Volume4D as one of the parameters in the request body.  RID qualifier uses only a 2D area in its observation request because that corresponds to how users would likely query a Display Provider.  For strategic coordination, the altitude and time bounds are more likely to be relevant.

interfaces/automated-testing/scd/injection.yaml
```diff
@@ -0,0 +1,629 @@
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/PutOperationalIntentReferenceParametersList'
+        required: true
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PutOperationalIntentReferenceParametersList'
```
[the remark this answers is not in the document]
hrishiballal: At this moment, the soon to be shared data generator can generate both paths (GeoJSON / KML / WKT) and flight volume (Volume4D). I think we should discuss this in detail. One option is to share the GeoJSON / KML and parameters, in which case it cannot be a Volume4D (?), or do you think we can send Volume4D with additional parameters? In my estimation, the response state is where the USS will tell us what happened.

interfaces/automated-testing/scd/injection.yaml
```diff
@@ -0,0 +1,629 @@
...
+    Endpoints exposed by the USS for interaction with strategic deconfliction test suite (`scd_qualifier`)
+
+
+components:
+  securitySchemes:
+    Authority:
+      type: oauth2
+      flows:
+        clientCredentials:
+          tokenUrl: https://token_vendor.com/oauth/token
```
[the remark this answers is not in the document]
hrishiballal: Thanks, yes I have now updated the reusable-objects in both the observation and injection yaml.

interfaces/automated-testing/scd/observation.yaml
```diff
@@ -0,0 +1,473 @@
+---
+openapi: "3.0.3"
+info:
+  title: "SCD Qualifier Observaton API"
+  description: |-
+    This interface must be implemented by every Service Provider wishing to be tested by the automated testing framework. The test suite calls these endpoints to verify compliance with the test suite.
```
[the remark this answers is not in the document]
hrishiballal: Thanks yes, I have now updated it.

interfaces/automated-testing/scd/injection.yaml
```diff
@@ -0,0 +1,629 @@
+---
+openapi: "3.0.3"
+info:
+  title: "SCD Qualifier Injection API"
+  description: |-
+    This interface is implemented by the service provider wishing to be tested by the Flight Authorization testing suite. The automated testing suite calls this interface to inject flight-related test data into the Service Provider system under test.
```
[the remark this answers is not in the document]
hrishiballal: Observation has now been removed. Resolving

interfaces/automated-testing/scd/scd_injection_observation.yaml
```diff
@@ -0,0 +1,792 @@
...
+            If an existing subscription is not specified in `subscription_id`,
+            and the operational intent is in the Activated, Nonconforming, or
+            Contingent state, then this field must be populated.  When this
+            field is populated, an implicit subscription will be created and
+            associated with this operational intent, and will generally be
+            deleted automatically upon the deletion of this operational intent.
+          anyOf:
+          - $ref: '#/components/schemas/ImplicitSubscriptionParameters'
+
+    OperationalIntentList:
```
BenjaminPelletier, line 611: Let's put a big comment separator here indicating that everything above this point is just copied from the standard SCD API, and only the stuff below this point is new for the automated testing API

interfaces/automated-testing/scd/scd_injection_observation.yaml
```diff
@@ -0,0 +1,792 @@
...
+      description: A list of operational intents
+      type: array
```
BenjaminPelletier, line 613: Best practice for data types is to [make them objects](https://github.com/interuss/dss/pull/554#discussion_r718957576).  Since an OperationalIntentList is only used one place, there is no need to define a separate data type for it.

interfaces/automated-testing/scd/scd_injection_observation.yaml
```diff
@@ -0,0 +1,792 @@
...
+      items: 
+        $ref: '#/components/schemas/OperationalIntent'
+    Key:
```
BenjaminPelletier, line 616: This definition belongs above the divider mentioned above since it is not new for the automated testing API

interfaces/automated-testing/scd/scd_injection_observation.yaml
```diff
@@ -0,0 +1,792 @@
...
+
+    CreateTestParameters:
+      description: Collection of fully-defined operational intents
+      type: object
+      properties:
+        test_id: 
+          description: ID provided when a test payload is submitted to the system under test
+          type: string
+          format: uuid 
+        operational_intents:
```
BenjaminPelletier, line 650: Following [the RID automated testing API](https://github.com/interuss/dss/blob/25ccb1a6a3e732eb0d57ad87c09f3be8b71e2465/interfaces/automated-testing/rid/injection.yaml#L133), we should have an injection ID for each operational intent requested to support USSs that can't easily accept arbitrary operational intent IDs (because they generate their own).

interfaces/automated-testing/scd/scd_injection_observation.yaml
```diff
@@ -0,0 +1,792 @@
...
+        area_of_interest:
+          type: object
+          $ref: '#/components/schemas/Volume4D'
+                  
+      
+    SCDProcessingResults: 
+      description: Results for the operational intents submitted to the create test endpoint
+      type: array
+      items:
+        anyOf:
```
BenjaminPelletier, line 673: We should not use polymorphism here because most code generation tools will not be able to render this data type.  Instead, there should be two separate fields (perhaps `success_result` and `not_authorized_result` of types `OperationalIntent` and `FlightNotAuthorizedResult`, respectively) where exactly one is populated, per the description.

interfaces/automated-testing/scd/scd_injection_observation.yaml
```diff
@@ -0,0 +1,792 @@
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateTestParameters'
+        required: true
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CreateTestParameters'
```
BenjaminPelletier, line 713: The request and response data types [should not be the same](https://github.com/interuss/dss/pull/554#discussion_r718958855), and this certainly isn't the response data type we want because the USS may not be able to create the specified operational intents.

interfaces/automated-testing/scd/scd_injection_observation.yaml
```diff
@@ -0,0 +1,792 @@
...
+      summary: Delete test
+      description: Remove all injected test data associated with this test from the host Service Provider's system under test.
+      security:
+        - TestAuth:
+            - scd.inject_test_data
+            
+      tags:
+        - Strategic Deconfliction
+
+  /scd_qualifier_data:
```
BenjaminPelletier, line 767: I think we would need to see the automated testing concept overall (perhaps with sequence and/or architecture diagram) to be sure, but I wouldn't expect this endpoint would be necessary.  In RID, we need an explicit observation endpoint because there is no standard API between the USS and Display Application/user.  In SCD, I don't think there is any corresponding interface that we want to tap into -- instead, the SCD automated testing can simply check with the DSS and USS directly using the ASTM-standard endpoints to determine if an operational intent was created correctly.

hrishiballal:
Thank you so much for the review and the feedback. I will make the required changes. Also, After a discussion with @barroco, I plan to "combine" two tests (the data format checks and this one) and open a PR into the FOCA branch and then we or he will will submit a new branch here. I hope that it is OK.

interfaces/automated-testing/scd/scd_injection_observation.yaml
hrishiballal: OK I have now removed the OperationalIntentList as a separate object.

interfaces/automated-testing/scd/scd_injection_observation.yaml
hrishiballal: OK I have now created a new `injection_id` in the submission payload.

interfaces/automated-testing/scd/scd_injection_observation.yaml
hrishiballal: OK I have now changed the Test create data set and additionally a new object to display response

interfaces/automated-testing/scd/scd_injection_observation.yaml
hrishiballal: OK understood, I will remove it from this review and in the test create response we can have the results of the creation request itself

interfaces/automated-testing/scd/scd_injection_observation.yaml
hrishiballal: I have now created two objects showing the success or failure response.

hrishiballal:
I have now made the changes to the API that accommodates all the all your feedback and changed the definitions to accept Volume4Ds instead of Operational Intents. We are going to combine the data format checks as well here.

hrishiballal:
Closing this since I have now created a new PR #644 , that captures the review.

PR #556, opened 2021-09-20
500 error handling during loop validation

pratibhagupta2109:


pkg/rid/server/isa_handler.go
```diff
@@ -221,7 +221,7 @@ func (s *Server) SearchIdentificationServiceAreas(
 
 	cu, err := geo.AreaToCellIDs(req.GetArea())
 	if err != nil {
-		return nil, stacktrace.Propagate(err, "Invalid area")
+		return nil, stacktrace.PropagateWithCode(err, dsserr.BadRequest, "Invalid area")
```
BenjaminPelletier, line 224: The problem the automated testing is catching here is that some invalid areas return a different code -- specifically, 413.  So, some (but not all) codepaths within AreaToCellIDs are producing `err` results with the code already propagated, but some are not.  I think this is the best level at which to assign those codes, so the fact that some lower-level layer is assigning 413 isn't ideal.  But, at this level, we'll need to check whether the error is an area-too-large error or a different error and assign 413 if it's an area-too-large error, and 400 otherwise.

pratibhagupta2109: e2e tests `test_get_isa_by_search_huge_area` and `test_get_sub_by_searching_huge_area` pass with assertion 413 code. New test cases are added to confirm above code. I checked getting 500 error by removing this code.

BenjaminPelletier (approved):

PR #563, opened 2021-09-27
Fix shell scripts to pass linter

barroco:
This pull request fixes shell scripts in order to pass shellcheck linter test.

Test command:
> make lint

Please note that there are few TODO comments to be discussed with reviewers.

### DIFF
--- a/build/dev/startup/scd_bootstrapper.sh
+++ b/build/dev/startup/scd_bootstrapper.sh
@@ -3,6 +3,8 @@
 # This startup script is meant to be invoked from within a Docker container
 # started by docker-compose_dss.yaml, not on a local system.
 
+# TODO: Is the test disabled on purpose?
+# shellcheck disable=SC2157
 if [ -z "ls /var/local-dss-data/scd_bootstrap > /dev/null 2>&1" ]; then
   echo "SCD DB is already set up; no need to run bootstrapper."
   exit 0
--- a/build/dev/startup/scd_bootstrapper.sh
+++ b/build/dev/startup/scd_bootstrapper.sh
@@ -3,9 +3,7 @@
 # This startup script is meant to be invoked from within a Docker container
 # started by docker-compose_dss.yaml, not on a local system.
 
-# TODO: Is the test disabled on purpose?
-# shellcheck disable=SC2157
-if [ -z "ls /var/local-dss-data/scd_bootstrap > /dev/null 2>&1" ]; then
+if ls /var/local-dss-data/scd_bootstrap > /dev/null 2>&1; then
   echo "SCD DB is already set up; no need to run bootstrapper."
   exit 0
 else
--- a/monitoring/rid_qualifier/host/run_locally.sh
+++ b/monitoring/rid_qualifier/host/run_locally.sh
@@ -14,6 +14,8 @@ cd "${BASEDIR}/../../.." || exit 1
 # for use with rid_qualifier
 AUTH="DummyOAuth(http://host.docker.internal:8085/token,uss1)"
 DSS="http://host.docker.internal:8082"
+# TODO: fix unused
+# shellcheck disable=SC2034
 AUD="host.docker.internal"
 PORT=8072
 RID_HOST="http://localhost:${PORT}"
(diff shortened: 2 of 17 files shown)


BenjaminPelletier:
I haven't had a chance to investigate why, but this PR currently seems to break the docker_e2e.sh integration test (verified on local machine as well)

barroco:
> I haven't had a chance to investigate why, but this PR currently seems to break the docker_e2e.sh integration test (verified on local machine as well)

The PR was targeting the `scd_api_update` and failing because of the unfinished implementation of the tests. 

I updated the PR to target `master`.

I had to add a [filter to the linter](https://github.com/interuss/dss/pull/563/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L44) to skip checks of the shell script in the `astm-utm` submodule. Let me know if you think that it would be preferable to propose the changes to the `astm-utm` repository instead. Please note that the script has been removed from the `astm-utm` repository in newer versions.

build/dev/startup/scd_bootstrapper.sh
```diff
@@ -3,6 +3,8 @@
 # This startup script is meant to be invoked from within a Docker container
 # started by docker-compose_dss.yaml, not on a local system.
 
+# TODO: Is the test disabled on purpose?
+# shellcheck disable=SC2157
```
BenjaminPelletier, line 7: I think this is also a bug -- we want to execute the command in that string, not check if it's non-empty :)

monitoring/rid_qualifier/host/run_locally.sh
```diff
@@ -14,6 +14,8 @@ cd "${BASEDIR}/../../.." || exit 1
 # for use with rid_qualifier
 AUTH="DummyOAuth(http://host.docker.internal:8085/token,uss1)"
 DSS="http://host.docker.internal:8082"
+# TODO: fix unused
+# shellcheck disable=SC2034
```
BenjaminPelletier, line 18: We can just remove this `AUD=` line since it's not used

monitoring/rid_qualifier/run_locally.sh
```diff
@@ -44,21 +44,22 @@ CONFIG='--config config.json'
...
+touch "$(pwd)/monitoring/rid_qualifier/report.json"
 
 docker build \
     -f monitoring/rid_qualifier/Dockerfile \
     -t interuss/dss/rid_qualifier \
-    --build-arg version=`scripts/git/commit.sh` \
+    --build-arg version="$(scripts/git/commit.sh)" \
     monitoring
 
+# shellcheck disable=SC2086
```
BenjaminPelletier, line 55: Why do we need to disable here?  Don't the quotes below fix the issue?

monitoring/tracer/run_poll_locally.sh
```diff
@@ -7,6 +7,8 @@ LOGS='--output-folder=/logs'
 KML_SERVER='--kml-server=https://example.com/kmlgeneration'
 KML_FOLDER='--kml-folder=test/localmock'
 MONITOR='--rid-isa-poll-interval=15 --scd-operation-poll-interval=15 --scd-constraint-poll-interval=15'
+# TODO: Fix unused
+# shellcheck disable=SC2034
```
BenjaminPelletier, line 11: This is a polling routine and doesn't host an endpoint so `PORT=` is definitely not needed; that line can be deleted instead.

monitoring/tracer/run_poll_locally.sh
```diff
@@ -16,12 +18,13 @@ echo Reminder: must be run from root repo folder
 docker build \
     -f monitoring/tracer/Dockerfile \
     -t interuss/dss/tracer \
-    --build-arg version=`scripts/git/commit.sh` \
+    --build-arg version="$(scripts/git/commit.sh)" \
     monitoring
 
+# shellcheck disable=SC2086
```
BenjaminPelletier, line 24: Don't the quotes below fix this issue?

BenjaminPelletier (commented):
Thanks for doing this!  A few comments below.

monitoring/rid_qualifier/run_locally.sh
barroco: If we quote the variable, the argument is passed as a single argument (escaping word splitting) and the python script fails to properly parse arguments.

monitoring/tracer/run_poll_locally.sh
barroco: Same as above; If we quote the variable as proposed by the linter, $TRACER_OPTIONS is passed as a single argument (escaping word splitting) and the python script fails to properly parse arguments.

BenjaminPelletier (approved):

PR #604, opened 2021-10-06
SCD Qualifier Flight Data / Operational Intents Generator

hrishiballal:
This branch generates 4D Flight Paths and 4D Volumes and Operational Intents per ASTM  for SCD testing


monitoring/scd_qualifier/utils.py
```diff
@@ -0,0 +1,117 @@
...
+  def __new__(cls, value):
+    if isinstance(value, str):
+      t = arrow.get(value).datetime
+    else:
+      t = value
+    str_value = str.__new__(cls, arrow.get(t).to('UTC').format('YYYY-MM-DDTHH:mm:ss.SSSSSS') + 'Z')
+    str_value.datetime = t
+    return str_value
+
+class LatLngPoint(ImplicitDict):
```
BenjaminPelletier, line 18: If this is mirroring [a type in the SCD API](https://github.com/astm-utm/Protocol/blob/f6e5ee5a84cbe56d1b39187aa5ba0f67a7d8f5ec/utm.yaml#L265), let's follow [the RID style](https://github.com/interuss/dss/blob/25ccb1a6a3e732eb0d57ad87c09f3be8b71e2465/monitoring/monitorlib/rid.py#L86) and mirror the type in [`scd.py`](https://github.com/interuss/dss/blob/master/monitoring/monitorlib/scd.py).  Same comment for other types in this file mirror types defined in the API.

monitoring/scd_qualifier/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    is_control: bool = 0
+    expected_result: str
+
+class GeoJSONFeature(ImplicitDict):
+    ''' A class to hold GeoJSON Feature flight plan, at this moment we are using a LineString but it could be a LineString or Polygon '''
+    geojson: Feature
+    
+class SCDVolume4D(ImplicitDict):
+    ''' A class to hold a volume 4D with path and time details and GeoJSON flight plan, USSP will have to transform flight plan to a  '''
+    flight_plan: GeoJSONFeature
```
BenjaminPelletier, line 85: What purpose does the `GeoJSONFeature` class serve?  Why wouldn't the `flight_plan` field simply be a `Feature` type?

monitoring/scd_qualifier/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    """ A class to hold configuration for developing treatment flight path """
+    intersect_space:bool = 0
+    
+class GeometryPayload(ImplicitDict):
+    ''' An object to hold generated flight path, is_control is a nomenclature used to see if the generated path is the first one '''
+    geometry: Union[LineString, Polygon]
+    is_control: bool
+    geometry_generation_rule: GeometryGenerationRule
+  
+class FlightAuthPayload(ImplicitDict):
```
BenjaminPelletier, line 101: Class names should usually be a succinct form of self-documentation, and that succinct form should match the long form.  In this case, we're calling the class a flight auth payload, but then we document it as flight authorization details.  I think the long form is probably more accurate, in which case the better class name would be FlightAuthDetails.  But, I think we can do even better than that -- these are actually the details of a flight authorization, these are the details of a request for flight authorization (depending on these details, the flight authorization may never be created).  So, I would recommend naming this class FlightAuthRequestDetails, or possibly just FlightAuthRequest.

monitoring/scd_qualifier/utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+class Volume4D(ImplicitDict):
+    '''A class to hold ASTM Volume4D objects '''
+    volume: Volume3D
+    time_start: StringBasedDateTime
+    time_end: StringBasedDateTime
+
+
+class OperationalIntentDetails(ImplicitDict):
+    """Class for holding details of an operational intent reference """
```
BenjaminPelletier, line 68: nit: "holding details of an operational intent" (the details are for the operational intent, not for the reference to the operational intent)

monitoring/scd_qualifier/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    volumes: List[Volume4D]
+    priority: int
+
+class VolumeGenerationRule(ImplicitDict):
+    """ A class to hold configuration for holding rule for volume generation """
```
BenjaminPelletier, line 73: What is a "holding rule"?

monitoring/scd_qualifier/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    intersect_altitude: bool = 0
```
BenjaminPelletier, line 74: What does the volume intersect (or not) in altitude? (same question for time)

monitoring/scd_qualifier/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    intersect_time: bool = 0
+    is_control: bool = 0
```
BenjaminPelletier, line 76: What does this field indicate/what will it be used for?

monitoring/scd_qualifier/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    operation_mode: str
+    operation_category:str
+    uas_class:str
+    identification_technologies:List =[]
+    connectivity_methods:List = []
+    endurance_minutes: int 
+    emergency_procedure_url: str
+    operator_id:str
+
+class SCDTestPayload(ImplicitDict):
```
BenjaminPelletier, line 113: This seems useful but doesn't seem to be used anywhere and doesn't appear in the API, so it doesn't seem like we should add it yet.

monitoring/scd_qualifier/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    time_start: StringBasedDateTime
+    time_end: StringBasedDateTime
+    altitude_lower: Altitude
+    altitude_upper: Altitude
+    
+class GeometryGenerationRule(ImplicitDict):
+    """ A class to hold configuration for developing treatment flight path """
+    intersect_space:bool = 0
+    
+class GeometryPayload(ImplicitDict):
```
BenjaminPelletier, line 95: "Payload" is the content within a transport mechanism; what is the transport mechanism here?  I think probably the "payload" part of this name probably doesn't belong.  I'm not sure what set of information actually is; maybe GeneratedGeometry?

monitoring/scd_qualifier/flight_volumes_generator.py
```diff
@@ -0,0 +1,307 @@
...
+        box = shapely.geometry.box(self.minx, self.miny, self.maxx, self.maxy)
+        area = abs(self.geod.geometry_area_perimeter(box)[0])
+
+        # Have a area less than 500m x 500m square and more than 300m x 300m square to ensure enough space for tracks
+        if (area) < 250000 and (area) > 90000:
+            return
+        else:
+            raise ValueError("The extents provided are not of the correct size, please provide extents that are less than 500m x 500m and more than 300m x 300m square")
+        
+    def generate_random_flight_path_polygon(self, generate_polygon:bool) -> Union[LineString, Polygon]:
```
BenjaminPelletier, line 100: Why combine two entirely separate behaviors into one method with behavior controlled by a bool flag?  It seems like there should be two separate methods instead.

BenjaminPelletier (commented):
Here is an initial set of comments, but there will be more.  To reduce the number of comments in the future, I would recommend [smaller pull requests](https://artsy.github.io/blog/2021/03/09/strategies-for-small-focused-pull-requests/) ([also](https://essenceofcode.com/2019/10/29/the-art-of-small-pull-requests/))  It would also be good to have an overview of the intended test process (probably as an initial PR prior to this one) to verify that this PR is implementing part of the process in a way that will fit into the rest of the process.

monitoring/scd_qualifier/flight_volumes_generator.py
```diff
@@ -0,0 +1,307 @@
...
+        self.control_flight_geometry: Union[LineString, Polygon] # the initial flight path or geometry against which subsequent flight paths are generated, this flag
+        
+        self.raw_geometries: List[GeometryPayload] # Object to hold polyons or linestrings, and the rule that generated the geometry (e.g. should this geometry intersect with the control)
+        self.now = arrow.now()        
+        self.geod = Geod(ellps="WGS84")
+        self.grid_cells : List[shapely.geometry.box] # When a bounding box is given, it is split into smaller boxes this object holds the grids
+        self.input_extents_valid()
+        self.generate_grid_cells()
+        
+    def generate_grid_cells(self):
```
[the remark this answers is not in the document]
hrishiballal: Thank you, this has been fixed now.

monitoring/scd_qualifier/flight_volumes_generator.py
```diff
@@ -0,0 +1,307 @@
...
+            line_polygon_feature = {'type': 'Feature', 'properties': {}, 'geometry': shapely.geometry.mapping(raw_geometry.geometry)}
+            scd_volume_4d = SCDVolume4D(flight_plan= line_polygon_feature,time_start = self.flight_volumes[path_id].time_start, time_end=self.flight_volumes[path_id].time_end, altitude_lower= self.flight_volumes[path_id]['volume']['altitude_lower'], altitude_upper= self.flight_volumes[path_id]['volume']['altitude_upper'] )
+            
+            path_file_name = '4d_path_%s.json' % str(path_id + 1)  # Avoid Zero based numbering
+            tracks_file_path = self.output_subdirectories[1] / path_file_name
+            with open(tracks_file_path, 'w') as f:
+                f.write(json.dumps(scd_volume_4d))
+
+    
+    def write_astm_volume_4d(self) -> None:
```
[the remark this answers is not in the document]
hrishiballal: Thanks for this, I am re-reviewing the API definition and I think that the test driver should submit Volume 4D and the USS should generate a operational intent, I am happy to stand corrected but I don think that the test driver should submit Operational Intents.

monitoring/scd_qualifier/flight_volumes_generator.py
```diff
@@ -0,0 +1,307 @@
...
+        # Create test_definition directory if it does not exist        
+        self.output_directory.mkdir(parents=True, exist_ok=True)
+        self.output_subdirectories = (Path(self.output_directory, 'astm_4d_volumes'), Path(self.output_directory, '4d_flight_geometries'),Path(self.output_directory, 'scd_rules'),)
+        for output_subdirectory in self.output_subdirectories:
+            output_subdirectory.mkdir(parents=True, exist_ok=True)
+        self.raw_geometries = raw_geometries
+        self.flight_volumes = flight_volumes
+        self.all_rules = all_rules
+
+    def write_4d_flight_geometries(self) -> None:
```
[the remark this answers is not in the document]
hrishiballal: The main reason this was written, was because of the manual checkout, one of my goals was to build the manual check out data with automated tools so that we start doing things simultaneously. This is not really required for fully automated tests. I will remove this section of code.

monitoring/scd_qualifier/flight_volumes_generator.py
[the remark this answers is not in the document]
hrishiballal: I have now made the changes so that the automated testing API does not accept Operational Intents but Volume 4D which was my intention, I think that the operational intents should be generated by the USS.

monitoring/scd_qualifier/utils.py
```diff
@@ -0,0 +1,117 @@
+from typing import List, Union
+from geojson import Feature
+from shapely.geometry import LineString, Polygon
+from monitoring.monitorlib.typing import ImplicitDict
+import arrow
+
+class StringBasedDateTime(str):
```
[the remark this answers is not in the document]
hrishiballal: Thank you, yes, I have now removed it.

monitoring/scd_qualifier/utils.py
hrishiballal: I have now moved the ASTM specific definitions to `scd.py` file.

monitoring/scd_qualifier/utils.py
hrishiballal: Thanks! I have now made this change.

monitoring/scd_qualifier/utils.py
hrishiballal: Basically there are three kinds of intersections that we want to test: 
- In space: two flight plans that overlap in space and also in height 
- In time: Two flight plans that intersect in space and time or intersect in space but not in time (e.g. i use a space at 9AM and the other party uses the same space at 915AM
- In altitude: We can two operations sharing a airspace at different altitudes and they should be approved as well (given enough vertical separation). 
While we will not test all three of them, I wanted to make sure we can generate volumes that handle these conditions.

monitoring/scd_qualifier/utils.py
hrishiballal: In a First In First Out system, the first plan regardless of anything will be accepted, I call this first operation as `control`, if you have a different name, please let me know. The idea is that we can have a baseline / control volume against which, intersections can be generated (as in the manual checkout). If you feel there is no need for such a setup, we can remove this not have this distinction between the first one vs. any other volume.

monitoring/scd_qualifier/utils.py
hrishiballal: OK I will remove it. We can add it when we combine the test 1 (operator data etc. that the FlightAuthPayload) carries in to this. My idea is that this structure is what should be sent to the USS who then processes it.

monitoring/scd_qualifier/utils.py
hrishiballal: Thanks yes, I have now renamed this.

monitoring/scd_qualifier/flight_volumes_generator.py
hrishiballal: Thank you for this, I have now split them in to two smaller methods.

monitoring/scd_qualifier/utils.py
hrishiballal: Thank you for this, I really struggled with the naming. The only thing here is that this is not a complete request, the thing that is missing here is the Volume4D, this class holds the "data-format" checks so for the moment, I have named it `FlightAuthPartialRequestDetails`, the idea is that we add the volume 4D and then send this for USS processing. See also this [other comment](https://github.com/interuss/dss/pull/604#discussion_r753081200).

monitoring/scd_qualifier/utils.py
hrishiballal: I have now removed this totally, I think we have the USS create the intent and we just send the operator / vehicle data and a volume 4D.

monitoring/scd_qualifier/utils.py
hrishiballal: This has now been removed, we will not be dealing with Operational Intents and stay at the level of Volume4D

PR #609, opened 2021-10-14
[scd] Fix empty OVNs

BenjaminPelletier:
While the intent of the SCD API was to make default values for fields interchangeable with omission of those fields, this intent failed with regard to OVNs because the [`minLength` was set to 16](https://github.com/astm-utm/Protocol/blob/f6e5ee5a84cbe56d1b39187aa5ba0f67a7d8f5ec/utm.yaml#L125), which means that an empty string is an invalid value for that field.  This PR works around that issue without modifying the API by adding a string value (of at least 16 characters) indicating omission of the OVN when references are read from the DSS by clients who don't own them.  A previously-failing test validates the corrected behavior.


barroco:
Thank you very much for preparing this PR. It seems that the OVN may be returned as an empty value on those places as well:
- `constraints_handler.go` on line 127.
- `subscriptions_handler.go` on line 189 and 210.
- `operational_intents_handler.go` on line 435 and 451.

Issue #346, opened 2020-07-14
Change remote ID database name from `defaultdb` to `rid`


BenjaminPelletier:
Currently, the remote ID capabilities of the DSS uses a database named `defaultdb` to store/communicate its backing information for remote ID.  A more proper name for this database would be `rid` since the DSS implements multiple sets of functionality.  All instances of `defaultdb` should be changed to `rid`.

PR #613, opened 2021-10-31
`defaultdb` renamed to `rid`

pratibhagupta2109:
This PR addresses the issue #346, i.e to change database name `defaultdb` to `rid` through migration. 
`defaultdb` Database at any previous version can be upgraded to the latest version post this code merge.

Following steps have been followed to check database upgrade from the repo root:

```
git checkout db-manager/test-upgrade
test/migrations/clear_db.sh
test/migrations/migrate_db.sh defaultdb 1.0.0
git checkout rid-db-rename
test/migrations/migrate_db.sh rid latest
```

The same process has been repeated to test the migration from versions 2.0.0, 3.0.0, 3.1.0, and 3.1.1.

Post migration end to end testing can be achieved by running `test/migrations/rid_db_post_migration_e2e.sh`


BenjaminPelletier (commented):
The continuous integration tests are failing on the `make test-cockroach` command, which runs some DB-related unit tests.

pkg/rid/application/application_test.go
```diff
@@ -58,6 +59,10 @@ func setUpStore(ctx context.Context, t *testing.T, logger *zap.Logger) (store.St
 			},
 		}, func() {}
 	}
+	if !(strings.Contains(*storeURI, "rid") || strings.Contains(*storeURI, "scd")) {
+		*storeURI = strings.Replace(*storeURI, "?sslmode", "/rid?sslmode", 1)
```
BenjaminPelletier, line 63: Is this the replacement we want when storeURI contains "scd"?  Shouldn't it be "/scd?sslmod" in that case rather than "?sslmode"?  Or possibly no replacement in that case?

pkg/rid/store/cockroach/store.go
```diff
@@ -59,6 +59,9 @@ type Store struct {
 
 // NewStore returns a Store instance connected to a cockroach instance via db.
 func NewStore(ctx context.Context, db *cockroach.DB, logger *zap.Logger) (*Store, error) {
+	if DatabaseName == "defaultdb" {
```
BenjaminPelletier, line 62: Shouldn't we change the DatabaseName on line 38 rather than here?

BenjaminPelletier (commented):
Currently, this PR leaves the db_schema folder named "defaultdb"; could we change that folder name to "rid" since it now holds the database steps for the rid database?  That database was previously named defaultdb in prior database versions (before 4.0.0), but as of this PR, there is no such database as defaultdb.  We definitely want to retain the ability to upgrade any previously-deployed systems to database version 4.0.0, but we don't need to keep the old folder naming (I think).

pratibhagupta2109:
> _No description provided._

build/deploy/db_schemas/defaultdb/000008_db_rename_to_rid.up.sql
```diff
@@ -0,0 +1,4 @@
+SET sql_safe_updates = false;
```
pratibhagupta2109: All tests are passing now. Please review.

test/docker_migration_test_step_2.sh
```diff
@@ -0,0 +1,145 @@
+set -eo pipefail
```
pratibhagupta2109, line 1: These two files are temporarily there with the PR, just for testing purpose.
I have taken following steps for testing:
1. Set `currentMajorSchemaVersion` to 3
2. Run test/docker_migration_test_step_1.sh to create db with older db_schema version.
3. Set `currentMajorSchemaVersion` to 4
4. Run test/docker_migration_test_step_2.sh

There seems some issue with my testing, sometime it creates random names for local-gateway container, but if I do not kill the containers, I can see on crdb console that database gets renamed to `rid` from `defaultdb`.

BenjaminPelletier (approved):

PR #620, opened 2021-11-16
uss_availability endpoints

pratibhagupta2109:


pratibhagupta2109:
after PR #630, this PR is also ready to be reviewed.

BenjaminPelletier (commented):
In general, we would like our systems to be as backwards-compatible as possible.  For this feature, we incremented the minor version of the database because we're just adding a feature -- the major database version remains the same, which means the code from now should be able to use the old DB version (perhaps with degraded functionality) and old code should be able to use the new DB version.  To check behavior of new code with old DB, I followed these steps:

1. Check out master branch
1. `build/dev/run_locally.sh down` (reset/delete Cockroach database)
1. `build/dev/run_locally.sh up` (Create a Cockroach database at v3.0.0)
1. Kill deployment (but don't `run_locally.sh down` -- the Cockroach database will be preserved at SCD version 3.0.0)
1. Check out this branch
1. `build/dev/run_locally.sh` (Run the new code using the pre-existing Cockroach database at v3.0.0)
1. `build/dev/probe_local_instance.sh`

The result is that test_set_uss_availability triggers a 500 response -- presumably because the availability database table doesn't exist.  It would be great if new code would work gracefully with old (but still within major version) DBs -- if the availability table doesn't exist, we should just always return Unknown with a blank version.  The above test will still fail once we return Unknown with a blank version, but it will fail on the `== 'Normal'` assertion rather than the DSS returning a 500 error code.

pkg/scd/store/cockroach/availability.go
```diff
@@ -63,7 +63,9 @@ func (u *repo) UpsertUssAvailability(ctx context.Context, s *scdmodels.UssAvaila
 func (u *repo) fetchAvailabilities(ctx context.Context, q dsssql.Queryable, query string, args ...interface{}) ([]*scdmodels.UssAvailabilityStatus, error) {
 	rows, err := q.QueryContext(ctx, query, args...)
 	if err != nil {
-		return nil, stacktrace.Propagate(err, "Error in query: %s", query)
+		if strings.Contains(err.Error(), "does not exist") {
```
BenjaminPelletier, line 66: If `err` is not `nil` and doesn't contain `"does not exist"` (for instance, if there were a SQL syntax error), it seems like execution will continue normally -- is this actually intended behavior?

pkg/scd/uss_availability_handler.go
```diff
@@ -2,34 +2,108 @@ package scd
...
+				Availability: ussa.Availability.String(),
+				Uss:          id.String()},
+			Version: ussa.Version.String(),
+		}
+		return nil
+	}
+
+	err := a.Store.Transact(ctx, action)
+	if err != nil {
+		if strings.Contains(err.Error(), "does not exist") {
```
BenjaminPelletier, line 53: nit: Let's add a comment explaining when this condition may happen (presumably when the availability table doesn't exist)

pkg/scd/uss_availability_handler.go
```diff
@@ -2,34 +2,108 @@ package scd
...
+	if ussID == "" {
+		return nil, stacktrace.NewErrorWithCode(dsserr.BadRequest, "ussID not provided.")
 	}
-	result := &scdpb.UssAvailabilityStatusResponse{
-		Status: &scdpb.UssAvailabilityStatus{
-			Availability: "Unknown",
-			Uss:          id},
+	// Retrieve USS availability status from request params
+	availability := params.GetAvailability()
+	if availability == "" {
```
BenjaminPelletier, line 72: Let's also check against the allowed values for availability -- right now, it seems like a USS could set `uss1`'s availability to `Purple`.  A good way to do this would probably be to write `scdmodels.UssAvailabilityStateFromString(string)` and have it return `(UssAvailabilityState, Error)`, and populate that error if the string didn't match one of the allowed values.  Bonus points if we parse without case sensitivity, but return a properly-capitalized value :)  The check for `if availability == ""` would presumably also go into `scdmodels.UssAvailabilityStateFromString`.

BenjaminPelletier (commented):
Thanks, this looks really close -- just a few comments

pkg/scd/store/cockroach/availability.go
pratibhagupta2109: This was a misplaced check. Thank you for finding it.

pkg/scd/uss_availability_handler.go
pratibhagupta2109: Done

BenjaminPelletier (approved):

Issue #612, opened 2021-10-25
Implement USS availability tracking in DSS


BenjaminPelletier:
The new ASTM API includes [endpoints](https://github.com/astm-utm/Protocol/blob/f6e5ee5a84cbe56d1b39187aa5ba0f67a7d8f5ec/utm.yaml#L3127) to track the availability of USSs.  These endpoints should be fully implemented in the DSS.

To implement these endpoints, we probably want to:

1. Add a failing test to prober to set a USS's availability and then retrieve it (both directly in the availability API and indirectly via operational intent & constraint references), but comment out the assertions which would fail (the ones verifying that the availability is as set rather than Unknown) and add TODOs
1. Add `store`/`repo` methods to get and set a USS's availability, but have the get method in store/cockroach/availability.go always return Unknown and have the set method do nothing
1. Add database calls to availability.go to correctly implement the availability methods, failing gracefully if the table doesn't exist (which it won't yet)
1. Create a new `scd_uss_availability` table in the SCD database linking a USS ID (string) to its availability state (string or enum) -- note this will be backwards-compatible, so this change should rev the minor (second) version number of the database schema
1. Uncomment test assertions and verify that functionality is working properly

Most of the steps above probably warrant their own PRs.

BenjaminPelletier:
Completed by #620

PR #623, opened 2021-11-19
pytest placeholders for uss_availability check

pratibhagupta2109:


monitoring/prober/scd/test_operation_simple_v0_3_17.py
```diff
@@ -210,6 +210,22 @@ def test_get_op_by_search_latest_time_included(ids, scd_api, scd_session):
   assert ids(OP_TYPE) in [x['id'] for x in resp.json()['operational_intent_references']]
 
 
+@default_scope(SCOPE_AA)
+@depends_on(test_create_op)
+def test_set_uss_availability(ids, scd_session2):
```
BenjaminPelletier, line 215: These two tests verify that the endpoints produce 200 responses, but I think we probably want to make tests a bit more rigorous than that.  After we have a placeholder implementation that does nothing upon set and always returns "Unknown" upon get (step 2 in #612), both of these tests will pass.  Ideally, we want tests that will not pass until the implementation is correct (but then comment out the assertions during development).  To check that, we should at least set a particular USS's availability to each of the possible states and verify that it is that state (via a get) after each set.  Another good test would probably be to verify that uninitialized USSs return Unknown -- to do that, we would get the availability for a USS that we know has never be set, and picking the USS ID using a UUID would probably work well for that.  Another test we might consider is one where we look at two different USS IDs and verify that changing the availability on one doesn't change the availability of the other.

pkg/scd/uss_availability_handler.go
```diff
@@ -1,12 +1,15 @@
 package scd
 
+import "log"
 import "context"
 import "github.com/interuss/dss/pkg/api/v1/scdpb"
 
 func (a *Server) GetUssAvailability(ctx context.Context, request *scdpb.GetUssAvailabilityRequest) (*scdpb.UssAvailabilityStatusResponse, error) {
-	panic("implement me")
+	log.Println("implement me")
+	return nil, nil
```
BenjaminPelletier, line 9: I believe this will cause the endpoint to return a success code -- we should make sure these endpoints still indicate some kind of error even if they no longer take down the entire server when they are hit.  To accomplish this, we can look at other places where that happens and determine that we should return some kind of error.  Probably returning `stacktrace.NewError("Not implemented")` should work.

monitoring/prober/scd/test_operation_simple_v0_3_17.py
pratibhagupta2109: Updated with expected result.

pkg/scd/uss_availability_handler.go
```diff
@@ -2,11 +2,12 @@ package scd
...
+import "github.com/interuss/stacktrace"
 
 func (a *Server) GetUssAvailability(ctx context.Context, request *scdpb.GetUssAvailabilityRequest) (*scdpb.UssAvailabilityStatusResponse, error) {
-	panic("implement me")
+	return nil, stacktrace.NewError("Not implemented")
 }
 
 func (a *Server) SetUssAvailability(ctx context.Context, request *scdpb.SetUssAvailabilityRequest) (*scdpb.UssAvailabilityStatusResponse, error) {
-	panic("implement me")
+	return nil, stacktrace.NewError("Not implemented")
```
pratibhagupta2109, line 12: This returns an error, but with 500 code.

BenjaminPelletier (approved):

PR #627, opened 2021-11-25
uss_availability repo methods 

pratibhagupta2109:


pkg/scd/models/availability.go
```diff
@@ -14,8 +12,9 @@ type UssAvailabilityState string
 
 // UssAvailabilityStatus models an uss availability status.
 type UssAvailabilityStatus struct {
-	Uss          dssmodels.Manager
 	Availability UssAvailabilityState
+	Uss          string
```
BenjaminPelletier, line 16: I think giving a strong type hint here with `dssmodels.Manager` is probably clearer -- it is a string underneath, but the additional type indicates to developers what kinds of values we expect in that string.

Less important, but I'd also probably keep the original ordering (`Uss` then `Availability`) -- this is essentially a versioned key-value pair and the Uss field is the key.

pkg/scd/repos/repos.go
```diff
@@ -53,6 +53,12 @@ type Subscription interface {
 	IncrementNotificationIndices(ctx context.Context, subscriptionIds []dssmodels.ID) ([]int, error)
 }
 
+type UssAvailability interface {
+	GetUssAvailability(ctx context.Context, id string) (*scdmodels.UssAvailabilityStatus, error)
```
BenjaminPelletier, line 57: Along with the comment above, I think typing `id` as a `dssmodels.Manager` will provide a clearer hint to developers in terms of usage.

pkg/scd/store/cockroach/availability.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+func (u *repo) UpsertUssAvailability(ctx context.Context, s *scdmodels.UssAvailabilityStatus) (*scdmodels.UssAvailabilityStatus, error) {
+	// todo: yet to implement
+	log.Println("Implement set uss availability!")
+	return nil, nil
+}
+
+// GetUssAvailability returns the Availability status identified by "id".
+func (u *repo) GetUssAvailability(ctx context.Context, id string) (*scdmodels.UssAvailabilityStatus, error) {
```
BenjaminPelletier, line 16: Same `dssmodels.Manager` type comment here

BenjaminPelletier (approved):

PR #628, opened 2021-11-26
uss_availability: define database functions

pratibhagupta2109:


pkg/scd/store/cockroach/availability.go
```diff
@@ -0,0 +1,122 @@
...
+	var payload []*scdmodels.UssAvailabilityStatus
+	for rows.Next() {
+		var (
+			u         = new(scdmodels.UssAvailabilityStatus)
+			updatedAt time.Time
+		)
+		err := rows.Scan(
+			&u.Uss,
+			&u.Availability,
+			&updatedAt,
```
BenjaminPelletier, line 78: It looks like the timestamp is scanned, but then we don't do anything with it.  How will `UssAvailabilityStatus.Version` get populated when reading from the database?

pkg/scd/store/cockroach/availability.go
```diff
@@ -0,0 +1,122 @@
...
+		return nil, sql.ErrNoRows
+	}
+	return availabilities[0], nil
+}
+
+// GetUssAvailability returns the Availability status identified by "id".
+func (u *repo) GetUssAvailability(ctx context.Context, ussID string) (*scdmodels.UssAvailabilityStatus, error) {
+	var ussAvailabilityQuery = `
+      SELECT
+        id, availability, updated_at
```
BenjaminPelletier, line 109: To make things easier if/when we want to change the columns in this table, let's use the pattern as above where we insert `availabilityFieldsWithoutPrefix` via `fmt.Sprintf`.

pkg/scd/store/cockroach/availability.go
```diff
@@ -0,0 +1,123 @@
...
+	if len(availabilities) > 1 {
+		return nil, stacktrace.NewError("Query returned %d availabilities when only 0 or 1 was expected", len(availabilities))
+	}
+	if len(availabilities) == 0 {
+		return nil, sql.ErrNoRows
+	}
+	return availabilities[0], nil
+}
+
+// GetUssAvailability returns the Availability status identified by "id".
```
BenjaminPelletier, line 107: nit: the name of the variable is now `ussID` rather than `id`

BenjaminPelletier (commented):
This PR needs to be resynced with master since recent PR merges.

I didn't catch this before, but the prober tests for this functionality should be in their own file (like, test_uss_availability.py) -- the tests are for the availability functionality rather than operational intent/operations functionality.

Also, as we add code, we should be able to start uncommenting test assertions as we go.  In general, all added code should have tests that verify it, and in this case we've actually already written those tests)  When we made the tests initially, we changed the `panic` in the implementation to instead return an error, which allowed us to make the calls to the availability endpoints from the test.  When we complete(d) [step 2](https://github.com/interuss/dss/issues/612), we should be able to uncomment the `assert resp.status_code == 200` lines, and the line that checks that an unknown USS indicates `Unknown` availability.  I don't think step 3 allows us to uncomment any more tests, but it also shouldn't regress the progress we've made so far (since we want this to be a minor database version update -- just adding functionality while being backwards compatible -- we need to fail gracefully if the availability table doesn't exist yet).  Step 4 should enable the uncommenting of all tests.

pratibhagupta2109:
To uncomment `assert resp.status_code == 200`, I should have implemented `pkg/scd/uss_availability_handler.go` along with  [step 2](https://github.com/interuss/dss/issues/612) which, as per my understanding, is going to be the complete [code ](https://github.com/interuss/dss/pull/620/files#diff-e6aa5c96c7de4a7d6abdcfb984c12b6a5ba21cd167813ad876a7b551d3c74488R1). Let me know if there is another way.

BenjaminPelletier (approved):

PR #632, opened 2021-11-29
[automated-testing/scd] UAS Flight Authorisation uas_serial_number

barroco:
First sequence diagram to describe the automated testing approach for checking operational intent submission format.


interfaces/automated-testing/scd/design/operational-intent-format/uas_flight_auth_uas_serial_number.plantuml
```diff
@@ -0,0 +1,15 @@
...
+title UAS Flight Auth uas_serial_number
+
+participant "Test Driver" as TD
+participant "USS under test" as USS
+
+== Invalid ==
+TD --> USS: PUT /inject/operational_intent\nwith invalid uas_serial_number
+TD <-- USS: 400 Bad Format
+
+== Valid ==
```
BenjaminPelletier, line 12: Just a note: I think this is a strong candidate for combining when more tests are defined (no reason to repeat injection of a fully-valid operational intent once for valid time range, again for valid geography, again for valid registration number, again for valid state, etc)

interfaces/automated-testing/scd/design/operational-intent-format/uas_flight_auth_uas_serial_number.plantuml
```diff
@@ -0,0 +1,15 @@
+@startuml
+center header Operational Intent Format
+title UAS Flight Auth uas_serial_number
+
+participant "Test Driver" as TD
+participant "USS under test" as USS
+
+== Invalid ==
+TD --> USS: PUT /inject/operational_intent\nwith invalid uas_serial_number
```
BenjaminPelletier, line 9: Operational intents don't have serial numbers; should this be a notional /inject/flight_authorisation endpoint rather than /inject/operational_intent?

barroco: It makes sense and simplify the fact that this endpoint is optional for USS which doesn't need to comply with U-Space regulation.

interfaces/automated-testing/scd/design/operational-intent-format/uas_flight_auth_uas_serial_number.plantuml
barroco: Sounds good to me. I will add the next tests in the same file. Should we envision splitting it in two sets of tests? On one end, ASTM requirements (time range, valid geography, etc.) and on the other hand, U-Space requirements related to the flight auth?

BenjaminPelletier (approved):

PR #633, opened 2021-11-30
[automated-testing/scd] Initial nominal interoperability test for strategic coordination

barroco:
This PR proposes a first sequence diagram to test interoperability of USSs with the strategic coordination role.


interfaces/automated-testing/scd/design/strategic-coordination-interoperability/nominal-test.plantuml
```diff
@@ -0,0 +1,28 @@
+@startuml
+center header Strategic Coordination Interoperability
+title Nominal test
+
+participant "Test Driver" as TD
+participant "USS1 under test" as USS1
+participant "USS2 under test" as USS2
+participant "DSS"
+
+== TD injects via USS1 non-conflicting Operational Intent 1 ==
```
hrishiballal, line 10: This is a architectural question we (you) need to decide, do we inject Operational Intents or do you want to send them Volume4Ds and then the USS creates Operational Intents, I know that it is a small / minor thing but the reason I ask is because [OperationalIntent contains ](http://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/astm-utm/Protocol/master/utm.yaml#operation/createOperationalIntentReference)details that the test driver / simulator may or may not know. My suggestion is that we submit Volume4D and let the USS handle creation of operational intents etc. Otherwise we may have to submit "partial" operational intents.

interfaces/automated-testing/scd/design/strategic-coordination-interoperability/nominal-test.plantuml
```diff
@@ -0,0 +1,28 @@
...
+
+TD --> USS1: PUT /inject/operational_intent
+TD <-- USS1: 200 OK - Operational Intent 1 created
+
+TD --> DSS: POST /dss/v1/operational_intent_references/query\nwith area_of_interest encompassing Operational Intent 1
```
hrishiballal, line 15: This is more of a question for both of you, do we really want the test driver to query the DSS (I don't have a problem), see the other comment, having a mock USS / reference implementation could be a better approach. I don't have a strong preference, but my gut feeling is that the test driver could be as simple as possible.

interfaces/automated-testing/scd/design/strategic-coordination-interoperability/nominal-test.plantuml
```diff
@@ -0,0 +1,28 @@
+@startuml
+center header Strategic Coordination Interoperability
+title Nominal test
```
hrishiballal, line 3: Suggestion to put a test number to this e.g. Nominal Test OPIN001 something like that

hrishiballal (commented):
Thank you for this and all the work, just a couple of comments re the structure

interfaces/automated-testing/scd/design/strategic-coordination-interoperability/nominal-test.plantuml
```diff
@@ -0,0 +1,28 @@
...
+
+TD --> USS1: GET /uss/v1/operational_intents/{entityid1}
+TD <-- USS1: 200 OK - returns Operational Intent 1
+
+...
+
+== TD injects via USS2 a conflicting Operational Intent (time, altitude, priority and area) ==
+
+TD --> USS2: PUT /inject/operational_intent
+TD <-- USS2: 409 Conflict - Operational Intent 2 rejected
```
BenjaminPelletier, line 26: I would suggest that the injection API return 200 here -- there was no error on the USS side; it appropriately decided that the requested flight could not be planned and gracefully indicated that fact to the "user" (qualifier).  Instead, we can have a response structure that contains information for a number of possible outcomes, including both "flight was successfully planned" and "flight could not be planned due to a conflict".

interfaces/automated-testing/scd/design/strategic-coordination-interoperability/nominal-test.plantuml
```diff
@@ -0,0 +1,28 @@
...
+
+TD --> DSS: POST /dss/v1/operational_intent_references/query\nwith area_of_interest encompassing Operational Intent 1
+TD <-- DSS: 200 OK - returns Operational Intent Reference 1 with entityid1 managed by USS1
+
+TD --> USS1: GET /uss/v1/operational_intents/{entityid1}
+TD <-- USS1: 200 OK - returns Operational Intent 1
+
+...
+
+== TD injects via USS2 a conflicting Operational Intent (time, altitude, priority and area) ==
```
BenjaminPelletier, line 23: We probably don't want to make a conflict on all 6 dimensions at the same time -- it would be better to propose many operational intents where each one conflicts on just one axis (to make sure the USS deconflicts on each axis).  In anticipation of that, I would drop the parenthetical here (just "TD injects via USS2 a conflicting Operational Intent"), or else use "or" instead of "and" when listing the possible conflict dimensions.

interfaces/automated-testing/scd/design/strategic-coordination-interoperability/nominal-test.plantuml
BenjaminPelletier: The test driver is acting as a simulated user attempting to utilize the system, so I don't think there should be anything about the flight it doesn't know (it is making up everything about the flight).  I don't think giving the USS just the Volume4D is sufficient for the USS to create an operational intent on behalf of the user (qualifier) -- the USS doesn't know what the priority of the user's flight is, for instance, nor do they know whether this is a flight being planned for the future (targeting Accepted state) or a flight about to take off (targeting Activated state) or a rogue flight (targeting Contingent state), nor do they know if this is a new flight or an update to an existing flight.  When the qualifier injects an instruction to create an operational intent, I think all of this information needs to be supplied one way or another.

interfaces/automated-testing/scd/design/strategic-coordination-interoperability/nominal-test.plantuml
BenjaminPelletier: In an ideal world, USSs would implement each of the appropriate standards and nothing else, then we would write an automated testing tool which would test that system.  The standards don't yet contain enough specification to accomplish this though, so we are forced to create some additional interfaces we need USSs to implement in order to be tested by our tools (hopefully these interfaces or something like them will be incorporated in later versions of the standards).  We should write these interfaces to be as simple as possible for USSs to implement while still allowing us to test the things we need to test.  In the case of RID, there is no interface defined in the standard that we could use to determine which flights a Display Provider sees, so we needed an observation interface to accomplish that goal.  For SCD, there is an interface defined in the standard that allows us to determine what flights (operational intents) exist in the system (query DSS for references, then query USSs for details).  We should definitely use this pre-existing interface to determine what operational intents are present in the system.  So, after we instruct a USS to create a flight that we expect to be created successfully, we can then definitively verify whether that operational intent was emplaced in the system by querying the DSS (and then following up for details with the USS).

TL;DR: Yes, we should definitely query the DSS to retrieve ground truth.  The only reason it was different for RID was because an equivalent interface didn't exist there.

interfaces/automated-testing/scd/design/strategic-coordination-interoperability/nominal-test.plantuml
hrishiballal: Thanks! That's great I will start working on the changes to the simulator and the API to accommodate / generate Operational Intents (instead of Volume4Ds).

interfaces/automated-testing/scd/design/strategic-coordination-interoperability/nominal-test.plantuml
hrishiballal: Thanks yes, see above comments, I will start making the changes.

interfaces/automated-testing/scd/design/strategic-coordination-interoperability/nominal-test.plantuml
barroco: OPIN001 is a reference used in the standard and it could be confusing. If you agree, I would suggest to introduce test indexes when we will add other use cases.

BenjaminPelletier (approved):

PR #640, opened 2021-12-03
[automated-testing/scd] Add Authorization

barroco:
> depends on #639

This PR adds authorization to the status endpoint.

### DIFF
--- a/interfaces/automated-testing/scd/scd.yaml
+++ b/interfaces/automated-testing/scd/scd.yaml
@@ -52,6 +52,8 @@ paths:
             The USS automated testing interface is activated and its status was retrieved successfully.
         '401':
           description: Bearer access token was not provided in Authorization header, token could not be decoded, or token was invalid.
+        '403':
+          description: The access token was decoded successfully but did not include a scope appropriate to this endpoint.
         '404':
           description: The USS automated testing interface is not activated.
       summary: Retrieve the status of the USS automated testing interface
--- a/interfaces/automated-testing/scd/scd.yaml
+++ b/interfaces/automated-testing/scd/scd.yaml
@@ -17,21 +17,7 @@ components:
             utm.inject_test_data: |-
               Client may inject test data into a USS for the purpose of conducting automated tests.
       description: |-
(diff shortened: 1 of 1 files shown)


interfaces/automated-testing/scd/scd.yaml
```diff
@@ -0,0 +1,72 @@
...
+  securitySchemes:
+    Authority:
+      type: oauth2
+      flows:
+        clientCredentials:
+          tokenUrl: https://auth.example.com/oauth/token
+          scopes:
+            utm.inject_test_data: |-
+              Client may inject test data into a USS for the purpose of conducting automated tests.
+      description: |-
```
BenjaminPelletier, line 19: I think it's less than ideal to copy the full content of SCD auth description here; instead, can we basically say that these are SCD auth tokens with extended scopes?  [Here](https://github.com/interuss/dss/blob/744b7a4733c785066802533b9aa80485fc68db62/interfaces/automated-testing/rid/injection.yaml#L176) is how the RID testing interface does it.

interfaces/automated-testing/scd/scd.yaml
```diff
@@ -0,0 +1,72 @@
...
+            - utm.inject_test_data
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/StatusResponse'
+          description: >-
+            The USS automated testing interface is activated and its status was retrieved successfully.
+        '401':
```
BenjaminPelletier, line 67: I don't know that we actually need to describe all the standard errors (though I know the ASTM standards try to do this), but if we do, 403 would be another response the client would expect to receive following the addition of authorization per this PR.

barroco: I added the 403. Making explicit the meaning of the standard error codes will make the specification more verbose but in the other end, I would expect that they will get properly implemented by USSs and therefore, may simplify debugging when running in the staging environment for instance. Do you think we need to add a response body or we can simplify and only specify the error codes as it is now?

BenjaminPelletier: I think it would be easier for USSs if we don't specify a response body (how things are currently).  I agree that making the meaning of certain error codes explicit is probably a good idea.

BenjaminPelletier (approved):
Looks ready to merge to me, but the PR is still currently marked "Draft"

PR #641, opened 2021-12-06
Fix Concurrency tests

pratibhagupta2109:


monitoring/prober/scd/test_operation_simple_heavy_traffic_concurrent.py
```diff
@@ -86,9 +85,9 @@ def _make_op_request_differ_in_altitude(idx):
 # Generate request with volumes that cover the circle area that centered at (-56, 178), with altitude 0 to 120
 # The operation lasts 9 mins and the time window is one after one per sequential idx change
 # The 2D area and altitude won't change with idx
-def _make_op_request_differ_in_time(idx):
+def _make_op_request_differ_in_time(idx, time_gap):
   delta = 10
-  time_start = datetime.datetime.utcnow() + datetime.timedelta(minutes=20) + datetime.timedelta(minutes=delta * idx)
+  time_start = datetime.datetime.utcnow() + datetime.timedelta(minutes=20 + time_gap) + datetime.timedelta(minutes=delta * idx)
```
BenjaminPelletier, line 90: It's confusing that time_start is calculated partly here (+delta*idx) and partly in the calling context (time_gap) -- let's either calculate it entirely outside this method (in which case only time_gap is needed and not idx) or entirely inside this method (in which case only idx is needed).  I think the latter probably makes the most sense.

pratibhagupta2109: let me know if there is still a chance of improvement.

BenjaminPelletier (approved):

PR #642, opened 2021-12-06
Initial Specification of /inject/flight endpoint for SCD automated tests

hrishiballal:
This is the first PR of series which will fully specify the SCD testing interfaces. 

The payloads for the requests, have been left out for the next PRs.

### DIFF
--- a/interfaces/automated-testing/scd/scd.yaml
+++ b/interfaces/automated-testing/scd/scd.yaml
@@ -45,14 +45,16 @@ components:
           description: >-
             The result of the flight submission
 
-              - `Planned`: The flight submission data was correct and the operational intent was successfully accepted by the DSS.
+              - `Planned`: The flight submission data was valid and the flight was successfully processed by the USS and is now authorized.
 
-              - `Rejected`: The flight submission data was incorrect and / or the USS was not able to successfully submit the operational intent to the DSS.
+              - `Rejected`: The flight submission data provided was invalid and/or could not be used to attempt to authorize the flight.
 
-              - `Conflict`: The flight submission data was correct but an operational intent could not be created because of conflicting operational intent.
+              - `ConflictWithFlight`: The flight submission data was valid, but the flight could not be authorized because of a disallowed conflict with another flight.
+
+              - `Failed`: The USS was not able to successfully authorize the flight due to a problem with the USS or a downstream system
 
           type: string
-          enum: [Planned, Rejected, Conflict]
+          enum: [Planned, Rejected, ConflictWithFlight, Failed]
           example: Planned
 
 paths:
--- a/interfaces/automated-testing/scd/scd.yaml
+++ b/interfaces/automated-testing/scd/scd.yaml
@@ -78,12 +78,19 @@ paths:
       summary: Retrieve the status of the USS automated testing interface
       description: Get the status of the USS automated testing interface
 
-  /inject/flight:
-    summary: Process operator details
+  /v1/flights/{flight_id}:    
     put:      
       security:
         - Authority:
             - utm.inject_test_data
+      parameters:
+        - name: flight_id
+          in: path
+          required: true
+          description: A UUID string identifying the injected flight.
+          schema:
+            type: string
+            format: uuid
       responses:
         '200':
           content:
--- a/interfaces/automated-testing/scd/scd.yaml
+++ b/interfaces/automated-testing/scd/scd.yaml
@@ -39,7 +39,7 @@ components:
     InjectFlightResponse:
       type: object
       required:
-        - results
+        - result
       properties: 
         result: 
           description: >-
--- a/interfaces/automated-testing/scd/scd.yaml
+++ b/interfaces/automated-testing/scd/scd.yaml
@@ -36,6 +36,25 @@ components:
           enum: [Starting, Ready]
           example: Ready
 
+    InjectFlightResponse:
+      type: object
+      required:
+        - results
+      properties: 
+        result: 
+          description: >-
+            The result of the flight submission
+
+              - `Planned`: The flight submission data was correct and the operational intent was successfully accepted by the DSS.
+
+              - `Rejected`: The flight submission data was incorrect and / or the USS was not able to successfully submit the operational intent to the DSS.
+
+              - `Conflict`: The flight submission data was correct but an operational intent could not be created because of conflicting operational intent.
+
+          type: string
+          enum: [Planned, Rejected, Conflict]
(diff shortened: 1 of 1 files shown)


interfaces/automated-testing/scd/scd.yaml
```diff
@@ -57,4 +76,26 @@ paths:
         '404':
           description: The USS automated testing interface is not activated.
       summary: Retrieve the status of the USS automated testing interface
-      description: Get the status of the USS automated testing interface
\ No newline at end of file
+      description: Get the status of the USS automated testing interface
+
+  /inject/flight:
```
BenjaminPelletier, line 81: The design sketch has `/inject/flight`, but I think this is the point where we want to look more closely at the precise address of this endpoint.  We should be consistent with other endpoints in this API, so this endpoint should be prefixed with `/v1`.  Also, we should use [RESTful style](https://restfulapi.net/resource-naming/) when naming the resources we are reading or writing -- the RESTful resource collection we're writing to here is "flights" (or possibly "test_flights".  Modeling after [the RID interface](https://github.com/interuss/dss/blob/744b7a4733c785066802533b9aa80485fc68db62/interfaces/automated-testing/rid/injection.yaml#L9), we should be able to PUT a particular "flight", so that leads to and endpoint address of `/v1/flights/{flight_id}`.  The other candidate would be POST `/v1/flights`.  The difference is that the client (USS qualifier) creates the flight ID for a PUT and the server (USS) creates the flight ID for a POST.  For technical idempotency reasons (the POST request would need a request_id parameter to be idempotent, so we may as well make that the flight_id via PUT), I recommend the PUT style.

So, I suggest we change the path to `/v1/flights/{flight_id}`.  Note that this will require the addition of a [`parameters` declaration](https://github.com/interuss/dss/blob/744b7a4733c785066802533b9aa80485fc68db62/interfaces/automated-testing/rid/injection.yaml#L10) in the endpoint.

interfaces/automated-testing/scd/scd.yaml
```diff
@@ -57,4 +76,26 @@ paths:
...
+    summary: Process operator details
+    put:
```
BenjaminPelletier, line 83: This isn't a reasonable endpoint without a request body, so let's at least put a TODO comment ([where the `requestBody` will go](https://github.com/interuss/dss/blob/744b7a4733c785066802533b9aa80485fc68db62/interfaces/automated-testing/rid/injection.yaml#L23) eventually) indicating that the request body will be added later.

interfaces/automated-testing/scd/scd.yaml
```diff
@@ -57,4 +76,26 @@ paths:
         '404':
           description: The USS automated testing interface is not activated.
       summary: Retrieve the status of the USS automated testing interface
-      description: Get the status of the USS automated testing interface
\ No newline at end of file
+      description: Get the status of the USS automated testing interface
+
+  /inject/flight:
+    summary: Process operator details
```
BenjaminPelletier, line 82: The summary on line 99 and description on line 100 are good, but I don't think that "Process operator details" is an accurate summary of the endpoint that will be `/v1/flights/{flight_id}` (nor the `/inject/flight` endpoint); I think we can just remove this summary line, allowing lines 99 & 100 to handle the documentation.

interfaces/automated-testing/scd/scd.yaml
```diff
@@ -36,6 +36,25 @@ components:
...
+    InjectFlightResponse:
+      type: object
+      required:
+        - results
+      properties: 
+        result: 
+          description: >-
+            The result of the flight submission
+
+              - `Planned`: The flight submission data was correct and the operational intent was successfully accepted by the DSS.
```
BenjaminPelletier, line 48: nit: I think the USS can check whether flight submission data is "valid", but it can't really check if it's "correct".

Also, I think we can be a bit more general in the success criteria.  I would recommend instead: `The flight submission data was valid and the flight was successfully processed by the USS and is now authorized.`

interfaces/automated-testing/scd/scd.yaml
```diff
@@ -36,6 +36,25 @@ components:
...
+
+              - `Rejected`: The flight submission data was incorrect and / or the USS was not able to successfully submit the operational intent to the DSS.
```
BenjaminPelletier, line 50: I think "Failed" (a problem with the USS or a downstream system) should probably be different from "Rejected" (a problem with the user's submission), so with same style of comments as above, I recommend: `The flight submission data provided was invalid and/or could not be used to attempt to authorize the flight.`

I think adding "Failed" separately is probably useful, so I'd recommend: `The USS was not able to successfully authorize the flight due to a problem with the USS or a downstream system.`

interfaces/automated-testing/scd/scd.yaml
```diff
@@ -36,6 +36,25 @@ components:
...
+
+              - `Conflict`: The flight submission data was correct but an operational intent could not be created because of conflicting operational intent.
```
BenjaminPelletier, line 52: Recommend: `The flight submission data was valid, but the flight could not be authorized because of a disallowed conflict with another flight.`  And, since it is also possible to have a conflict with a constraint, let's make the enum value a bit more specific; e.g., `ConflictWithFlight`.

interfaces/automated-testing/scd/scd.yaml
hrishiballal: Thank you, I have now updated the YAML.

interfaces/automated-testing/scd/scd.yaml
hrishiballal: Thanks! I have now added the todo.

interfaces/automated-testing/scd/scd.yaml
hrishiballal: Thanks this is now removed.

interfaces/automated-testing/scd/scd.yaml
hrishiballal: Thanks! I have now added this.

interfaces/automated-testing/scd/scd.yaml
hrishiballal: Thanks! I have now made changes to this.

interfaces/automated-testing/scd/scd.yaml
hrishiballal: I have now added the ConflictWithFlight response

BenjaminPelletier (approved):

PR #647, opened 2021-12-08
[scripts] Add why_go_dep script

BenjaminPelletier:
This script is a small tool to track down which modules are resulting in indirect dependencies.  It is intended specifically to identify why older versions of a particular package with a security issue are being imported.

If a user does not have [GraphViz](https://graphviz.org/) installed/available on their system, the textual graph file (go_mod_graph.gv) is still written and can be visualized by online GraphViz tools such as [WebGraphviz](http://webgraphviz.com/).

### DIFF
--- a/scripts/why_go_dep.py
+++ b/scripts/why_go_dep.py
@@ -1,0 +1,91 @@
+import subprocess
+import sys
+from typing import Dict, Set, Tuple
+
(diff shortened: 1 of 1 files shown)


BenjaminPelletier:
Duplicate of #480

PR #652, opened 2021-12-12
[automated/scd] Add U-Space reference for volumes and priority

barroco:
This PR adds reference for volumes and priority fields required to submit a valid U-Space UAS flight authorisation request.

### DIFF
--- a/interfaces/automated-testing/scd/scd.yaml
+++ b/interfaces/automated-testing/scd/scd.yaml
@@ -260,6 +260,9 @@ components:
         - Activated
         - Nonconforming
         - Contingent
+    ################################################################################
+    ####################### End of ASTM-standard definitions #######################
+    ################################################################################
     Priority:
       description: >-
         Ordinal priority of the operational intent, as defined by the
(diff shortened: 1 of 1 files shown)


interfaces/automated-testing/scd/scd.yaml
```diff
@@ -260,6 +260,9 @@ components:
         - Activated
         - Nonconforming
         - Contingent
+    ################################################################################
+    ####################### End of ASTM-standard definitions #######################
+    ################################################################################
     Priority:
```
BenjaminPelletier, line 266: `Priority` is still an ASTM-specified data type so I don't think we want to override it here.  Plus, `Priority` is used in `OperationalIntentTestInjection` which will be specified for any USS implementing the ASTM standard, regardless of whether they're checking out for U-space.

interfaces/automated-testing/scd/scd.yaml
```diff
@@ -270,19 +273,22 @@ components:
         conflict with a lower-priority operational intent.  The regulator
         specifies whether an operational intent may create a conflict with other
         operational intents of the same priority.
+
+        For U-Space UAS flight authorisation requests, priority is required
```
BenjaminPelletier, line 277: I don't think we need this note here -- since uss_qualifier is the one that will be populating this field, we can decide how it is populated in the implementation.  The `Priority` field that is populated by the USS is the one described by the ASTM standard and which doesn't (can't) have this note.  Instead, we'll validate its value using a check in uss_qualifier if the testing configuration has specified the U-space profile.

interfaces/automated-testing/scd/scd.yaml
```diff
@@ -298,6 +304,9 @@ components:
         volumes:
           description: >-
             Nominal volumes, as would be reported by a USS's operational_intents endpoint.
+
+            For U-Space UAS flight authorisation requests, volumes are required by ANNEX IV
```
BenjaminPelletier, line 308: Same comment as for `Priority`; uss_qualifier is populating this field, so I don't think we need to specify requirements on how it must be populated here.  And, `volumes` are required anyway per original line 290

interfaces/automated-testing/scd/scd.yaml
barroco: I see your point. It makes sense.

interfaces/automated-testing/scd/scd.yaml
barroco: I agree. I will discard this PR then.

interfaces/automated-testing/scd/scd.yaml
barroco: I agree

interfaces/automated-testing/scd/scd.yaml
barroco: I wanted to capture that information somewhere for completeness without referring to the implementation since I am planning to retire this repository: https://github.com/swiss-foca/Protocol. Where would you recommend to put it?

barroco:
As discussed with @BenjaminPelletier , closing this PR. The information regarding the U-Space regulation will be captured in a design doc and this will be the subject of a new pr.

PR #658, opened 2021-12-17
[uss_qualifier] Add SCD entrypoint and update config schema

barroco:
This PR adds a placeholder for SCD automated tests execution and proposes an update of the configuration schema to handle multiple services.

It assumes the locale will be the same for all services under test.

### DIFF
--- a/monitoring/uss_qualifier/scd/test_executor.py
+++ b/monitoring/uss_qualifier/scd/test_executor.py
@@ -1,0 +1,18 @@
+from monitoring.uss_qualifier.scd.utils import SCDQualifierTestConfiguration
+from monitoring.uss_qualifier.utils import is_url
+
+
+def validate_configuration(test_configuration: SCDQualifierTestConfiguration):
+    try:
+        for injection_target in test_configuration.injection_targets:
+            is_url(injection_target.injection_base_url)
+    except ValueError:
+        raise ValueError("A valid url for injection_target must be passed")
+
+def run_scd_tests(locale: str, test_configuration: SCDQualifierTestConfiguration,
+                  auth_spec: str):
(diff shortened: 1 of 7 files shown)


monitoring/uss_qualifier/scd/test_executor.py
```diff
@@ -0,0 +1,18 @@
...
+    try:
+        for injection_target in test_configuration.injection_targets:
+            is_url(injection_target.injection_base_url)
+    except ValueError:
+        raise ValueError("A valid url for injection_target must be passed")
+
+def run_scd_tests(locale: str, test_configuration: SCDQualifierTestConfiguration,
+                  auth_spec: str):
+    # TODO Replace with actual implementation
+    if locale is 'che':
```
BenjaminPelletier, line 15: I think we should explicitly declare (in the configuration) whether or not to run U-space tests rather than inferring based on `locale`, but we can adjust that later.

monitoring/uss_qualifier/utils.py
```diff
@@ -0,0 +1,26 @@
+from typing import Optional
+from urllib.parse import urlparse
+
+from monitoring.monitorlib.typing import ImplicitDict
+from monitoring.uss_qualifier.rid.utils import RIDQualifierTestConfiguration
+from monitoring.uss_qualifier.scd.utils import SCDQualifierTestConfiguration
+
+
+class USSQualifierTestConfiguration(ImplicitDict):
+  locale: str
```
BenjaminPelletier, line 10: I would suggest that the sole purpose of this parameter be to locate external files to be used with this qualifier run.

BenjaminPelletier (approved):

PR #661, opened 2022-01-03
[crdb] Extract connectTo method to prepare for reuse

BenjaminPelletier:
Currently, there is a private `connectTo` method in the grpc-backend that produces a CockroachDB connection, but hides the fact that the the connection parameters for that connection come from flags defined in an entirely different package.  This PR changes the `ConnectTo` method to accept only the full, explicit parameter set, makes the method public in the main `cockroach` package to prepare for reuse in other areas, and makes the origin of the rest of the connection parameters more clear in the grpc-backend.


barroco (approved):

barroco:
Looks good to me. Thanks for this improvement!

PR #662, opened 2022-01-04
[db-manager] Rewrite db-manager to remove go-migrate dependency

BenjaminPelletier:
The [db-manager](https://github.com/interuss/dss/tree/master/cmds/db-manager) tool in this repository was originally written to use the [go-migrate](https://github.com/golang-migrate/migrate) package, which manages database schema versioning and migrations with integer step labels.  Later, we found that we needed [semantic versioning](https://semver.org/) to differentiate between backwards-compatible database schema versions and non-compatible versions, so we augmented our migrations with a separate semantic version identifier.  Our semantic version identifier reduces the unique functionality provided by `go-migrate` to essentially just 1) executing the appropriate SQL files and 2) providing an indication of interrupted migrations.  Meanwhile, go-migrate has now been the cause of two separate groups of security vulnerabilities due to its indirect dependencies, at least [one of which](https://github.com/golang-migrate/migrate/issues/569) is still not fixed.

This PR reworks db-manager to drop the go-migrate dependency and instead be a standalone migration tool using only the same database driver used by grpc-backend.  The SQL files are also renamed to improve clarity regarding what they do (e.g., 1.up.xxx.sql -> upto.v1.xxx.sql + 1.down.xxx.sql -> downfrom.v1.xxx.sql), which has been a point of confusion for multiple developers.  Despite removing the major go-migrate dependency, adding more comments than the original and adding a bonus capability to just print the current schema version for a particular database, db-manager's code length has still been reduced by about 10% in this PR.

This PR also adds build/dev/migrate_local_db.sh, which allows developers to easily migrate between different database versions on their local DSS instance (this script was used to test the behavior of the db-manager rewrite).

The only known functionality lost in this PR is the ability to detect an incomplete migration and refuse to attempt additional migration steps, but that functionality can be fairly easily added in the future if it is anticipated to be useful (I do not believe it is likely to be useful in any current or near-future deployment regime).


build/deploy/db_schemas/rid.libsonnet
```diff
@@ -1,19 +1,16 @@
...
-    "000004_drop_cells_table.down.sql": importstr "rid/000004_drop_cells_table.down.sql",
-    "000004_drop_cells_table.up.sql": importstr "rid/000004_drop_cells_table.up.sql",
-    "000005_bump_version.down.sql": importstr "rid/000005_bump_version.down.sql",
-    "000005_bump_version.up.sql": importstr "rid/000005_bump_version.up.sql",
-    "000006_add_writer_column.down.sql": importstr "rid/000006_add_writer_column.down.sql",
-    "000006_add_writer_column.up.sql": importstr "rid/000006_add_writer_column.up.sql",
-    "000007_add_index_by_time_subscriptions.down.sql": importstr "rid/000007_add_index_by_time_subscriptions.down.sql",
-    "000007_add_index_by_time_subscriptions.up.sql": importstr "rid/000007_add_index_by_time_subscriptions.up.sql",
-    "000008_db_rename_to_rid.up.sql": importstr "rid/000008_db_rename_to_rid.up.sql"
+    "upto-v1.0.0-remove_isa_subs_tables.sql": importstr "rid/upto-v1.0.0-remove_isa_subs_tables.sql",
```
barroco, line 3: The name of the file seems wrong. Shouldn't it be `upto-v1.0.0-create_isa_subs_tables.sql` ?

build/deploy/db_schemas/rid.libsonnet
```diff
@@ -1,19 +1,16 @@
...
+    "upto-v3.0.0-no_change.sql": importstr "rid/upto-v3.0.0-no_change.sql",
+    "upto-v3.1.0-add_writer_column.sql": importstr "rid/upto-v3.1.0-add_writer_column.sql",
+    "upto-v3.1.1-add_index_by_time_subscriptions.sql": importstr "rid/upto-v3.1.1-add_index_by_time_subscriptions.sql",
+    "upto-v4.0.0-rename_defaultdb_to_rid.sql": importstr "rid/upto-v4.0.0-rename_defaultdb_to_rid.sql",
+    "downfrom-v4.0.0-move_rid_to_defaultdb.sql": importstr "rid/downfrom-v4.0.0-move_rid_to_defaultdb.sql",
+    "downfrom-v3.1.1-remove_index_by_time_subscriptions.sql": importstr "rid/downfrom-v3.1.1-remove_index_by_time_subscriptions.sql",
+    "downfrom-v3.1.0-remove_writer_column.sql": importstr "rid/downfrom-v3.1.0-remove_writer_column.sql",
+    "downfrom-v3.0.0-no_change.sql": importstr "rid/downfrom-v3.0.0-no_change.sql",
+    "downfrom-v2.0.0-remove_inverted_indices.sql": importstr "rid/downfrom-v2.0.0-remove_inverted_indices.sql",
+    "downfrom-v1.0.0-create_isa_subs_tables.sql": importstr "rid/downfrom-v1.0.0-create_isa_subs_tables.sql",
```
barroco, line 14: The name of the file seems wrong. Shouldn't it be downfrom-v1.0.0-remove_isa_subs_tables.sql ?

barroco (approved):
Thanks @BenjaminPelletier for this great contribution. There is just a small remark other than that it looks good to me

build/deploy/db_schemas/rid.libsonnet
BenjaminPelletier: Good catch!  Fixed.

PR #657, opened 2021-12-17
[uss_qualifier/scd] Add data definitions

hrishiballal:
Added data definitions for SCD Test injection endpoints in to match the [scd test suite api](https://github.com/interuss/dss/blob/master/interfaces/automated-testing/scd/scd.yaml)

PR #669, opened 2022-01-07
SCD Qualifier Simulator: Interface definitions between test executor and simulator

hrishiballal:
This PR introduces data structures for the interface between test executor and the simulator.

### DIFF
--- a/monitoring/uss_qualifier/scd/data_interfaces.py
+++ b/monitoring/uss_qualifier/scd/data_interfaces.py
@@ -2,13 +2,10 @@ from typing import List
 from monitoring.monitorlib.typing import ImplicitDict
 from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
 
-
-
 class RequiredResults(ImplicitDict):
     ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest) '''
-    expected_response: str 
-    authorisation_data_fields_to_evaluate: List[str]  # In the 
-    operational_intent_processing_result: str
+    expected_result: str 
+    # Why we expected a response of rejected (add a dict / enum)
 
 class TestInjectionRequiredResult(ImplicitDict):
     test_injection: InjectFlightRequest
--- a/monitoring/uss_qualifier/scd/data_interfaces.py
+++ b/monitoring/uss_qualifier/scd/data_interfaces.py
@@ -1,12 +1,32 @@
-from typing import List
+from typing import Optional
 from monitoring.monitorlib.typing import ImplicitDict
 from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
+import datetime
 
-class RequiredResults(ImplicitDict):
+class Issue(ImplicitDict):
+    ''' A class to hold a message that the test executor can provide to the USS in cases when the USS provides a response that is not same as the expected result. '''
+    timestamp: Optional[str]
+
+    subject: Optional[str]
+    ''' Identifier of the subject of this issue, if applicable. This may be a UAS serial number, or any field of other object central to the issue. '''
+
+    summary: str
+    '''Human-readable summary of the issue'''
+
+    details: str
+    '''Human-readable description of the issue'''
+
+    def __init__(self, **kwargs):
+        super(Issue, self).__init__(**kwargs)
+        if 'timestamp' not in kwargs:
+            self.timestamp = datetime.datetime.utcnow().isoformat()
+
+class RequiredResult(ImplicitDict):
     ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest) '''
-    expected_result: str 
-    # Why we expected a response of rejected (add a dict / enum)
+    expected_result: str
+    issue_details: Optional[Issue]
+           
 
 class TestInjectionRequiredResult(ImplicitDict):
     test_injection: InjectFlightRequest
-    required_result: RequiredResults
+    required_result: RequiredResult
--- a/monitoring/uss_qualifier/scd/data_interfaces.py
+++ b/monitoring/uss_qualifier/scd/data_interfaces.py
@@ -1,0 +1,15 @@
+from typing import List
+from monitoring.monitorlib.typing import ImplicitDict
+from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
+
+
+
+class RequiredResults(ImplicitDict):
+    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest) '''
+    expected_response: str 
+    authorisation_data_fields_to_evaluate: List[str]  # In the 
+    operational_intent_processing_result: str
+
+class TestInjectionRequiredResult(ImplicitDict):
+    test_injection: InjectFlightRequest
+    required_result: RequiredResults
--- a/monitoring/uss_qualifier/scd/data_interfaces.py
+++ b/monitoring/uss_qualifier/scd/data_interfaces.py
@@ -28,9 +28,9 @@ class KnownIssueFields(ImplicitDict):
     '''Human-readable description of the issue'''
 
 class KnownResponses(ImplicitDict):
-    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest). Holds information about the "result" response provided by the test interface of the USS under test. Per the SCD Testing API this should be one of "Planned", "Rejected", "ConflictWithFlight" or "Failed". In the case where the USS under test provides a result that is not the same as expected result then the test driver can provide a message detailing why the expected result was populated the way it was. This should help the USS under test debug specific parts of their internal systems. '''
+    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest). This calls holds information about the "result" response provided by the test interface of the USS under test. Per the SCD Testing API this should be one of "Planned", "Rejected", "ConflictWithFlight" or "Failed". In the case where the USS under test provides a result that is not the same as expected result then the test driver can provide a message detailing why the expected result was populated the way it was. This should help the USS under test debug specific parts of their internal systems. '''
     acceptable_results: List[str] 
-    incorrect_result_details: Optional[Dict[Result, KnownIssueFields]]
+    incorrect_result_details: Optional[Dict[str, KnownIssueFields]]
     response_wait_time_secs: int # amount of time the test driver should wait before expecting a response
     response_timeout_message: KnownIssueFields
 
--- a/monitoring/uss_qualifier/scd/data_interfaces.py
+++ b/monitoring/uss_qualifier/scd/data_interfaces.py
@@ -1,30 +1,25 @@
 from typing import Optional, List, Dict
 from monitoring.monitorlib.typing import ImplicitDict
 from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
-from monitoring.monitorlib import fetch
-from monitoring.uss_qualifier.rid import Severity
+from monitoring.uss_qualifier.common_data_definitions import Severity
 
-class Result(object):
+### Begin Definitions around users and test executor 
+
+class Result(ImplicitDict):
     ''' A class to hold the expected output from data processing and the test interface of the USS under test '''
-    Planned = 'Planned'   
-    Rejected = 'Rejected' 
-    ConflictWithFlight = 'ConflictWithFlight'    
-    Failed = 'Failed'
+    processing_outcome: str
         
 class KnownIssueFields(ImplicitDict):
     ''' A class to hold a message that the test executor can provide to the USS in cases when the USS provides a response that is not same as the expected result. '''    
     test_code: str
-    """Code corresponding to check generating this issue"""
+    '''Code corresponding to check generating this issue'''
 
     relevant_requirements: List[str] = []
-    """Requirements that this issue relates to"""
+    '''Requirements that this issue relates to'''
 
     severity: Severity
-    """How severe the issue is"""
+    '''How severe the issue is'''
 
-    injection_target: str
-    """Issue is related to data injected into this target. 
-    """
     subject: Optional[str]
     ''' Identifier of the subject of this issue, if applicable. This may be a UAS serial number, or any field of other object central to the issue. '''
 
@@ -34,18 +29,19 @@ class KnownIssueFields(ImplicitDict):
     details: str
     '''Human-readable description of the issue'''
 
-    queries: List[fetch.Query]
-    """Description of HTTP requests relevant to this issue"""
-
 class RequiredResult(ImplicitDict):
     ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest) '''
     acceptable_results: List[str] # Holds information about the "result" response provided by the test interface of the USS under test. Per the SCD Testing API this should be one of "Planned", "Rejected", "ConflictWithFlight" or "Failed"
     incorrect_result_details: Optional[Dict[Result, KnownIssueFields]] # In the case where the USS under test provides a result that is not the same as expected result then the test driver can provide a message detailing why the expected result was populated the way it was. This should help the USS under test debug specific parts of their internal systems. 
 
+### End of definitions aronud users and test executor
+
+### Definitions around interface between scd data generation and test executor 
+  
 class InjectionTarget(ImplicitDict):
-    ''' A class to hold details of a USS under test '''
+    ''' A class to hold details of the USS under test '''
+    uss_sequence_id: int
     name: str
-    uss_base_url: str
 
 class SCDTestConfiguration(ImplicitDict):
     ''' A class to hold test configuration, injection target data  '''
@@ -56,12 +52,14 @@ class FlightInjectionAttempt(ImplicitDict):
     test_injection: InjectFlightRequest
     required_result: RequiredResult
 
-class TestDescription(ImplicitDict):
-    ''' A class to hold different test structures '''
+class TestDetails(ImplicitDict):
+    ''' A class to hold details of different test structures '''
     name: str # e.g. a Nominal test or Flight Authorisation test
     response_wait_time_secs: int # amount of time the test driver should wait before expecting a response
 
 class AutomatedTest(ImplicitDict):
     ''' A class to hold injection attempts and test definitions '''
-    description: TestDescription # Details about the type of test it is
+    description: TestDetails # Details about the type of test it is
     injection_attempts: List[FlightInjectionAttempt]
+
+###  End of Definitions around interface between scd data generation and test executor 
--- a/monitoring/uss_qualifier/scd/data_interfaces.py
+++ b/monitoring/uss_qualifier/scd/data_interfaces.py
@@ -3,7 +3,7 @@ from monitoring.monitorlib.typing import ImplicitDict
 from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
 import datetime
 
-class Issue(ImplicitDict):
+class SCDTestIssue(ImplicitDict):
     ''' A class to hold a message that the test executor can provide to the USS in cases when the USS provides a response that is not same as the expected result. '''
     timestamp: Optional[str]
 
@@ -17,15 +17,14 @@ class Issue(ImplicitDict):
     '''Human-readable description of the issue'''
 
     def __init__(self, **kwargs):
-        super(Issue, self).__init__(**kwargs)
+        super(SCDTestIssue, self).__init__(**kwargs)
         if 'timestamp' not in kwargs:
             self.timestamp = datetime.datetime.utcnow().isoformat()
 
 class RequiredResult(ImplicitDict):
     ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest) '''
     expected_result: str
-    issue_details: Optional[Issue]
-           
+    incorrect_result_details: Optional[SCDTestIssue]
 
 class TestInjectionRequiredResult(ImplicitDict):
     test_injection: InjectFlightRequest
--- a/monitoring/uss_qualifier/scd/data_interfaces.py
+++ b/monitoring/uss_qualifier/scd/data_interfaces.py
@@ -3,8 +3,6 @@ from monitoring.monitorlib.typing import ImplicitDict
 from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
 from monitoring.uss_qualifier.common_data_definitions import Severity
 
-### Begin Definitions around users and test executor 
-
 class Result(ImplicitDict):
     ''' A class to hold the expected output from data processing and the test interface of the USS under test '''
     processing_outcome: str
@@ -29,34 +27,27 @@ class KnownIssueFields(ImplicitDict):
     details: str
     '''Human-readable description of the issue'''
 
-class RequiredResult(ImplicitDict):
-    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest) '''
-    acceptable_results: List[str] # Holds information about the "result" response provided by the test interface of the USS under test. Per the SCD Testing API this should be one of "Planned", "Rejected", "ConflictWithFlight" or "Failed"
-    incorrect_result_details: Optional[Dict[Result, KnownIssueFields]] # In the case where the USS under test provides a result that is not the same as expected result then the test driver can provide a message detailing why the expected result was populated the way it was. This should help the USS under test debug specific parts of their internal systems. 
+class KnownResponses(ImplicitDict):
+    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest). Holds information about the "result" response provided by the test interface of the USS under test. Per the SCD Testing API this should be one of "Planned", "Rejected", "ConflictWithFlight" or "Failed". In the case where the USS under test provides a result that is not the same as expected result then the test driver can provide a message detailing why the expected result was populated the way it was. This should help the USS under test debug specific parts of their internal systems. '''
+    acceptable_results: List[str] 
+    incorrect_result_details: Optional[Dict[Result, KnownIssueFields]]
+    response_wait_time_secs: int # amount of time the test driver should wait before expecting a response
+    response_timeout_message: KnownIssueFields
 
-### End of definitions aronud users and test executor
-
-### Definitions around interface between scd data generation and test executor 
-  
 class InjectionTarget(ImplicitDict):
     ''' A class to hold details of the USS under test '''
     uss_sequence_number: int
-    name: str
 
 class FlightInjectionAttempt(ImplicitDict):
     ''' A class to hold details of the test injection, the injection target and the expected result of processing the flight request '''
     test_injection: InjectFlightRequest
-    required_result: RequiredResult
+    known_responses: KnownResponses
     injection_target: InjectionTarget
 
 class TestDetails(ImplicitDict):
     ''' A class to hold details of different test structures '''
     name: str # e.g. a Nominal test or Flight Authorisation test
-    response_wait_time_secs: int # amount of time the test driver should wait before expecting a response
 
 class AutomatedTest(ImplicitDict):
     ''' A class to hold injection attempts and test definitions '''
-    description: TestDetails # Details about the type of test it is
     injection_attempts: List[FlightInjectionAttempt]
-
-###  End of Definitions around interface between scd data generation and test executor 
--- a/monitoring/uss_qualifier/scd/data_interfaces.py
+++ b/monitoring/uss_qualifier/scd/data_interfaces.py
@@ -40,17 +40,14 @@ class RequiredResult(ImplicitDict):
   
 class InjectionTarget(ImplicitDict):
     ''' A class to hold details of the USS under test '''
-    uss_sequence_id: int
+    uss_sequence_number: int
     name: str
 
-class SCDTestConfiguration(ImplicitDict):
-    ''' A class to hold test configuration, injection target data  '''
-    injection_targets: List[InjectionTarget]
-
 class FlightInjectionAttempt(ImplicitDict):
-    ''' A class to hold details of the test injection and the expected result of processing the flight request '''
+    ''' A class to hold details of the test injection, the injection target and the expected result of processing the flight request '''
     test_injection: InjectFlightRequest
     required_result: RequiredResult
+    injection_target: InjectionTarget
 
 class TestDetails(ImplicitDict):
     ''' A class to hold details of different test structures '''
--- a/monitoring/uss_qualifier/scd/data_interfaces.py
+++ b/monitoring/uss_qualifier/scd/data_interfaces.py
@@ -30,7 +30,7 @@ class KnownResponses(ImplicitDict):
     
 
 class InjectionTarget(ImplicitDict):
-    ''' A class to hold sequential assignment of the USS as per the SCD test API '''
+    ''' A class to hold the role of the USS under test '''
     uss_role: str
 
 class FlightInjectionAttempt(ImplicitDict):
--- a/monitoring/uss_qualifier/scd/data_interfaces.py
+++ b/monitoring/uss_qualifier/scd/data_interfaces.py
@@ -3,10 +3,6 @@ from monitoring.monitorlib.typing import ImplicitDict
 from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
 from monitoring.uss_qualifier.common_data_definitions import Severity
 
-class Result(ImplicitDict):
-    ''' A class to hold the expected output from data processing and the test interface of the USS under test '''
-    processing_outcome: str
-        
 class KnownIssueFields(ImplicitDict):
     ''' A class to hold a message that the test executor can provide to the USS in cases when the USS provides a response that is not same as the expected result. '''    
     test_code: str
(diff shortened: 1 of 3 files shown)


monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,31 @@
+from typing import Optional
+from monitoring.monitorlib.typing import ImplicitDict
+from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
+import datetime
+
+class SCDTestIssue(ImplicitDict):
```
BenjaminPelletier, line 6: I would expect this class to look very similar to the [RID Issue](https://github.com/interuss/dss/blob/10f5192c17695566347eba3e52843ce1c48d06dc/monitoring/uss_qualifier/rid/reports.py#L22) and a key field there is [`queries`](https://github.com/interuss/dss/blob/10f5192c17695566347eba3e52843ce1c48d06dc/monitoring/uss_qualifier/rid/reports.py#L59).  Another key field would be which USS the InjectFlightRequest was sent to.  These fields cannot be populated by the simulator, so the simulator cannot store RequiredResults that contain Issues, so this data structure should be something else -- like KnownIssueFields (no "SCD" prefix needed because this data structure is defined in the module `[...].scd.data_interfaces`.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,31 @@
...
+    ''' A class to hold a message that the test executor can provide to the USS in cases when the USS provides a response that is not same as the expected result. '''
+    timestamp: Optional[str]
```
BenjaminPelletier, line 8: This field is not appropriate for a data structure that will store the output of the simulator because the time the simulator generated the set of `TestInjectionRequiredResult`s is irrelevant.  The time the issue occurred (during test execution by test executor) is definitely relevant and should be included on the SCD Issue, but let's generally try to be architecturally consistent with [existing work](https://github.com/interuss/dss/blob/10f5192c17695566347eba3e52843ce1c48d06dc/monitoring/uss_qualifier/rid/reports.py#L22) and put the SCD Issue in `reports.py` when we are ready to define it (probably not this PR).

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,31 @@
...
+        super(SCDTestIssue, self).__init__(**kwargs)
+        if 'timestamp' not in kwargs:
+            self.timestamp = datetime.datetime.utcnow().isoformat()
+
+class RequiredResult(ImplicitDict):
+    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest) '''
+    expected_result: str
+    incorrect_result_details: Optional[SCDTestIssue]
+
+class TestInjectionRequiredResult(ImplicitDict):
```
BenjaminPelletier, line 29: As currently named, "TestInjection" reads as an adjective to "RequiredResult"; to improve readability, let's change this name to `TestInjectionAndRequiredResult`.  Or, semantically, a better name for this class would probably be something like `FlightInjectionAttempt`.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,31 @@
...
+    '''Human-readable description of the issue'''
+
+    def __init__(self, **kwargs):
+        super(SCDTestIssue, self).__init__(**kwargs)
+        if 'timestamp' not in kwargs:
+            self.timestamp = datetime.datetime.utcnow().isoformat()
+
+class RequiredResult(ImplicitDict):
+    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest) '''
+    expected_result: str
```
BenjaminPelletier, line 26: There may be some tests in the future where the USS may have multiple implementation options and therefore may reasonably produce one of multiple different results (e.g., 'Failed' or 'ConflictWithFlight') -- this might be the case, for instance, if there were two problems with the InjectFlightRequest (one which would cause 'Failed' and one which would cause 'ConflictWithFlight').  So, let's be more general here and change the field to `acceptable_results` with a type of `List[str]`.

BenjaminPelletier, line 26: All fields should be documented in this relatively-small PR describing key interface data structures only

monitoring/uss_qualifier/scd/data_interfaces.py
BenjaminPelletier, line 29: How will test executor know which of the USSs under test to attempt this injection on?

BenjaminPelletier, line 29: All classes should be documented in this relatively-small PR describing key interface data structures only

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,31 @@
...
+    incorrect_result_details: Optional[SCDTestIssue]
+
+class TestInjectionRequiredResult(ImplicitDict):
+    test_injection: InjectFlightRequest
+    required_result: RequiredResult
```
BenjaminPelletier, line 31: nit: let's have [a line break as the last character in the file](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file)

BenjaminPelletier (commented):
uss_qualifier will have multiple tests (two of them already defined fairly explicitly), each requiring multiple injection attempts.  This PR currently lacks information regarding how that necessary information will be provided to test executor (and therefore what the simulator must produce) -- it describes how a single injection attempt will be represented, but not how a test with multiple attempts will be represented, nor how a full session will be represented (though if `AutomatedTest` were defined as containing the multiple injection attempts which constitute a test, we could infer that a full test session would be represented by a bunch of `AutomatedTest` files in a particular folder).  At a minimum, there should be a data structure describing a test (like the [nominal test](https://github.com/interuss/dss/tree/master/interfaces/automated-testing/scd/design/astm-strategic-coordination) or [flight authorisation test](https://github.com/interuss/dss/tree/master/interfaces/automated-testing/scd/design/u-space)), and that data structure should at least contain the list of flight injection attempts (`TestInjectionRequiredResult` here) required to accomplish that test.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks, yes I  have now added a line break

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: I have added a docstring here now.

hrishiballal: Thanks for this I will add configuration class for your review.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks! I have now updated the Requiredresult to have comments / description

hrishiballal: Thanks yes I have now update it.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thank you, I will rename to FlightInjectionAttempt

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thank you for this, I have now removed the timestamp. Re moving to reports.py we can tackle this later but I noticed that it sits in the rid folder, we may need to create a new common report file. I agree that is for later.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thank you for this, I have now renamed KnownIssueFields included more fields from the RID Issue. I have also added the relevant_requirements to this for the moment, my thinking is that here we would reference the Annex 4 of the regulation. I am not sure we can remove it if it is not required.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def __init__(self, **kwargs):
+        super(SCDTestIssue, self).__init__(**kwargs)
+        if 'timestamp' not in kwargs:
+            self.timestamp = datetime.datetime.utcnow().isoformat()
+
+class RequiredResult(ImplicitDict):
+    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest) '''
+    expected_result: str
+    incorrect_result_details: Optional[SCDTestIssue]
```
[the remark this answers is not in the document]
hrishiballal: Thank you for this, I have now updated the file to have a Dict for the incorrect results for your review.

monitoring/uss_qualifier/scd/data_interfaces.py
BenjaminPelletier: The thing I was trying to get at here was that the simulator will probably generate scenarios where injection 1 should go to USS 1, injection 2 to USS 2, injection 3 to USS 1, etc -- the simulator can't know the actual identities of USS 1, USS 2, etc, nor their URLs (which would be defined in a configuration file for a specific test run), but it probably needs a way to tell test executor to which USS out of a supported group of USSs to send each particular injection.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -1,12 +1,30 @@
...
 from monitoring.monitorlib.typing import ImplicitDict
 from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
-import datetime
+from monitoring.monitorlib import fetch
+from monitoring.uss_qualifier.rid import Severity
 
-class SCDTestIssue(ImplicitDict):
-    ''' A class to hold a message that the test executor can provide to the USS in cases when the USS provides a response that is not same as the expected result. '''
-    timestamp: Optional[str]
+class Result(object):
```
BenjaminPelletier, line 7: There are many ways that enums could be treated/expressed, but let's be consistent.  We previously discussed [this](https://github.com/interuss/dss/pull/657#discussion_r775062762) and [this](https://github.com/interuss/dss/pull/657#discussion_r773356609).

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -1,12 +1,30 @@
-from typing import Optional
+from typing import Optional, List, Dict
 from monitoring.monitorlib.typing import ImplicitDict
 from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
-import datetime
+from monitoring.monitorlib import fetch
+from monitoring.uss_qualifier.rid import Severity
```
BenjaminPelletier, line 5: We should not import from rid.  If there are common data structures, those should be moved to a common location, and this PR would be an appropriate place to perform that move.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -16,16 +34,34 @@ class SCDTestIssue(ImplicitDict):
     details: str
     '''Human-readable description of the issue'''
 
-    def __init__(self, **kwargs):
-        super(SCDTestIssue, self).__init__(**kwargs)
-        if 'timestamp' not in kwargs:
-            self.timestamp = datetime.datetime.utcnow().isoformat()
+    queries: List[fetch.Query]
```
BenjaminPelletier, line 37: How is the simulator going to populate this field?

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -1,12 +1,30 @@
...
+    test_code: str
+    """Code corresponding to check generating this issue"""
 
+    relevant_requirements: List[str] = []
+    """Requirements that this issue relates to"""
+
+    severity: Severity
+    """How severe the issue is"""
+
+    injection_target: str
```
BenjaminPelletier, line 25: How is the simulator going to populate this field?

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -16,16 +34,34 @@ class SCDTestIssue(ImplicitDict):
...
-        if 'timestamp' not in kwargs:
-            self.timestamp = datetime.datetime.utcnow().isoformat()
+    queries: List[fetch.Query]
+    """Description of HTTP requests relevant to this issue"""
 
 class RequiredResult(ImplicitDict):
     ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest) '''
-    expected_result: str
-    incorrect_result_details: Optional[SCDTestIssue]
+    acceptable_results: List[str] # Holds information about the "result" response provided by the test interface of the USS under test. Per the SCD Testing API this should be one of "Planned", "Rejected", "ConflictWithFlight" or "Failed"
```
BenjaminPelletier, line 42: Let's try to be consistent with style within a single file at a minimum.  Above classes use triple-single-quote pseudo-docstrings on the line below the declaration whereas this class uses comments after the line.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -16,16 +34,34 @@ class SCDTestIssue(ImplicitDict):
...
+    incorrect_result_details: Optional[Dict[Result, KnownIssueFields]] # In the case where the USS under test provides a result that is not the same as expected result then the test driver can provide a message detailing why the expected result was populated the way it was. This should help the USS under test debug specific parts of their internal systems. 
+
+class InjectionTarget(ImplicitDict):
```
BenjaminPelletier, line 45: Let's make a clear section divider between 1) data types used in the interface between the simulator (or other test definition/data generation tool) and the text executor and 2) data types used in the interface between the user and the test executor.  Those two interfaces are very different -- we shouldn't have data types from interface 2 (InjectionTarget + SCDTestConfiguration) interspersed with data types for interface 1.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -16,16 +34,34 @@ class SCDTestIssue(ImplicitDict):
...
+    ''' A class to hold details of the test injection and the expected result of processing the flight request '''
     test_injection: InjectFlightRequest
-    required_result: RequiredResult
\ No newline at end of file
+    required_result: RequiredResult
+
+class TestDescription(ImplicitDict):
+    ''' A class to hold different test structures '''
+    name: str # e.g. a Nominal test or Flight Authorisation test
+    response_wait_time_secs: int # amount of time the test driver should wait before expecting a response
```
BenjaminPelletier, line 62: Why is a behavior part of a class labeled "Description"?

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks! I have now modified the data structure to have a `sequence_number`, we can change it as necessary,

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thank you, I have now removed the Severity to a new file and referenced it in both the files.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks yes, good point, I have now made it all single quotes.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: My initial thinking was that this would be specified in the simulator but will remove this from the data structure

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks! I will remove this field.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks I have now added the separators.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: I have now renamed the Description -> Details to accommodate the behaviour

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks! I have now made it consistent with the other PR data definitions.

monitoring/uss_qualifier/scd/data_interfaces.py
BenjaminPelletier: As of 39cbc58, this comment has not been addressed.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,62 @@
...
+    summary: str
+    '''Human-readable summary of the issue'''
+
+    details: str
+    '''Human-readable description of the issue'''
+
+class RequiredResult(ImplicitDict):
+    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest) '''
+    acceptable_results: List[str] # Holds information about the "result" response provided by the test interface of the USS under test. Per the SCD Testing API this should be one of "Planned", "Rejected", "ConflictWithFlight" or "Failed"
+    incorrect_result_details: Optional[Dict[Result, KnownIssueFields]] # In the case where the USS under test provides a result that is not the same as expected result then the test driver can provide a message detailing why the expected result was populated the way it was. This should help the USS under test debug specific parts of their internal systems.
```
BenjaminPelletier, line 35: This is [not serializable to JSON](https://stackoverflow.com/questions/9304528/why-json-allows-only-string-to-be-a-key), which is the purpose of `ImplicitDict`.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,62 @@
...
+  
+class InjectionTarget(ImplicitDict):
+    ''' A class to hold details of the USS under test '''
+    uss_sequence_number: int
+    name: str
+
+class FlightInjectionAttempt(ImplicitDict):
+    ''' A class to hold details of the test injection, the injection target and the expected result of processing the flight request '''
+    test_injection: InjectFlightRequest
+    required_result: RequiredResult
```
BenjaminPelletier, line 49: This field does not contain a required result; it contains a description of what to do in the case of different results.  Perhaps `known_responses` or `response_reactions`.  The class should be renamed as well.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,62 @@
+from typing import Optional, List, Dict
+from monitoring.monitorlib.typing import ImplicitDict
+from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
+from monitoring.uss_qualifier.common_data_definitions import Severity
+
+### Begin Definitions around users and test executor
```
BenjaminPelletier, line 6: None of the three classes in this section belong in a separate section from the others because they are all constitute the interface between SCD data generation and test executor -- if we look at a set of generated data that test executor can load, it will contain instances of all these types.

[This comment](https://github.com/interuss/dss/pull/669#discussion_r782463072) was because some test executor configuration parameters (`uss_base_url`) were being included in this file.  The simulator could not provide those inputs (instead, they would be provided at test executor runtime in the test configuration), so they needed to be strongly segregated.  It does not appear that we have that strong segregation need any more.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,62 @@
...
+    injection_target: InjectionTarget
+
+class TestDetails(ImplicitDict):
+    ''' A class to hold details of different test structures '''
+    name: str # e.g. a Nominal test or Flight Authorisation test
+    response_wait_time_secs: int # amount of time the test driver should wait before expecting a response
+
+class AutomatedTest(ImplicitDict):
+    ''' A class to hold injection attempts and test definitions '''
+    description: TestDetails # Details about the type of test it is
```
BenjaminPelletier, line 59: `description` is no longer an appropriate field name

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,62 @@
...
+class FlightInjectionAttempt(ImplicitDict):
+    ''' A class to hold details of the test injection, the injection target and the expected result of processing the flight request '''
+    test_injection: InjectFlightRequest
+    required_result: RequiredResult
+    injection_target: InjectionTarget
+
+class TestDetails(ImplicitDict):
+    ''' A class to hold details of different test structures '''
+    name: str # e.g. a Nominal test or Flight Authorisation test
+    response_wait_time_secs: int # amount of time the test driver should wait before expecting a response
```
BenjaminPelletier, line 55: The `FlightInjectionAttempt` class contains nearly all information about an attempt to inject a flight into a USS: it contains a description of the flight that should be injected, to which USS that flight should be injected, and what test executor should do based on the response it receives from that injection.  Why would we put one aspect of that flight injection attempt in a separate class like this?  Instead, it should be part of `FlightInjectionAttempt`.  The place where would probably make most sense would probably be in the class currently named `RequiredResult` (to be renamed; see other comment) since that class contains information about what test executor should do if it gets various responses, and no response is a response in that context.  The test executor should also be told what to do if it hits this timeout (and thus receives no response) -- the way we're accomplishing that for other types of responses is via a `KnownIssueFields` instance.

BenjaminPelletier (changes requested):

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks for this noted, I misunderstood your comment, I have now made the changes.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,62 @@
...
+
+    details: str
+    '''Human-readable description of the issue'''
+
+class RequiredResult(ImplicitDict):
+    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest) '''
+    acceptable_results: List[str] # Holds information about the "result" response provided by the test interface of the USS under test. Per the SCD Testing API this should be one of "Planned", "Rejected", "ConflictWithFlight" or "Failed"
+    incorrect_result_details: Optional[Dict[Result, KnownIssueFields]] # In the case where the USS under test provides a result that is not the same as expected result then the test driver can provide a message detailing why the expected result was populated the way it was. This should help the USS under test debug specific parts of their internal systems. 
+
+### End of definitions aronud users and test executor
```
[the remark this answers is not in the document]
hrishiballal: Thanks! this is fixed now.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,62 @@
...
+
+### Definitions around interface between scd data generation and test executor 
+  
+class InjectionTarget(ImplicitDict):
+    ''' A class to hold details of the USS under test '''
+    uss_sequence_number: int
+    name: str
```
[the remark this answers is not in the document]
hrishiballal: Thank you for this, I have removed the name and just kept the `uss_sequence_number`

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks, I have now changed it to known_responses and also the definition.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks for this, I have removed the comments separating the definitions

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks I have removed it, originally I was thinking there `nominal` or `flight_authorisation` could be here

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thank you I have added a new timeout response field as a `KnownIssueFields` instance.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks for this, I have now fixed it, I plan to look at this again tomorrow before marking as ready for review.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,53 @@
+from typing import Optional, List, Dict
+from monitoring.monitorlib.typing import ImplicitDict
+from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
+from monitoring.uss_qualifier.common_data_definitions import Severity
+
+class Result(ImplicitDict):
```
BenjaminPelletier, line 6: Is this used anywhere?

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,53 @@
...
+    summary: str
+    '''Human-readable summary of the issue'''
+
+    details: str
+    '''Human-readable description of the issue'''
+
+class KnownResponses(ImplicitDict):
+    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest). This data structure holds information about the "result" response provided by the test interface of the USS under test and details of how long the executor should wait for a response. Per the SCD Testing API an acceptable result should be one of "Planned", "Rejected", "ConflictWithFlight" or "Failed". In the case where the USS under test provides a result that is not the same as expected result then the test driver can provide a message detailing why the expected result was populated the way it was. This should help the USS under test debug specific parts of their internal systems. '''
+    acceptable_results: List[str] 
+    incorrect_result_details: Optional[Dict[str, KnownIssueFields]]
```
BenjaminPelletier, line 33: No need to create an edge case to distinguish between `{}` and `null`; let's just remove `Optional` and therefore always provide an empty `dict` (rather than `null` or omitting) when there are no incorrect result details.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,53 @@
...
+    response_wait_time_secs: int
```
BenjaminPelletier, line 34: Seconds seems unnecessarily granular; let's specify milliseconds (`msecs` or `ms`) instead.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,53 @@
...
+    ''' A class to hold sequential assignment of the USS as per the SCD test API '''
+    uss_sequence_number: int
+
+class FlightInjectionAttempt(ImplicitDict):
+    ''' A class to hold details of the test injection, the injection target and the expected result of processing the flight request '''
+    test_injection: InjectFlightRequest
+    known_responses: KnownResponses
+    injection_target: InjectionTarget
+
+class TestDetails(ImplicitDict):
```
BenjaminPelletier, line 47: Is this used anywhere?

monitoring/uss_qualifier/scd/data_interfaces.py
BenjaminPelletier, line 34: [All fields should be documented in this relatively-small PR describing key interface data structures only](https://github.com/interuss/dss/pull/669#discussion_r780368720)

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,53 @@
...
+
+class KnownResponses(ImplicitDict):
+    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest). This data structure holds information about the "result" response provided by the test interface of the USS under test and details of how long the executor should wait for a response. Per the SCD Testing API an acceptable result should be one of "Planned", "Rejected", "ConflictWithFlight" or "Failed". In the case where the USS under test provides a result that is not the same as expected result then the test driver can provide a message detailing why the expected result was populated the way it was. This should help the USS under test debug specific parts of their internal systems. '''
+    acceptable_results: List[str] 
+    incorrect_result_details: Optional[Dict[str, KnownIssueFields]]
+    response_wait_time_secs: int 
+    response_timeout_message: KnownIssueFields
+
+class InjectionTarget(ImplicitDict):
+    ''' A class to hold sequential assignment of the USS as per the SCD test API '''
```
BenjaminPelletier, line 38: What part of the SCD test API does "as per the SCD test API" refer to?  It seems like the choice of identifying an injection target using an integer sequence number is being made here, not in the API.  `FlightInjectionAttempt.injection_target` needs to refer to a particular USS, but test executor will perform the mapping between "particular USS" and the URL/identity of that USS.  A sequence number is only one way to uniquely identify a particular USS in an `AutomatedTest`, and it is arguably not as good as using a more human-legible role name.  For instance, `InjectionTarget`'s sole field could be `uss_role: str` so that an automated test definition would look like

```json
"injection_attempts": [
  {
    "injection_target": {
      "uss_role": "Good USS"
    }
    "test_injection": {...},
    "known_responses": {...}
  },
  {
    "injection_target": {
      "uss_role": "Querying USS"
    }
    "test_injection": {...},
    "known_responses": {...}
  },
  {
    "injection_target": {
      "uss_role": "Good USS"
    }
    "test_injection": {...},
    "known_responses": {...}
  }
]
```

rather than

```json
"injection_attempts": [
  {
    "injection_target": {
      "uss_sequence_number": 1
    }
    "test_injection": {...},
    "known_responses": {...}
  },
  {
    "injection_target": {
      "uss_sequence_number": 2
    }
    "test_injection": {...},
    "known_responses": {...}
  },
  {
    "injection_target": {
      "uss_sequence_number": 1
    }
    "test_injection": {...},
    "known_responses": {...}
  }
]
```

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,53 @@
...
+    uss_sequence_number: int
+
+class FlightInjectionAttempt(ImplicitDict):
+    ''' A class to hold details of the test injection, the injection target and the expected result of processing the flight request '''
+    test_injection: InjectFlightRequest
+    known_responses: KnownResponses
+    injection_target: InjectionTarget
```
BenjaminPelletier, line 45: nit: The logical flow of the fields in this class would be either "A test injection is sent to the injection target and then known responses are expected", in which case the field order should be `test_injection`, `injection_target`, `known_responses`, or "The injection target is sent a test injection and then known responses are expected", in which case the field order should be `injection_target`, `test_injection`, `known_responses`.  `injection_target` should probably not be the last field defined for this reason.

monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -0,0 +1,53 @@
...
+
+class TestDetails(ImplicitDict):
+    ''' A class to hold details of different test structures '''
+    name: str # e.g. a Nominal test or Flight Authorisation test
```
BenjaminPelletier, line 49: [We want consistent comment style](https://github.com/interuss/dss/pull/669#discussion_r782460219) x3

BenjaminPelletier (changes requested):

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thank you, I have now renamed it o `msecs`

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thank you, ok I have now removed `Optional`

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: I was thinking this can be used in the simulator to generate / hold results, but I will remove it.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Same as the other comment, I was hoping that I will use it in the simulator but I will remove it.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks! I will remove this one and the timeout message.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks yes! It makes much more sense the way you have suggested. The sequence 1, 2 was referring to the [block diagram](https://github.com/interuss/dss/tree/master/interfaces/automated-testing/scd/design/astm-strategic-coordination) for nominal test. Your suggested way is much better.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: This has now been removed. Thanks!

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: This has been dropped / removed from the PR since subsequently.

BenjaminPelletier: Are all fields documented?  "Documented" means that there is a comment associated with the field explaining what the field is/does.  I don't see comments for any of the fields in KnownResponses, InjectionTarget, FlightInjectionAttempt, or AutomatedTest.  All fields *are* documented for KnownIssueFields.

BenjaminPelletier (approved):
I'm concerned about some of the comments capturing *how* classes currently achieve their purpose rather than *what* their purpose is, but I think we can follow up on that in a separate PR.

monitoring/uss_qualifier/scd/data_interfaces.py
hrishiballal: Thanks for this, I have now, commented every fields in the file

hrishiballal:
Thanks! Yes lets talk about the _how_ in our call, I am happy open additional PRs to cover that as well based on your guidance

PR #675, opened 2022-01-20
[prober] Test ISAs and Subscriptions with barely overlapping areas

pratibhagupta2109:


BenjaminPelletier:
With a3227f4, the areas are now overlapping more than we want I think -- the original edge-sharing was really close to what we want, but we just want a very slight amount of true overlap rather than having a mathematical overlap of 0 area.  A good amount of overlap would be order of 1 centimeter.  Since 360 degrees of latitude or longitude equals the circumference of the earth (40e8 cm), moving a point 1 cm equates to 1/40e8 * 360 ~= 1e-7 degrees.  So, for instance, we might move latitude -23.6298 to -23.6297999

pratibhagupta2109:
> 

As per my understanding from the above description, I see doing -1e-7 to -23.6407 instead of +1e-7. Let me know if it is not the case.

BenjaminPelletier (approved):

PR #676, opened 2022-01-20
[uss_qualifier/scd] Adjust comments and formatting on interface between test executor and simulator

BenjaminPelletier:
Following up on #669, this PR adjusts formatting to follow [PEP 257](https://www.python.org/dev/peps/pep-0257/) and adjusts documentation descriptions per [this comment](https://github.com/interuss/dss/pull/669#pullrequestreview-858520616).

### DIFF
--- a/monitoring/uss_qualifier/scd/data_interfaces.py
+++ b/monitoring/uss_qualifier/scd/data_interfaces.py
@@ -3,52 +3,59 @@ from monitoring.monitorlib.typing import ImplicitDict
 from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightRequest
 from monitoring.uss_qualifier.common_data_definitions import Severity
 
+
 class KnownIssueFields(ImplicitDict):
-    ''' A class to hold a message that the test executor can provide to the USS in cases when the USS provides a response that is not same as the expected result. '''    
+    """Information, which can be defined at the time of test design, about a problem detected by an automated test when a USS provides a response that is not same as the expected result"""
     test_code: str
-    '''Code corresponding to check generating this issue'''
+    """Code corresponding to check generating this issue"""
 
     relevant_requirements: List[str] = []
-    '''Requirements that this issue relates to'''
+    """Requirements that this issue relates to"""
 
     severity: Severity
-    '''How severe the issue is'''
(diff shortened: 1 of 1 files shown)


monitoring/uss_qualifier/scd/data_interfaces.py
```diff
@@ -3,52 +3,59 @@
...
-    ''' A class to evaluate results / response to an injection of test flight data (TestFlightRequest). This data structure holds information about the "result" response provided by the test interface of the USS under test and details of how long the executor should wait for a response. Per the SCD Testing API an acceptable result should be one of "Planned", "Rejected", "ConflictWithFlight" or "Failed". In the case where the USS under test provides a result that is not the same as expected result then the test driver can provide a message detailing why the expected result was populated the way it was. This should help the USS under test debug specific parts of their internal systems. '''
+    """Mapping of the flight injection attempt's USS response to test outcome"""
     acceptable_results: List[str] 
-    ''' Acceptable strings that the USS under test can report as the result of processing the test data. '''
+    """Acceptable InjectFlightResponse `result` values that will result in the flight injection attempt being considered successful if the USS under test reports one of them as the result of attempting to inject the requested flight."""
```
hrishiballal, line 31: Consider: Acceptable values in the `result` data field of InjectFlightResponse. The flight injection attempt will be considered successful if the USS under test reports one of these as the result of test data processing and / or  attempting to inject the flight.

hrishiballal (approved):

monitoring/uss_qualifier/scd/data_interfaces.py
BenjaminPelletier: Mostly adopted, but after removing "test data processing and / or".  The injection endpoint is semantically an endpoint to trigger creation of a flight with the given characteristics.  The data provided has to be processed, but that is incidental to the goal of treating the InjectFlightRequest as an attempt by a user to create a flight with those specifications.  If a USS had a bespoke routine (one not involved in the user-created flight creation pipeline) to first process the incoming test data before forwarding it on to the subsystem that creates the flight, that would essentially be cheating the automated test since logic not involved in the production system would be responsible for the success of some of the automated tests.

barroco (approved):

Issue #616, opened 2021-11-13
Migrate from lib/pq to jackc/pgx


BenjaminPelletier:
[The Postgres client we are using](https://github.com/lib/pq) is in maintenance mode and suggests that [pgx](https://github.com/jackc/pgx) is more actively developed.  We should migrate to the more actively-developed client.

See #467 for an old, un-merged PR that attempted to do this.

pratibhagupta2109:
There are few concerns reported by other developers regarding the migration of` lib/pg` to` jackc/pgx `as listed in below mentioned comment and issue:
https://github.com/golang-migrate/migrate/issues/512#issuecomment-787008491
https://github.com/lib/pq/issues/1022

Please review if these are applicable to our migration too.

BenjaminPelletier:
Those are both great links to know about.  On a cursory read, my guess is that the problems may have primarily stemmed from user error, but I'm definitely not sure.  To mitigate our risk, it would be great to do as little work as possible to get pgx working on enough of the DSS to allow us to benchmark its performance (and verify it doesn't have lots of errors).  I think a good performance benchmark will be one or more of our "heavy concurrent" tests -- we can (perhaps temporarily) add a couple lines to those tests to note the clock before the test starts and after it completes, then run the test ~5 times with pq then ~5 times with pgx and compare the performance.  If pgx performs at least, say, 90% as well as pq and doesn't have errors, then switching to a maintained (and recommended) project is probably a good strategy.  If we see errors or substantial performance problems, then we'll know to stick with pq.

BenjaminPelletier:
Resolved by #679

PR #679, opened 2022-01-22
[crdb] Migrate postgres pq to pgx

pratibhagupta2109:
As pq library is in maintenance mode, it is suggested to use [pgx](https://github.com/jackc/pgx) for postgres connection.

There is a significant performance improvement noted by the use of pgx over the standard database/sql, which is shown through unit tests run below:

Time taken by concurrent ops using database/sql:
Screen Shot 2022-01-27 at 7 48 32 am

Time taken by concurrent ops using pgx:
Screen Shot 2022-01-27 at 7 52 41 am


BenjaminPelletier (commented):
Wow, this migration was very involved -- thanks for getting it done!  One more thing I think we want to do before merging this PR is to make sure this new driver doesn't [introduce problems](https://github.com/interuss/dss/issues/616#issuecomment-1006239857).  It looks like we're not getting any errors, so that's good and verified by our existing prober tests.  The other thing we should check is [whether queries take noticeably longer with the new driver](https://github.com/interuss/dss/issues/616#issuecomment-1006912871).  The clearest way to do that check would probably be to create a separate, small PR that adds a bit of code to print/log how long queries took (on average) in the heavy concurrent tests.  After merging that PR, we could simply run it to get a baseline.  Then, we could merge that change into this branch and run it again to see the performance with the new driver.

BenjaminPelletier (approved):

PR #680, opened 2022-01-26
[dummy-oauth] Add API definition for Dummy OAuth

BenjaminPelletier:
We currently have a webserver in cmds/dummy-oauth that generates DSS-consumable access tokens for development purposes.  Its API is currently implicit in the Go code implementing the handler.  This PR makes the API explicit and also adds a README explaining the tool and intended usage.  Not only will this make the tool more usable and approachable by new users, but it also prepares for using the new interfaces/openapi-to-go-server tool to autogenerate the webserver part of Dummy OAuth.

### DIFF
--- a/cmds/dummy-oauth/main.go
+++ b/cmds/dummy-oauth/main.go
@@ -19,7 +19,7 @@ import (
 
 var (
 	address = flag.String("addr", ":8085", "address")
-	keyFile = flag.String("private_key_file", "build/test-certs/oauth.key", "oauth private key file")
+	keyFile = flag.String("private_key_file", "build/test-certs/auth2.key", "oauth private key file")
 )
 
 // TODO(steeling): add other parameters so we can test expired tokens, invalid tokens, etc.
--- a/interfaces/dummy-oauth/dummy-oauth.yaml
+++ b/interfaces/dummy-oauth/dummy-oauth.yaml
@@ -27,18 +27,21 @@ paths:
         description: Fully-qualified domain name where the service for which this access token will be used is hosted.  The `aud` claim will be populated with this value.
         schema:
           type: string
+        example: uss.example.com
       - name: scope
         in: query
         required: true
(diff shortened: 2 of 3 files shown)


cmds/dummy-oauth/main.go
```diff
@@ -19,7 +19,7 @@ import (
 
 var (
 	address = flag.String("addr", ":8085", "address")
-	keyFile = flag.String("private_key_file", "build/test-certs/oauth.key", "oauth private key file")
+	keyFile = flag.String("private_key_file", "build/test-certs/auth2.key", "oauth private key file")
```
barroco, line 22: What is the reason for changing the default key?

interfaces/dummy-oauth/dummy-oauth.yaml
```diff
@@ -0,0 +1,63 @@
...
+paths:
+  /token:
+    get:
+      parameters:
+      - name: intended_audience
+        in: query
+        required: true
+        description: Fully-qualified domain name where the service for which this access token will be used is hosted.  The `aud` claim will be populated with this value.
+        schema:
+          type: string
```
barroco, line 29: May I suggest to add value examples?

cmds/dummy-oauth/main.go
barroco: Perhaps, it would makes sense to document the purpose of the files in build/test-certs/*

barroco (approved):
Small suggestions in comments.

cmds/dummy-oauth/main.go
BenjaminPelletier: As far as I know, `oauth.key` (and `oauth.crt`) are not used anywhere in the repo and should probably be deleted (especially since `oauth.crt` isn't actually the public key for `oauth.key`!).  The `auth2` key pair is what is used in all the local/dummy/development configurations, so it's almost always what a user wants to use.  I've added #682 to track separately from this PR.

interfaces/dummy-oauth/dummy-oauth.yaml
BenjaminPelletier: Good call; done.

PR #686, opened 2022-01-27
[docs] Add missing doc.go files

pratibhagupta2109:


pkg/rid/store/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package store provides the interface to obtain Repos, to interact with
```
BenjaminPelletier, line 1: "Package store defines interfaces that every RID Store implementation, acting on an RID Repository, must implement"

pkg/scd/models/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package models bundles up types and functions used for performing
```
BenjaminPelletier, line 1: Similar suggestion as for pkg/rid/models

pkg/scd/store/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package store provides the interface to obtain Repos, to interact with
```
BenjaminPelletier, line 1: Similar suggestion as for pkg/rid/store

pkg/version/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package version provides types and functions for getting current
+// version.
```
BenjaminPelletier, line 2: "...getting the version of the executable defined at build time."

BenjaminPelletier (approved):

PR #692, opened 2022-01-31
[core-service] Do not panic when TotalConns==0

### LINKED ISSUE #691: core-service panics when database connection is still good
#679 included a change from `pingDB` to `getDBStats`.  One purpose of both routines was to verify a good connection to the database, or else kill the core-service process so that the Kubernetes manager could restart it and (hopefully) restore the database connection.  `pingDB` used `PingContext` to verify connectivity whereas `getDBStats` checked `TotalConns == 0`.  We verified that `TotalConns == 0` occurs when the connection to the database is broken, but it turns out that also occurs when the connection to the database is idle.  Because of this, core-service will die after tens of minutes of inactivity even though everything is still working.  The simplest fix to this problem is to only warn when `TotalConns == 0`, and do not attempt to kill core-service upon bad database connection.  Instead, if the motivating issue is detected again, the database client should be improved to attempt to restore the connection upon initial failure without returning an error (this will be necessary at higher request volume any way).

BenjaminPelletier:
This PR fixes #691 in the near-mid term.  Long term, we will want to check our database client's behavior.

### DIFF
--- a/cmds/core-service/main.go
+++ b/cmds/core-service/main.go
@@ -74,7 +74,7 @@ func getDBStats(ctx context.Context, db *cockroach.DB, databaseName string) {
 	stats["MaxConns"] = strconv.Itoa(int(statsPtr.MaxConns()))
 	stats["TotalConns"] = strconv.Itoa(int(statsPtr.TotalConns()))
 	if stats["TotalConns"] == "0" {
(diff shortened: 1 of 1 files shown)


barroco (approved):

barroco:
Looking at #691 description, I support the idea of checking our database client's behavior taking into account the work in #688 in a future iteration.

PR #701, opened 2022-02-04
[docs/automated-testing] Restructure scd automated-testing interface documentation

BenjaminPelletier:
Our approach to SCD automated testing has become more solid since first creating the documentation in interfaces/automated-testing/scd/design, and this PR aims to restructure that documentation more formally to more tightly link documentation and implementation.

Following this PR, I envision that each test group folder (astm-strategic-coordination and u-space right now) contains one subfolder per "test", and a "test" maps directly to an instance of [AutomatedTest](https://github.com/interuss/dss/blob/64b02294e2fa1e76d0d3f95693fed85760a7c257/monitoring/uss_qualifier/scd/data_interfaces.py#L55).  Therefore, it makes sense to identify an AutomatedTest according to this structure -- so, the IDs of the two tests we have currently defined should be `astm-strategic-coordination/nominal-planning` and `u-space/flight-authorisation-validation`.  This means that the documentation for a test can be linked automatically by appending the ID of the test to the base URL of https://github.com/interuss/dss/tree/master/interfaces/automated-testing/scd/design/  This also suggests a very clear interface for test definition files: each AutomatedTest should be located at `<LOCALE>/<TEST_ID>.json` -- so, for the two tests currently defined: `CHE/astm-strategic-coordination/nominal-planning.json` and `CHE/u-space/flight-authorisation-validation.json`.  Also, the actor names in the sequence diagrams can map directly to the [`uss_role`](https://github.com/interuss/dss/blob/64b02294e2fa1e76d0d3f95693fed85760a7c257/monitoring/uss_qualifier/scd/data_interfaces.py#L39) in the AutomatedTest definition.

Based on the paradigm above, reading of SCD automated test files is also updated in test_executor.py.


barroco (approved):
👍

PR #702, opened 2022-02-04
[uss_qualifier/scd] Add flight deletion capability to SCD automated testing

BenjaminPelletier:
Currently, there is no way to instruct a USS to delete a flight created for an automated test.  This means that the system cannot be effectively cleared following the completion of a test.  This PR fixes that problem by adding a "delete flight" operation in the SCD automated testing API.

This PR also adjusts the definition of an AutomatedTest to specify generic "test steps" (rather than requiring every step to be a flight injection attempt) and includes the deletion of a flight as an action that can be defined within an automated test.  This would be useful to, e.g., check the caching behavior of a USS by injecting flight 1 into USS1, injecting flight 2 which is nearby but not conflicting into USS2, deleting flight 1 from USS1, then injecting flight 3 which conflicts with now-absent flight 1 into USS2.

My expectation is that the test driver should always clean up after its tests -- that is, the test driver should call "delete flight" on every flight injection request that resulted in Planned in the course of the test.  So, the AutomatedTest definitions would not need to specify deletion of a created flight at the end of the test in general -- this capability of prescribing flight deletions would only be used in special cases like the paragraph above.

Finally, mock USS is updated to support this new deletion operation.

### DIFF
--- a/interfaces/automated-testing/scd/scd.yaml
+++ b/interfaces/automated-testing/scd/scd.yaml
@@ -439,7 +439,7 @@ components:
 
           type: string
           enum: [Planned, Rejected, ConflictWithFlight, Failed]
-          example: Planned
+          example: ConflictWithFlight
         notes:
           description: >-
             Human-readable explanation of the observed result.  This explanation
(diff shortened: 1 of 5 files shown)


barroco (approved):
👍

PR #703, opened 2022-02-05
[uss_qualifier] Add cross-process access to in-memory databases

BenjaminPelletier:
The simple in-memory "databases" used by the ridsp and riddp mock USS modules are not cross-process -- that is, if gunicorn spawns multiple workers to serve mock USS requests, updating the "database" in worker 1 will not be reflected in worker 2.  This was acceptable in ridsp and riddp, but not in scdsc because scdsc will sometimes call itself while handling a request.  This PR makes "databases" for all mock USS modules cross-process, meaning that all gunicorn workers will share the same database, using a new class named SynchronizedValue.

Note that #700 is required to successfully test these changes.


monitoring/mock_uss/riddp/routes_observation.py
```diff
@@ -108,7 +109,9 @@ def display_data() -> Tuple[str, int]:
     # Construct detailed flights response
     response = observation_api.GetDisplayDataResponse(flights=[
       _make_flight_observation(f, view) for f in validated_flights])
-    db.flights = flight_info
+    with db as tx:
+        for k, v in flight_info.items():
+            tx[k] = v
```
barroco, line 114: Shouldn't it be tx.flights[k] ?

barroco (approved):

monitoring/mock_uss/riddp/routes_observation.py
BenjaminPelletier: Yes, excellent catch -- we don't have an automated test that would have revealed this bug yet

PR #705, opened 2022-02-08
[db-manager] Add db-manager API

BenjaminPelletier:
This PR proposes a webserver interface for db-manager according to the general plan described [here](https://lists.interussplatform.org/g/dss/topic/making_db_manager_a_service/88980821).  If/once this PR is merged, I'll add a `--webserver` flag to db-manager which, when specified, will cause it to start a webserver implementing this interface rather than performing a single database schema migration (though the single migration can still be performed by specifying different flags).

### DIFF
--- a/interfaces/db-manager/db-manager.yaml
+++ b/interfaces/db-manager/db-manager.yaml
@@ -1,0 +1,202 @@
+openapi: 3.0.2
+info:
+  title: Database Manager
+  version: 1.0.0
+  description: >-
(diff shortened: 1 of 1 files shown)


BenjaminPelletier:
Upon further investigation regarding deployment approaches, it appears that running one-off jobs should probably not need to be as difficult as it currently happens to be.  So, it isn't worth the additional risk and overhead an API presents to enable it, so we won't pursue the approach started in this PR.

PR #706, opened 2022-02-08
[crdb] Code cleaning post pgx migration

pratibhagupta2109:


pkg/cockroach/cockroach.go
```diff
@@ -145,19 +145,6 @@ func Dial(ctx context.Context, connParams ConnectParameters) (*DB, error) {
 	}, nil
 }
```
pratibhagupta2109, line 147: I tried removing config, `err := pgxpool.ParseConfig(uri)` at line 125, but `pgxpool.ConnectConfig(ctx, config) `accepts `pgxpool.Config` type parameter, which is returned by `pgxpool.ParseConfig(uri)`.

BenjaminPelletier: `pgxpool.ParseConfig(uri)` is one way to obtain a [`pgxpool.Config`](https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool#Config), but another way is to simply instantiate one directly; e.g.:

```go
config := &pgxpool.Config{
    ConnConfig: &pgx.ConnConfig{
        Host: p.Host,
        Port: p.Port,
        ...
    },
    MaxConns: int32(connParams.MaxOpenConns),
    MaxConnIdleTime: (time.Duration(connParams.MaxConnIdleSeconds) * time.Second),
    ...
}
```

(Note that pgx.ConnConfig inherits from [pgconn.Config](https://pkg.go.dev/github.com/jackc/pgconn#Config))

pratibhagupta2109: If I create pgxpool.Config object directly, I get `panic: config must be created by ParseConfig` error from https://github.com/jackc/pgx/blob/master/conn.go#L193.
So I have created the config by passing connection params in [DSN format](https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool#ParseConfig).
It works fine for e2e tests. Please check if works for real params too.

BenjaminPelletier (approved):

PR #696, opened 2022-02-02
[crdb/schemas] Update SCD DB down migrate scripts.

pratibhagupta2109:
downfrom* scripts have been tested by running following commands in sequence:
```
test/migrations/clear_db.sh
test/migrations/migrate_db.sh scd latest
test/migrations/migrate_db.sh scd <older-version>
```

example:
`test/migrations/migrate_db.sh scd 2.0.0`

PR #708, opened 2022-02-10
[dev tools] Expand and document local SCD data injection tools

BenjaminPelletier:
#696 added tools to inject data into an SCD database for the purpose of verifying database down migration behavior.  This PR simplifies the usage of those tools and generalizes them for potential other purposes as well.  It also documents the tools' purpose and usage in the README.

This PR also splits out the two types of lint (Go and shell) so that the shell lint (much quicker) can be run independently from the Go lint.


barroco (approved):
👍

PR #710, opened 2022-02-14
[uss_qualifier/scd] Test executor

barroco:
This PR implements an initial version of the test executor for SCD qualification. Reporting and findings gathering will follow in another PR. 

- The test runner includes a simple Dry mode to evaluate test_definitions.
- The PR introduces two valid test_definitions for test purposes. Those should be removed when #685 will be ready.

Test (Make sure dependencies are correctly started. See disclaimer in console when running scripts): 
> cd monitoring/uss_qualifier
> ./run_locally_scd.sh

### DIFF
--- a/monitoring/mock_uss/scdsc/database.py
+++ b/monitoring/mock_uss/scdsc/database.py
@@ -10,7 +10,7 @@ from monitoring.monitorlib.typing import ImplicitDict
 class FlightRecord(ImplicitDict):
     """Representation of a flight in a USS"""
     op_intent_injection: scd_injection_api.OperationalIntentTestInjection
-    flight_authorisation: Optional[scd_injection_api.FlightAuthorisationData]
+    flight_authorisation: scd_injection_api.FlightAuthorisationData
     op_intent_reference: scd.OperationalIntentReference
 
 
--- a/monitoring/mock_uss/scdsc/routes_injection.py
+++ b/monitoring/mock_uss/scdsc/routes_injection.py
@@ -117,7 +117,7 @@ def inject_flight(flight_id: str) -> Tuple[str, int]:
     record = database.FlightRecord(
         op_intent_reference=result.operational_intent_reference,
         op_intent_injection=req_body.operational_intent,
-        flight_authorisation=req_body.flight_authorisation if 'flight_authorisation' in req_body else None)
+        flight_authorisation=req_body.flight_authorisation)
     with db as tx:
         tx.flights[flight_id] = record
 


monitoring/monitorlib/scd_automated_testing/scd_injection_api.py
```diff
@@ -38,14 +38,15 @@ class FlightAuthorisationData(ImplicitDict):
 class InjectFlightRequest(ImplicitDict):
     ''' A class to hold the details of a test injection payload '''
     operational_intent: OperationalIntentTestInjection
-    flight_authorisation: FlightAuthorisationData
+    flight_authorisation: Optional[FlightAuthorisationData]
```
BenjaminPelletier, line 41: I agree that the first thought should probably be that if the uss_qualifier isn't checking for U-space compliance, the flight authorisation information may not be necessary.  However, I can imagine a circumstance in which some USSs have designed their systems with U-space in mind, yet we don't want to check for U-space compliance.  In that case, it would still be very useful to the USSs to have U-space information provided to them since they're designed to use it.  And, it's no problem for uss_qualifier to generate this mock data, so I'd suggest we provide all information all the time from uss_qualifier.  Concretely, this means we would provide flight_authorisation even if the uss_qualifier wasn't going to validate anything related to that information, and even if that information wasn't relevant in the locality for which the test was being conducted (e.g., US).  In those cases, it's still easy for the USS to ignore the flight_authorisation field.

So, I'd recommend we keep this field as non-Optional.

monitoring/monitorlib/scd_automated_testing/scd_injection_api.py
```diff
@@ -38,14 +38,15 @@ class FlightAuthorisationData(ImplicitDict):
...
 
 
 class InjectFlightResult(str, Enum):
     Planned = 'Planned'
     Rejected = 'Rejected'
     ConflictWithFlight = 'ConflictWithFlight'
     Failed = 'Failed'
+    DryRun = 'DryRun'
```
BenjaminPelletier, line 49: This is the API USSs under test will see, so I don't think we want an option to respond with "DryRun" from them

monitoring/uss_qualifier/scd/executor/target.py
```diff
@@ -0,0 +1,51 @@
...
+
+from monitoring.monitorlib import infrastructure, auth
+from monitoring.monitorlib.clients.scd_automated_testing import create_flight, delete_flight
+from monitoring.monitorlib.scd_automated_testing.scd_injection_api import InjectFlightResult, \
+    DeleteFlightResult, InjectFlightResponse, DeleteFlightResponse
+from monitoring.uss_qualifier.rid.utils import InjectionTargetConfiguration
+from monitoring.uss_qualifier.scd.data_interfaces import FlightInjectionAttempt
+
+
+class TestTarget():
```
BenjaminPelletier, line 12: I'd recommend implementing the concept of a "dry run" almost entirely within this class.  One approach could be to have a `dry_run` flag that, when set, causes this class not to make any calls to the actual target, and instead always assume failure.  But actually, that's not quite what we want because often times, inject_flight should return a result other than Failed -- so maybe a `dry_run` flag isn't enough, and instead we could add something like `mocked_responses` which is a list of `Tuple[InjectFlightRequest, InjectFlightResponse]` and the appropriate response is looked up there according to the request received (instead of calling out to the real client).  But that seems to be getting pretty complicated -- it actually seems simpler to just spin up a few mock_uss instances and run the uss_qualifier normally.  Do we really need to be able to perform dry runs instead of running normally against mocks?

monitoring/uss_qualifier/scd/test_definitions/CHE/automated_test/automated_test_1.json
```diff
@@ -0,0 +1,659 @@
+{
+    "name": "Nominal Planning Test",
```
BenjaminPelletier, line 2: If this is a draft of the the nominal planning test, could we put it in `CHE/astm-strategic-coordination/nominal-planning.json` to mirror [the documentation](https://github.com/interuss/dss/tree/master/interfaces/automated-testing/scd/design/astm-strategic-coordination/nominal-planning)?

monitoring/uss_qualifier/scd/test_definitions/CHE/automated_test/automated_test_1.json
```diff
@@ -0,0 +1,659 @@
...
+                    }
+                },
+                "injection_target": {
+                    "uss_role": "Second USS"
+                },
+                "name": "za0q1nwk"
+            }
+        },
+        {
+            "name": "Delete first injected flight",
```
BenjaminPelletier, line 653: I'd recommend that we don't need this step -- uss_qualifier should attempt clean up after every test, so if there is still an injected flight that hasn't been deleted, it should automatically detect that and try to fix it (no need to explicitly say to delete it since the implicit process will handle it).  Same with other automated test.

monitoring/uss_qualifier/scd/test_definitions/CHE/automated_test/automated_test_2.json
```diff
@@ -0,0 +1,116 @@
+{
+    "name": "Flight Authorisation validation test",
```
BenjaminPelletier, line 2: If this is a draft of the the flight authorisation data format test, could we put it in `CHE/u-space/flight-authorisation-validation.json` to mirror [the documentation](https://github.com/interuss/dss/tree/master/interfaces/automated-testing/scd/design/u-space/flight-authorisation-validation)?

monitoring/uss_qualifier/scd/executor/runner.py
```diff
@@ -0,0 +1,105 @@
...
+        print("[SCD]     Result: SUCCESS")
+        return True
+
+    def print_targets_state(self):
+        print("[SCD] Targets States:")
+        for name, target in self.targets.items():
+            print(f"[SCD]   - {name}: {target.created_flight_ids}")
+
+    def print_test_plan(self):
+        """Runs the automated test in dry mode to validate and print the test plan."""
```
BenjaminPelletier, line 99: If we keep this, I think it would benefit from some additional documentation.  What does it mean to print the test plan?  It seems like what the test executor does is going to depend on how the USSs respond, so what does it do if they're not responding?

monitoring/uss_qualifier/scd/executor/runner.py
```diff
@@ -0,0 +1,105 @@
...
+        if 'inject_flight' in step:
+            return self.targets[step.inject_flight.injection_target.uss_role]
+        elif 'delete_flight' in step:
+            return self.get_managing_target(step.delete_flight.flight_name)
+        else:
+            raise NotImplementedError("Unsupported step. A Test Step shall contain either a inject_flight or a delete_flight object.")
+
+    # TODO: Use this method as a canvas to create findings and move the evaluation at the end of or outside the execution.
+    @staticmethod
+    def evaluate_inject_flight_response(attempt: FlightInjectionAttempt, resp: InjectFlightResponse, dry=False) -> bool:
```
BenjaminPelletier, line 79: I'd recommend not conditioning logic here on whether it's a "dry run"

monitoring/uss_qualifier/scd/executor/runner.py
```diff
@@ -0,0 +1,105 @@
...
+        if 'inject_flight' in step:
+            print("[SCD]     Step: Inject flight {} to {}".format(step.inject_flight.name, target.name))
+            resp = target.inject_flight(step.inject_flight, dry=dry)
+            # TODO: Move evaluation at the end of the execution
+            TestRunner.evaluate_inject_flight_response(step.inject_flight, resp, dry=dry)
+        elif 'delete_flight' in step:
+            print("[SCD]     Step: Delete flight {} to {}".format(step.delete_flight.flight_name, target.name))
+            target.delete_flight(step.delete_flight.flight_name, dry=dry)
+        else:
+            print("[SCD] Warning: no action defined for step {}".format(step.name))
```
BenjaminPelletier, line 59: It seems like this error should be fatal (not just print a warning)

monitoring/uss_qualifier/scd/executor/runner.py
```diff
@@ -0,0 +1,105 @@
...
+            # TODO implement reporting
+            self.print_targets_state()
+            raise RuntimeError("[SCD] Error: Unable to identify the target managing flight {}".format(
+                step.inject_flight.name if 'inject_flight' in step else step.delete_flight.flight_name
+            ))
+
+        if 'inject_flight' in step:
+            print("[SCD]     Step: Inject flight {} to {}".format(step.inject_flight.name, target.name))
+            resp = target.inject_flight(step.inject_flight, dry=dry)
+            # TODO: Move evaluation at the end of the execution
```
BenjaminPelletier, line 53: Can we just resolve this TODO right now?

BenjaminPelletier (commented):
This all looks pretty good to me with the exception of the concept of a dry run -- I don't think I understand how that should work and how it helps, as compared with just running normally with an appropriate number of instances of mock_uss

monitoring/monitorlib/scd_automated_testing/scd_injection_api.py
barroco: I completely removed the dry run since it doesn't bring value at this point in time. It will be the topic of another PR if useful.

monitoring/monitorlib/scd_automated_testing/scd_injection_api.py
barroco: Thank you for the clarification, I reverted this change and updated the test_definitions.

monitoring/uss_qualifier/scd/executor/target.py
barroco: I started to look into mocking the responses as you proposed. I will keep it for another PR if this is ok for you.

monitoring/uss_qualifier/scd/test_definitions/CHE/automated_test/automated_test_1.json
barroco: Done

monitoring/uss_qualifier/scd/executor/runner.py
barroco: I removed it since the mocking approach seems more appropriate.

monitoring/uss_qualifier/scd/executor/runner.py
barroco: No more dry run.

monitoring/mock_uss/scdsc/database.py
```diff
@@ -10,7 +10,7 @@
 class FlightRecord(ImplicitDict):
     """Representation of a flight in a USS"""
     op_intent_injection: scd_injection_api.OperationalIntentTestInjection
-    flight_authorisation: scd_injection_api.FlightAuthorisationData
+    flight_authorisation: Optional[scd_injection_api.FlightAuthorisationData]
```
BenjaminPelletier, line 13: nit: I'd make this non-optional in accordance with [this comment](https://github.com/interuss/dss/pull/710#discussion_r806115352)

monitoring/mock_uss/scdsc/routes_injection.py
```diff
@@ -118,7 +117,7 @@ def inject_flight(flight_id: str) -> Tuple[str, int]:
     record = database.FlightRecord(
         op_intent_reference=result.operational_intent_reference,
         op_intent_injection=req_body.operational_intent,
-        flight_authorisation=req_body.flight_authorisation)
+        flight_authorisation=req_body.flight_authorisation if 'flight_authorisation' in req_body else None)
```
BenjaminPelletier, line 120: nit: if flight_authorisation is made non-optional per comment above, then this change isn't needed either

monitoring/uss_qualifier/scd/executor/target.py
BenjaminPelletier: Definitely -- I think smaller, separable PRs are best in general, subject to the fixed overhead of creating the PR and getting it reviewed

monitoring/uss_qualifier/scd/executor/test_executor.py
```diff
@@ -0,0 +1,77 @@
...
+                    flight_authorisation=None
+                ),
+                known_responses=KnownResponses(
+                    acceptable_results=[InjectFlightResult.Planned],
+                    incorrect_result_details={}
+                ),
+                injection_target=InjectionTarget(uss_role=SecondMoverRole)
+            )
+        ),
+        TestStep(
```
BenjaminPelletier, line 48: nit: I don't [think](https://github.com/interuss/dss/pull/710#discussion_r806135052) we need this step

BenjaminPelletier (approved):
I have a few nits below, but this looks good to merge regardless of if/how they're resolved.  Thanks!

monitoring/uss_qualifier/scd/executor/test_executor.py
barroco: Since TestStep supports deleting flights as part of scenarios (even if not used yet), I think we should keep it in the tests.

monitoring/mock_uss/scdsc/routes_injection.py
barroco: Done thanks!

PR #713, opened 2022-02-17
[mock_uss/scd] Validate flight authorisation fields in mock USS

BenjaminPelletier:
The algorithms and assumptions for serial number and operator registration number are from #685 and I do not have a way to validate their correctness.


barroco (approved):
Many thanks, looks great. In the description you are referring to #710 but I think you meant #685.
I think it would be valuable in the short future to add unit tests with examples of valid and invalid values. Let's discuss this as part of #685.

BenjaminPelletier:
Yes, #685; comment updated, thanks.  Definitely unit tests for those two constructs would be good.

PR #715, opened 2022-02-19
[ci] Clear image cache and print logs in case of failure

barroco:
This PR improves the script used to test the scd qualifier by forcing the images of the mocks to be built without cache and it prints the container logs in case of starting failure.

Note: USS Qualifier Tests check is failing because of the issue addressed by PR #716.

### DIFF
--- a/monitoring/uss_qualifier/scripts/test_docker.sh
+++ b/monitoring/uss_qualifier/scripts/test_docker.sh
@@ -10,6 +10,8 @@ echo "Ensure the environment is clean"
 echo "============="
 docker-compose -f docker-compose_qualifier_mocks.yaml -p $project_name down
 
+echo "Rebuild images without cache"
+docker-compose -f docker-compose_qualifier_mocks.yaml -p $project_name build --no-cache
 
 echo "Start mocks"
 echo "============="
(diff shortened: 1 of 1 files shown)


monitoring/uss_qualifier/scripts/test_docker.sh
```diff
@@ -10,6 +10,8 @@ echo "Ensure the environment is clean"
 echo "============="
 docker-compose -f docker-compose_qualifier_mocks.yaml -p $project_name down
 
+echo "Rebuild images without cache"
+docker-compose -f docker-compose_qualifier_mocks.yaml -p $project_name build --no-cache
```
BenjaminPelletier, line 14: This will make running tests locally take a lot longer in general -- could we add this as an option that can be triggered when needed, or perhaps just put it in the documentation?

barroco: Sure, I added a check for the `$CI` environment variable which is [automatically populated by the github actions](https://docs.github.com/en/actions/learn-github-actions/environment-variables)

BenjaminPelletier (approved):

PR #717, opened 2022-02-20
[uss_qualifier/scd] report

barroco:
This PR adds the following improvement to the uss qualifier scd component:
1. Capture of interactions and issues in the test runner 
2. Generation a single report per invocation of the SCD part of uss_qualifier (`report_scd.json`)
3. Adds qualifier_scd version to the report using `uss_qualifier_scd` as release component name.

### DIFF
--- a/monitoring/uss_qualifier/run_locally_scd.sh
+++ b/monitoring/uss_qualifier/run_locally_scd.sh
@@ -64,6 +64,7 @@ docker run ${docker_args} --name uss_qualifier \
   -e SCD_QUALIFIER_OPTIONS="${SCD_QUALIFIER_OPTIONS}" \
   -e PYTHONBUFFERED=1 \
   -v "${REPORT_FILE}:/app/monitoring/uss_qualifier/report.json" \
+  -v "${REPORT_SCD_FILE}:/app/monitoring/uss_qualifier/report_scd.json" \
   -v "$(pwd):/app" \
   interuss/uss_qualifier \
   python main.py $SCD_QUALIFIER_OPTIONS
--- a/monitoring/uss_qualifier/run_locally_scd.sh
+++ b/monitoring/uss_qualifier/run_locally_scd.sh
@@ -49,6 +49,7 @@ docker build \
     -f monitoring/uss_qualifier/Dockerfile \
     -t interuss/uss_qualifier \
     --build-arg version="$(scripts/git/commit.sh)" \
+    --build-arg qualifier_scd_version="$(scripts/git/version.sh uss_qualifier_scd --long)" \
     monitoring
 
 if [ "$CI" == "true" ]; then
--- a/monitoring/uss_qualifier/scd/executor/executor.py
+++ b/monitoring/uss_qualifier/scd/executor/executor.py
@@ -8,7 +8,7 @@ from typing import Dict, List
 from monitoring.monitorlib.locality import Locality
 from monitoring.monitorlib.typing import ImplicitDict
 from monitoring.uss_qualifier.scd.configuration import SCDQualifierTestConfiguration
-from monitoring.uss_qualifier.scd.data_interfaces import AutomatedTest, TestStep
+from monitoring.uss_qualifier.scd.data_interfaces import AutomatedTest, TestStep, AutomatedTestContext
 from monitoring.uss_qualifier.scd.executor.runner import TestRunner, TestRunnerError
 from monitoring.uss_qualifier.scd.executor.target import TestTarget
 from monitoring.uss_qualifier.scd.reports import Report
@@ -55,6 +55,7 @@ def validate_configuration(test_configuration: SCDQualifierTestConfiguration):
 
 def combine_targets(targets: List[TestTarget], steps: List[TestStep]) -> typing.Iterator[Dict[str, TestTarget]]:
     """Gets combination of targets assigned to the uss roles specified in steps"""
+
     injection_steps = filter(lambda step: 'inject_flight' in step, steps)
 
     # Get unique uss roles in injection steps in deterministic order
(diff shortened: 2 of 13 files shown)


monitoring/uss_qualifier/scd/reports.py
```diff
@@ -57,28 +58,49 @@ class Interaction(ImplicitDict):
...
-    """ID of test for which this interaction was performed"""
+    context: AutomatedTestContext
+    """Context in which this interaction was performed"""
 
     test_step: int
-    """Step of test for which this interaction was performed"""
+    """Step of test for which this interaction was performed.
+    0-based indexed.
+    Special values:
+     * -1: capture occurred during test initialization
```
BenjaminPelletier, line 68: Instead of encoding two different things in one field, could we break them out separately?  The `test_step` field could become `step_index` (or something that could potentially be used in initialization and cleanup also), and then we could add a `test_phase` field indicating `Initialization`, `Test`, and `Cleanup`.

monitoring/uss_qualifier/scd/reports.py
```diff
@@ -57,28 +58,49 @@ class Interaction(ImplicitDict):
...
     """ID of this interaction (used to refer to this interaction from an issue)"""
 
-    test_id: str
-    """ID of test for which this interaction was performed"""
+    context: AutomatedTestContext
+    """Context in which this interaction was performed"""
 
     test_step: int
-    """Step of test for which this interaction was performed"""
+    """Step of test for which this interaction was performed.
```
BenjaminPelletier, line 65: In terms of readability of the reports, what do you think about either having an additional [`test_step_name`](https://github.com/interuss/dss/blob/dd807cc51dd85e00809ccdae656304aab9bc1faa/monitoring/uss_qualifier/scd/data_interfaces.py#L66) field, or else replacing this index with that name?  I'm not particularly fond of duplicating information in general, but the `context` does have both `test_id` and `test_name` which are technically redundant (`test_name` can be determined from `test_id`), but I think that still works well so it seems reasonable to also have the technically-redundant index and name both in the report.

monitoring/uss_qualifier/scd/executor/executor.py
```diff
@@ -73,17 +76,53 @@ def format_combination(combination: Dict[str, TestTarget]) -> List[str]:
 
 
 def run_scd_tests(locale: Locality, test_configuration: SCDQualifierTestConfiguration,
-                  auth_spec: str):
+                  auth_spec: str) -> bool:
     automated_tests = load_scd_test_definitions(locale)
     configured_targets = list(map(lambda t: TestTarget(t.name, t, auth_spec), test_configuration.injection_targets))
+    report = Report(
+            qualifier_version=os.environ.get("SCD_VERSION", "unknown"),
```
BenjaminPelletier, line 83: It might be nice to have run_locally_scd populate this reasonably to get a more complete/accurate test of the system

monitoring/uss_qualifier/scd/executor/executor.py
```diff
@@ -73,17 +76,53 @@ def format_combination(combination: Dict[str, TestTarget]) -> List[str]:
...
+            except TestRunnerError as e:
+                print("[SCD] TestRunnerError: {} Issue: {} Related interactions: {}".format(e, e.issue.details, e.issue.interactions))
+            finally:
+                runner.teardown()
+
+            executed_test_run_count = executed_test_run_count + 1
+            should_exit = len(report.findings.critical_issues()) > 0
+            if should_exit:
+                print ("[SCD] Critical issues found during test. Interrupting test sequence. {}".format(report.findings))
+                break
```
BenjaminPelletier, line 115: We may want to consider what is still doable under what failures.  A critical failure of one test will certainly stop that test, but other tests may still be possible (we do expect to be able to re-run this entire executable after all).

BenjaminPelletier (approved):
This looks great -- thanks!  I left some comments, but your discretion whether we address them in this PR or a later one.  I think the code is good to merge as-is, and we can make additional improvements later.

monitoring/uss_qualifier/run_locally_scd.sh
```diff
@@ -64,6 +65,7 @@ docker run ${docker_args} --name uss_qualifier \
   -e SCD_QUALIFIER_OPTIONS="${SCD_QUALIFIER_OPTIONS}" \
   -e PYTHONBUFFERED=1 \
   -v "${REPORT_FILE}:/app/monitoring/uss_qualifier/report.json" \
```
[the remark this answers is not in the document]
barroco: Sure, I was planning to address this in a future PR.

monitoring/uss_qualifier/scd/reports.py
barroco: Sounds great. done.

monitoring/uss_qualifier/scd/executor/executor.py
barroco: I am sorry, I don't understand the comment, could you please elaborate? Currently, this variable is set as an argument at [build time](https://github.com/swiss-foca/dss/blob/scd_automated_testing/reports/monitoring/uss_qualifier/run_locally_scd.sh#L52).

monitoring/uss_qualifier/scd/executor/executor.py
barroco: Can we agree on the following severities description?
- **Critical**: Error is unrecoverable and left the system dirty preventing subsequent correct test run execution. For instance, inability to clean a created flight which will conflict with another run with a different combination of targets.
- **High**: Error interrupts a test run and doesn't impact subsequent test runs. For instance, the test failed and all created can safely be clean up during teardown by the test driver.
- **Low**: Error doesn't impact the test run and further test steps can be executed without impact.

monitoring/uss_qualifier/scd/reports.py
barroco: test_step_name added.

monitoring/uss_qualifier/scd/executor/executor.py
BenjaminPelletier: In the report, I'm currently seeing `"qualifier_version": "interuss/uss_qualifier_scd/v0.0.0",` in the report.  When I wrote this comment, I was thinking I would see 0.1.0 somewhere, but upon further thought, I shouldn't see that because only `dss` is at 0.1.0.  But, I would still expect to see `-<commit hash>`.  That doesn't necessarily need to be solved in this PR though.

monitoring/uss_qualifier/scd/executor/executor.py
BenjaminPelletier: Yes, those look good to me.  I'd expect we would augment the [existing descriptions](https://github.com/interuss/dss/blob/067e550199182e59b447843395f1faee488d07cc/monitoring/uss_qualifier/common_data_definitions.py#L7) with these descriptions.  I'd also suggest **Medium**: The system does not meet requirements, but further test steps can likely be executed without impact.

monitoring/uss_qualifier/scd/executor/executor.py
barroco: 👍 Solved by #719

monitoring/uss_qualifier/scd/executor/executor.py
barroco: #720

PR #718, opened 2022-02-22
[uss_qualifier/scd] Local instructions

barroco:
This PR adds instructions and helper scripts to run configurable automated tests locally  from a service provider point of view. Additionally it changes the module version to uss_qualifier instead of uss_qualifier_scd.

### DIFF
--- a/monitoring/uss_qualifier/run_locally_scd.sh
+++ b/monitoring/uss_qualifier/run_locally_scd.sh
@@ -12,11 +12,11 @@ else
 fi
 cd "${BASEDIR}/../.." || exit 1
 
-echo '#########################################################################'
-echo '## NOTE: A prerequisite for running this command locally is to have    ##'
-echo '## a running instance of the uss_qualifier SCD system mock             ##'
-echo '## (../mock_uss/run_locally_scdsc.sh) including related dependencies.  ##'
-echo '#########################################################################'
+echo '####################################################################################'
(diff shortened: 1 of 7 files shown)


monitoring/uss_qualifier/run_locally_scd.sh
```diff
@@ -12,11 +12,11 @@ else
 fi
 cd "${BASEDIR}/../.." || exit 1
 
-echo '#########################################################################'
-echo '## NOTE: A prerequisite for running this command locally is to have    ##'
-echo '## a running instance of the uss_qualifier SCD system mock             ##'
-echo '## (../mock_uss/run_locally_scdsc.sh) including related dependencies.  ##'
-echo '#########################################################################'
+echo '####################################################################################'
+echo '## NOTE: Prerequisites to run this command are:                                   ##'
```
BenjaminPelletier, line 16: nit: strictly speaking, I think the previous documentation was probably better -- at some point in the future, we might add a mode to the mock USS that makes it function without a DSS, and in that case there would be no need for a DSS.  When the user refers to run_locally_scdsc.sh, that will say what the dependencies for that tool are.

TL;DR: I don't know that we need to list indirect dependencies here.  But I actually don't have a strong opinion; I would be ok with these changes since they do make things more explicit for the user.

BenjaminPelletier (approved):

PR #719, opened 2022-02-22
[version] Append commit hash to default version 0.0.0

barroco:
We identified that if a module was never released, the version generated defaults to v0.0.0. To make itmeaningful, this PR adds the commit hash to the fallback version name. That will generate useful version name for unreleased projects such as the `uss_qualifier`.

### DIFF
--- a/scripts/git/version.sh
+++ b/scripts/git/version.sh
@@ -43,10 +43,8 @@ COMMIT=$(git rev-parse --short HEAD)
 # If no version was found, use default v0.0.0.
 if [[ -z "$LAST_VERSION" ]]; then
   LAST_VERSION="v0.0.0-$COMMIT"
-fi
-
 # Check if there are some commits on top of the tag by checking if an abbrev part is present.
-if [[ "$LAST_VERSION" == *"-"* ]]; then
+elif [[ "$LAST_VERSION" == *"-"* ]]; then
   # Remove abbrev part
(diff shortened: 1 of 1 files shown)


scripts/git/version.sh
```diff
@@ -37,14 +37,14 @@ LAST_VERSION_TAG=$(git describe --abbrev=1 --tags --match="${UPSTREAM_ORG}/${COM
 LAST_VERSION=${LAST_VERSION_TAG##*/}
 #echo "LAST_VERSION: $LAST_VERSION"
 
+# Current commit
+COMMIT=$(git rev-parse --short HEAD)
+
 # If no version was found, use default v0.0.0.
 if [[ -z "$LAST_VERSION" ]]; then
-  LAST_VERSION="v0.0.0"
+  LAST_VERSION="v0.0.0-$COMMIT"
```
BenjaminPelletier, line 45: Should the `if [[ "$LAST_VERSION" == *"-"* ]]; then` be an else-if clause to this block?  It seems like maybe we're setting LAST_VERSION to a value which will trigger that second block so we're adding the commit, then removing the commit, then adding the commit (if I read things correctly)

BenjaminPelletier (approved):

scripts/git/version.sh
barroco: Good catch, thank you.

PR #726, opened 2022-02-27
ENAC_UPC_Team

shuanet:
Simulated USSP directory generated fpr API testing with a token request python script used as a template


BenjaminPelletier (commented):
Could you elaborate a bit more on the purpose of this PR?  We generally have our monitoring & diagnostic tools in the `monitoring` folder; is there a reason `Sim_USSP` should be at the root folder level rather than `monitoring`?

Also, what is the purpose of `Sim_USSP`?  Could we add a README.md to make this clear to others?  It looks like both files currently just get an access token from the dummy OAuth server.  We have a fairly robust auth package in `monitorlib` that allows the retrieval of access tokens from many different servers according to an auth spec, as demonstrated by `monitoring/get_access_token.py` -- it seems like we would probably want to reuse that functionality in a new tool.

It appears the `flight-blender` submodule was not specified correctly (no URL), and that's what's causing the continuous integration tests to fail.  Could you clarify why we are importing flight-blender as a submodule?  And, given its import, why it is imported at the root folder of the repository?

In the past, we generally haven't committed workspace configurations to the repository as they can vary by user to user, but I'm not opposed to a standard VSCode configuration if it would be widely applicable to most VSCode users leveraging this repository.  However, we would definitely want any workspace configuration committed to the repository to be compatible with the codebase requirements.  It looks like, for instance, this configuration targets Python 3.10 while most `monitoring` Dockerfiles target 3.8.  It also looks like this workspace configuration specifies many tool dependencies that aren't used by the repository currently.  Could we align the workspace configuration to the repository needs?

shuanet:
No you should not merge it. We are using that directory for learning about
the code working ny doing some functionalities by ourselves. This PR was
not intentional and cause of some mistake of ours.

Sorry for the inconveniance.

BenjaminPelletier:
Closing PR as mistakenly created (see comments)

PR #729, opened 2022-03-03
[docs/deployment] Update deployment documentation and make-certs

BenjaminPelletier:
This PR updates deployment documentation based on following it for a recent deployment.  One theme of the changes is that "cluster" was overloaded referring sometimes to a Kubernetes cluster and sometimes to a CockroachDB cluster, and now the latter has been changed to "DSS pool" and "DSS instance" (according to context).  Also, the ability to use InterUSS Docker images straight from Docker Hub is added.  Various other clarifications and improvements are made as well.

Also, make-certs previously required an explicit CA-override flag even when there were no CA certs to override -- this behavior is fixed so that it behaves as expected creating certs for the first time.


barroco (approved):
👍 Changes look good
Please note that I didn't test the deployment on GCP.

PR #730, opened 2022-03-07
Links in Qualifier webapp readme updated

pratibhagupta2109:


monitoring/uss_qualifier/webapp/README.md
```diff
@@ -1,7 +1,7 @@
 # RID Qualifier WebApp
```
pratibhagupta2109, line 1: Is this app still known as RID Qualifier or it should be changed to `USS Qualifier`?

BenjaminPelletier: Should be USS Qualifier -- thanks!

monitoring/uss_qualifier/webapp/README.md
```diff
@@ -55,7 +55,7 @@ Resulting files will be generated under `monitoring/uss_qualifier/test_definitio
 
 ## User Config information
 
-rid_qualifier host needs a configuration object and an auth spec (single string e.g. `NoAuth()`) in order to run. Configuration object is a simple nested json string which should have at least a list of `injection_targets` and `observers`. A sample configuration object can be found [here](/monitoring/uss_qualifier/run_locally.sh) at line 20.
+uss_qualifier host needs a configuration object and an auth spec (single string e.g. `NoAuth()`) in order to run. Configuration object is a simple nested json string which should have at least a list of `injection_targets` and `observers`. A sample configuration object can be found [here](/monitoring/uss_qualifier/run_locally.sh#L26).
```
pratibhagupta2109, line 58: looks like structure of configuration object has changed in the past. USS qualifier code expects `injection_targets` and `observers` keys available in config object, but according to https://github.com/interuss/dss/blob/master/monitoring/uss_qualifier/run_locally.sh#L27, these keys are moved under `rid` key.

BenjaminPelletier: Yes, that's true -- we should update webapp to handle this.

pratibhagupta2109: If I update webapp to handle it (as done in this PR commented code in forms.py), `RIDQualifierTestConfiguration` fails to validate the configuration with following error:
`Required field "observers" not specified in RIDQualifierTestConfiguration`, as this configuration expects old format and it is used at other places too.

BenjaminPelletier (approved):

PR #731, opened 2022-03-08
[uss_qualifier/scd] Check for injected operational intents in UTM system

BenjaminPelletier:
This PR has uss_qualifier double check with the DSS and USS (acting as a USS itself) that the operational intent a USS reported injecting actually appears properly in the UTM system whenever a USS properly indicates Planned as their response to a flight injection attempt.

As a consequence of debugging some behaviors observed during development of this change, this PR also adds some improved error messages and mock USS behaviors and fixes a problem where issues discovered through TestRunnerErrors were not added to the report.


barroco (approved):
👍 Many thanks for this improvement!

PR #733, opened 2022-03-09
[automated-testing/scd] Add nominal planning with higher priority test

barroco:
This PR adds to the automated testing of SCD the test case which checks the ability of a USP to plan an operational intent with a higher priority despite the conflict in space and time with an existing operational intent.


BenjaminPelletier (approved):
This looks good, but I would encourage us to expand the test a bit to verify that first-mover rejects an attempt to activate the flight after it has been pre-empted -- this can be in another PR :)

PR #738, opened 2022-03-11
[prober] Fix, simplify, and improve workspace cleaning actions

BenjaminPelletier:
Prior to this PR, cleaning of the workspace (deleting any previous instances of resources to be manipulated in the test) was unnecessarily verbose (same logic copy/pasted in many different files) and sometimes incorrect (API-dependence not respected in test_constraint_simple).  This PR creates a few generic actions in actions.py that are sensitive to API, and then calls them from all workspace cleaning pre-tests.  This fixes a 405 (method not supported) error when running test_constraint_simple on live instances under dirty conditions.


barroco (approved):
👍

PR #739, opened 2022-03-14
[uss_qualifier/webapp] OpenAPI definition for uss_qualifier webapp interface: Initial check-in

pratibhagupta2109:


interfaces/uss_qualifier/uss_qualifier.yaml
```diff
@@ -0,0 +1,113 @@
...
+      required:
+      - task_status
+      properties:
+        task_status:
+          type: string
+          description: >-
+            Returns current status of the background task.
+          example: 'Started|Finished'
+        task_result:
+          type: object
```
BenjaminPelletier, line 63: Just a note: eventually (not in this PR), we'll want to formalize this schema

interfaces/uss_qualifier/uss_qualifier.yaml
```diff
@@ -0,0 +1,113 @@
...
+    RunTests:
+      type: object
+      required:
+      - flight_records
+      - auth_spec
+      - user_config
+      properties:
+        flight_records:
+          description: >-
+            Comma separated list of flight record file names. Files should be uploaded by upload method before running the tests.
```
BenjaminPelletier, line 19: We'll also need to define the endpoint(s) where these files can be uploaded -- this can be in a separate PR.

BenjaminPelletier (approved):

PR #748, opened 2022-03-28
[mock_uss] Fixes for Mock USS / RID run_locally

hrishiballal:
This PR fixes the issue when `run_locally_ridddp.sh` and `run_locally_ridsp.sh` is run. Without these fixes I am getting a error attached below. 
image

### DIFF
--- a/monitoring/mock_uss/run_locally_riddp.sh
+++ b/monitoring/mock_uss/run_locally_riddp.sh
@@ -25,7 +25,7 @@ docker run ${docker_args} --name mock_uss_riddp \
   -e MOCK_USS_TOKEN_AUDIENCE="${AUD}" \
   -e MOCK_USS_SERVICES="riddp" \
   -p ${PORT}:5000 \
-  -v "$(pwd)/build/test-certs:/var/test-certs:ro" \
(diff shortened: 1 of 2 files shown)


monitoring/mock_uss/run_locally_riddp.sh
```diff
@@ -25,7 +25,7 @@ docker run ${docker_args} --name mock_uss_riddp \
   -e MOCK_USS_TOKEN_AUDIENCE="${AUD}" \
   -e MOCK_USS_SERVICES="riddp" \
   -p ${PORT}:5000 \
-  -v "$(pwd)/build/test-certs:/var/test-certs:ro" \
```
BenjaminPelletier, line 28: Let's keep the read-only indicator

BenjaminPelletier (commented):
Looks good, but one comment (applies in two places)

monitoring/mock_uss/run_locally_riddp.sh
hrishiballal: Thanks! I have now added the indicator.

BenjaminPelletier (approved):

PR #749, opened 2022-03-28
[uss_qualifier/webapp] Implementation of upload files endpoints through Open API 

pratibhagupta2109:


interfaces/uss_qualifier/uss_qualifier.yaml
```diff
@@ -124,6 +143,56 @@ paths:
       description: A task ID of an initiated task.
       schema:
         $ref: '#/components/schemas/TaskID'
+  /api/flight-records-upload/json:
```
BenjaminPelletier, line 146: Let's stay consistent with dashes versus underscores and use underscores since they can be used in variable names.  And, the means by which the flight records are acquired shouldn't matter, so let's change this to just `/api/flight_records/json` (and the corresponding variant below to `/api/flight_records/kml`)

BenjaminPelletier (approved):

PR #758, opened 2022-04-07
[deployment_manager] Add working ingress to hello_world deployment manager example

BenjaminPelletier:


monitoring/deployment_manager/systems/test/hello_world.py
```diff
@@ -46,7 +48,7 @@ def _define_webserver_deployment() -> k8s.V1Deployment:
 def _define_webserver_service() -> k8s.V1Service:
     spec = k8s.V1ServiceSpec(
         selector={'app': APP_NAME},
-        type='NodePort',
+        type='LoadBalancer',
         ports=[
```
barroco, line 52: It seems that the ports values are swapped.

target_port should be the port exposed by the container. In this case 8080.
port should be the port of the service expected by the load balancer.

barroco (approved):
Please see comments before merging.

monitoring/deployment_manager/systems/test/hello_world.py
BenjaminPelletier: This is confusing -- from [here](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1ServicePort.md) (which reflects [here](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec)):

`target_port`: Number or name of the port to access on the pods targeted by the service.

`port`: The port that will be exposed by this service.

The port previously exposed by the container was 80 (`container_port=80` in deployment definition), so it was correct that `target_port` should match that value, so the correct value for `target_port` was 80 (not 8080).

I was very confused about how a lot of this worked so I changed the example to use a different port number for every value.  The binary in the container now listens on 8001, and this is the pod's container port.  The Service listens on 8002 (`V1ServicePort.port=8002`) and forwards requests to 8001 (`V1ServicePort.target_port=8001`).  The ingress listens on ports 80 and 443 and routes HTTP(S) requests to the named service (which declares that it listens on 8002).  While `minikube tunnel` is running, the ingress can be accessed at http://localhost:80 and https://localhost:443 in a browser and the service can be accessed at http://localhost:8002 in a browser.  Port 8001 is not visible outside the cluster.

I've updated the documentation to capture this information.

PR #759, opened 2022-04-07
[uss_qualifier/webapp] Endpoint `/api/tests/` specifications

pratibhagupta2109:


interfaces/uss_qualifier/uss_qualifier.yaml
```diff
@@ -1,198 +1,263 @@
...
+                "200":
+                    content:
+                        application/json:
+                            schema:
+                                $ref: "#/components/schemas/BackgroundTaskRunResponse"
+                    description: >-
+                        Returns task details of the task launched in the background to process KML files.
+                "400":
+                    description: Failed to upload files.
+    /api/tests:
```
BenjaminPelletier, line 229: We already have a resource collection named `test_runs` that we can POST to in order to create (trigger) a new test run to begin -- if we want to enumerate the resources in that collection, I would expect the endpoint to be `GET /api/test_runs`, which would return a list of the `test_run` resources.  The details for a particular resource would then be available at `GET /api/test_runs/{test_id}` (rather than `GET /api/tests/{test_id}`)

interfaces/uss_qualifier/uss_qualifier.yaml
```diff
@@ -1,198 +1,263 @@
...
+                    example: "A task has been started in the background."
+                    description: >-
+                        Status of currently running task.
+                user_id:
+                    type: string
+                    description: >-
+                        User ID of user running the tests.
+                specification:
+                    $ref: "#/components/schemas/TestRunSpecification"
+        TestsRunResponse:
```
BenjaminPelletier, line 70: I would expect that the representation of a `test_run` resource should be essentially the same whether it is being reported as a consequence of creation, or being queried specifically.  I think `CreateTestRunResponse` is basically what we want here (though we should probably rename it `TestRunResponse` to logically use it in both places), after adding an optional `outputs` field (Or perhaps `report`?  Are there are outputs apart from the report?) which would be populated only after the test run completed.

pratibhagupta2109:
The changes in this PR are covered by #769

PR #764, opened 2022-04-19
[uss_qualifier/scd] Add planning_time and dynamic dates

hsrodriguez:
Added planning time to the json file and the data interfaces. 
Calculate the reference time, start time and end time based on the current time and planning time


monitoring/uss_qualifier/scd/executor/test_steps/inject_flight.py
```diff
@@ -18,6 +19,15 @@
 def execute(runner, step: TestStep, step_ref: TestStepReference, target: TestTarget) -> None:
     print("[SCD]     Step: Inject flight {} to {}".format(step.inject_flight.name, target.name))
     try:
+        reference_time = datetime.utcnow() + timedelta(seconds=pytimeparse.parse(step.inject_flight.planning_time))
```
BenjaminPelletier, line 22: This is different adjustment logic than I would have expected.  My expectation is that we would offset all volume timestamps (time_start and time_end) by the same timedelta, and that timedelta would be now - reference_time + planning_time.  Adding that timedelta makes it so that a volume timestamp equal to reference_time in the automated test definition would become (now + planning time) after adjustment.

BenjaminPelletier, line 22: Small note: because planning_time is a StringBasedTimeDelta, the timedelta value can be read simply as [`step.inject_flight.planning_time.timedelta`](https://github.com/interuss/dss/blob/9f6a372d1ada89873361872c9f37fcc9e6bab175/monitoring/monitorlib/typing.py#L222) rather than `timedelta(seconds=pytimeparse.parse(step.inject_flight.planning_time))`

monitoring/uss_qualifier/scd/test_definitions/CHE/u-space/flight-authorisation-validation-1.json
```diff
@@ -55,11 +56,11 @@
                                     }
                                 },
                                 "time_start": {
-                                    "value": "2023-02-11T09:03:05.359502+00:00",
+                                    "value": "2022-04-19T09:03:05.359502+00:00",
```
BenjaminPelletier, line 59: Since the reference_time is 2023-02-12T10:34:14.681217+00:00, I would expect time_start to be adjusted to be 10 months *before* the time uss_qualifier is run (since 2022-04-19T09:03:05.359502+00:00 is 10 months before 2023-02-12T10:34:14.681217+00:00); that doesn't seem like what we want

monitoring/uss_qualifier/scd/executor/test_steps/inject_flight.py
```diff
@@ -18,6 +19,15 @@
...
+        step.inject_flight.reference_time = reference_time.isoformat() + "Z"
+        for volume in step.inject_flight.test_injection.operational_intent.volumes:            
+            time_start = reference_time + timedelta(minutes=3)
```
BenjaminPelletier, line 25: I would not expect any magic numbers (3 minutes and 5 minutes in this case) in the adjustment algorithm

BenjaminPelletier (commented):
To make sure we're on the same page, here's an illustration of what I would expect:

At some point, someone runs a simulator or otherwise generates test data -- let's say they do this at 2022-03-01T10:00:00.  I would expect the simulator to set the reference_time to 2022-03-01T10:00:00 (the time of generation), and then make volumes with realistic time bounds assuming the flight nominally begins at 2022-03-01T10:00:00.  This might mean, for instance, that the time bounds would be:
* vol1: 2022-03-01T10:00:00 to 2022-03-01T10:06:00
* vol2: 2022-03-01T10:04:00 to 2022-03-01T10:09:00

Later, someone will want to run uss_qualifier, which will attempt to inject this flight into a USS.  If uss_qualifier is run at 2022-05-06T8:30:00 and we have allowed 0:05:00 of planning_time, then I would expect the volumes to be adjusted as follows:
* vol1: 2022-05-06T8:35:00 to 2022-05-06T8:41:00
* vol2: 2022-05-06T8:39:00 to 2022-05-06T8:44:00
vol1's first timestamp (2022-05-06T8:35:00) is equal to now (2022-05-06T8:30:00) + 0:05:00 because it was originally equal to reference_time.  vol1's second timestamp (2022-05-06T8:41:00) is equal to now + 0:05:00 + 0:06:00 because it was originally equal to reference_time + 0:06:00.

monitoring/uss_qualifier/scd/executor/test_steps/inject_flight.py
```diff
@@ -18,6 +20,23 @@
 def execute(runner, step: TestStep, step_ref: TestStepReference, target: TestTarget) -> None:
     print("[SCD]     Step: Inject flight {} to {}".format(step.inject_flight.name, target.name))
     try:
+        t_test = datetime.now(timezone.utc)
+        
+        for volume in step.inject_flight.test_injection.operational_intent.volumes:
+            volume.time_start.value = volume.time_start.value
+            volume.time_end.value = volume.time_end.value
+
+            t_delta = (t_test - step.inject_flight.reference_time.datetime) + (step.inject_flight.reference_time.datetime - volume.time_start.value.datetime)
```
BenjaminPelletier, line 29: This appears to time-shift each volume by a different amount, which is different from what I would expect.  Summarizing my [example](https://github.com/interuss/dss/pull/764#pullrequestreview-946326993):

* reference_time = 2022-03-01T10:00:00
* planning_time = 0:05:00
* vol1 (test definition) = 2022-03-01T10:00:00 to 2022-03-01T10:06:00
* vol2 (test definition) = 2022-03-01T10:04:00 to 2022-03-01T10:09:00
* t_test = 2022-05-06T8:30:00
* vol1 (ready to inject) = 2022-05-06T8:35:00 to 2022-05-06T8:41:00
* vol2 (ready to inject) = 2022-05-06T8:39:00 to 2022-05-06T8:44:00

I notice a few things about the current algorithm:
1. The expression on this line simplifies to `t_test - volume.time_start.value.datetime` (we both add and subtract `step.inject_flight.reference_time.datetime`)
2. `t_test` will pretty much always be after reference_time so `step.inject_flight.planning_time.timedelta` will always get added to `t_delta`
3. This means `t_start_adjusted` will always be `volume.time_start.value.datetime + t_test - volume.time_start.value.datetime + step.inject_flight.planning_time.timedelta`, which equals `t_test + step.inject_flight.planning_time.timedelta`

This seems to imply that we will instead observe the following adjusted times:
* vol1 (ready to inject) = 2022-05-06T8:35:00 to 2022-05-06T8:41:00
* vol2 (ready to inject) = 2022-05-06T8:35:00 to 2022-05-06T8:40:00

I don't think we want that -- we've gone from vol2 overlapping the end of vol1 and extending beyond it to vol2 ending before vol1 ends.

I think the adjustment algorithm should be pretty straightforward: shift every timestamp on every volume by the same amount.  That amount should be whatever is required to cause a reference_time timestamp to be shifted to t_test + planning_time.

BenjaminPelletier (approved):

PR #768, opened 2022-05-04
webapp: Using Restful api for webapp forms

pratibhagupta2109:


pratibhagupta2109:
#770 is a refined version of this PR.

PR #773, opened 2022-05-20
[mock_uss] Add RID misbehaviors to mock_uss

BenjaminPelletier:
Currently, there are two forms of mock systems for testing the RID portion of uss_qualifier:
* mock_uss, which behaves similarly to a real USS including use of a DSS
* The uss_qualifier RID mock, which does not require a DSS

An important characteristic the RID portion of uss_qualifier currently lacks is the ability to determine which USS is likely the cause of an interoperability problem.  The reason that this is the case currently is that the RID portion of uss_qualifier relies solely on the injection and observation interfaces, and so only observes that an injected flight does not reach the observation interface -- it cannot determine whether the problem between the injection and observation was due to the Service Provider or the Display Provider.  If uss_qualifier first probed a Service Provider directly (acting as a Display Provider), it could determine whether the incorrect transmission of a particular injected flight to the observation interface was due to the Service Provider (in which case the problem would be observed when probing the Service Provider directly) or the Display Provider (in which case probing the Service Provider directly would yield no problems, but probing the observation interface would indicate a problem).

In order to add this feature to the RID portion of the uss_qualifier, the mock USS(s) must use the DSS.  Since the uss_qualifier RID mock does not use the DSS, we should remove that mock from the codebase as inferior to mock_uss.  The RID qualifier mock is slightly easier to use in that it does not require a DSS instance to run, but this is a negligible advantage when a local DSS instance can be deployed with a single command.  Meanwhile, maintaining two separate mock systems that accomplish the same goal requires a lot of development attention; it would be better to spend that attention refining and testing a single mock instead.

Before removing the uss_qualifier RID mock, we should transfer any capabilities it currently has but mock_uss does not currently have.  The main capability is misbehaviors: settings that can be changed to cause the mock to misbehave in various ways (which should be detected by automated tests).  This PR transfers that misbehavior capability, as well as some clustering code and a few other small things, to mock_uss.

Following this PR, I intend to remove the uss_qualifier RID mock in favor of mock_uss.

### DIFF
--- a/monitoring/mock_uss/ridsp/behavior.py
+++ b/monitoring/mock_uss/ridsp/behavior.py
@@ -1,0 +1,53 @@
+from typing import Optional
+
+from monitoring.monitorlib.rid_automated_testing.injection_api import TestFlight
+from monitoring.monitorlib.rid import RIDFlight
+from monitoring.monitorlib.typing import ImplicitDict
+
+
+FEET_PER_METER = 1 / 0.3048
+
+
+class ServiceProviderBehavior(ImplicitDict):
+    switch_latitude_and_longitude_when_reporting: Optional[bool] = False
+    use_agl_instead_of_wgs84_for_altitude: Optional[bool] = False
+    use_feet_instead_of_meters_for_altitude: Optional[bool] = False
+
+
+def adjust_reported_flight(flight: TestFlight, reported_flight: RIDFlight, behavior: ServiceProviderBehavior) -> RIDFlight:
+    """Adjust how a flight is reported based on the SP behavior"""
+
+    adjusted = ImplicitDict.parse(reported_flight, RIDFlight)
+    if behavior.switch_latitude_and_longitude_when_reporting:
+        if adjusted.has_field_with_value('current_state'):
+            lng = adjusted.current_state.position.lat
+            adjusted.current_state.position.lat = adjusted.current_state.position.lng
+            adjusted.current_state.position.lng = lng
+        if adjusted.has_field_with_value('recent_positions'):
+            for p in adjusted.recent_positions:
+                lng = p.position.lat
+                p.position.lat = p.position.lng
+                p.position.lng = lng
+
+    if behavior.use_agl_instead_of_wgs84_for_altitude:
+        ground_altitude = min(t.position.alt for t in flight.telemetry)
+        if adjusted.has_field_with_value('current_state'):
+            if adjusted.current_state.has_field_with_value('height'):
+                adjusted.current_state.position.alt = adjusted.current_state.height.distance
+            else:
+                adjusted.current_state.position.alt -= ground_altitude
+
+        if adjusted.has_field_with_value('recent_positions'):
+            for p in adjusted.recent_positions:
+                p.position.alt -= ground_altitude
+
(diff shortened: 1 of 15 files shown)


monitoring/uss_qualifier/README.md
```diff
@@ -6,4 +6,6 @@ The USS Qualifier automates testing standards compliance and interoperability of
 service providers. It currently includes the following components:
 
 - [Remote ID](./rid/README.md): ASTM F3411-19 - Remote ID and Tracking
+    - See [run_locally_rid.sh](run_locally_rid.sh)
 - [ASTM strategic coordination and UAS Flight Authorisation](./scd/README.md): ASTM WK63418 and U-Space Regulation
+    - See [run_locally_rid.sh](run_locally_scd.sh)
```
barroco, line 11: Label should be run_locally_**scd**.sh

monitoring/mock_uss/ridsp/behavior.py
```diff
@@ -0,0 +1,53 @@
...
+        ground_altitude = min(t.position.alt for t in flight.telemetry)
+        if adjusted.has_field_with_value('current_state'):
+            if adjusted.current_state.has_field_with_value('height'):
+                adjusted.current_state.position.alt = adjusted.current_state.height.distance
+            else:
+                adjusted.current_state.position.alt -= ground_altitude
+
+        if adjusted.has_field_with_value('recent_positions'):
+            for p in adjusted.recent_positions:
+                p.position.alt -= ground_altitude
```
barroco, line 42: Shouldn't we check for height here as well? (similar to line 35)

monitoring/mock_uss/ridsp/behavior.py
```diff
@@ -0,0 +1,53 @@
...
+
+    if behavior.use_feet_instead_of_meters_for_altitude:
+        if adjusted.has_field_with_value('current_state'):
+            adjusted.current_state.position.alt *= FEET_PER_METER
+            if adjusted.current_state.has_field_with_value('height'):
+                adjusted.current_state.height.distance *= FEET_PER_METER
+        if adjusted.has_field_with_value('recent_positions'):
+            for p in adjusted.recent_positions:
+                p.position.alt *= FEET_PER_METER
```
barroco, line 51: May be worth updating the height of the position as well here.

barroco (approved):
Looks very good, thank you! Few details in comments

monitoring/uss_qualifier/README.md
BenjaminPelletier: Thanks, done

monitoring/mock_uss/ridsp/behavior.py
BenjaminPelletier: Recent positions don't have height and we don't want to set everything to current_state's height.  We could probably adjust these differently/more intelligently, but especially since we don't actually even populate height data in injection data right now, I think that can be a future feature.

monitoring/mock_uss/ridsp/behavior.py
BenjaminPelletier: Recent positions don't contain height

PR #781, opened 2022-05-29
[monitoring] Rename DSSTestSession -> UTMClientSession

BenjaminPelletier:
`DSSTestSession` was named a very long time ago when the only thing it did was provide access to the DSS.  Now, it is used for communicating with other parts of the UTM ecosystem in addition to the DSS.  For clarity, this PR renames the session a `UTMClientSession` to better reflect its actual use and capabilities.

Also, add uss_qualifier `make format` and `make lint` targets to root `make format` and `make lint` targets, respectively.


monitoring/monitorlib/infrastructure.py
```diff
@@ -201,23 +205,23 @@ def wrapper_default_scope(*args, **kwargs):
       # Change <DSSTestSession>.default_scopes to scopes for all DSSTestSession arguments
```
barroco, line 205: Looks like there are few remaining occurrence of DSSTestSession in comments. (9 occurences in [monitorlib/infrastrucure.py](https://github.com/interuss/dss/blob/566fc0b3f50f2460709cc3db1cf73bc20ddbd17e/monitoring/monitorlib/infrastructure.py) and one in [introduction_to_repository.md](https://github.com/interuss/dss/blob/master/introduction_to_repository.md?plain=1#L37))

barroco (approved):
Thanks for taking the time to do it and keep it in a separate PR :) Just a small detail in comment.

monitoring/monitorlib/infrastructure.py
BenjaminPelletier: Good catch -- thanks

PR #794, opened 2022-06-29
[monitoring/prober] Remove references to obsolete SCD API

BenjaminPelletier:
Historically, DSS capabilities for SCD were used in demonstrations circa 2020 before ASTM F3548-21 was fully developed or published.  These demonstrations used a working version of the API.  All of these demonstrations have long since been completed and the old API version is not a recognized standard version of the SCD API (instead, v1.0.0 is the first published version of that API).  So, we no longer need to support tests for this API version.

This PR removes all references to the old working version of the SCD API in prober tests.  A future PR will rename the `API_0_3_17` constant to `API_1_0_0` to reflect the actual published API version (which is programmatically equivalent to the unofficial working version v0.3.17).


monitoring/prober/scd/resources/op_400_version1.json
```diff
@@ -53,7 +53,7 @@
       }
     }
   ],
-  "oldVersion": 1,
+  "old_version": 1,
```
barroco, line 56: This is the only occurence of a test change in this PR. Is it on purpose?

barroco (approved):
Thank you for doing it! One small remark in comment.

PR #799, opened 2022-07-06
[geo-awareness] Automated testing inject definition

barroco:
This PR proposes a definition of interface to inject, activate and deactivate a test Geozones dataset on a USS.
Please note that the observation part ie /checklist endpoint will be the subject of another PR.


interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -41,9 +58,55 @@ components:
           type: string
           example: v0.0.1-445ad3
 
+    InjectGeozonesRequest:
+      type: object
+      properties:
+        source:
```
BenjaminPelletier, line 64: We'll probably add other sources in the future that aren't HTTPS sources (e.g., AMQP), so let's name this field `https_source`

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -61,5 +124,83 @@ paths:
...
+                $ref: '#/components/schemas/InjectGeozonesResponse'
+          description: >-
+            The Geo-Awareness dataset is handled by the USS and its status was retrieved successfully.
+        '401':
+          description: Bearer access token was not provided in Authorization header, token could not be decoded, or token was invalid.
+        '403':
+          description: The access token was decoded successfully but did not include a scope appropriate to this endpoint.
+        '404':
+          description: The Geozones dataset has been successfully deleted or didn't exist.
+      operationId: GetInjectGeozones
```
BenjaminPelletier, line 183: Let's either simplify this to `GetGeozone` (singular, since we're only getting one of them with this operation) or else change the resource path name to `injected_geozones` rather than just `geozones` (`/v1/injected_geozones/{injected_geozone_id}` -> GetInjectedGeozone).  RESTful endpoints should be based on a common resource name ("geozone" or "injected geozone") which has simple actions associated with it.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -61,5 +124,83 @@ paths:
...
+    delete:
+      security:
+        - Authority:
+            - utm.inject_test_data
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/InjectGeozonesResponse'
```
BenjaminPelletier, line 196: Since we aren't actually performing an Inject operation here, this data type name doesn't seem appropriate (though I agree we should reuse the same data structure).  How about changing InjectGeozonesResponse -> GeozoneResponse?

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -61,5 +124,83 @@ paths:
...
+          description: >-
+            Request has been correctly handled and the Geozones dataset will be deleted by the USS. 
+            Status of the dataset shall be checked using the GET method until it returns a 404.
+        '401':
+          description: Bearer access token was not provided in Authorization header, token could not be decoded, or token was invalid.
+        '403':
+          description: The access token was decoded successfully but did not include a scope appropriate to this endpoint.
+      operationId: DeleteGeozones
```
BenjaminPelletier, line 204: I think we probably want just singular here (DeleteGeozone)

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -41,9 +58,55 @@ components:
...
+            
+            - `Deleting`: the Geozones dataset is being deactivated and deleted.
+            
+            - `Rejected`: the Geozones dataset was rejected because the dataset is invalid.
+            
+            - `Error`: the Geozones dataset activation or deactivation failed. The description field is required in this case.
+          type: string
+          enum: [Starting, Ready]
+          example: Ready
+        description:
```
BenjaminPelletier, line 100: Personally, I would recommend `message`, but not important

interfaces/automated-testing/geo-awareness/design/sequence.plantuml
```diff
@@ -0,0 +1,37 @@
...
+
+...
+
+TD -> U: GET /geozones/{geozones_id}
+TD <- U: 200 Ok - Geozones ready\n{"result": "Ready"}
+
+== Query the USS for Geo-Awareness information ==
+
+TD -> U: GET /checklist
+TD <- U: 200 Ok - Returns the list of Geo-Awareness information\nfor a specific Volume4D.
```
BenjaminPelletier, line 22: How about "Returns a list of advisories and restrictions (when applicable) for the specified flight"?  (basically, I think a key point here is that the USSP won't simply regurgitate the information they ingested, but rather interpret it as it pertains to the flight and we should expect the resulting actions the USSP should take WRT the operator)

barroco: Sounds good to me. Thank you for the input.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
barroco: I was thinking to use the same field with different types of objects and therefore add another ref at line 67 for AMQP for instance. Isn't it correct as is? Happy to change if you think multiple keys is a better option.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -41,9 +58,55 @@ components:
...
+
+            - `Ready`: the Geozones dataset has been successfully imported and the USS can receive test requests.
+            
+            - `Deleting`: the Geozones dataset is being deactivated and deleted.
+            
+            - `Rejected`: the Geozones dataset was rejected because the dataset is invalid.
+            
+            - `Error`: the Geozones dataset activation or deactivation failed. The description field is required in this case.
+          type: string
+          enum: [Starting, Ready]
```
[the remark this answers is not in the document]
barroco: Thank you for spotting it.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -41,9 +58,55 @@ components:
...
+        source:
+          anyOf:
+            - $ref: '#/components/schemas/GeozonesHttpsSource'
+
+    GeozonesHttpsSource:
+      type: object
+      required:
+        - url
+      properties:
+        url:
```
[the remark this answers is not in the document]
barroco: Done

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
barroco: Sounds good.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
barroco: You are actually highlighting a doubt I had when working on the PR because the reference we inject points to a json file which contains a set a of geozones. So it is not entirely correct to use geozone as a resource since the entity is actually a set of geozone. What do you think would be the best approach?

barroco (commented):
Thank you @BenjaminPelletier for your review. I did a first pass of fixes. Please see this comment which will help solve the remaining issues.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
BenjaminPelletier: OpenAPI allows a field to contain AnyOf multiple types (so the field can contain data which may be one of multiple different types) because it's based on JSON validation, but we generally want to avoid that because lots of code generation tools do not like polymorphic data types.  A more common-denominator approach is to have each field contain only one data type, and then specify just one of the field options.  An example of this is `outline_circle` and `outline_polygon` in `Volume3D` of the UTM API (rather than just `outline` with AnyOf `Circle` or `Polygon`).  Since the data type for this field is (currently) `GeozonesHttpsSource`, I think `https_source` is the most appropriate name for the field.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
BenjaminPelletier: Yeah, that makes sense.  Maybe the resource name should be `geozone_source` or `geozone_set` or `geozone_collection`?  (so, e.g., GET /v1/geozone_sources/{source_id} -> GetGeozoneSource)

barroco: Changed to `/v1/geozone_sources/{geozone_source_id}` and `GetGeozoneSource`

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -41,9 +58,61 @@ components:
...
+          type: string
+          format: url
+          description: The URL at which the Geo-Awareness information shall be downloaded from.
+          example: https://caa.example.com/geozones.json
+        format:
+          type: string
+          description: The Media Type format of the response expected from this source. Only application/json is supported for the moment.
```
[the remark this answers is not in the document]
barroco: @BenjaminPelletier Thanks for the feedback. I modified it with a single value enum for the moment: [ED-269]
I don't have a strong opinion if we should already provision this field for GeoJSON now or wait for the requirement to come. Please let me know as well if a simple string would be preferable based on your experience.

barroco:
@BenjaminPelletier thank you for the review. Please note that I addressed your comments in the last commit and took the opportunity to improve some descriptions to better fit the changes. Please, let me know what do you think.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -61,5 +130,83 @@ paths:
...
+    delete:
+      security:
+        - Authority:
+            - utm.inject_test_data
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GeozoneResponse'
```
BenjaminPelletier, line 202: GeozoneSourceResponse?

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -61,5 +130,83 @@ paths:
...
+    get:
+      security:
+        - Authority:
+            - utm.inject_test_data
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GeozoneResponse'
```
BenjaminPelletier, line 180: GeozoneSourceResponse?

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -41,9 +58,61 @@ components:
           type: string
           example: v0.0.1-445ad3
 
+    GeozoneSourceRequest:
```
BenjaminPelletier, line 61: nit: I'd suggest `CreateGeozoneSourceRequest` since just `GeozoneSourceRequest` sounds like maybe we're using this to request a GeozoneSource (we're actually doing the opposite and trying to make one rather than get one)

BenjaminPelletier (approved):

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
barroco: Sounds very good, thanks

PR #811, opened 2022-09-19
[uss_qualifier/rid] Use injected flights from test response rather than from request

BenjaminPelletier:
Currently, the `rid` module of `uss_qualifier` discards the response from the test injection rather than using the flights (as injected) in the response.  This PR fixes that issue.


barroco (approved):
👍

PR #812, opened 2022-09-20
[uss_qualifier/rid] Add RID version selector to uss_qualifier rid module

BenjaminPelletier:
Currently, the `rid` module of `uss_qualifier` only tests RID according to F3411-19 and not the latest published version F3411-22a.  This PR extends uss_qualifier to test according to either version of the standard.


barroco (approved):
👍

PR #818, opened 2022-10-12
[uss_qualifier/geo-awareness] Update and generalize geo-awareness test API

BenjaminPelletier:
This PR makes some changes to the geo-awareness test API to:

1. Make it more broadly applicable to geo-awareness in non-U-space/ED-269 contexts
2. Clarify/unify some concepts (especially "selection of Geozones based on filter criteria")
3. Harmonize terminology between "UAS Zone" and "Geozone" (selecting "Geozone" as the more generic term)
4. Remove `/v1` prefixes
5. Make a few other smaller adjustments.

### DIFF
--- a/interfaces/automated-testing/geo-awareness/geo-awareness.yaml
+++ b/interfaces/automated-testing/geo-awareness/geo-awareness.yaml
@@ -152,9 +152,9 @@ components:
           $ref: '#/components/schemas/UomDimensions'
         verticalReferenceType:
           $ref: '#/components/schemas/VerticalReferenceType'
-        verticalReference:
+        height:
           description: >-
-            Height above vertical reference datum, in units of uomDimensions.
+            Height above vertical reference datum indicated in `verticalReferenceType`, in units of `uomDimensions`.
           type: number
           default: 0
         longitude:
--- a/interfaces/automated-testing/geo-awareness/geo-awareness.yaml
+++ b/interfaces/automated-testing/geo-awareness/geo-awareness.yaml
@@ -202,12 +202,12 @@ components:
             - $ref: '#/components/schemas/Position'
         after:
           description: >-
-            If specified, only select Geozones which are defined at or after this time.
+            If specified, only select Geozones which encompass at least some times at or after this time.
           type: string
           format: date-time
         before:
           description: >-
-            If specified, only select Geozones which are defined at or before this time.
+            If specified, only select Geozones which encompass at least some times at or before this time.
           type: string
           format: date-time
         ed269:
@@ -241,7 +241,7 @@ components:
             type: boolean
             description: >-
               True if and only if one or more applicable Geozones were selected according to the selection criteria of the corresponding check.
-            default: false
+          default: []
 
 paths:
   /status:
--- a/interfaces/automated-testing/geo-awareness/geo-awareness.yaml
+++ b/interfaces/automated-testing/geo-awareness/geo-awareness.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.2
 info:
   title: Geo-Awareness Automated Test Interfaces
-  version: 0.0.1
+  version: 0.1.0
   license:
     name: Apache License v2.0
     url: https://github.com/interuss/dss/blob/master/LICENSE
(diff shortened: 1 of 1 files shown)


interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -147,57 +147,100 @@ components:
...
       properties:
         uomDimensions:
           $ref: '#/components/schemas/UomDimensions'
         verticalReferenceType:
           $ref: '#/components/schemas/VerticalReferenceType'
         verticalReference:
+          description: >-
+            Height above vertical reference datum, in units of uomDimensions.
           type: number
+          default: 0
```
BenjaminPelletier, line 159: Some serialization protocols can't differentiate between omission and the default value of a field (Protobuf).  IMO, this was a very questionable design decision, but they are in use so it's generally best to accommodate their limitations when practical.  Best practice for this is generally to make otherwise-required fields optional with a default value that matches the standard default value for that data type.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -147,57 +147,100 @@ components:
...
         - verticalReferenceType
-        - verticalReference
-        - longitude
-        - latitude
       properties:
         uomDimensions:
           $ref: '#/components/schemas/UomDimensions'
         verticalReferenceType:
           $ref: '#/components/schemas/VerticalReferenceType'
         verticalReference:
```
BenjaminPelletier, line 155: Is the field name `verticalReference` used in ED-269?  If not, this should probably be something like `heightAboveVerticalReference` rather than simply `verticalReference` (since the reference datum is actually named in `verticalReferenceType` rather than this field)

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -147,57 +147,100 @@ components:
...
           type: number
+          default: 0
         latitude:
+          description: >-
+            Latitude, degrees north of the equator.
           type: number
+          default: 0
 
-    UASZonesCheckRequest:
+    GeozonesCheckRequest:
```
BenjaminPelletier, line 171: I've changed all instances of "UAS Zone" to "Geozone" as I think "Geozone" is more generic (ED-269 UASZones are special cases of generic Geozones).

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -147,57 +147,100 @@ components:
...
       type: object
       required:
-        - position
-        - uSpaceClass
-        - acceptableRestrictions
-        - startDateTime
-        - endDateTime
+        - checks
```
BenjaminPelletier, line 174: Since I think we'll have a number of tests where we'll want to perform a number of different checks, this saves a bunch of HTTP handshaking to perform multiple checks in a single HTTP call.  The response now has a corresponding array of results.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -147,57 +147,100 @@ components:
...
-            $ref: '#/components/schemas/Restriction'
-        startDateTime:
+            $ref: '#/components/schemas/GeozonesCheck'
+
+    GeozonesCheck:
+      type: object
+      required:
+        - filterSets
+      properties:
+        filterSets:
```
BenjaminPelletier, line 186: This field name is intended to make it clear that the USSP should be filtering its geo-awareness data according to the criteria provided in order to make the final true/false determination.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -147,57 +147,100 @@ components:
...
+      description: >-
+        Set of filters to select only a subset of Geozones.  Only Geozones which are applicable to all specified filters within this filter set should be selected.
+      type: object
+      properties:
+        position:
+          description: >-
+            If specified, only select Geozones encompassing this position.
+          anyOf:
+            - $ref: '#/components/schemas/Position'
+        after:
```
BenjaminPelletier, line 203: I think this is clearer than `startDateTime` because nothing starts at this time -- we are interested in some Geozones that start before this time, and some Geozones that start after this time.  The important thing is that we only look for Geozones that some temporal part that extends past this time.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -147,57 +147,100 @@ components:
...
+            If specified, only select Geozones which are defined at or after this time.
           type: string
           format: date-time
-        endDateTime:
+        before:
+          description: >-
+            If specified, only select Geozones which are defined at or before this time.
           type: string
           format: date-time
+        ed269:
```
BenjaminPelletier, line 213: By grouping filter criteria by requirement source, we can add filter criteria for other contexts alongside existing filter criteria without affecting existing filter criteria.  When the replacement for ED-269 is release, we can leave this field and the tests built for it while still stepping up to and supporting the new ED-xxx by simply adding a new filter criteria group like `edXXX`.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -147,57 +147,100 @@ components:
...
           description: >-
-            True if there is one or more applicable UASZone under the requested conditions.
-        error:
-          description: Human-readable explanation of the result for debugging purpose only. This field is required when the check do not complete.
+            True if and only if one or more applicable Geozones were selected according to the specified selection criteria.
+          default: false
 
 paths:
-  /v1/status:
+  /status:
```
BenjaminPelletier, line 247: Although other APIs include a version prefix like this, I think we shouldn't do that going forward.  USSPs are free to make the base URL whatever they want, so mandating a shared prefix like this is unnecessary and overly-constraining.  I suspect a common desire would be to include the API version in the base URL, and that API version is a full semantic version (like 0.1.0) rather than simply "v1" (there are multiple "v1"s).

BenjaminPelletier (commented):
I've added some contextual comments to help with review.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
barroco: No indeed you are right, it uses the following fields:
- lowerLimit
- lowerVerticalReference
- upperLimit
- upperVerticalReference

`verticalReference` with type `VerticalReferenceType`
and
`heighAboveVerticalReference` with type number would be correct.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
barroco: Sounds good to me.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
barroco: This is a great improvement to the initial design. Thank you!

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
```diff
@@ -147,57 +147,104 @@ components:
       required:
         - uomDimensions
         - verticalReferenceType
-        - verticalReference
-        - longitude
-        - latitude
```
barroco, line 152: Why did you remove the requirement on those fields?

barroco (approved):
Thank you for the improvements and detailed comments. Please check my answers and question below.

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
BenjaminPelletier: [To accommodate serializers that can't differentiate between omission and the default value](https://github.com/interuss/dss/pull/818#discussion_r993964212)

interfaces/automated-testing/geo-awareness/geo-awareness.yaml
BenjaminPelletier: I've now updated the `verticalReference` field name.

PR #824, opened 2022-10-13
[python] Format monitorlib

barroco:
Similarly to #823, This PR adds :
- a Makefile to lint and format the monitoring/monitorlib package
- a step to cover monitorlib.
- Format the code of monitorlib.

Due to the formatting of the package, the PR is pretty large. See first commit for points 1 and 2. The second commit is the result of make format.

### DIFF
--- a/.github/workflows/monitoring-tests.yml
+++ b/.github/workflows/monitoring-tests.yml
@@ -19,6 +19,10 @@ jobs:
           docker images
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Run Monitorlib tests
+        run: |
+          cd monitoring/monitorlib
+          make test
       - name: Run Mock USS tests
         run: |
           cd monitoring/mock_uss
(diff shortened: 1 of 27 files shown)


BenjaminPelletier (approved):
Looking at this now, do we actually need/want a separate Makefile for `monitorlib`, or could we just format all contents of `monitoring` (including `monitorlib`) directly with a single Makefile in `monitoring`?  If there's a good reason to format `monitorlib` separately from other `monitoring` components, then this PR looks good to me.

barroco:
My intention was to format package by package starting with the ones which are currently tested by the CI to build confidence about the impact of the formatting using black. We can address the other packages, which are not covered by the CI, next time before starting on on it. Eventually, we will remove package's makefiles and replace them with a single make file at the root of `monitoring`. If you agree I would propose the following sequence:
1. uss_qualifier - ✅ 
2. mock_uss - ✅ 
3. monitorlib
4. prober
5. On next modification, the other packages: atproxy, deployment_manager, interoperability, tracer, loadtest

BenjaminPelletier:
Sounds good to me -- still approved :)

PR #825, opened 2022-10-13
[uss_qualifier] Implement resources, move RID injection targets to resources

BenjaminPelletier:
This PR establishes an implementation framework for test resources (along with documentation in uss_qualifier/resources/README.md) and moves RID's "injection targets" from the "RID configuration" into this resources framework.  Future PRs will move all or nearly all of the RID configuration data structure into resource specifications in a similar way.

### DIFF
--- a/monitoring/uss_qualifier/run_locally_rid.sh
+++ b/monitoring/uss_qualifier/run_locally_rid.sh
@@ -24,18 +24,37 @@ monitoring/build.sh || exit 1
 CONFIG_LOCATION="monitoring/uss_qualifier/config_run_locally_rid.json"
 CONFIG='--config config_run_locally_rid.json'
 
-AUTH='--auth DummyOAuth(http://host.docker.internal:8085/token,uss_qualifier)'
+AUTH_SPEC='DummyOAuth(http://host.docker.internal:8085/token,uss_qualifier)'
+AUTH_FLAG="--auth ${AUTH_SPEC}"
 
 echo '{
     "locale": "CHE",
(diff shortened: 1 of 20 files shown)


monitoring/uss_qualifier/run_locally_rid.sh
```diff
@@ -24,18 +24,37 @@ monitoring/build.sh || exit 1
...
+AUTH_FLAG="--auth ${AUTH_SPEC}"
 
 echo '{
     "locale": "CHE",
+    "resources": {
+      "resource_declarations": {
+        "utm_auth": {
+          "resource_type": "communications.AuthAdapter",
+          "specification": {
+            "environment_variable_containing_auth_spec": "AUTH_SPEC"
```
barroco, line 37: Do we need to keep it from the environment? Perhaps it could simply be included directly.

barroco (approved):

monitoring/uss_qualifier/run_locally_rid.sh
BenjaminPelletier: Yes, since this is a test configuration with no sensitive information, we could put it directly in the config file and there's an option to do that in the specification.  But, my thought is that most real-world applications will have sensitive data in the auth spec and people generally won't want to put that data directly in the configuration (which is sometimes published fairly broadly), so I figured I'd set an example here.  That's the same logic that lead the auth spec to be passed via a flag in the initial implementation rather than being put directly in the configuration.

PR #827, opened 2022-10-14
[uss_qualifier/rid] Move RID flight data to resource

BenjaminPelletier:
Similar to #825 and #826, this PR moves acquisition of injectable RID flight data to a test resource instead of the current bespoke logic.

This PR also starts tracking rendered simulated flight data to promote auditability and stability of that data.  The existing simulator does not currently directly produce this data format, but can be updated in a later PR (the flight data resource specification can be adjusted to enable on-the-fly data generation in addition to loading from file).

It should be noted that uss_qualifier's webapp will need to be updated after this architecture update is complete.  Piecewise updates seemed inefficient given the current usage level.

This PR also adds missing data for TestFlightDetails (`aircraft_type` field).


barroco:
> The existing simulator does not currently directly produce this data format

Is it correct that the only difference is the missing aircraft_type or is there any other difference?

barroco (approved):

BenjaminPelletier:
> Is it correct that the only difference is the missing aircraft_type or is there any other difference?

No, the existing simulator writes N separate files, each with the `FullFlightRecord` schema whereas now there is a single file with multiple records according to the `FlightRecordCollection` schema.  It shouldn't be too hard to update, but I think we'll want to do it a little differently than it is currently and didn't want to lump that work into the same PR.

PR #832, opened 2022-10-16
[uss_qualifier] Add standard report format and capabilities

BenjaminPelletier:
The PR primarily adds a new standard report format that all test scenarios will migrate to in the future.  The TestScenario base class is upgraded to support declarations (from its subclasses) regarding its activities (start scenario, start test case, start test step, log failed check, etc) which are recorded in the standard report format automatically.  The RID nominal behavior scenario is updated to use these new TestScenario capabilities (mostly in display_data_evaluator), plus an additional previously-missing check is added for the injection step.


monitoring/uss_qualifier/scenarios/scenario.py
```diff
@@ -22,17 +33,247 @@ class TestScenario(ABC):
...
+
+    def begin_test_scenario(self) -> None:
+        if self._scenario_report is not None:
+            raise RuntimeError(
+                f"Test scenario `{self.me()}` had already begun when begin_test_scenario was called"
+            )
+        self._make_scenario_report()
+
+    def begin_test_case(self, name: str) -> None:
+        if self._scenario_report is None:
```
barroco, line 71: nit: Those checks may perhaps be extracted to a dedicated method / decorator to limit repetition and improve readability.

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -306,11 +346,20 @@ def _evaluate_normal_observation(
                 pass  # TODO: Check position, altitude, flight details, etc
 
     def _evaluate_area_to_large_observation(
```
barroco, line 348: nit: Unrelated to your changes but it seems to be a typo in the function name (`too` instead of `to` large?)

barroco (approved):

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
BenjaminPelletier: Good catch -- fixed

monitoring/uss_qualifier/scenarios/scenario.py
BenjaminPelletier: Yeah, I'm open to making that kind of change (perhaps in a different PR) -- the challenge is that the error messages are mostly slightly different from each other and I think good, precise error messages are important

PR #833, opened 2022-10-17
[uss_qualifier] Add test suites

BenjaminPelletier:
This PR implements the concept of test suites in uss_qualifier, using ASTM NetRID per F3411-19 as the first example.


monitoring/uss_qualifier/reports/report.py
```diff
@@ -123,3 +123,41 @@ class TestScenarioReport(ImplicitDict):
...
+
+class TestSuiteActionReport(ImplicitDict):
+    test_suite: Optional["TestSuiteReport"]
+    """If this action was a test suite, this field will hold its report"""
+
+    test_scenario: Optional[TestScenarioReport]
+    """If this action was a test scenario, this field will hold its report"""
+
+    def successful(self) -> bool:
+        if "test_suite" in self and self.test_suite is not None:
```
barroco, line 136: We may want to check here for invalid report which would contain test_suite and test_scenario keys so test_scenario failures don't go unnoticed in case of future wrong class usage..

monitoring/uss_qualifier/suites/suite.py
```diff
@@ -0,0 +1,198 @@
...
+    """Name of the test suite"""
+
+    resources: Dict[ResourceID, ResourceType]
+    """Enumeration of the resources used by this test suite"""
+
+    actions: List[TestSuiteActionDeclaration]
+    """The actions to take when running the test suite.  Components will be executed in order."""
+
+    @staticmethod
+    def load(suite_type: str) -> "TestSuiteDefinition":
```
barroco, line 76: `TestSuiteDefinition` type is between quotes. Is it wanted?

barroco (approved):

monitoring/uss_qualifier/suites/suite.py
BenjaminPelletier: Yes, it essentially allows forward declaration of types.  If this type wasn't in quotes and we don't have `from __future__ import annotations`, then Python will have an error since TestSuiteDefinition is not defined at this point.  Another way to address the issue is to include that __future__ statement, but I've seen a couple cases where that changed ImplicitDict behavior in unexpected ways and I didn't want to get that all ironed out at the moment.

monitoring/uss_qualifier/reports/report.py
BenjaminPelletier: Done

monitoring/uss_qualifier/suites/suite.py
barroco: Nice, thank you for the explanation.

PR #839, opened 2022-10-18
[uss_qualifier] Add action generator

BenjaminPelletier:
Currently, test runs can only use fixed definitions for test suites which means it's impossible to, for instance, run a particular test scenario for each participant in a list of participants in a participant-list resource.  This PR resolves this issue by adding the concept of action generators.  An action generator is a third type of action a test suite may contain (in addition to test scenarios and other test suites) and they present the capability of dynamically generating other test suite actions.  This PR contains a simple demonstration action generator for repeating single test suite action, and it is tested by running the simulation data generation suite twice.


barroco (approved):
👍

PR #840, opened 2022-10-18
[uss_qualifier] Add flight intents resource

BenjaminPelletier:
The PR primarily adds a flight intents resource that provides data constituting user intents to create flights (mostly `InjectFlightRequest` from the SCD injection API).  This part of the full `AutomatedTest` definition is extracted out of the JSON definition files.  To be updated in a near-future PR, this PR manually adds the flight intents into the AutomatedTest object when combining with the new resources, but appropriate TODOs are added to fix that temporary measure.

This effort necessitated a dependency cleanup such that now data definitions are split into different files/modules than the business logic, especially with regard to resources (so, now there is a new `resources/definitions.py` with content moved from `resources/resource.py`).


monitoring/uss_qualifier/resources/resource.py
```diff
@@ -35,85 +38,75 @@ def is_type(self, resource_type: str) -> bool:
...
-        constructor_signature = inspect.signature(resource_type.__init__)
-        specification_type = None
-        constructor_args = {}
-        for arg_name, arg in constructor_signature.parameters.items():
-            if arg_name == "self":
-                continue
-            if arg_name == "specification":
-                specification_type = arg.annotation
+    resources_created = 1
+    while resources_created > 0:
```
barroco, line 47: Do we still need this loop ? resources_created doesn't seem to be called and I don't see control flow statements for the while

barroco (approved):
Please have a quick look to my comments before merging.

monitoring/uss_qualifier/resources/resource.py
BenjaminPelletier: Heh, multiple mistakes apparently added up to tests passing :)  Yes, the idea here is to loop through all the resources and only create ones that don't have any unmet dependencies, then repeat that loop until no resources were created (we're finished or stuck).  We do need this condition when the logic is corrected (which is done now, hopefully).

monitoring/uss_qualifier/resources/resource.py
```diff
@@ -35,85 +38,75 @@ def is_type(self, resource_type: str) -> bool:
...
-                    'Resource "{}" was not found in the resource pool when trying to create {} resource'.format(
-                        self.dependencies[arg_name], self.resource_type
-                    )
-                )
-            constructor_args[arg_name] = resource_pool[self.dependencies[arg_name]]
-        if specification_type is not None:
-            constructor_args["specification"] = ImplicitDict.parse(
-                self.specification, specification_type
+            unmet_dependencies = sum(
+                1 if d in resource_pool else 0 for d in declaration.dependencies
```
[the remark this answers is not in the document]
BenjaminPelletier: Yes, this is one of multiple mistakes in this logic sequence; hopefully fixed now

PR #841, opened 2022-10-18
[tanka][schema-manager] add missing command on schema manager

marcadamsge:


BenjaminPelletier (approved):
I'm a little confused about why this worked before without these commands (or why it works now with them added), but seems reasonable and tests pass :)

marcadamsge:
> I'm a little confused about why this worked before without these commands (or why it works now with them added), but seems reasonable and tests pass :)

I'm deploying the dss on a local k3s instance, I had to add this to make the schema-manager job pass. I'm new to tanka though, so it could just be that I missed something somewhere 🙂

barroco (approved):
Before 6e67fecc82ec6113a51d2a9470158a94e2b348ad db-manager was running with a different image (VAR_SCHEMA_MANAGER_IMAGE_NAME) than the DSS and probably with a default command. Since then, the schema-manager is on the same image as the DSS, and therefore may require a different startup command.

Issue #822, opened 2022-10-13
Remove gRPC layer


barroco:
The current server exposes a gRPC interface, which handles HTTP requests using a separate HTTP gateway. Exposing an HTTP interface directly will substantially reduce debugging effort and improve insight into errors while reducing system complexity. We already have a tool that produces a complete HTTP server with handler stubs directly from an OpenAPI YAML; this task is to use and expand that tool to entirely replace our gRPC server infrastructure. An important aspect of this change is to update the deployment Jsonnet files and verifying the functionality.

Rationale and design notes: https://github.com/interuss/dss/wiki/Update-DSS-webserver-architecture

- [x] Phase 1: openapi-to-go-server
- [ ] Phase 2: Migrate DSS webserver
- [ ] Phase 3: Update the deployment Jsonnet files and verify the functionality. 

**References**
- Open API to go server tool: https://github.com/interuss/dss/tree/master/interfaces/openapi-to-go-server
- Integration tests: https://github.com/interuss/dss/tree/master/monitoring/prober

mickmis:
Here follows a tentative plan that also gives some insight on the approach I'm taking and points where I might need clarifications. Feedback greatly appreciated!

# Plan
- [x] **Generate interfaces and models from OpenAPI definitions**
  - [x] Generate go source files in:
    - `pkg/api/rid_v1`
    - `pkg/api/rid_v2`
    - `pkg/api/scd_v1`
  - [x] Integrate generation in Makefile
- [ ] **Replace handlers**
  For each API, there is a `Server` struct that currently implements the gRPC server interface. Each of those handlers is to be replaced by its REST counterpart generated by the tool.
  - [x] [RID v1](https://github.com/interuss/dss/tree/18db4b7c3e203e0c92fa435a3e6b7adf6d6d06fc/pkg/rid/server/v1)
  - [ ] [RID v2](https://github.com/interuss/dss/tree/18db4b7c3e203e0c92fa435a3e6b7adf6d6d06fc/pkg/rid/server/v2)
  - [ ] [SCD v1](https://github.com/interuss/dss/tree/18db4b7c3e203e0c92fa435a3e6b7adf6d6d06fc/pkg/scd)
- [ ] **Replace models conversion**
  - [x] [RID v1](https://github.com/interuss/dss/blob/18db4b7c3e203e0c92fa435a3e6b7adf6d6d06fc/pkg/rid/models/api/v1/conversions.go)
[...]

PR #848, opened 2022-10-20
#822 gRPC layer removal: refactor RID v1 API handlers and error handling, logging and validation middlewares

mickmis:
Here is a first batch of modifications for the removal of the gRPC layer.

It contains the following modifications:
- Add Makefile target for generation of API files
- Refactor error handling:
  - remove error handling middleware
  - let the handlers generate the appropriate final HTTP error
  - they do so by using an helper function that functionally replaces the middleware
  - the helper function generates a friendly error message with the error ID for the client
- Refactoring logging:
  - adapt the existing logging middleware to be used on all request
  - with a parameter controlled through CLI that allows to dump the requests and responses in the logs
  - some fields are added to cover all the info previously included by the grpc logger
- remove the validation middleware
- RID v1 API
  - update models and models conversion from proto to REST
  - adapt server unit tests

Do note the following:
- this branch in this state does not compile nor pass tests because of missing implementation (that is coming in a next PR) 
- but the version I have locally passes the prober tests (for RID API v1)
- the generated files include modifications of the generator tool that will be the subject of a next PR


BenjaminPelletier (approved):
Looks good to me!

mickmis:
> Looks good to me!

Do you mind merging it?
Because I'm making the PRs from the Orbitalize repo I can't make more than one at a time.
Thanks!

PR #849, opened 2022-10-20
[tanka][prometheus] prometheus storage class

marcadamsge:
similarly to https://github.com/interuss/dss/pull/845 , make the prometheus storage class configurable

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -305,8 +305,8 @@ a PR to that effect would be greatly appreciated.
           pool, including newly joined instances. See CockroachDB's note on the
           [join flag](https://www.cockroachlabs.com/docs/stable/start-a-node.html#flags).
 
-    1.  `VAR_CRDB_STORAGE_CLASS`: Kubernetes Storage Class to use for CockroachDB
-        volumes. You can check your cluster's possible values with
+    1.  `VAR_STORAGE_CLASS`: Kubernetes Storage Class to use for CockroachDB and
+        Prometheus volumes. You can check your cluster's possible values with
         `kubectl get storageclass`.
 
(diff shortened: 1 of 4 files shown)


build/README.md
```diff
@@ -305,8 +305,8 @@ a PR to that effect would be greatly appreciated.
           pool, including newly joined instances. See CockroachDB's note on the
           [join flag](https://www.cockroachlabs.com/docs/stable/start-a-node.html#flags).
 
-    1.  `VAR_CRDB_STORAGE_CLASS`: Kubernetes Storage Class to use for CockroachDB
-        volumes. You can check your cluster's possible values with
+    1.  `VAR_STORAGE_CLASS`: Kubernetes Storage Class to use for CockroachDB and
```
BenjaminPelletier, line 308: Could we suggest reasonable defaults here?  Something like "On Google Cloud, a reasonable value would be 'standard'.  On AWS, a reasonable value would be 'xxx'."  Basically, we want to make the procedure cloud-agnostic (which it is), but we also want to make it as accessible as possible, so breaking out how to complete a particular step in various clouds is valuable too I think (when that varies by cloud).

BenjaminPelletier (approved):

build/README.md
marcadamsge: Good idea, I'll add this 👍

marcadamsge: I'm not sure about the AWS EKS default storage classes. It seems from their doc that it's `gp2` (https://docs.aws.amazon.com/eks/latest/userguide/storage-classes.html)

barroco (approved):

PR #850, opened 2022-10-20
[tanka][ingress] optional ingress

marcadamsge:
The idea is to have one generic ingress configuration per cloud/k8s distribution to get started. The configuration is done with one variable that would take values in `gke`, `azure`, `aws` or `none` to select the target cloud. For each cloud, there would be one object with custom variables:

```
gkeIngress: {
  ipName: 'VAR_GKE_IP_NAME',
},
azureIngress: {
  ...
}
awsIngress: {
  ...
}
```

### DIFF
--- a/build/deploy/examples/minimum/main.jsonnet
+++ b/build/deploy/examples/minimum/main.jsonnet
@@ -19,9 +19,13 @@ local metadata = metadataBase {
     storageClass: 'VAR_CRDB_STORAGE_CLASS',
   },
   gateway+: {
-    ipName: 'VAR_INGRESS_NAME',
     image: 'VAR_DOCKER_IMAGE_NAME',
     hostname: 'VAR_APP_HOSTNAME',
+    ingress: 'none', // <-- This string value is VAR_INGRESS
+    // uncomment if ingress == 'gke'
+    // gkeIngress: {
+    //  ipName: 'VAR_GKE_IP_NAME',
+    // },
     traceRequests: true,
   },
   backend+: {
--- a/build/README.md
+++ b/build/README.md
@@ -309,10 +309,6 @@ a PR to that effect would be greatly appreciated.
         volumes. You can check your cluster's possible values with
         `kubectl get storageclass`.
 
-    1.  `VAR_INGRESS_NAME`: If using Google Kubernetes Engine, set this to the
-        the name of the gateway static IP address created above (e.g.,
-        `CLUSTER_NAME-gateway`).
-
     1.  `VAR_DOCKER_IMAGE_NAME`: Full name of the docker image built in the
         section above.  `build.sh` prints this name as the last thing it does
         when run with `DOCKER_URL` set.  It should look something like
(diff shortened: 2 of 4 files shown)


marcadamsge:
That's a breaking change though. I'm not sure if that's ok? @BenjaminPelletier @barroco 
If not I could try setting `gke` as the default cloud. I didn't do it there because I'm not sure if there's a nice way to set that `ipName` value from `metadata.gateway.gkeIngress.ipName` with a fallback on `metadata.gateway.ipName` and have the `ipName: error 'must supply ip name',` assertion at the same time

build/deploy/examples/minimum/main.jsonnet
```diff
@@ -19,9 +19,13 @@ local metadata = metadataBase {
     storageClass: 'VAR_CRDB_STORAGE_CLASS',
   },
   gateway+: {
-    ipName: 'VAR_INGRESS_NAME',
     image: 'VAR_DOCKER_IMAGE_NAME',
     hostname: 'VAR_APP_HOSTNAME',
+    ingress: 'none', // <-- This string value is VAR_INGRESS
```
BenjaminPelletier, line 24: I imagine there will be other things we would condition on cloud type; could we set the cloud type only once somewhere and then use that value here (and other places later)?

BenjaminPelletier (commented):
I think breaking changes are ok as long as we can effectively communicate to users that they've happened.  Is there a good, standard way to keep track of things that should go into the next release's changelog?  Maybe some GitHub magic I don't know about currently?

If there isn't such a thing, we could bootstrap something by adding `/build/changes.txt` and add a "Changes since version 0.4.0" heading, and then we'd add a new heading each release.

marcadamsge:
> I think breaking changes are ok as long as we can effectively communicate to users that they've happened. Is there a good, standard way to keep track of things that should go into the next release's changelog? Maybe some GitHub magic I don't know about currently?
> 
> If there isn't such a thing, we could bootstrap something by adding `/build/changes.txt` and add a "Changes since version 0.4.0" heading, and then we'd add a new heading each release.

A changelog is a good idea 👍 I'm unaware of such a github feature, I tried googling around but didn't find anything. I've just seen there's some kind of changelog standard https://keepachangelog.com/en/1.0.0/ and a tool for linting it https://github.com/brightcove/kacl . It might be a good idea to use these and add a linting step in the CICD, what do you think?

BenjaminPelletier:
Starting a deployment changelog seems like a good idea, especially for the changes we have coming up.  kacl seems abandoned though, and we may not need automated enforcement.  If we can just write down specific expectations, I'll bet we can manage well manually.

marcadamsge:
> Starting a deployment changelog seems like a good idea, especially for the changes we have coming up. kacl seems abandoned though, and we may not need automated enforcement. If we can just write down specific expectations, I'll bet we can manage well manually.

I added a changelog here: https://github.com/interuss/dss/pull/854

barroco:
@marcadamsge, looking at the new approach, I am not sure we will need this as is. I will close this PR for the moment. Let's keep the branch and reopen it when we add a new cloud if it is still relevant.

PR #852, opened 2022-10-21
#822 gRPC layer removal: refactoring of auth layer and merge main function

mickmis:
Second batch of modifications:

- refactor authentication and authorization layer
  - mainly due to the fact that the generation tool does not exactly use a middleware for auth, but invoke the auth object configured which will generate an `AuthorizationResult` object that will then be checked by the handlers
  - note the change in signature of `Authorize` function defined by the generation tool, this modification will be the subject of an upcoming PR
  - some cleanup of code not needed anymore 
  - adapt handlers for RID v1 API to:
    - check auth result and respond appropriately
    - retrieve `owner` from the `AuthorizationResult.ClientID` field
- merge the `main()` functions 
  - set up of middlewares

## Important note about the refactoring of the auth layer
From my understanding the previous implementation was not implementing properly the standard wrt the scopes needed for the endpoints, notably when several scopes are listed for an endpoint. 

Example for the endpoint `/v1/dss/subscriptions/{id}` (RID v1 API):
```yaml
      security:
      - AuthFromAuthorizationAuthority:
        - dss.read.identification_service_areas
        - dss.write.identification_service_areas
```
The previous implementation used to require only one of the two scopes, while referring to [the OpenAPI documentation](https://swagger.io/docs/specification/authentication/), this mandates having both scopes in the token.
In order to require only one of the two the syntax as used by the endpoint `/dss/subscriptions/{id}` (RID v2 API) must be used:
```yaml
      security:
      - Authority:
        - rid.display_provider
      - Authority:
        - rid.service_provider
```

In order to properly implement the specifications I had to change a bit the generation tool, this will be the subject of an upcoming PR.
This change induces modifications in the e2e prober tests (in an upcoming PR), because some calls do not have the proper scopes.


BenjaminPelletier:
> The previous implementation used to require only one of the two scopes, while referring to [the OpenAPI documentation](https://swagger.io/docs/specification/authentication/), this mandates having both scopes in the token.

It's true that the v1 API does not properly specify scopes, but the intent and expectation is the previous behavior: that only one of those two scopes is required.  If it helps, we can change augmented.yaml to contain the fix in the upstream repo (on a separate v1 branch).

BenjaminPelletier:
We can change the subrepo to point here for v1:
https://github.com/uastech/standards/blob/fix/auth-scopes/remoteid/augmented.yaml

mickmis:
> It's true that the v1 API does not properly specify scopes, but the intent and expectation is the previous behavior: that only one of those two scopes is required.

Alright thanks, that is good to know.

> We can change the subrepo to point here for v1

I've updated the branch to point to this new version, and re-generated the source files.

If it's now OK with you, it would be great to merge this PR so that I make the next one :) Thanks!

BenjaminPelletier (approved):

PR #855, opened 2022-10-24
[uss_qualifier] Add GraphViz visualization of a test run

BenjaminPelletier:
The nesting of suites, action generators, and scenarios in uss_qualifier can be complicated, along with which resources are used where.  This PR adds a small tool to visualize the breakdown of a test run via the data contained in its report.

Screen Shot 2022-10-24 at 9 24 16 AM

### DIFF
--- a/monitoring/uss_qualifier/visualize_configuration.py
+++ b/monitoring/uss_qualifier/visualize_configuration.py
@@ -1,0 +1,36 @@
+#!env/bin/python3
+
+import argparse
+import json
+import os
+import sys
+
+from implicitdict import ImplicitDict
+from monitoring.uss_qualifier.reports.graphs import make_graph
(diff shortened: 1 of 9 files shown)


monitoring/uss_qualifier/visualize_configuration.py
```diff
@@ -0,0 +1,36 @@
+#!env/bin/python3
```
barroco, line 1: Just a small suggestion, it would be nice to add a script in `bin/` to run this inside a container.

barroco (approved):

monitoring/uss_qualifier/visualize_configuration.py
BenjaminPelletier: Sounds good -- done.

PR #865, opened 2022-10-27
[uss_qualifier] Add tested requirements table generator

BenjaminPelletier:
The primary purpose of uss_qualifier is to determine which requirements which participants do or don't meet.  This PR adds a simply summarization of the full `TestRunReport` consisting of a table showing each participant's compliance percentage for each requirement tested.

It also changes the output generation paradigm to use the same primary uss_qualifier entry point.  The report can be generated by running uss_qualifier, or it can be loaded from a previous run's file.  Then, various outputs can be requested, including (currently) a GraphViz summary and the aforementioned tested requirements table.  All generation options are added to `run_locally.sh` and therefore exercised by the CI.

Also, documentation loading is reduced by caching loaded documentation for scenario type.


barroco (commented):
Failures seem to not be captured and report is indicating 100% success even when errors are introduced. Could you please have look?

BenjaminPelletier:
I've reproduced the behavior you mentioned and will look into it; setting this PR to draft in the mean time.

BenjaminPelletier:
Ok, issue addressed; thanks for raising.

That probably does highlight the fact that the test coverage on our test coverage is not as high as our first-level test coverage :)  Specifically, we don't exercise the failure code branches of uss_qualifier as much as the success ones.  I don't think that's necessarily a fundamental problem as there will always be more levels in the infinite regress of testing testing tools, but I think it's helpful to keep in mind.

monitoring/uss_qualifier/reports/templates/tested_requirements.html
```diff
@@ -0,0 +1,15 @@
+<html>
+<body>
+    <table>
```
barroco, line 3: In case of failure, all requirements will not appear. It is clear in the name of the file but would be good to make it explicit on the page itself.

```suggestion
    <h1>Tested requirements</h1>
    <table>
```

barroco (approved):
Looks good to me. Discovered a problem with the client hanging forever when running in docker and the ussp under test is shut down. I will make a proposal of fix in another PR since the problem is unrelated to this PR.

monitoring/uss_qualifier/reports/templates/tested_requirements.html
BenjaminPelletier: Sounds good; note added

PR #869, opened 2022-10-28
[uss_qualifier] Update documented requirements

BenjaminPelletier:
This PR cleans up and adds requirements for all of the checks in our currently-defined test scenarios.  In preparation for a more formal programmatic definition of requirements, a package-style format is adopted for requirement IDs.  Links to reference material relating to requirements are removed with the assumption that the requirements will be more explicitly defined (in a single place) in a future PR.

To avoid duplication and better align documentation with implementation, this PR also adds a feature where a test scenario documentation file may link to another file defining a test step -- this corresponds with calling a shared function to perform a common test step.  The two `test_steps.py` files contain examples of these shared test steps.

The `validate_shared_operational_intent` common test step is moved from the original single `test_steps.py` file into a `test_steps.py` in the `astm.utm` package since that test step is only valid for ASTM F3548 tests whereas the other functions in the original `test_steps.py` are suitable for any flight planning test (e.g., U-space Flight Authorisation validation).

### DIFF
--- a/monitoring/uss_qualifier/scenarios/README.md
+++ b/monitoring/uss_qualifier/scenarios/README.md
@@ -60,7 +60,7 @@ If the entire test step heading is enclosed in a link, the contents of that link
 
 Each check a test step performs that may result in a finding/issue must be documented via a subsection of the parent test step, named with a " check" suffix (example: `#### Successful injection check`).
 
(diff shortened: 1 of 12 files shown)


barroco (approved):
Looks good to me!

PR #873, opened 2022-10-28
[uss_qualifier] Add skeleton for UFT message signing test

BenjaminPelletier:
Following a discussion regarding other contributors' plans for implementing a test for UFT-style message signing, this PR adds the placeholders for this test according to the new uss_qualifier framework.

The assumption for the later implementation of this test is that an instance of InterUSS's mock_uss will be hosted at a known location and will be augmented with an optional capability to record information about message signing in its interactions with other USSs.  Control for this capability (when enabled) will be through standard HTTP calls to the mock USS.  This PR adds two `TestScenario`s intended to actuate that control when it is available.

To execute uss_qualifier with this new functionality, run `monitoring/uss_qualifier/run_locally.sh configurations.dev.faa.uft.local_message_signing` after deploying the normal prerequisites (`build/dev/run_locally.sh` to bring up a local DSS instance, `monitoring/mock_uss/run_locally_scdsc.sh` to bring up an SCD-enabled mock USS).  Doing this will run the new uss_qualifier configuration `configurations/dev/faa/uft/local_message_signing.yaml`.  This configuration includes a resource declaration for the mock USS using the new `MockUSSResource` resource type, which produces a `MockUSSClient`.  This new resource, along with the other resources needed for F3548-21 testing, are provided to the new `suites/faa/uft/message_signing.yaml` test suite which wraps the `astm.utm.f3548_21` test suite with a new `scenarios.faa.uft.StartMessageSigningReport` test scenario from `scenarios/faa/uft/message_signing_start.py` and a new `scenarios.faa.uft.FinalizeMessageSigningReport` test scenario from `scenarios/faa/uft/message_signing_finalize.py`.  The required documentation for these two scenarios is included as well.

To accommodate the desire to have mock_uss involved in all F3548-21 scenarios and to never have mock_uss involved more than once in any F3548-21 scenario, this PR also adds `FlightPlannerCombinationSelectorResource` which controls the behavior of `FlightPlannerCombinations` action generator when present.  Incidental to that addition, this PR also cleans up some resource-passing behavior with regard to action generators and introduces the ability to declare action generator resources as optional.


monitoring/uss_qualifier/scenarios/faa/uft/message_signing_start.md
```diff
@@ -0,0 +1,27 @@
+# Start message signing test scenario
+
+This test scenario instructs a mock USS to begin capturing message signing data.
+
+## Resources
+
+### mock_uss
```
barroco, line 7: I am not aware of the specifics of UFT signing but as I understand it, there should be at some point a UFT Service mocked by mock_uss, perhaps naming the resource accordingly would be better than using `mock_uss` ? (eg UFT service)

barroco (approved):
Looks good to me, just a small suggestion below.

monitoring/uss_qualifier/scenarios/faa/uft/message_signing_start.md
BenjaminPelletier: The way it sounds like message signing testing will be implemented is that InterUSS's mock_uss (the one in `monitoring`) will acquire an optional capability to 1) validate incoming signed messages (both requests of incoming calls and responses to outgoing calls) and 2) log message signing performance.  Like we currently have `ridsp`, `riddp`, `scdsc`, and `geoawareness` capability sets in mock_uss, this future PR would add `messagesigning` (or similar) which would have endpoints supporting instruction to start logging message signing performance and to finalize message signing logging.  None of the USSs under test would interact with the `messagesigning` endpoints -- instead, they would interact with other endpoints (specifically, the ASTM SCD endpoints from `scdsc`) which would use and log message signing while the `messagesigning` logging capability was active.  So, there wouldn't be a separate UFT service exactly -- instead, this is just expecting mock_uss to acquire additional controllable behaviors.  Basically, this would be similar to the `routes_behavior.py` routes for `ridsp`, only not specific to just one service.

barroco: Understood, sounds good to me. thanks

monitoring/uss_qualifier/configurations/dev/faa/uft/local_message_signing.yaml
```diff
@@ -9,8 +9,17 @@ resources:
         flight_planners:
         - participant_id: uss1
           injection_base_url: http://host.docker.internal:8074/scdsc
+        - participant_id: uss2
```
punamverma, line 12: If we want mock_uss to be the second uss in flight planner combination, then would we need uss2?

BenjaminPelletier: The way this PR is currently set up, all the F3548-21 tests will be run while mock_uss is logging, but only USS combinations 1) including mock_uss and 2) not having mock_uss in more than one role will be used.  With uss1, uss2, and mock_uss all listed as flight planners here, that means we'll see the following combinations for the nominal planning scenario:

| First-mover USS | Blocked USS |
|---|---|
| uss1 | mock_uss |
| uss2 | mock_uss |
| mock_uss | uss1 |
| mock_uss | uss2 |

Basically, uss1 and uss2 will both be tested with mock_uss present.  Putting uss1 and uss2 just in the First-mover USS role would be incomplete as they would make no outgoing calls to mock_uss.

punamverma: Sound good.
As we ll have only on uss under test at a time for message signing. 
So, we will probably go for a combination like -
uss1 | mock_uss 
mock_uss | uss1.

Hope that will be ok?

punamverma (approved):

monitoring/uss_qualifier/configurations/dev/faa/uft/local_message_signing.yaml
BenjaminPelletier: Yes -- to test just one USS at a time, only uss1 and mock_uss would be listed as flight planners in the configuration.

PR #878, opened 2022-10-31
[uss_qualifier] Add requirement declarations

BenjaminPelletier:
This PR adds programmatically-verified requirement documentation.

When a test scenario performs a check, it should generally specify a requirement that is not met when that check fails.  (Though specifying such a requirement is not currently mandatory).  If a check fails, the impacted USS should be able to clearly determine what requirement they have not fulfilled.  This PR ensures that each referenced requirement is documented to enable that discovery (even though the content of the requirements imposed by InterUSS is not populated in this PR).

These documentation requirements are documented in requirements/README.md and noted in and linked from scenarios documentation (scenarios/README.md).

In the second commit of this PR, the output of tested_requirements.html is grouped by requirement set (`--role_requirements`), and untested requirements for that requirement set are listed as well.

### DIFF
--- a/monitoring/uss_qualifier/requirements/astm/f3548/v21.md
+++ b/monitoring/uss_qualifier/requirements/astm/f3548/v21.md
@@ -1,0 +1,167 @@
+# ASTM F3548-21
+
+_Standard Specification for UAS Traffic Management (UTM) UAS Service Supplier (USS) Interoperability_
+
+For information on these requirements, refer to [the ASTM standard F3548-21](https://www.astm.org/f3548-21.html).
+
+## Common Requirements
+
+* <tt>GEN0005</tt>
+* <tt>GEN0010</tt>
+* <tt>GEN0015</tt>
+* <tt>GEN0100</tt>
+* <tt>GEN0105</tt>
+* <tt>GEN0200</tt>
+* <tt>GEN0300</tt>
+* <tt>GEN0305</tt>
+* <tt>GEN0310</tt>
+* <tt>GEN0400</tt>
+* <tt>GEN0405</tt>
+* <tt>GEN0500</tt>
+
+## Operational Intent Creation and Modification
+
+* <tt>OPIN0005</tt>
+* <tt>OPIN0010</tt>
+* <tt>OPIN0015</tt>
+* <tt>OPIN0020</tt>
+* <tt>OPIN0025</tt>
+* <tt>OPIN0030</tt>
+* <tt>OPIN0035</tt>
+* <tt>OPIN0040</tt>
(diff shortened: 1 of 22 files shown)


monitoring/uss_qualifier/requirements/astm/f3548/v21.md
```diff
@@ -0,0 +1,167 @@
...
+* <tt>CMSA0215</tt>
+* <tt>CMSA0300</tt>
+* <tt>CMSA0305</tt>
+* <tt>CMSA0310</tt>
+* <tt>CMSA0315</tt>
+* <tt>CMSA0320</tt>
+* <tt>CMSA0325</tt>
+* <tt>CMSA0330</tt>
+* <tt>USS0105</tt>
+* <tt>USS0105</tt>
```
barroco, line 91: For Strategic Conflict Detection Service, you removed duplicated USS0105 (should be required 3 times) whereas here it is mentioned twice. Is there a particular reason?

barroco (approved):
Looks good to me 👍

monitoring/uss_qualifier/requirements/astm/f3548/v21.md
BenjaminPelletier: The intent was to not have duplicates -- will remove.

monitoring/uss_qualifier/requirements/astm/f3548/v21.md
```diff
@@ -0,0 +1,167 @@
...
+* <tt>GEN0015</tt>
+* <tt>GEN0100</tt>
+* <tt>GEN0105</tt>
+* <tt>GEN0200</tt>
+* <tt>GEN0300</tt>
+* <tt>GEN0305</tt>
+* <tt>GEN0310</tt>
+* <tt>GEN0400</tt>
+* <tt>GEN0405</tt>
+* <tt>GEN0500</tt>
```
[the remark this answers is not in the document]
BenjaminPelletier: I know the checklist lists GEN0500 in those two services, but it is defined in section 5.2 (specifically 5.2.7.1) which was the rationale behind putting it here.  I think the checklists in the standard may not be fully accurate, so I've tried to make a few changes in this definition document so that each requirement is only specified once (this would be important if we were documenting what the requirement was, even though we wouldn't do this with the standards since the standard is the source of truth for that).  Now most requirements should appear in the section of the standard in which they are defined.  A few (USS*) are lumped into other sections rather than breaking them out into another "USS-USS Interfaces" section where they are actually defined in the standard because I think that's more helpful, but that could be changed in the future if it causes problems.

PR #879, opened 2022-10-31
[uss_qualifier] Set timeout to uss qualifier client requests

barroco:
When the uss_qualifier runs inside a container, connection to missing host may hang forever. This PR adds a default timeout to the test client to catch this use case. 15 seconds seemed reasonable in our context.

### DIFF
--- a/monitoring/monitorlib/infrastructure.py
+++ b/monitoring/monitorlib/infrastructure.py
@@ -18,7 +18,7 @@ ALL_SCOPES = [
 
 EPOCH = datetime.datetime.utcfromtimestamp(0)
 TOKEN_REFRESH_MARGIN = datetime.timedelta(seconds=15)
-
+CLIENT_TIMEOUT = 15 # seconds
 
 class AuthAdapter(object):
     """Base class for an adapter that add JWTs to requests."""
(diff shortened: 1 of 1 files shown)


monitoring/monitorlib/infrastructure.py
```diff
@@ -18,6 +18,7 @@
 
 EPOCH = datetime.datetime.utcfromtimestamp(0)
 TOKEN_REFRESH_MARGIN = datetime.timedelta(seconds=15)
+CLIENT_TIMEOUT = 15  # seconds
```
BenjaminPelletier, line 21: The case where we may need more than 15 seconds is with manual `atproxy` testing -- calls to `atproxy` block indefinitely expecting a worker to query `atproxy` and find the pending request, fulfill it, and post back to `atproxy` with the result.  Doing that manually would usually take more than 15 seconds.  If we could increase this to 60 seconds, I think that use case would be satisfied as well.

barroco: Thanks for mentioning it. Value changed.

BenjaminPelletier (approved):

Issue #462, opened 2020-10-27
DSS returns 200 even when operation reference is created the first time


MAbr53:
According to ASTM specs, DSS must return 200 when operation reference is updated, but it must return 201 when new operation reference is created. In fact, it always returns 200, even when operation reference is created the first time.

Found while testing for UAMX3-582. To reproduce in UAMX3-582 branch:

1. Start: operator-api, conformance-checker, datastore, fims-az-sig, dc-submitter, jaegertracing, tokenprovider, postgres, and interuss-local/dss in a docker
2. Run systest: ./run.sh dockerized gov.nasa.nup.NupBasicTest.testSuccessfulOperationLifecycle
3. Examine the operator-api log-file. You can also inspect jagger traces for PsuExchangeUtils.send that clearly show "actual_http_response": 200 in psuExchangeBody log.

BenjaminPelletier:
Fixed by #905

PR #880, opened 2022-11-01
#822 gRPC layer removal: adaptations of the prober tests

mickmis:
Seventh batch of modifications, those changes enable all tests to be validated with the new implementation.

- change from POST to GET the method to retrieve the token: because the generated code used to not check for the HTTP method, the POST call used to work, while this was actually not according to the OpenAPI definition of the dummy oauth provider
- aux API: do not retrieve anymore the version through the `as_string` subfield, which is a construct inherited from gRPC
- in all APIs: remove the error message exact comparison: now we have an error ID embedded, so we only do a substring check
- in all APIs: some validation messages changed due to the fact that it is not possible anymore for some fields to be missing (when they are specified as required they will be present but empty in the go models)
- in SCD API: some calls for creating resources used to return 200 when success, while the specs specify 201
- in RID v2 API: adapt scopes used in some endpoints, which were not implemented as per specs in the previous implementation (please check that this is indeed correct and not something that was intended as such but not specified properly like the scope issue for RID v1 API we has discussed)

After this, 2 more upcoming PRs before everything is done (modulo if I forget something): 1. fix for the path prefix; 2. cleanup and documentation

### DIFF
--- a/monitoring/prober/rid/v2/test_isa_expiry.py
+++ b/monitoring/prober/rid/v2/test_isa_expiry.py
@@ -14,7 +14,7 @@ ISA_TYPE = register_resource_type(347, 'ISA')
 
 
 def test_ensure_clean_workspace_v2(ids, session_ridv2):
-  resp = session_ridv2.get('{}/{}'.format(ISA_PATH, ids(ISA_TYPE)), scope=SCOPE_SP)
+  resp = session_ridv2.get('{}/{}'.format(ISA_PATH, ids(ISA_TYPE)), scope=SCOPE_DP)
   if resp.status_code == 200:
     version = resp.json()['service_area']['version']
     resp = session_ridv2.delete('{}/{}/{}'.format(ISA_PATH, ids(ISA_TYPE), version), scope=SCOPE_SP)
--- a/monitoring/prober/rid/v2/test_isa_expiry.py
+++ b/monitoring/prober/rid/v2/test_isa_expiry.py
@@ -14,7 +14,7 @@ ISA_TYPE = register_resource_type(347, 'ISA')
 
 
 def test_ensure_clean_workspace_v2(ids, session_ridv2):
-  resp = session_ridv2.get('{}/{}'.format(ISA_PATH, ids(ISA_TYPE)), scope=SCOPE_DP)
+  resp = session_ridv2.get('{}/{}'.format(ISA_PATH, ids(ISA_TYPE)), scope=SCOPE_SP)
   if resp.status_code == 200:
     version = resp.json()['service_area']['version']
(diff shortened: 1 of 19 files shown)


monitoring/prober/rid/v2/test_isa_expiry.py
```diff
@@ -14,7 +14,7 @@
 
 
 def test_ensure_clean_workspace_v2(ids, session_ridv2):
-  resp = session_ridv2.get('{}/{}'.format(ISA_PATH, ids(ISA_TYPE)), scope=SCOPE_SP)
+  resp = session_ridv2.get('{}/{}'.format(ISA_PATH, ids(ISA_TYPE)), scope=SCOPE_DP)
```
BenjaminPelletier, line 17: This is a bug in the OpenAPI -- Service Providers should be able to retrieve ISA information.  That change is now in `updated.yaml`, so the uastech/standards subrepo can be updated to [here](https://github.com/uastech/standards/commit/dd4016b09fc8cb98f30c2a17b5a088fb2995ab54) (and this should stay as `SCOPE_SP`)

monitoring/prober/rid/v2/test_isa_simple.py
```diff
@@ -80,7 +80,8 @@ def test_get_isa_by_id(ids, session_ridv2):
 
 @default_scope(SCOPE_SP)
 def test_update_isa(ids, session_ridv2):
-  resp = session_ridv2.get('{}/{}'.format(ISA_PATH, ids(ISA_TYPE)))
+  resp = session_ridv2.get('{}/{}'.format(ISA_PATH, ids(ISA_TYPE)), scope=SCOPE_DP)
```
BenjaminPelletier, line 83: Same as test_is_expiry.py -- this should stay as SCOPE_SP after `updated.yaml` is updated

monitoring/prober/scd/test_constraint_simple.py
```diff
@@ -108,7 +108,7 @@ def test_create_constraint(ids, scd_api, scd_session):
   assert resp.status_code == 403, resp.content
 
   resp = scd_session.put('/constraint_references/{}'.format(id), json=req, scope=SCOPE_CM)
-  assert resp.status_code == 200, resp.content
+  assert resp.status_code == 201, resp.content
```
BenjaminPelletier, line 111: I'm looking forward to finally fixing the category of problem from #462 :)  (the 201 is a solid improvement)

monitoring/prober/rid/v2/test_isa_simple.py
mickmis: Now fixed

monitoring/prober/rid/v2/test_isa_expiry.py
mickmis: Now fixed

PR #881, opened 2022-11-02
#822 gRPC layer removal: add option for path prefix and use it for RID v2 API

mickmis:
Batch 8, a PR that is normally-sized for once.
This adds an option to the OpenAPI generation tool to add a prefix to the paths of the endpoints. This is useful for the RID v2 API. This is achieved by adding an optional syntax on the CLI.


BenjaminPelletier (approved):

BenjaminPelletier:
Not sure why the last two automated checks aren't running but I think they're irrelevant for the merge decision at this point so I'm going to ignore them :)

PR #895, opened 2022-11-12
Adding mock uss report recording feature

punamverma:
1. Added report recording feature in mock uss.
2. Added report start and end endpoints in mock uss, to be used by uss_qualifier tests while running faa aft message signing tests.
3. Had to reduce gunicorn workers to 1, as report (global reprt) is used as a global variable for report recording in report_settings.py file. With multiple gunicorn workers, seems like the global variable was not being passed correctly, and we couldn't record all interactions. Not sure how to solve it at the moment. But, we can open an issue to resolve it later. Or, if somebody could help resolve it. Thanks

### DIFF
--- a/monitoring/mock_uss/scdsc/muss_reports.py
+++ b/monitoring/mock_uss/scdsc/muss_reports.py
@@ -1,0 +1,90 @@
+import datetime, json
+from typing import Dict, List, Optional
+from implicitdict import ImplicitDict
+from monitoring.monitorlib import fetch
+from pathlib import Path
+
+InteractionID = str
+
+
+class MockUssTestContext(ImplicitDict):
+    test_name: str
+    """Name of the test"""
+
+    test_case: str
+    """Testcase describing the received request, action and the expected result"""
+
+
+class Issue(ImplicitDict):
+    timestamp: Optional[str]
+    """Time the issue was discovered"""
+
+    context: MockUssTestContext
+    """The context describing what is being tested"""
+
+    uss_role: str
+    """Role USS was performing in the test when the issue occurred"""
+
+    target: str
+    """Issue is related to this USS/DSS"""
+
+    summary: str
+    """Human-readable summary of the issue"""
+
+    details: str
+    """Human-readable description of the issue"""
+
+    interactions: List[InteractionID]
+    """Description of interactions relevant to this issue"""
+
+    def __init__(self, **kwargs):
+        super(Issue, self).__init__(**kwargs)
+        if "timestamp" not in kwargs:
+            self.timestamp = datetime.datetime.utcnow().isoformat()
+
+
+class Interaction(ImplicitDict):
+    interaction_id: InteractionID
+    """ID of this interaction (used to refer to this interaction from an issue)"""
+
+    purpose: str
+    """Intended purpose of the interaction - eg. to return the appropriate response"""
+
+    context: MockUssTestContext
+    """The context describing what is being tested"""
+
+    query: fetch.Query
+    """Interaction performed (flight injection, DSS query, USS query, etc)"""
+
+
+class Findings(ImplicitDict):
+    issues: List[Issue] = []
+    interactions: List[Interaction] = []
+
+    def add_interaction(self, interaction: Interaction):
+        self.interactions.append(interaction)
+
+    def add_issue(self, issue: Issue):
+        self.issues.append(issue)
+
+    def critical_issues(self) -> List[Issue]:
+        return list(filter(lambda issue: issue.severity.Critical, self.issues))
+
+    def __repr__(self):
+        return "[{} issues in {} interactions]".format(
+            len(self.issues), len(self.interactions)
+        )
+
+class MussReport(ImplicitDict):
+    findings: Findings = Findings()
+
+    def save(self):
+        filepath = "./report/report_mock_uss_scdsc_messagesigning.json"
+        with open(filepath, "w") as f:
+            f.write(json.dumps(self, indent=4, default=str))
+        print("[Mock USS] Report saved to {}".format(filepath))
+
+
+    def reset(self):
+        self.findings.issues = []
+        self.findings.interactions = []
--- a/monitoring/mock_uss/scdsc/report_settings.py
+++ b/monitoring/mock_uss/scdsc/report_settings.py
@@ -1,0 +1,9 @@
+from monitoring.mock_uss.scdsc.muss_reports import MussReport
+from monitoring.mock_uss.scdsc.muss_report_recorder import MussReportRecorder
+
+reprt = MussReport()
+reprt_recorder = MussReportRecorder(reprt)
+
+def reset():
+    global reprt
+    reprt.reset()
--- a/monitoring/mock_uss/scdsc/routes_injection.py
+++ b/monitoring/mock_uss/scdsc/routes_injection.py
@@ -5,6 +5,10 @@ import uuid
 import flask
 import requests.exceptions
 import yaml
+import json
+from datetime import datetime
+
+from loguru import logger
 
 from monitoring.monitorlib import scd, versioning
 from monitoring.monitorlib.clients import scd as scd_client
@@ -27,6 +31,10 @@ from monitoring.mock_uss.auth import requires_scope
 from monitoring.mock_uss.scdsc import database
 from monitoring.mock_uss.scdsc.database import db
 from monitoring.monitorlib.uspace import problems_with_flight_authorisation
+from monitoring.monitorlib.clients.scd import OperationError
+from monitoring.mock_uss.scdsc import report_settings
+from loguru import logger
+import flask
 
 
 def query_operational_intents(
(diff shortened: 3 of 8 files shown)


monitoring/mock_uss/start.sh
```diff
@@ -20,6 +20,6 @@ cp health_check.sh /app
 # Start mock_uss server on port 5000
 gunicorn \
     --preload \
-    --workers=4 \
+    --workers=1 \
```
BenjaminPelletier, line 23: More than 1 worker are needed for certain operations; why would we want to reduce to 1 here?

monitoring/mock_uss/scdsc/routes_injection.py
```diff
@@ -251,24 +273,29 @@ def delete_flight(flight_id: str) -> Tuple[str, int]:
...
         )
-    scd_client.notify_subscribers(
-        resources.utm_client,
-        result.operational_intent_reference.id,
-        None,
-        result.subscribers,
-    )
-
+    for subscriber in result.subscribers.copy():
+        if subscriber.uss_base_url == 'http://host.docker.internal:8074/mock/scd' :
```
BenjaminPelletier, line 277: Why would we not want to notify these subscribers?

monitoring/mock_uss/scdsc/routes_injection.py
```diff
@@ -27,6 +31,10 @@
 from monitoring.mock_uss.scdsc import database
 from monitoring.mock_uss.scdsc.database import db
 from monitoring.monitorlib.uspace import problems_with_flight_authorisation
+from monitoring.monitorlib.clients.scd import OperationError
+from monitoring.mock_uss.scdsc import report_settings
+from loguru import logger
```
BenjaminPelletier, line 36: Duplicate import (see line 11)

monitoring/mock_uss/scdsc/routes_injection.py
```diff
@@ -70,8 +77,21 @@ def query_operational_intents(
 @requires_scope([SCOPE_SCD_QUALIFIER_INJECT])
 def scdsc_injection_status() -> Tuple[str, int]:
     """Implements USS status in SCD automated testing injection API."""
-    return flask.jsonify({"status": "Ready", "version": versioning.get_code_version()})
+    return flask.jsonify({'status': 'Ready', 'version': versioning.get_code_version()})
+
+@webapp.route('/scdsc/v1/startreport', methods=['POST'])
+@requires_scope([SCOPE_SCD_QUALIFIER_INJECT])
+def scdsc_injection_start_reporter() -> Tuple[str, int]:
```
BenjaminPelletier, line 84: The `scdsc` functionality is meant to support the scd.yaml interface.  Since these endpoints are not part of that interface (and often not used when using that interface's set of capabilities), let's put these endpoints in a new/different set of mock_uss capabilities -- perhaps named `msgsigning` or similar.

monitoring/mock_uss/scdsc/routes_injection.py
```diff
@@ -52,11 +60,10 @@ def query_operational_intents(
 
     updated_op_intents = []
     for op_intent_ref in get_details_for:
-        updated_op_intents.append(
-            scd_client.get_operational_intent_details(
-                resources.utm_client, op_intent_ref.uss_base_url, op_intent_ref.id
-            )
+        op_int, resp = scd_client.get_operational_intent_details(
```
BenjaminPelletier, line 63: [get_operational_intent_details](https://github.com/interuss/dss/blob/03ec9e86979def793da35a9e07aadcad14435684/monitoring/monitorlib/clients/scd.py#L79) seems to return an OperationalIntent, not a tuple, so this change seems incorrect

monitoring/mock_uss/scdsc/muss_reports.py
```diff
@@ -0,0 +1,90 @@
...
+from pathlib import Path
+
+InteractionID = str
+
+
+class MockUssTestContext(ImplicitDict):
+    test_name: str
+    """Name of the test"""
+
+    test_case: str
```
BenjaminPelletier, line 14: It doesn't seem like mock_uss is going to know what test case is currently being run by uss_qualifier; do we need/want this field?  Same with test_name and many of the fields in Issue.  Can we just record the interactions and their message signing statuses?

I would imagine uss_qualifier could add message signing evaluation information into the main report if mock_uss would return report content in its "stop message signing" endpoint or similar.

monitoring/mock_uss/scdsc/report_settings.py
```diff
@@ -0,0 +1,9 @@
+from monitoring.mock_uss.scdsc.muss_reports import MussReport
+from monitoring.mock_uss.scdsc.muss_report_recorder import MussReportRecorder
+
+reprt = MussReport()
```
BenjaminPelletier, line 4: This data object won't span multiple workers like it seems like it needs to.  I think this should be inside a "database" (notional, in-memory) which all workers can read and write reliably -- see [riddp](https://github.com/interuss/dss/blob/03ec9e86979def793da35a9e07aadcad14435684/monitoring/mock_uss/riddp/routes_observation.py#L137), [ridsp](https://github.com/interuss/dss/blob/b36e2988b4567d5392f84b4abfd5d1e5b0d22387/monitoring/mock_uss/ridsp/routes_injection.py#L63), and [scdsc](https://github.com/interuss/dss/blob/b36e2988b4567d5392f84b4abfd5d1e5b0d22387/monitoring/mock_uss/scdsc/routes_injection.py#L61) for examples of how to do this.

monitoring/mock_uss/scdsc/routes_injection.py
```diff
@@ -70,8 +77,21 @@ def query_operational_intents(
 @requires_scope([SCOPE_SCD_QUALIFIER_INJECT])
 def scdsc_injection_status() -> Tuple[str, int]:
     """Implements USS status in SCD automated testing injection API."""
-    return flask.jsonify({"status": "Ready", "version": versioning.get_code_version()})
+    return flask.jsonify({'status': 'Ready', 'version': versioning.get_code_version()})
+
+@webapp.route('/scdsc/v1/startreport', methods=['POST'])
```
BenjaminPelletier, line 82: Let's make these endpoints [more RESTful](https://stackoverflow.blog/2020/03/02/best-practices-for-rest-api-design/#h-use-nouns-instead-of-verbs-in-endpoint-paths).  I think that would mean making the endpoint `/reports`, then POSTing to that endpoint would create a new report with a particular ID which could then be retrieved (GET by ID) or stopped (PUT/PATCH by ID).

monitoring/mock_uss/scdsc/routes_injection.py
punamverma: This is the mock_uss itself. We were getting POSTs to self in mock_uss. Whereas, we were looking for POSTs from the other USS (USS-under-test) to mock_uss.

monitoring/mock_uss/scdsc/routes_injection.py
punamverma: Will fix.

monitoring/mock_uss/scdsc/routes_injection.py
punamverma: Will do.

monitoring/mock_uss/scdsc/routes_injection.py
punamverma: That's right. We made it a tuple, in our message_signing code. Wanted to keep that as a part of another PR. This line got overlooked while breaking the PR . Will need to fix.

monitoring/mock_uss/scdsc/muss_reports.py
punamverma: We have given different values for the test_name and test_case for the interactions and issues in mock_uss reports. 
We want to report the issues, just interactions will not help find the issues easily.

monitoring/mock_uss/scdsc/routes_injection.py
BenjaminPelletier: Perhaps self-interactions are not interesting for message signing, but mock_uss is used for many other purposes apart from message signing.  The general expectation is that notifications will be sent to all subscribers and the message signing component is free to ignore incoming notifications from itself if that is the appropriate thing to do with regard to message signing.  But, message signing should not change the baseline behavior of mock_uss.

monitoring/mock_uss/scdsc/muss_reports.py
BenjaminPelletier: How will these values be populated though?  mock_uss is a separate executable from uss_qualifier and so therefore does not generally know what uss_qualifier is doing (what test scenario/case/step/etc uss_qualifier is on).

One approach could be to simply collect message signing information in mock_uss after being instructed to start doing so, and then transfer that information back to the client (uss_qualifier) when instructed to stop collecting message signing information.  With that set of information, uss_qualifier could then incorporate that information in its single standard report.

monitoring/mock_uss/scdsc/report_settings.py
punamverma: Yes, that s why we reduced the workers to 1. But, as you said above, we need to keep multiple workers. We will look into how to remove using the examples that you mentioned.

monitoring/mock_uss/scdsc/routes_injection.py
punamverma: Will do.

BenjaminPelletier:
With the merge of #938, this PR should be reopened on the [`monitoring` repo](https://github.com/interuss/monitoring) when appropriate.

PR #896, opened 2022-11-13
Removing ClearAreaRequest field from ClearAreaResponse as per given api

punamverma:
ClearAreaRequest needs to be removed from ClearAreaResponse, as per the yaml spec.

### DIFF
--- a/monitoring/monitorlib/scd_automated_testing/scd_injection_api.py
+++ b/monitoring/monitorlib/scd_automated_testing/scd_injection_api.py
@@ -111,7 +111,6 @@ class ClearAreaOutcome(ImplicitDict):
 
 
 class ClearAreaResponse(ImplicitDict):
-    request: ClearAreaRequest
     outcome: ClearAreaOutcome
 
 
--- a/monitoring/monitorlib/scd_automated_testing/scd_injection_api.py
+++ b/monitoring/monitorlib/scd_automated_testing/scd_injection_api.py
@@ -111,6 +111,7 @@ class ClearAreaOutcome(ImplicitDict):
 
 
 class ClearAreaResponse(ImplicitDict):
+    request: Optional[ClearAreaRequest]
     outcome: ClearAreaOutcome
 
 


BenjaminPelletier (changes requested):
Implementations are generally allowed to add fields not specified in the API, which is what is being done here.  The API specifies what information must be present (and therefore what a client may depend on).  It is sometimes useful to provide information beyond that.

If this PR was merged as-is, [this line](https://github.com/interuss/dss/blob/b36e2988b4567d5392f84b4abfd5d1e5b0d22387/monitoring/mock_uss/scdsc/routes_injection.py#L297) and [this line](https://github.com/interuss/dss/blob/b36e2988b4567d5392f84b4abfd5d1e5b0d22387/monitoring/mock_uss/scdsc/routes_injection.py#L346) would break.  We apparently don't have automated tests that cover those failures currently, but they can be discovered by searching for usages of `ClearAreaResponse` in the repo.

punamverma:
> If this PR was merged as-is, [this line](https://github.com/interuss/dss/blob/b36e2988b4567d5392f84b4abfd5d1e5b0d22387/monitoring/mock_uss/scdsc/routes_injection.py#L297) and [this line](https://github.com/interuss/dss/blob/b36e2988b4567d5392f84b4abfd5d1e5b0d22387/monitoring/mock_uss/scdsc/routes_injection.py#L346) would break. We apparently don't have automated tests that cover those failures currently, but they can be discovered by searching for usages of `ClearAreaResponse` in the repo.

In fact those line broke when the clear request was sent y uss_qualifier, and then in mock_uss's response there was no ClearAreaRequest field found in ClearAreaResponse. Hence, I removed it, as the spec also didn't require that field. With this change the clear area flow now doesn't fail.

BenjaminPelletier:
Ah, ok.  The better way to parse a field when present is to mark it optional -- so:

```python
class ClearAreaResponse(ImplicitDict):
    request: Optional[ClearAreaRequest]
    outcome: ClearAreaOutcome
```

monitoring/monitorlib/scd_automated_testing/scd_injection_api.py
```diff
@@ -111,7 +111,6 @@ class ClearAreaOutcome(ImplicitDict):
 
 
 class ClearAreaResponse(ImplicitDict):
-    request: ClearAreaRequest
     outcome: ClearAreaOutcome
```
BenjaminPelletier, line 114: If this were changed to:

```python
class ClearAreaResponse(ImplicitDict):
    request: Optional[ClearAreaRequest]
    outcome: ClearAreaOutcome
```

...then this PR would look good to me.

punamverma: Will do.

BenjaminPelletier (approved):

PR #893, opened 2022-11-10
Message signing

punamverma:
1. Added reporting feature that records interactions and issues for mock uss.
2. Added a well known endpoint for public key of mock uss.
3. Removed a wrong field in ClearAreaOutcome, which was causing test failures.
4. Added message signing validation for requests and responses received by mock uss.
5. Added message signing for requests and responses sent by mock uss.
6. Added logging to mock uss using loguru.

PR #897, opened 2022-11-13
Adding a well-known endpoint for public key of mock uss

punamverma:
For FAA UFT message signing needs to be done between uss-uss communication. The public key needs to be provided at  a well-known endpoint. This PR implements the endpoint. Mock keys will be added to the keys folder. Waiting for mock keys from FAA, as the USSes need to validate the keys using SCVP validation tool. Or else, we will add both mock keys, and add a requirement for USSes to disable SCVP validation when running interuss tests.

### DIFF
--- a/monitoring/messagesigning/Dockerfile
+++ b/monitoring/messagesigning/Dockerfile
@@ -1,0 +1,19 @@
+# start by pulling the python image
+FROM python:3.8-alpine
+
+RUN mkdir -p /app/monitoring/messagesigning
+RUN mkdir -p /app/monitoring/monitorlib
+COPY ./requirements.txt /app/monitoring/messagesigning/requirements.txt
+COPY mockuss_well_known_public_key.py /app/monitoring/messagesigning/routes_well_known.py
+# switch working directory
+WORKDIR /app/monitoring/messagesigning
+
+# install the dependencies and packages in the requirements file
+RUN pip install -r requirements.txt
+
+COPY . /app/monitoring/messagesigning
+
+# configure the container to run in an executed manner
+ENTRYPOINT [ "python" ]
+
+CMD ["mockuss_well_known_public_key.py"]
--- a/monitoring/messagesigning/README.md
+++ b/monitoring/messagesigning/README.md
@@ -1,0 +1,25 @@
+# Message signing reporting/testing set up
+
+There is an environment variable titled `MESSAGE_SIGNING`. It must be set to `"true"` within `run_locally.sh` and `run_locally_scdsc.sh`. This will ensure that outgoing requests are signed and incoming requests, as well as received responses, have their message signing headers validated.
+## Setting up the `well-known` mock server
+
+### Keys
+Keys are under the folder - `monitoring/messagesigning/keys`. This folder should have two keys. Your `uft_priv.pem` key named as `mock_priv.pem`, and your public key used for valdiation named `mock_pub.der`. It is essential these two keys exist in that directory before running any message signing analysis.
+
+### Run well known server
+
+``./run_well_known.sh``  
+Public key will now be accessible at: http://host.docker.internal:8077/mock/scd/.well-known/uas-traffic-management/mock_pub.der
+
+## Creating the message signing report
+1. Inside of `monitoring/uss_qualifier`, run `./run_locally.sh configurations.dev.faa.uft.local_message_signing`
+2. Consult the created report within `monitoring/mock_uss/report/report_mock_uss_scdsc_messagesigning.json`
+
+## Analyzing the report
+No issues within the report signify that all incoming requests and received responses were successfully validated. There are several kinds of issues that could be generated. These include:  
+1. Missing message signing headers. If responses or incoming requests do not have all required message signing headers, an issue detailing which request/response was missing the headers will be recorded.
+2. Invalid content-digest message signing headers. If the `sha-512` hash value within the `content-digest` header does not match the one that `mock_uss` generates, an issue detailing the difference will be recorded. This is most likely a difference in `json` strings in the request/response bodies that get hashed. Check how your USS is generating the hash, vs how `mock_uss` is. For request/response bodies that are not null, `mock_uss` will take the `json.dumps(payload)` value and hash that.
+3. Invalid signatures. This kind of issue means the signature within the `x-utm-message-signature` header was not valid. This could happen for several reasons, it is bessed to check the logs within the `mock_uss` container for more information. 
+
+### Other info
+The `mock_uss` container will sign it's responses. It's important to check that it's signed responses can be validated by your USS. It's also important to test that the outgoing signed requests can be validated by your USS. In addition to this happy path testing, you can test that your USS rejects invalid requests, by turning the environment variable `MESSAGE_SIGNING` to "false". This will disable the signing of outgoing requests, and you should be able to see your USS handles the missing message signing headers request accordingly. It is also possible to leave the `MESSAGE_SIGNING` var to true, but use a faulty private key to sign. USS's should reject these faulty requests with a 403 status code.
--- a/monitoring/messagesigning/config.py
+++ b/monitoring/messagesigning/config.py
@@ -1,0 +1,7 @@
+import os
+
+
+class Config:
+    BASE_PATH = os.path.dirname(__file__)
+    PRIVATE_KEY_PATH = "{}/keys/mock_priv.pem".format(BASE_PATH)
+    PUBLIC_KEY_PATH = "{}/keys/mock_pub.der".format(BASE_PATH)
--- a/monitoring/messagesigning/keys/mock_priv.pem
+++ b/monitoring/messagesigning/keys/mock_priv.pem
@@ -1,0 +1,1 @@
+Private Key for Mock Uss goes here
--- a/monitoring/messagesigning/mockuss_well_known_public_key.py
+++ b/monitoring/messagesigning/mockuss_well_known_public_key.py
@@ -1,0 +1,20 @@
+import os
+import flask
+from loguru import logger
+from config import Config
+
+webapp = flask.Flask(__name__)
+
+
+@webapp.route("/mock/scd/.well-known/uas-traffic-management/mock_pub.der", methods=["GET"])
+def get_public_key():
+    """Implements notifyOperationalIntentDetailsChanged in ASTM SCD API."""
+    public_key_file_location = Config.PUBLIC_KEY_PATH
+
+    logger.info("Retreiving public key file from {}".format(public_key_file_location))
+
+    return flask.send_file(public_key_file_location)
+
+if __name__ == "__main__":
+    port = int(os.environ.get('PORT', 8077))
+    webapp.run(debug=True, host='0.0.0.0', port=port)
--- a/monitoring/messagesigning/requirements.txt
+++ b/monitoring/messagesigning/requirements.txt
@@ -1,0 +1,5 @@
+flask==1.1.2
+loguru
+itsdangerous==2.0.1 # Version 2.1.0 is not compatible with flask 1.1.2.
+Jinja2==3.0.3
+Werkzeug==2.0.3
--- a/monitoring/messagesigning/run_well_known.sh
+++ b/monitoring/messagesigning/run_well_known.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+docker image build --no-cache -t well-known .
+
+docker run -p 8077:8077 --name well-known -d well-known
--- a/monitoring/messagesigning/keys/mock_pub.der
+++ b/monitoring/messagesigning/keys/mock_pub.der
@@ -1,0 +1,1 @@
+Public cert for Mock Uss goes in this file name


monitoring/messagesigning/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# start by pulling the python image
```
barroco, line 1: By moving your code to the mock_uss module, this file will not be required anymore. 
Instead, create a `run_locally_[SERVICE_NAME].sh` similar to `[run_locally_scdsc.sh](https://github.com/interuss/dss/blob/master/monitoring/mock_uss/run_locally_scdsc.sh)`.

Based on the design document requested in #893, we will be able to agree on the actual SERVICE_NAME and if we need a new script or should update the existing one.

For your information, the docker image which will be used is defined in [`monitoring/Dockerfile`](https://github.com/interuss/dss/blob/master/monitoring/Dockerfile)
Please do not modify the docker image itself. 
If you need to use the keys defined in the configuration below, please use a volume mount in the shell script like [this](https://github.com/interuss/dss/blob/master/monitoring/mock_uss/run_locally_scdsc.sh#L30)

monitoring/messagesigning/run_well_known.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
barroco, line 1: This file should not be required anymore since it will be replaced by the new `mock_uss/run_locally_[SERVICE_NAME].sh`.

monitoring/messagesigning/requirements.txt
```diff
@@ -0,0 +1,5 @@
+flask==1.1.2
```
barroco, line 1: This file should not be required anymore since dependencies are provided by the [monitorlib requirements](https://github.com/interuss/dss/blob/master/monitoring/monitorlib/requirements.txt).

monitoring/messagesigning/requirements.txt
```diff
@@ -0,0 +1,5 @@
...
+loguru
```
barroco, line 2: I see in your other PRs that you used extensively this library, is it a requirement for the feature you are contributing or can we live without it ?
We are making our best to limit the number of dependencies to limit the security risks and size of our docker images. New additions must be motivated.

monitoring/messagesigning/config.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+
+class Config:
+    BASE_PATH = os.path.dirname(__file__)
+    PRIVATE_KEY_PATH = "{}/keys/mock_priv.pem".format(BASE_PATH)
```
barroco, line 6: Those configuration input should be moved to `mock_uss/config.py`. Again, depending on the design document requested in https://github.com/interuss/dss/pull/893, we can agree on the name of those configuration flags.

monitoring/messagesigning/mockuss_well_known_public_key.py
```diff
@@ -0,0 +1,20 @@
+import os
+import flask
+from loguru import logger
+from config import Config
+
+webapp = flask.Flask(__name__)
+
+
+@webapp.route("/mock/scd/.well-known/uas-traffic-management/mock_pub.der", methods=["GET"])
```
barroco, line 9: Is the well-known endpoint service-specific or a general feature USSs will need to enable for all services including Remote ID for instance?
In any case, this implementation will need to move to the [`mock_uss`](https://github.com/interuss/dss/tree/master/monitoring/mock_uss) module, either in a new service folder or in [mock_uss/routes.py](https://github.com/interuss/dss/blob/master/monitoring/mock_uss/routes.py) if this endpoint is unique for a USS.
Please note that the activation of this endpoint should be subject to a configuration flag disabled by default.

monitoring/messagesigning/keys/mock_priv.pem
```diff
@@ -0,0 +1 @@
+Private Key for Mock Uss goes here
```
barroco, line 1: A script to generate the keys would be required to allow the end to end testing.

monitoring/messagesigning/README.md
```diff
@@ -0,0 +1,25 @@
+# Message signing reporting/testing set up
```
barroco, line 1: The sections in this file should be move to mock_uss/README.md in a `Message signing` section.

monitoring/messagesigning/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+There is an environment variable titled `MESSAGE_SIGNING`. It must be set to `"true"` within `run_locally.sh` and `run_locally_scdsc.sh`. This will ensure that outgoing requests are signed and incoming requests, as well as received responses, have their message signing headers validated.
```
barroco, line 3: I don't see the new configuration flag in Config class below. You may want to add it?

monitoring/messagesigning/README.md
```diff
@@ -0,0 +1,25 @@
...
+3. Invalid signatures. This kind of issue means the signature within the `x-utm-message-signature` header was not valid. This could happen for several reasons, it is bessed to check the logs within the `mock_uss` container for more information. 
+
+### Other info
+The `mock_uss` container will sign it's responses. It's important to check that it's signed responses can be validated by your USS. It's also important to test that the outgoing signed requests can be validated by your USS. In addition to this happy path testing, you can test that your USS rejects invalid requests, by turning the environment variable `MESSAGE_SIGNING` to "false". This will disable the signing of outgoing requests, and you should be able to see your USS handles the missing message signing headers request accordingly. It is also possible to leave the `MESSAGE_SIGNING` var to true, but use a faulty private key to sign. USS's should reject these faulty requests with a 403 status code.
```
barroco, line 25: I would recommend to move this section in a new subsection called `Usage` in the new `Message signing` section in `mock_uss/README.md`.

monitoring/messagesigning/README.md
```diff
@@ -0,0 +1,25 @@
...
+### Run well known server
+
+``./run_well_known.sh``  
+Public key will now be accessible at: http://host.docker.internal:8077/mock/scd/.well-known/uas-traffic-management/mock_pub.der
+
+## Creating the message signing report
+1. Inside of `monitoring/uss_qualifier`, run `./run_locally.sh configurations.dev.faa.uft.local_message_signing`
+2. Consult the created report within `monitoring/mock_uss/report/report_mock_uss_scdsc_messagesigning.json`
+
+## Analyzing the report
```
barroco, line 18: This section should not be part of this PR.

monitoring/messagesigning/README.md
```diff
@@ -0,0 +1,25 @@
...
+### Keys
+Keys are under the folder - `monitoring/messagesigning/keys`. This folder should have two keys. Your `uft_priv.pem` key named as `mock_priv.pem`, and your public key used for valdiation named `mock_pub.der`. It is essential these two keys exist in that directory before running any message signing analysis.
+
+### Run well known server
+
+``./run_well_known.sh``  
+Public key will now be accessible at: http://host.docker.internal:8077/mock/scd/.well-known/uas-traffic-management/mock_pub.der
+
+## Creating the message signing report
+1. Inside of `monitoring/uss_qualifier`, run `./run_locally.sh configurations.dev.faa.uft.local_message_signing`
```
barroco, line 15: This section should not be part of this PR.

barroco (changes requested):
Thank you @punamverma for breaking down the PR as recommended by @BenjaminPelletier. Please see my comments on how to better integrate with the current code base, namely, move this new endpoint inside the [mock_uss](https://github.com/interuss/dss/tree/master/monitoring/mock_uss) module and docker image.

monitoring/messagesigning/requirements.txt
kpkeenan12: `loguru` is a nice logging library, we started using it because when calling `print`, we wouldn't see that output. `loguru` started logging our output and we saw it was highly configurable.

monitoring/messagesigning/mockuss_well_known_public_key.py
kpkeenan12: mock_uss will sign its responses according to the latest [spec](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-message-signatures-11). This endpoint provides the public key needed to validate these signatures. Other USSs will not necessarily need this endpoint, so long as when they sign their requests/responses, in their `x5u` field, they provide a publicly accessible endpoint that other USSs can use to validate their signatures.

monitoring/messagesigning/keys/mock_priv.pem
kpkeenan12: This is tricky. You can gen a pub/priv keypair with `pycryptodome` to use for testing, but these keys will not pass the FAA SCVP validation. Because of this, we asked FAA to give us mock keys that would pass their cert eval, they are located in `monitoring/messagesigning/keys`. This requires the least amount of work since by default these valid keys are going to be used. Based on other comments in this PR, we will add an endpoint to make the keypair to be used configurable, so that negative tests can be run etc, but no need to have to gen a keypair since a valid pair already exists.

monitoring/messagesigning/README.md
kpkeenan12: This is an environment variable as of right now, set similarly to other env vars (`MOCK_USS_DSS_URL`, etc)

BenjaminPelletier: Note the discussion [here](https://github.com/interuss/dss/pull/899#discussion_r1023171932); it would probably be a good idea to align on design by completing #899 before working on PRs that refer to wider designs not included in the PR (like this one)

monitoring/messagesigning/keys/mock_priv.pem
BenjaminPelletier: It would be great to complete #899 to get a good wholistic view of how each component PR will fit into the whole.  For instance, this is a question of what kind of test data will be provided where.  There are a number of different possibilities (including build/test-certs and monitoring/uss_qualifier/test_data) depending on the overall approach.  But, I don't think we've identified a reason why a `monitoring/messagesigning` folder needs to exist -- it seems like it was initially created separately from mock_uss because of [unfamiliarity with mock_uss architecture](https://github.com/interuss/dss/pull/897#discussion_r1022024581).  Using multiple workers and the SynchronizedValue tool used in other parts of mock_uss, this probably doesn't need to be a separate monitoring tool.  It would be great to discuss in #899 (see [architecture suggestion](https://github.com/interuss/dss/pull/899#discussion_r1023154837)).

kpkeenan12: The idea behind making `messagesigning` it's own package in `monitoring` is it won't just be used by `mock_uss`. `mock_uss` uses it to sign responses and validate incoming request http signatures, but it's also used by `monitoring.monitorlib.infrastructure.AuthAdapter` to sign outgoing requests to the USS under test (to ensure that individual USSs are correctly validating request signatures). I will work on the requested changes for `mock_uss` but i still think `messagesigning` should be its own package since it's being shared across components.

kpkeenan12: Linking the [PR](https://github.com/interuss/dss/pull/893) containing the `messagesigning` code

BenjaminPelletier: The best place to put code shared across multiple pieces of monitoring infrastructure is the `monitoring/monitorlib` folder -- either in a single .py file if it's small enough, or, e.g., `monitoring/monitorlib/messagesigning`.  That still leaves the question of where test data should be placed and that would probably depend on what would use the test data and how.

`monitoring.monitorlib.infrastructure.AuthAdapter` (presumably the specific [`SignedRequest` implementation](https://github.com/interuss/dss/blob/37fbff5c070cd899a8e4e1e18ee9ff78def93a22/monitoring/monitorlib/auth.py#L252)) signs outgoing requests to an auth server, but not a USS under test; it would be helpful to clarify expectations, hopefully in #899.

I'm worried about linking to open PRs within a PR.  It's important to have [small and focused PRs](https://github.com/interuss/dss/blob/master/CONTRIBUTING.md#general-principles) and a small and focused PR should not depend on other work in progress.  Linking to another open PR within a PR seems to be suggesting that understanding this PR depends on understanding the linked PR, and that suggests that this PR is not small and focused.  PRs need to be reviewed (and updated, when necessary) and merged one at a time.  I would suggest that #899 is the best candidate for first PR to complete and merge.  #891 has a number of open comments, but updating the dummy auth API YAML does seem like a well-scoped set of changes for a single PR that wouldn't need a design document (like the one in #899).

kpkeenan12: I will focus on #899. To clarify my above comment, I didn't mean `SignedRequest`, I meant the overall parent `AuthAdapter`. To add the required message signing headers, I had to edit the `add_headers` method. This means that all outgoing requests are signed technically, not just to the USS under test.

kpkeenan12: Could you please clarify what you mean by test data? We're reusing test data that already exists. Its basically re-using already existing testing scenarios, and running message signing analysis on the requests/responses associated with that scenario.

BenjaminPelletier:
Closing this PR in favor of #927, though this PR will still be available and could be reopened if appropriate.

PR #899, opened 2022-11-15
Message signing test design

punamverma:
1. Added a diagram for the test harness being used for message signing tests as per UFT requirements.
2. Added a README that describes the setup and steps needed to run uss_qualifer tests with message signing.
3. Removed the links to references that are not yet in the master branch.


monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,79 @@
+# UFT Message Signing tests for SCD using Uss_Qualifier
+
+UTM Field Test (UFT) has a requirement for USSes to sign their message requests
+and responses in the SCD flow using IATF certificates provided by FAA.
+
+This test suite helps to test the message signing by extending [suites.astm.utm.f3548_21](../../../astm/utm/f3548_21.yaml)
+with scenarios that trigger reporting of message signing in the interactions with mock_uss.
```
barroco, line 7: Is it expected that all USSs capture the report and make it available to the FAA or is it going to be generated only by a specific component of the system ?
Is there a plan to run on a distributed environment or only locally by each USS organization ?

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,79 @@
...
+## Test Harness Setup
+
+The test setup includes running the following components on your local machine, as also shown in the [diagram](./InterUss%20Test%20Harness%20With%20Message%20Signing.png)
+
+1. Uss Qualifier - This is the test driver that injects the operations for the SCD flow tests in USSes.
+2. Dummy Auth - This is a dummy auth server that provides tokens for uss-to-uss and uss-to-dss communication. Runs as a dockerized container at port 8085.
+3. DSS - This is a local DSS running in dockerized container, and is available at port 8082.
+4. Mock USS - This is a mock implementation of a USS running as a dockerized container at port 8074. The message signing reports are generated using mock_uss.
+5. USS-under-Test - This is the USS that needs to be tested.
+6. Uss Qualifier Interface. USSes need to develop an interface for their USS
```
barroco, line 18: Looking at the diagram, I think the concept of Test Harness is a bit different for other automated testing interfaces we have been developing. 
In this repository a `Test Harness` is a service implementing the test interface and responsible to map the test driver requests into the USSP proprietary systems. This service is implemented and provided by the USS under test.

You can see an example here for Remote Identification.

So on the diagram above, as far as I understand, I would recommend to rename the purple box to `test harness`.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,79 @@
...
+to interface with the test harness. Uss_Qualifer will inject operations through this interface. The spec to
+implement is - [Strategic Coordination Test Data Injection](https://github.com/interuss/automated_testing_interfaces/blob/fa3a5f544161c408f50255630a23b670c74a67d1/scd/v1/scd.yaml)
```
barroco, line 20: For my understanding, why the mock uss has a different interface than the uss under test? Based on previous experience, I could imagine that this report could be requested from any USS, including mock_uss as a reference component in the system?

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,79 @@
...
+   ./run_locally.sh configurations.dev.faa.uft.local_message_signing
+   ```
+
+## Results
+
+The USS Qualifier tests run produces two reports.
+1. SCD tests report is generated under [uss_qualifier](../../../../../uss_qualifier).
+The report file name is - report.json
+2. Message signing test report is generated under [mock_uss/report]
+The report file name is - report_mock_uss_scdsc_messagesigning.json.
```
barroco, line 58: Just another suggestion based on the experience with other projects, wouldn't it be possible to expose the report on an endpoint so the USS qualifier can retrieve the result from mock_uss and possibly, any USS under test if the interface includes such endpoint ?
That would allow the USS qualifier to aggregate and analyze centrally the results.
The main benefit of this will be that the test could be performed on publicly exposed mock_uss and USSs under test without requiring access to their file system.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+There will be 3 runs of tests, with the above two reports for each run.
```
barroco, line 60: Runs of tests could be renamed as test scenarios to match the nomenclature of the repository.

barroco (commented):
Dear @punamverma, thank you very much for taking the time to document this. This is very valuable context. The most important to me is that the nomenclature is correct to make sure concepts are aligned with the rest of the repository.
You will find some questions as well on the design to see if we can help you leverage what we have learned from other projects.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+This test suite helps to test the message signing by extending [suites.astm.utm.f3548_21](../../../astm/utm/f3548_21.yaml)
+with scenarios that trigger reporting of message signing in the interactions with mock_uss.
+
+## Test Harness Setup
+
+The test setup includes running the following components on your local machine, as also shown in the [diagram](./InterUss%20Test%20Harness%20With%20Message%20Signing.png)
+
+1. Uss Qualifier - This is the test driver that injects the operations for the SCD flow tests in USSes.
+2. Dummy Auth - This is a dummy auth server that provides tokens for uss-to-uss and uss-to-dss communication. Runs as a dockerized container at port 8085.
```
BenjaminPelletier, line 14: I'd suggest that the auth source doesn't need to be Dummy Auth in the general case -- could we enable the general case while also describing the specific case by saying something like:

> 2. Auth - This auth server providers access tokens and the public key for token validation for USS-to-USS and USS-to-DSS communication.  In a local deployment of the test infrastructure, this can be supplied by an instance of dummy auth running as a dockerized container exposing port 8085, as per build/dev/run_locally.sh.

Same comment for DSS and Mock USS.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
BenjaminPelletier: I agree that the term "test harness" is being used in a different way in this document; I'd recommend simply renaming instances "Test harness" in this document to "Test setup" -- so, for instance, this:

```
## Test Harness Setup

The test setup includes [...]
```

...would simply change to this:

```
## Test Setup

The test setup includes [...]
```

...which is what the body text already says any way.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,79 @@
...
+to interface with the test harness. Uss_Qualifer will inject operations through this interface. The spec to
+implement is - [Strategic Coordination Test Data Injection](https://github.com/interuss/automated_testing_interfaces/blob/fa3a5f544161c408f50255630a23b670c74a67d1/scd/v1/scd.yaml)
+
+The main idea behind the tests is that mock_uss will consume and validate all the requests and responses from USS-under-test.
+
+Note - As different USSes have different implementations, it could happen that you might need to create a subscription in the area of the flights.
+
+## Steps to run the test
+
+1. Set your UssQualifier Interface implementation url in the [configuration file ](../../../../configurations/dev/faa/uft/local_message_signing.yaml )to run the UFT message signing tests.
```
BenjaminPelletier, line 28: I'd recommend having the user first copy configurations/dev/faa/uft/local_message_signing.yaml to, e.g., configurations/personal/message_signing.yaml before making personalized changes

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,79 @@
...
+The property to set is `resources.resource_declarations.flight_planners.specification.flight_planners.participant_id`
+2. Run DSS and dummy-oauth using the script [run_locally.sh](../../../../../../build/dev/run_locally.sh)
+    ```bash
+    ./run_locally.sh
+    ```
+3. Run mock_uss using the script [run_locally_scdsc.sh](../../../../../mock_uss/run_locally_scdsc.sh)
+    ```bash
+   ./run_locally_scdsc.sh
+    ```
+   Note - This script has an environment variable MESSAGE_SIGNING. It is set to true to turn message signing validation on in mock uss.
```
BenjaminPelletier, line 38: The way other sets of functionality are enabled in mock_uss is to specify the functionality in MOCK_USS_SERVICES (e.g., [scdsc](https://github.com/interuss/dss/blob/877e8db6a8be7f487b08ca334a544a3da3610897/monitoring/mock_uss/run_locally_scdsc.sh#L28), [riddp](https://github.com/interuss/dss/blob/877e8db6a8be7f487b08ca334a544a3da3610897/monitoring/mock_uss/run_locally_riddp.sh#L26), [ridsp](https://github.com/interuss/dss/blob/877e8db6a8be7f487b08ca334a544a3da3610897/monitoring/mock_uss/run_locally_ridsp.sh#L28)); could we run mock_uss from `run_locally_msgsigning.sh`, which would specify `MOCK_USS_SERVICES="scdsc,msgsigning"`?

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,79 @@
...
+4. Run well-known server with public key for mock_uss. [run_well_known.sh]
```
BenjaminPelletier, line 39: Why would this be a separate executable rather than a set of mock_uss functionality?

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,79 @@
...
+A valid set of private/public keys will be provided under [keys folder]
+A USS should pass all the uss_qualifier tests in this suite.
+As well as the message_signing report should have no issues reported. A USS will provide both of these reports.
+The message_signing report includes interactions and issues between the mock_uss and the USS-under-test.
+No issues indicate the USS-under-test message-signed all its requests and responses.
+
+For analysis of the message signing reports, please refer to [Analyzing the message signing report in README.md]
+
+### Negative tests -
+Replace the private/public keys with invalid key pair, the message signing by mock_uss will be invalid.
```
BenjaminPelletier, line 72: This seems like a lot of manual work for an "automated" test; why wouldn't we expose a control endpoint for mock_uss that allows uss_qualifier to specify a keypair to use and then have 3 different test scenarios for Positive, Negative, and No message signing?  The first MVP could of course provide support for only one of these scenarios without that control endpoint, and then the test could be built out to support that control endpoint as a second phase.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
punamverma: At the moment the plan is to run the components locally by USSes for generating the reports - both uss_qualifier and message_signing. I am not sure if any USS has run the uss_qualifier tests in a distributed environment in the past. Probably, Ben would know about it. Eventually, I think these tests should be run in a distributed environment, ready for any USS in production to come and get tested.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
punamverma: I agree with that. Currently, to deliver for the scheduled events for UFT, we were able to make the flow for positive tests only. We need to put in more time and effort to get the negative tests in. However, till we are not ready with those tests, we thought that for the negative tests, we can ask the USSes to run it again with wrong key pairs. But, you are right, if time was not a constraint, we would not ask for these manual steps.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
punamverma: When we started developing, we were told to keep the message_signing reports separate from the scd flow reports. That's how we ended up with a separate report. We will need to work more on the design to be able to associate the uss-uss communication with the uss-qualifier test reports.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
punamverma: I can change that.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
punamverma: We will need to fix the global variable for report, as you suggested in the other PR. That will help resolve number of gunicorn workers issue. After that we ll be able to provide this as a mock_uss functionality.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
punamverma: Can I call it `test scenarios`, as Ben mentioned in the comment below?

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+UTM Field Test (UFT) has a requirement for USSes to sign their message requests
+and responses in the SCD flow using IATF certificates provided by FAA.
+
+This test suite helps to test the message signing by extending [suites.astm.utm.f3548_21](../../../astm/utm/f3548_21.yaml)
+with scenarios that trigger reporting of message signing in the interactions with mock_uss.
+
+## Test Harness Setup
+
+The test setup includes running the following components on your local machine, as also shown in the [diagram](./InterUss%20Test%20Harness%20With%20Message%20Signing.png)
```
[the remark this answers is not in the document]
punamverma: Sure. We can work on this. As I understand, scd.yaml interface is already implemented and baked into mock_uss. We can take out the start/stop endpoints as a separate interface in mock_uss. 
If you would want to integrate the message signing logs into report.json, do you mean the message_signing logs will be POSTed to uss_qualifier, which means uss_qualifier would need an interface for that? Or the other way is after test run, the uss_qualifier GETs the message signing logs and has a way to associate each of the message signing logs to a specific uss_qulaifier test? That would mean more work on the mock_uss interface design. Which of the above ways are you thinking it should be?  Thanks.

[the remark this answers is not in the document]
BenjaminPelletier: > As I understand, scd.yaml interface is already implemented and baked into mock_uss

Yes, mock_uss already implements scd.yaml -- it's valuable to clarify that "Inject operations" goes through that interface to clarify that uss_qualifier doesn't need to know that it's talking to a mock_uss instance in order to inject operations.  uss_qualifier *does* need to know that it's talking to a mock_uss when starting and stopping message signing logging since most USSs won't have that capability.

> after test run, the uss_qualifier GETs the message signing logs

I think that approach will work well (to be more specific, "[after the F3548-21 test suite](https://github.com/interuss/dss/blob/939110bb448665ced9bdcbf1161daba74ae4d13a/monitoring/uss_qualifier/suites/faa/uft/message_signing.yaml#L25)").

> uss_qualifier [...] has a way to associate each of the message signing logs to a specific uss_qulaifier test

uss_qualifier already has this -- [the message signing test suite](https://github.com/interuss/dss/blob/master/monitoring/uss_qualifier/suites/faa/uft/message_signing.yaml) performs three actions: [start message signing logging](https://github.com/interuss/dss/blob/fb41083a0e8d1185e5b0706c9ac8b28a814143fa/monitoring/uss_qualifier/scenarios/faa/uft/message_signing_start.py#L16), [F3548-21](https://github.com/interuss/dss/blob/939110bb448665ced9bdcbf1161daba74ae4d13a/monitoring/uss_qualifier/suites/faa/uft/message_signing.yaml#L16), [stop message signing logging](https://github.com/interuss/dss/blob/fb41083a0e8d1185e5b0706c9ac8b28a814143fa/monitoring/uss_qualifier/scenarios/faa/uft/message_signing_finalize.py#L16).  When uss_qualifier runs that last scenario (where it would obtain the logs from mock_uss under the proposed architecture), it knows what test it's on.  Logging a check from within a test scenario automatically attaches this association without any effort required from the developer.  Any general dict-like metadata can [already be attached to a FailedCheck](https://github.com/interuss/dss/blob/939110bb448665ced9bdcbf1161daba74ae4d13a/monitoring/uss_qualifier/reports/report.py#L47).  We could also add the ability to attach the same kind of open-ended data to a [PassedCheck](https://github.com/interuss/dss/blob/939110bb448665ced9bdcbf1161daba74ae4d13a/monitoring/uss_qualifier/reports/report.py#L51) if needed.

> That would mean more work on the mock_uss interface design.

We're already adding endpoints to mock_uss for starting and stopping message signing; this would either be a third endpoint like those, or else some additional data on the second endpoint.  These endpoints probably don't need to be put in YAML since we don't have any YAML for mock_uss endpoints currently because it's just a mock not intended for use outside InterUSS tooling -- YAML is useful to clearly define interfaces for external users of that interface, but we don't have external users here.

BenjaminPelletier:
I am closing this PR in favor of #922, but this PR will still be available for reference and can even be re-opened in the future if appropriate.

PR #901, opened 2022-11-16
Enhance openapi-to-go-server tool

mickmis:
Those are enhancements of the `openapi-to-go-server` tool extracted from the work on #822:

- Revisit authorizer interface: the previous interface would not allow to properly implement verification of scopes as per the OpenAPI standard.
- Add HTTP method in routing: the HTTP method was not taken into account previously in the routing, due to this if endpoints has the same paths supporting several methods, the request would be routed towards the wrong handler.
- Fix: improperly-generated code when a response body is empty (missing package)
- Fix: in the generated server code, after routing and before invoking the handler, the context passed to the handler was `context.Background()`, i.e. the context of the request was not properly passed to the handler
- Add the ability to set the option `omitempty` of the Go JSON Marshalling library (and update the generated APIs and examples).
- Add an option to add a prefix to the paths of the endpoints. This is achieved by adding an optional syntax on the CLI.
- In the example generated code:
  - Fix: dockerfile build
  - Fix: script to use the previously-generated image
  - Regen and adapt generated code in the repo
- In the dummy-oauth server:
  - Regen and adapt generated code in the repo
- In the prober tests:
  - Change from POST to GET the method to retrieve the token: because the generated code used to not check for the HTTP method, the POST call used to work, while this was actually not according to the OpenAPI definition of the dummy oauth provider


interfaces/openapi-to-go-server/example/api/rid/interface.gen.go
```diff
@@ -7,54 +7,57 @@ import (
...
-			{RequiredScopes: []string{"dss.read.identification_service_areas"}},
+	SearchSubscriptionsSecurity = []api.AuthorizationOption{
+		{
+			"AuthFromAuthorizationAuthority": {"dss.read.identification_service_areas"},
 		},
 	}
-	GetSubscriptionSecurity = map[string]api.SecurityScheme{
-		"AuthFromAuthorizationAuthority": []api.AuthorizationOption{
-			{RequiredScopes: []string{"dss.read.position_reporting_entities", "dss.write.position_reporting_entities"}},
+	GetSubscriptionSecurity = []api.AuthorizationOption{
```
BenjaminPelletier, line 40: This output is what we want, but when I run `make apis` on this branch, I get the two-scope rendering here -- I think that's because the interfaces/rid/v1 subrepo reference hasn't been moved to [975ec682a4c976abfe8ee29772768940d79044e0](https://github.com/uastech/standards/tree/fix/auth-scopes).  Could we either update that subrepo reference (if doing so doesn't ripple out to other parts of the repo), or else update this file to reflect the output of `make apis` at the current subrepo reference?

BenjaminPelletier (commented):
Thanks for splitting this out from the #822 branch -- it looks great!  We had previously discussed possibly splitting the long comment lines auto-generated by openapi-to-go-server based on single strings (without line breaks) in newer versions of the OpenAPI source, but it does look like doing that would be non-trivial, so I'll say never mind on that (at least for this PR).

Just one request below before we merge.

interfaces/openapi-to-go-server/example/api/rid/interface.gen.go
mickmis: > or else update this file to reflect the output of make apis at the current subrepo reference

I will go down this road to keep the consistency of the committed code.

mickmis: Done. I will include the update to the generated code in the other PR :)

BenjaminPelletier (approved):

PR #905, opened 2022-11-23
[dss/architecture] Remove gRPC layer

mickmis:
This removes the gRPC layer. See issue #822 

See all previous PRs that led to this one for more details:
- #848 
- #852 
- #858
- #870 
- #872 
- #877 
- #880 
- #881 
- #882 
- #887 
- #903

PR #907, opened 2022-11-23
[deployment] Fix deployment issues

BenjaminPelletier:
When performing a full Google Cloud DSS deployment from scratch to test #905, I discovered a few pre-existing deployment issues that needed to be addressed.  This PR contains the fixes I found necessary to successfully deploy the DSS in #905 and successfully complete a prober run on the deployed DSS.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,4 +37,6 @@ COPY --from=build /go/bin/http-gateway /usr/bin
 COPY --from=build /go/bin/core-service /usr/bin
 COPY --from=build /go/bin/db-manager /usr/bin
 COPY --from=build /go/bin/dlv /usr/bin
+COPY build/jwt-public-certs /jwt-public-certs
+COPY build/test-certs /test-certs
 HEALTHCHECK CMD cat service.ready || exit 1
--- a/build/README.md
+++ b/build/README.md
@@ -71,8 +71,7 @@ choice.  All major cloud providers have a docker registry service, or you can
 set up your own.
 
 To use the prebuilt InterUSS Docker images (without building them yourself), use
-`docker.io/interuss/dss` for `VAR_DOCKER_IMAGE_NAME` and
-`docker.io/interuss/db-manager` for `VAR_SCHEMA_MANAGER_IMAGE_NAME`.
+`docker.io/interuss/dss` for `VAR_DOCKER_IMAGE_NAME`.
 
 To build these images (and, optionally, push them to a docker registry):
 
@@ -95,7 +94,7 @@ endpoint.
 1. Use the [`build.sh` script](./build.sh) in this directory to build and push
    an image tagged with the current date and git commit hash.
 
-1. Note the two VAR_* values printed at the end of the script.
+1. Note the VAR_* value printed at the end of the script.
 
 ## Deploying a DSS instance via Kubernetes
 
@@ -331,11 +330,12 @@ a PR to that effect would be greatly appreciated.
 
     1.  `VAR_PUBLIC_KEY_PEM_PATH`: If providing a .pem file directly as the
         public key to validate incoming access tokens, specify the name of this
-        .pem file here as `/public-certs/YOUR-KEY-NAME.pem` replacing
+        .pem file here as `/jwt-public-certs/YOUR-KEY-NAME.pem` replacing
         YOUR-KEY-NAME as appropriate.  For instance, if using the provided
         [`us-demo.pem`](./jwt-public-certs/us-demo.pem), use the path
-        `/public-certs/us-demo.pem`.  Note that your .pem file must have been
-        copied into [`jwt-public-certs`](./jwt-public-certs) in an earlier step.
+        `/jwt-public-certs/us-demo.pem`.  Note that your .pem file must have
+        been copied into [`jwt-public-certs`](./jwt-public-certs) in an earlier
+        step.
 
         - If providing an access token public key via JWKS, provide a blank
           string for this parameter.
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -37,4 +37,6 @@ COPY --from=build /go/bin/http-gateway /usr/bin
 COPY --from=build /go/bin/core-service /usr/bin
 COPY --from=build /go/bin/db-manager /usr/bin
 COPY --from=build /go/bin/dlv /usr/bin
+COPY build/jwt-public-certs /jwt-public-certs
+COPY build/test-certs /test-certs
```
barroco, line 41: I would expect that this would be mounted at runtime but I guess it doesn't hurt to add it to the container.

build/README.md
```diff
@@ -331,11 +330,12 @@ a PR to that effect would be greatly appreciated.
...
         public key to validate incoming access tokens, specify the name of this
-        .pem file here as `/public-certs/YOUR-KEY-NAME.pem` replacing
+        .pem file here as `/jwt-public-certs/YOUR-KEY-NAME.pem` replacing
         YOUR-KEY-NAME as appropriate.  For instance, if using the provided
         [`us-demo.pem`](./jwt-public-certs/us-demo.pem), use the path
-        `/public-certs/us-demo.pem`.  Note that your .pem file must have been
-        copied into [`jwt-public-certs`](./jwt-public-certs) in an earlier step.
+        `/jwt-public-certs/us-demo.pem`.  Note that your .pem file must have
+        been copied into [`jwt-public-certs`](./jwt-public-certs) in an earlier
+        step.
```
barroco, line 338: or mounted at runtime using a volume.

build/README.md
```diff
@@ -404,9 +404,9 @@ a PR to that effect would be greatly appreciated.
 1.  If joining an existing pool, share your CRDB node addresses with the
     operators of the existing DSS instances.  They will add these node addresses
     to JoinExisting where `VAR_CRDB_EXTERNAL_NODEn` is indicated in the minimum
-    example, and then perform another rolling restart of their CockroachDB pods:
+    example, and then update their deployment:
 
-    `kubectl rollout restart statefulset/cockroachdb --namespace $NAMESPACE`
+    `tk apply workspace/$CLUSTER_CONTEXT`
```
barroco, line 409: I would recommend to keep the rollout restart as an hint in case the tk apply was done at the wrong time.

build/deploy/db_schemas/scd.libsonnet
```diff
@@ -1,12 +1,12 @@
...
-    "downfrom-v2.0.0-remove_api_1_0_0_support.sql": importstr "rid/downfrom-v2.0.0-remove_api_1_0_0_support.sql",
-    "downfrom-v1.0.0-remove_initial_version.sql": importstr "rid/downfrom-v1.0.0-remove_initial_version.sql",
+    "upto-v1.0.0-create_initial_version.sql": importstr "scd/upto-v1.0.0-create_initial_version.sql",
+    "upto-v2.0.0-support_api_1_0_0.sql": importstr "scd/upto-v2.0.0-support_api_1_0_0.sql",
+    "upto-v3.0.0-add_inverted_indices.sql": importstr "scd/upto-v3.0.0-add_inverted_indices.sql",
+    "upto-v3.1.0-create_uss_availability.sql": importstr "scd/upto-v3.1.0-create_uss_availability.sql",
+    "downfrom-v3.1.0-remove_uss_availability.sql": importstr "scd/downfrom-v3.1.0-remove_uss_availability.sql",
+    "downfrom-v3.0.0-remove_inverted_indices.sql": importstr "scd/downfrom-v3.0.0-remove_inverted_indices.sql",
+    "downfrom-v2.0.0-remove_api_1_0_0_support.sql": importstr "scd/downfrom-v2.0.0-remove_api_1_0_0_support.sql",
+    "downfrom-v1.0.0-remove_initial_version.sql": importstr "scd/downfrom-v1.0.0-remove_initial_version.sql",
```
barroco, line 10: Nice catch, we noticed it last week as well.
Since the migration is highly coupled with the code base, wouldn't it make sense to add it to the container image instead of mounting it at runtime ?
This can be addressed later.

barroco (approved):
Looks good to me. See some suggestions.

Dockerfile
BenjaminPelletier: The problem is that we don't have any instructions on how to mount them at runtime (which is non-trivial using pre-built Tanka/Kubernetes deployment configurations and procedures) and our current build documentation assumes they are already available in the container.  I might be in favor of adding instructions and Tanka plumbing to mount the appropriate public key as long as it didn't add too much complexity, but until we have that, I think these keys/certs need to be added to the image to support the current procedure and documentation.

barroco: Sounds good to me. I will make a proposal as part of the work on the infrastructure if I identify an elegant solution.

build/README.md
BenjaminPelletier: Done.

build/README.md
BenjaminPelletier: There is a rolling restart command identical to this in step "If joining an existing DSS pool" (third substep), but it's not appropriate here -- the instructions in this step are to change JoinExisting, which is a value in the Tanka main.jsonnet.  That change won't get adopted by performing a rolling restart because, until `tk apply`, that change is merely a local change.

build/deploy/db_schemas/scd.libsonnet
BenjaminPelletier: Agreed; this is static and coupled and should probably be part of the built image.  Added #908.

build/README.md
barroco: Make sense 👍

Issue #917, opened 2022-12-05
Problem when pull/ clone the repo on windows


Dutzel:
**Describe the bug**
Its currently not possible to clone or pull from the repo on windows.

The log shows: 

`git -c diff.mnemonicprefix=false -c core.quotepath=false --no-optional-locks fetch --no-tags origin

git -c diff.mnemonicprefix=false -c core.quotepath=false --no-optional-locks pull --rebase origin master
From https://github.com/interuss/dss
 * branch            master     -> FETCH_HEAD

error: The following untracked working tree files would be overwritten by merge:
	interfaces/aux_/aux.yaml
Please move or remove them before you merge.
Aborting
Updating e5bf429..ed87121
`

**To Reproduce**
1. pull the repo

**Expected behavior**
Repo should be pulled.

**Screenshots**

**Desktop (please complete the following information):**
 - Windows 11

mickmis:
Have you tried doing a clean clone from scratch?
If you kept the repo after pulling in the new `aux` now removed with #914, I do not think git will handle this well because Windows probably hides the folder from all processes.
So I would expect that pulling the latest commits is unfortunately not enough to recover.

Dutzel:
Hey mickmis, 
thank you very much for taking this issue. 
Now normal clowning works, but the cloned repo contains several problems (see picture 1)

![image](https://user-images.githubusercontent.com/17947422/205915138-03f4ce0f-afc2-403a-8d9d-6d4d77ffcd6e.png)

There are still invalid paths (e.g. "interfaces/aux_/aux.yaml"). If I then try to delete it via the git shell, the same error as in my git tool snippet below is displayed.

![image](https://user-images.githubusercontent.com/17947422/205915919-aba9cc6b-9beb-4e29-8256-d637f02391b9.png)

mickmis:
Indeed, I believed only folders would be an issue, but files named `aux` with any extension also are problematic.
In PR #918 we fix this and ensure the problem does not arise again through automatic checks.

barroco:
PR #918 has been merged and should fix the issue. @Dutzel, please comment on this issue if you still experience problems.

Dutzel:
Thank you very much :) Pulling works now

PR #918, opened 2022-12-06
[ci] Check that the project can be successfully clone in all environments

barroco:
This PR adds a check to make sure the repository can be successfully cloned on all environments and provide an automated check to address #917.


barroco:
As expected, the Operating Systems check fails on Windows. See #917

mickmis:
IMO very good idea to add this check, the aux issue of #917 is not the first time it happens: https://github.com/interuss/dss/issues?q=aux

mickmis:
I've included in the branch the renaming of the `aux.yaml` to `aux_.yaml` which fixes the issue of #917 
(@barroco maybe the PR should be renamed to reflect the change)

BenjaminPelletier (approved):

PR #919, opened 2022-12-06
[uss_qualifier] Add new geo-awareness CIS scenario

barroco:
This PR adds to the uss qualifier a first version of a ED-269 format check for Geo-Awareness CIS dataset based on the future Switzerland dataset format.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/eurocae/ed269/source_data_model.py
+++ b/monitoring/uss_qualifier/scenarios/eurocae/ed269/source_data_model.py
@@ -1,0 +1,53 @@
+import json
+from typing import List
+from implicitdict import ImplicitDict, StringBasedDateTime
+from monitoring.uss_qualifier.common_data_definitions import Severity
+from monitoring.uss_qualifier.resources.eurocae.ed269.source_document import (
+    SourceDocument,
+)
+from monitoring.uss_qualifier.scenarios.scenario import TestScenario
+from uas_standards.eurocae_ed269 import UASZoneVersion
+
+
+# TODO: When the format is confirmed, this should be moved to uas_standards.eurocae_ed269
+class ED269SchemaFile(ImplicitDict):
+    formatVersion: str
+    createdAt: StringBasedDateTime
+    UASZoneList: List[UASZoneVersion]
+
+
+class SourceDataModelValidation(TestScenario):
+    source_document: SourceDocument
+
+    def __init__(self, source_document: SourceDocument):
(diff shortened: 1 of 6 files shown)


monitoring/uss_qualifier/scenarios/eurocae/ed269/source_data_model.py
```diff
@@ -0,0 +1,54 @@
...
+
+        self.record_note(
+            "Document",
+            f"Ready at {self.source_document.specification.url}",
+        )
+
+        self.begin_test_case("ED-269 data model compliance")
+        self.begin_test_step("Valid source")
+
+        with self.check("Valid JSON", [self.source_document.specification.url]):
```
BenjaminPelletier, line 37: Note that plain ValueError (or other error) will cause the test to fail with error rather than producing a FailedCheck; it would probably be good to catch errors expected when the test infrastructure is working and the system under test does not meet requirements -- that would differentiate them from test infrastructure failures.

monitoring/uss_qualifier/scenarios/eurocae/ed269/source_data_model.py
```diff
@@ -0,0 +1,54 @@
...
+
+        with self.check(
+            "Valid schema and values", [self.source_document.specification.url]
+        ) as check:
+            try:
+                ImplicitDict.parse(data, ED269SchemaFile)
+            except ValueError as e:
+                check.record_failed(
+                    summary="Invalid format error",
+                    severity=Severity.Critical,
```
BenjaminPelletier, line 48: nit: We don't have these criticality levels clearly and crisply defined yet, but I've been using Severity.High for "this test can't continue, but the other scenarios in the test suite are probably fine to run"

BenjaminPelletier (approved):
Looks good; just 2 comments

monitoring/uss_qualifier/scenarios/eurocae/ed269/source_data_model.py
barroco: Thanks, last commit catches `json.decoder.JSONDecodeError` error.

monitoring/uss_qualifier/scenarios/eurocae/ed269/source_data_model.py
barroco: Changed to `Severity.High`.

BenjaminPelletier (approved):

PR #922, opened 2022-12-07
[mock_uss/uss_qualifier] Message signing test design

kpkeenan12:
Duplicate of https://github.com/interuss/dss/pull/899, but within the NASA fork.

### DIFF
--- a/monitoring/uss_qualifier/suites/faa/uft/design/README.md
+++ b/monitoring/uss_qualifier/suites/faa/uft/design/README.md
@@ -29,7 +29,7 @@ Note - As different USSes have different implementations, it could happen that y
 the UFT message signing tests. If personal changes are needed, copy this yaml
 file to a personal configuration file in the [personal configuration folder](../../../../configurations/personal), and edit this file instead.
 The property to set is `resources.resource_declarations.flight_planners.specification.flight_planners.participant_id`
-2. Run DSS and dummy-oauth using the script [run_locally.sh](build/dev/run_locally.sh)
+2. Run DSS and dummy-oauth using the script [run_locally.sh](../../../../../../build/dev/run_locally.sh)
     ```bash
     build/dev/run_locally.sh
     ```
@@ -41,13 +41,13 @@ The property to set is `resources.resource_declarations.flight_planners.specific
    1. The auth server used by the UTM ecosystem under test (dummy auth at http://localhost:8085/token or http://host.docker.internal:8085/token in a local deployment of the test infrastructure).
    2. A DSS instance supporting SCD in the UTM ecosystem under test (DSS at http://localhost:8082 or http://host.docker.internal:8085 in a local deployment of the test infrastructure).
 5. Run the uss_qualifier interface for your USS.
-6. Run uss_qualifier tests using script [run_locally.sh](monitoring/uss_qualifier/run_locally.sh) with config
+6. Run uss_qualifier tests using script [run_locally.sh](../../../../../../monitoring/uss_qualifier/run_locally.sh) with config
     ```bash
    monitoring/uss_qualifier/run_locally.sh configurations.dev.faa.uft.local_message_signing
    ```
 
 ## Results
-SCD tests report is generated under [uss_qualifier](monitoring/uss_qualifier).
+SCD tests report is generated under [uss_qualifier](../../../../../../monitoring/uss_qualifier).
 The message signing results will be in the report created for the overall run - report.json. Failed message signing checks will show up as `FailedChecks` within the `FinalizeMessageSigningReport` test scenario.  
 
 ### Positive tests -


monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,87 @@
...
+3. Invalid signature - After the content-digest check is performed, the `utm-message-signature` is analyzed to ensure that it is a valid signature. To do this check, the signature base is created, and the public key served in the `x5u` field within the `x-utm-jws-header` is retreived. Using the value in the `utm-message-signature` header field, the public key, and the recreated signature base, it is determined whether or not the signature is valid.
+
+
+### Positive tests -
+A valid set of private/public keys are provided for message signing analysis under the [monitioring/monitorlib/messagesigning/keys] folder. The valid key pair, `mock_faa_priv.pem`/`mock_faa_pub.der`, is used by default. The public key is served under the mock_uss endpoint `/mock/scd/.well-known/uas-traffic-management/mock_pub.der`, and can be retreived by the USS under test in order for it to validate the mock_uss responses. This public key was provided by the FAA and will pass SCVP validation. To ensure that this valid keypair is used, make sure that the `USE_VALID_KEY_PAIR` in `run_locally_msgsigning.sh` is set to "true".
```
BenjaminPelletier, line 57: How about:

```suggestion
A set of private/public keys are provided for use by mock_uss in message signing analysis under the [monitioring/mock_uss/messagesigning/keys] folder. This key pair, `mock_faa_priv.pem`/`mock_faa_pub.der`, is used by mock_uss by default. The public key is served under the mock_uss endpoint `/mock/msgsigning/.well-known/uas-traffic-management/pub.der`, and can be retrieved by the USS under test in order for it to validate the mock_uss responses. This public key was provided by the FAA and will pass SCVP validation for the UFT activity.
```

I've removed "valid" as I don't think that's well-defined here.  I think "key pair" already implies that the public key matches the private key, and we say the public key will be recognized by SCVP in UFT; are there any other characteristics that make a key pair "valid"?

Since it seems that the only purpose of this sample key pair will be for mock_uss's use, it seems like they should be stored in the mock_uss folder.  Or, if they will be used for one or more other purposes, what will those purposes be?

I've changed `mock_pub.der` to simply `pub.der` because the usage intent of the content (to validate message signatures) shouldn't depend on the content itself (whether the certificate is a mock, a real one, an invalid one, etc).

Since the `.well-known/uas-traffic-management/pub.der` endpoint is specific to the `msgsigning` mock_uss capability, it should be hosted under that prefix rather than `scd`.

I've removed the last sentence because we previously say this key pair will be used by default, so there should be no need to have a separate setting to select it.  Moreover, since we will want to change key pairs during test to validate behavior with invalid key pairs (based on previous discussions), key pair selection should be accomplished via `msgsigning` endpoint rather than environment variable.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+
+### Positive tests -
+A valid set of private/public keys are provided for message signing analysis under the [monitioring/monitorlib/messagesigning/keys] folder. The valid key pair, `mock_faa_priv.pem`/`mock_faa_pub.der`, is used by default. The public key is served under the mock_uss endpoint `/mock/scd/.well-known/uas-traffic-management/mock_pub.der`, and can be retreived by the USS under test in order for it to validate the mock_uss responses. This public key was provided by the FAA and will pass SCVP validation. To ensure that this valid keypair is used, make sure that the `USE_VALID_KEY_PAIR` in `run_locally_msgsigning.sh` is set to "true". 
+
+
+A USS should pass all the uss_qualifier tests in this suite.
+As well as the message_signing report should have no issues reported. A USS will provide both of these reports.
```
BenjaminPelletier, line 61: With the changes above, I believe this entire sentence can be removed -- the issues will be captured as FailedChecks in the main uss_qualifier report and there won't be a separate message_signing report.  Same for the following sentence.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,87 @@
...
+The message_signing report includes interactions and issues between the mock_uss and the USS-under-test.
+No issues indicate the USS-under-test message-signed all its requests and responses.
+
+
+### Negative tests -
+Replace the private/public keys with invalid key pair by setting the `USE_VALID_KEY_PAIR` in `run_locally_msgsigning.sh` to "false". This will set the keypair to be `mock_priv.pem`/`mock_pub.der`. Using this keypair will lead to invalid signatures, and the `mock_pub.der` will not pass SCVP validation.
+The USS-under-test will respond with 403 to all requests from mock_uss. The Uss Qualifier tests will not pass.
+The message signing section of the report would show `403` in interactions with mock_uss.
```
BenjaminPelletier, line 69:
```suggestion
### Negative tests
In the future, this test suite will be expanded to instruct mock_uss to replace the private/public keys with an invalid key pair by calling the appropriate endpoint. This will set the keypair to be `mock_priv.pem`/`mock_pub.der`. Using this keypair will lead to invalid signatures on mock_uss's interactions, and the `mock_pub.der` will not pass SCVP validation.
The USS-under-test should respond with 403 to all requests from mock_uss. The Uss Qualifier tests will require this to happen and only pass if the correct behavior is demonstrated.
The message signing section of the report would show `403` in interactions with mock_uss.
```

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,87 @@
...
+This test suite helps to test the message signing by extending [suites.astm.utm.f3548_21](../../../astm/utm/f3548_21.yaml)
+with scenarios that trigger reporting of message signing in the interactions with mock_uss.
+
+## Test Setup
+
+The test setup includes running the following components on your local machine, as also shown in the [diagram](./InterUss_Test_Harness_With_Message_Signing.png)
+
+1. Uss Qualifier - This is the test driver that injects the operations for the SCD flow tests in USSes.
+2. Auth - This auth server provides access tokens and the public key for token validation for USS-to-USS and USS-to-DSS communication. In a local deployment of the test infrastructure, this can be supplied by an instance of dummy auth running as a dockerized container exposing port 8085, as per build/dev/run_locally.sh.
+3. DSS - This is a local DSS running in dockerized container, and is available at port 8082.
```
BenjaminPelletier, line 15: Along the same lines of [the comment for auth](https://github.com/interuss/dss/pull/899#discussion_r1023160627), could this be generalized as well?

```suggestion
3. DSS - This is a DSS instance supporting SCD in the UTM environment defined by the auth server.  In a local deployment of the test infrastructure, this can be supplied by the local DSS instance running as a dockerized container exposing port 8082, as per build/dev/run_locally.sh.
```

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,87 @@
...
+2. Run DSS and dummy-oauth using the script [run_locally.sh](../../../../../../build/dev/run_locally.sh)
+    ```bash
+    ./run_locally.sh
+    ```
+3. Run mock_uss using the script [run_locally_msgsigning.sh](../../../../../mock_uss/run_locally_msgsigning.sh)
+    ```bash
+   ./run_locally_msgsigning.sh
+    ```
+4. Prepare your USS to run with
+   1. dummy-oauth for getting tokens. Depending on your setup - localhost(http://localhost:8085/token) or dockerized (http://host.docker.internal:8085/token). The server has a GET and POST endpoint for getting tokens.
```
BenjaminPelletier, line 39: Per line 14 above, the auth is not necessarily dummy-oauth; how about:

```suggestion
   1. The auth server used by the UTM ecosystem under test (dummy auth at http://localhost:8085/token or http://host.docker.internal:8085/token in a local deployment of the test infrastructure).
```

monitoring/uss_qualifier/suites/faa/uft/design/README.md
kpkeenan12: ` are there any other characteristics that make a key pair "valid"?`

The only thing I can think of outside of the public key passing `SCVP` validation would be the private key needs to be one that can be used to sign with the [RSASSA-PKCS1-v1_5 w/ SHA-256 algorithm](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-11.html#name-rsassa-pkcs1-v1_5-using-sha)


`Since it seems that the only purpose of this sample key pair will be for mock_uss's use, it seems like they should be stored in the mock_uss folder. Or, if they will be used for one or more other purposes, what will those purposes be?`

The public key will only be used by `mock_uss` when it signs its responses, and the private key will be used to sign outgoing requests to the USS under test by the `AuthAdapter`. (There was some [discussion](https://github.com/interuss/dss/pull/897#discussion_r1041432715) on if this is the right place for this code, when the time comes to implement this actual feature we can revisit this discussion then, I would think. But because it's currently used in two different places I don't think the keys should be strictly under the `mock_uss` package. Showing its use in our current implementation [here](https://github.com/nasa/dss/blob/uss-qual-test-ext-msg-sign/monitoring/mock_uss/scdsc/routes_scdsc.py#L165) and [here](https://github.com/nasa/dss/blob/uss-qual-test-ext-msg-sign/monitoring/monitorlib/infrastructure.py#L61) for your reference.

kpkeenan12: Forgot to include [this](https://github.com/nasa/dss/blob/uss-qual-test-ext-msg-sign/monitoring/messagesigning/message_signer.py#L41). The `signer` referenced is shown in more detail. Implementation will change based on these design modifications, but the private key will still be used in multiple places.

monitoring/mock_uss/run_locally_msgsigning.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
kpkeenan12: Done

monitoring/mock_uss/run_locally_msgsigning.sh
```diff
@@ -0,0 +1,34 @@
...
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+"${SCRIPT_DIR}/run_locally_base.sh"
+
+AUTH="DummyOAuth(http://host.docker.internal:8085/token,uss1)"
+DSS="http://host.docker.internal:8082"
+PUBLIC_KEY="/var/test-certs/auth2.pem"
+AUD=${MOCK_USS_TOKEN_AUDIENCE:-localhost,host.docker.internal}
+
+PORT=8074
```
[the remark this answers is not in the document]
kpkeenan12: Done

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,87 @@
...
+and responses in the SCD flow using IATF certificates provided by FAA.
+
+This test suite helps to test the message signing by extending [suites.astm.utm.f3548_21](../../../astm/utm/f3548_21.yaml)
+with scenarios that trigger reporting of message signing in the interactions with mock_uss.
+
+## Test Setup
+
+The test setup includes running the following components on your local machine, as also shown in the [diagram](./InterUss_Test_Harness_With_Message_Signing.png)
+
+1. Uss Qualifier - This is the test driver that injects the operations for the SCD flow tests in USSes.
```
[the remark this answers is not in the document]
kpkeenan12: Done

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+Note - As different USSes have different implementations, it could happen that you might need to create a subscription in the area of the flights.
+
+## Steps to run the test
+
+1. Set your UssQualifier Interface implementation url in the [configuration file ](../../../../configurations/dev/faa/uft/local_message_signing.yaml )to run the UFT message signing tests. If personal changes are needed, copy this yaml file to /../../../configurations/personal/message_signing.yaml, and edit this file instead.
+The property to set is `resources.resource_declarations.flight_planners.specification.flight_planners.participant_id`
+2. Run DSS and dummy-oauth using the script [run_locally.sh](../../../../../../build/dev/run_locally.sh)
+    ```bash
+    ./run_locally.sh
```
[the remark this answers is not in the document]
kpkeenan12: Done

monitoring/uss_qualifier/suites/faa/uft/design/README.md
```diff
@@ -0,0 +1,87 @@
...
+5. Run Uss Qualifier Interface for your USS.
+6. Run USS Qualifier tests using script [run_locally.sh](../../../../../uss_qualifier/run_locally.sh) with config
+    ```bash
+   ./run_locally.sh configurations.dev.faa.uft.local_message_signing
+   ```
+
+## Results
+SCD tests report is generated under [uss_qualifier](../../../../../uss_qualifier).
+The message signing results will be in the report created for the overall run - report.json - under the `report.messagesigning` section. This section will have an object titled `findings`, with two sections. The first section, `interactions`, is a list of all the interactions associated with each test. An example of an `interaction` is an incoming request from the USS under test, or an outgoing request to the USS under test. The second section, `issues`, will hold any message signing issues generated by the analysis of incoming requests and received responses. Some examples of issues that could be discovered are listed and detailed below. The current specification used for http message signing can be found [here](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-message-signatures-11).
```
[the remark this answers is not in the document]
kpkeenan12: README should reflect this now.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
kpkeenan12: Done

monitoring/uss_qualifier/suites/faa/uft/design/README.md
BenjaminPelletier: Ok, so it sounds like the base `AuthAdapter` class will be modified to add additional headers to outgoing messages as part of the authorization being applied by the AuthAdapter -- I think that makes sense (though we'll want to change the AuthAdapter description to clarify its broader scope when those changes are made).  It would probably be helpful to write a summary of how that will be incorporated, but not necessary for this PR.  If I understand correctly, then these example auth specs:

* `NoAuth()`
* `ServiceAccount(https://example.com/token, ~/credentials/account.json)`
* `DummyOAuth(http://localhost:8085/token, sub=fake_uss)`

...would become something like these (when wanting to use UFT-style message signing):

* `NoAuth(message_signing=UFT(key_path=/auth/uss1.key, cert_url=https://uss1.example.com/uss1.der))`
* `ServiceAccount(https://example.com/token, ~/credentials/account.json, message_signing=UFT(key_path=/auth/uss1.key, cert_url=https://uss1.example.com/uss1.der))`
* `DummyOAuth(http://localhost:8085/token, sub=fake_uss, message_signing=UFT(key_path=/auth/uss1.key, cert_url=https://uss1.example.com/uss1.der))`

...and probably `SignedRequest` would then be modified to let the message signing functionality of the base AuthAdapter replace the custom logic currently in `SignedRequest` such that the auth spec

`SignedRequest(https://example.com/oauth/token, client_id=uss1.com,
   key_path=/auth/uss1.key, cert_url=https://uss1.com/uss1.der)`

...would likely become something like

* `FIMSRequest(https://example.com/oauth/token, message_signing=UFT(key_path=/auth/uss1.key, cert_url=https://uss1.com/uss1.der))`

If that understanding isn't correct, can we add something to the documentation describing the actual intention?  If it is correct, then it sounds like we will be using `mock_faa_priv.pem`/`mock_faa_pub.der` in roughly the same way as we use `build/test-certs/auth2.key`/`build/test-certs/auth2.pem` -- can we plan to put these two keys in test-certs then?  Revised suggestion:

> A set of private/public keys are provided for use by mock_uss in message signing analysis under the [build/test-certs/message-signing] folder. This key pair, `mock_faa_priv.pem`/`mock_faa_pub.der`, is used by mock_uss (when requested from an instance with the capability enabled) to both sign responses to incoming messages, and to sign requests of outgoing messages. When the message signing mock_uss capability is enabled, the public key is served under the mock_uss endpoint `/mock/msgsigning/.well-known/uas-traffic-management/pub.der`, and can be retrieved by the USS under test in order for it to validate the mock_uss responses, per UFT message signing requirements. This public key was provided by the FAA and will pass SCVP validation for the UFT activity.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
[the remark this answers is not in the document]
BenjaminPelletier: FYI there are 3 instances of "Uss Qualifier" and 1 instance of "Uss_Qualifer" left (but it's rare for nit-picking comments like this to block PR approval)

monitoring/uss_qualifier/suites/faa/uft/design/README.md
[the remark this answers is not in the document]
BenjaminPelletier: The comment here was intended to have the bash commands listed how they would be for someone in the root of the repo; so, e.g.,:

```suggestion
    build/dev/run_locally.sh
```

BenjaminPelletier (commented):
Awesome, this is looking really good to me -- I think there's just one substantive comment remaining [here](https://github.com/interuss/dss/pull/922#discussion_r1044903204), plus the broken links need to be fixed.  The two nit-picking comments ([here](https://github.com/interuss/dss/pull/922#discussion_r1044905757), [here](https://github.com/interuss/dss/pull/922#discussion_r1044906632)) could be addressed if you want.

Links need to work when viewed in GitHub which means the correct path relative to the .md file being viewed needs to be in the link.  So, for instance, `monitoring/uss_qualifier/suites/astm/utm/f3548_21.yaml` should actually be `../../../astm/utm/f3548_21.yaml`.

monitoring/uss_qualifier/suites/faa/uft/design/README.md
kpkeenan12: That is the intention, making sure to save this comment to come back to when we actually do the work. The only thing I'd disagree with is the naming of `SignedRequest`, I actually think we can keep that name. `FimsRequest` is a bit too narrow. It won't just be the FIMS server that would received signed requests, but also other USSs. `SignedRequest`, in my opinion, encapsulates this intention well.  Updated this paragraph with a minor edit, mock_uss signs it responses but it is not the module that signs outgoing requests, so I made sure to specify that `AuthAdapter` will be taking care of the signing of outgoing requests. Pushing shortly

BenjaminPelletier: Nothing in this PR depends on `SignedRequest` vs `FIMSRequest` so I'll resolve the thread (feel free to unresolve to continue discussion, even after this PR's submission, or we can discuss further when we get to that later PR), but just FYI: my thought around the difference was that that particular AuthAdapter will no longer sign requests, right?  Message signatures will be done at the base `AuthAdapter` level?  In that case, it seems like there is no "signed request", but rather simply a request to FIMS (a request formatted the way FIMS wants it to be formatted), and the required signature will be attached at the AuthAdapter level, the same way message signatures could be attached (or not attached) to any of the other particular concrete AuthAdapter subclasses.

BenjaminPelletier (commented):
I think all substantive issues have been resolved; as soon as the CI passes, I think I would be ready to approve and merge this PR.

BenjaminPelletier (approved):

PR #924, opened 2022-12-11
[uss_qualifier] Add activation to priority planning scenario

BenjaminPelletier:
This PR partially completes the unfinished nominal planning priority test scenario by activating the high-priority flight.

This required behavioral changes in mock_uss's scd functionality, and some logging was added to help troubleshoot, though it should be helpful in the future as well.  The changes to the inject_flight handler appear very large if whitespace is not omitted because most of the routine is placed into a try/finally block to ensure that the locked flight is always released.  Flight locking is added in order to mutate the flight in a process-safe way.

A single nominal_planning_priority configuration is added to run just that one scenario without the need for a suite.  This was useful for debugging behavior, but it will also serve as the first example of how to run a single scenario without a test suite.

Also, some conditional-continue logic is removed because now test scenarios will automatically stop if a High or Critical Severity failed check occurs, so that behavior doesn't need to be written into each individual scenario.

A few additional smaller cleanups are also included.

### DIFF
--- a/monitoring/mock_uss/scdsc/routes_injection.py
+++ b/monitoring/mock_uss/scdsc/routes_injection.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+import time
 from typing import List, Tuple
 import uuid
 
@@ -92,6 +93,7 @@ def scd_capabilities() -> Tuple[str, int]:
 @requires_scope([SCOPE_SCD_QUALIFIER_INJECT])
 def inject_flight(flight_id: str) -> Tuple[str, int]:
     """Implements flight injection in SCD automated testing injection API."""
+    print(f"[inject_flight:{flight_id}] Starting handler")
     try:
         json = flask.request.json
         if json is None:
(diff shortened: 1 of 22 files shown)


monitoring/mock_uss/scdsc/routes_injection.py
```diff
@@ -111,105 +114,162 @@ def inject_flight(flight_id: str) -> Tuple[str, int]:
...
+                existing_flight = tx.flights[flight_id]
+                if not existing_flight.locked:
+                    print(
+                        f"[inject_flight:{flight_id}] Existing flight locked for update"
+                    )
+                    existing_flight.locked = True
+                    break
+            else:
+                print(f"[inject_flight:{flight_id}] Request is for a new flight")
+                existing_flight = None
```
barroco, line 131: I think with this implementation, there is a risk that two threads enter in the next section in parallel in the case of a new flight and creates conflicting/overriding actions. Creating a temporary locked record in the database may prevent the problem.

barroco (approved):
Looks great, please one remark before merging.

barroco (approved):
Nevermind, the dss was not up and after checking, the timeout is properly triggered.

monitoring/mock_uss/scdsc/routes_injection.py
BenjaminPelletier: Makes sense; I've added that placeholder and also ensured the other two endpoints respect this flight ownership concept (including the new placeholder).

PR #927, opened 2022-12-13
[mock_uss] Add public key endpoint

kpkeenan12:
Duplicate of #897, but within the NASA fork.

Adds an endpoint to retrieve a public key that can be used by USS's under test to validate response signatures. This key was provided by the FAA and will pass SCVP validation.


kpkeenan12:
Hi @BenjaminPelletier ! Thanks for all the help with #922. Going to use this PR to finish the work from #897 but with all the changes we talked about in the design #922. Marking this one as draft for now but when you can, please close #897.

BenjaminPelletier:
No problem :)  Sounds good and we'll take a look when this PR is marked ready for review.

monitoring/mock_uss/scdsc/routes_scdsc.py
```diff
@@ -65,6 +70,21 @@ def make_uss_report():
 
     return flask.jsonify({"message": "Not yet implemented"}), 500
 
+
+if "msgsigning" in SERVICES:
```
BenjaminPelletier, line 74: I would [expect](https://github.com/interuss/dss/tree/master/monitoring/uss_qualifier/suites/faa/uft/design#positive-tests--) this endpoint to be hosted as part of a [new](https://github.com/interuss/dss/pull/922#discussion_r1042878898) `msgsigning` [service capability](https://github.com/interuss/dss/blob/17c4f01bd787014bcba0cc6942131192398834ae/monitoring/mock_uss/__init__.py#L25) of mock_uss (similar to [`geoawareness`](https://github.com/interuss/dss/tree/master/monitoring/mock_uss/geoawareness), [`riddp`](https://github.com/interuss/dss/tree/master/monitoring/mock_uss/riddp), [`ridsp`](https://github.com/interuss/dss/tree/master/monitoring/mock_uss/ridsp), and [`scdsc`](https://github.com/interuss/dss/tree/master/monitoring/mock_uss/scdsc)).

monitoring/mock_uss/config.py
```diff
@@ -40,3 +40,6 @@ class Config(object):
     DSS_URL = os.environ.get(ENV_KEY_DSS, None)
     BEHAVIOR_LOCALITY = Locality(os.environ.get(ENV_KEY_BEHAVIOR_LOCALITY, "CHE"))
     CODE_VERSION = os.environ.get(KEY_CODE_VERSION, "Unknown")
+    CERT_PATH = "/var/test-certs"
```
BenjaminPelletier, line 43: This class is where we define configurable variables; we wouldn't want to put a constant here.

I think it probably would make sense to specify a CERT_BASE_PATH as a configurable variable like this, and this would be a reasonable default value.  However, to make it configurable, we'd want to follow the design pattern for the other configurable variables which, to accommodate Flask's somewhat wonky configuration definition and ensure traceability, includes:

1. Define environment variable of the configuration variable (probably `ENV_KEY_CERT_BASE_PATH = "{}_CERT_BASE_PATH".format(ENV_KEY_PREFIX)` to match previous style)
2. Define a constant for the config key name to avoid magic strings in different files (probably `KEY_CERT_BASE_URL = "CERT_BASE_URL"`)
3. Add the configuration variable the config object, loading the value from the environment variable (probably `CERT_BASE_URL = os.environ.get(ENV_KEY_CERT_BASE_URL, "/var/test-certs")`)

At that point, the value can be used at runtime with [`webapp.config.get(KEY_CERT_BASE_URL)`](https://github.com/interuss/dss/blob/17c4f01bd787014bcba0cc6942131192398834ae/monitoring/mock_uss/ridsp/routes_injection.py#L48).

monitoring/mock_uss/config.py
```diff
@@ -40,3 +40,6 @@ class Config(object):
...
+    PRIVATE_KEY_PATH = "{}/messagesigning/mock_faa_priv.pem".format(CERT_PATH)
```
BenjaminPelletier, line 44: Since we want to [eventually be able to change these paths at runtime](https://github.com/interuss/dss/tree/master/monitoring/uss_qualifier/suites/faa/uft/design#negative-tests--), these shouldn't be items set in the static configuration of mock_uss.  Until we add the infrastructure to change these paths dynamically, I think it's fine to hardcode a path value in the handler with a TODO note pointing to the documentation saying what will happen eventually.

I would expect that eventually the `msgsigning` service capability would acquire its own "database" (like all the other service capabilities have) and one of the "tables" would be a setting of which key pair to use.

monitoring/mock_uss/scdsc/routes_scdsc.py
```diff
@@ -65,6 +70,21 @@ def make_uss_report():
...
+
+    @webapp.route(
+        "/mock/msgsigning/.well-known/uas-traffic-management/pub.der", methods=["GET"]
+    )
+    def get_public_key():
+        public_key_file_location = Config.PUBLIC_KEY_PATH
```
BenjaminPelletier, line 80: Based on the comments above, how about either

```suggestion
        # TODO: Implement runtime changes to certificates (see monitoring/uss_qualifier/suites/faa/uft/design/README.md)
        public_key_file_location = "/var/test-certs/messagesigning/mock_faa_pub.der"
```

or

```suggestion
        # TODO: Implement runtime changes to certificates (see monitoring/uss_qualifier/suites/faa/uft/design/README.md)
        public_key_file_location = os.path.join(webapp.config.get(config.KEY_CERT_BASE_PATH), "messagesigning/mock_faa_pub.der"
```

(depending on whether the CERT_BASE_PATH is made configurable in this PR)

kpkeenan12: I went with the latter!

monitoring/mock_uss/msgsigning/database.py
```diff
@@ -0,0 +1,19 @@
...
+from monitoring.monitorlib.multiprocessing import SynchronizedValue
+from implicitdict import ImplicitDict
+
+# TODO Use this database to dynamically alter which key pair to use for message signing activities.
+
+
+class Database(ImplicitDict):
+    """Simple in-memory pseudo-database tracking whether or not to use a valid key pair for message signing activities."""
+
+    use_valid_keypair: bool = True
```
BenjaminPelletier, line 13: This will work, but I'd suggest that maybe we could allow for much more flexibility without almost any additional effort:

```suggestion
    public_key_name: str = "messagesigning/mock_faa_priv.der"
    private_key_name: str = "messagesigning/mock_faa_priv.pem"
```

With this, a future endpoint to change the keypair could accept any keypair available to mock_uss without additional code, and the file names could be fully resolved without any additional hardcoding.

monitoring/mock_uss/msgsigning/routes_msgsigning.py
```diff
@@ -0,0 +1,18 @@
...
+import os
+
+
+@webapp.route(
+    "/mock/msgsigning/.well-known/uas-traffic-management/pub.der", methods=["GET"]
+)
+def get_public_key():
+    public_key_file_location = os.path.join(
+        webapp.config.get(config.KEY_CERT_BASE_PATH), "messagesigning/mock_faa_pub.der"
+    )
```
BenjaminPelletier, line 14: If the comment above is adopted, this could be changed to:

```suggestion
    public_key_file_location = os.path.join(
        webapp.config.get(config.KEY_CERT_BASE_PATH),
        db.value.public_key_name
    )
```

...and this line would also be needed:

```python
from monitoring.mock_uss.msgsigning.database import db
```

monitoring/mock_uss/msgsigning/routes_msgsigning.py
```diff
@@ -0,0 +1,18 @@
...
+
+    logger.info("Retreiving public key file from {}".format(public_key_file_location))
```
BenjaminPelletier, line 16: nit: Retrieving :)

monitoring/mock_uss/msgsigning/__init__.py
```diff
@@ -0,0 +1,6 @@
+from monitoring.mock_uss import config, SERVICE_MESSAGESIGNING
+
+if config.Config.CERT_BASE_PATH is None:
+    raise ValueError(
+        f"CERT_BASE_PATH is required for the {SERVICE_MESSAGESIGNING} service"
+    )
```
BenjaminPelletier, line 6: Since we're giving a default value to Config.CERT_BASE_PATH, I don't think this can ever be None.  We could check for non-emptiness:

```suggestion
if not config.Config.CERT_BASE_PATH:
    raise ValueError(
        f"Environment variable {config.ENV_KEY_CERT_BASE_PATH} may not be blank for the {SERVICE_MESSAGESIGNING} service"
    )
```

monitoring/mock_uss/msgsigning/routes_msgsigning.py
```diff
@@ -0,0 +1,31 @@
...
+    logger.info("Retrieving public key file from {}".format(public_key_file_location))
+
+    return flask.send_file(public_key_file_location)
+
+
+@webapp.route("/mock/msgsigning/set-keypair", methods=["POST"])
+def set_keypair():
+    # TODO: Update keypair to use via a received JSON object -> {"public_key": "/path", "private_key": "/path"}
+    return flask.jsonify({"message": "Not yet implemented"}), 501
+
```
kpkeenan12, line 26: General question, would having an endpoint to dynamically alter the keypair mean a change in the [uas_standards](https://github.com/interuss/uas_standards) project? Something like `KeyPairRequest`?

BenjaminPelletier: We don't currently have a standard interface for the mock_uss -- for instance, the [SP](https://github.com/interuss/dss/blob/master/monitoring/mock_uss/ridsp/routes_behavior.py) and [DP](https://github.com/interuss/dss/blob/master/monitoring/mock_uss/riddp/routes_behavior.py) behavior endpoints are not listed in any interfaces since they're only used internally within InterUSS tools.  Since mock_uss is the only USS that would need to swap out certificates, I don't think we would need to communicate this interface outside the InterUSS domain (so, same category as those behavior endpoints linked above).  When a test scenario requires a [MockUSSResource](https://github.com/interuss/dss/blob/master/monitoring/uss_qualifier/resources/interuss/mock_uss.py), the implied contract in providing that resource is that the referenced server actually is a current instance of mock_uss, so we can assume we (uss_qualifier) can use any of the mock_uss endpoints there regardless of whether they're published in YAML somewhere.  So, all I think we need to do to support dynamic mock_uss keypairs is to add a simple endpoint that will look a lot like the behavior endpoints linked above.  uss_qualifier can then use that endpoint when it is provided a MockUSSResource.  [Small PRs are good](https://github.com/interuss/dss/blob/master/CONTRIBUTING.md#general-principles) so I'd generally recommend resisting the urge to tack on additional functionality to a PR (saving it for the next PR instead), but I don't think it would be completely unreasonable for this PR to include that functionality.  Again though, smaller PRs are better, so saving that until a follow-up PR would be my personal recommendation.

[The automated_testing_interfaces repository](https://github.com/interuss/automated_testing_interfaces) contains the interfaces InterUSS created that USSs need to implement in order to be tested in certain ways by InterUSS tools -- so, for instance, that's where the `scd.yaml` API that flight planner USSs need to implement is canonically located.  I don't think a USS under test would need to support the ability to dynamically alter its keypair to be effectively tested by InterUSS, so we probably wouldn't add an endpoint/capability to support that in `scd.yaml`, but of course we're open to arguments/explanations otherwise.

[The uas_standards repository](https://github.com/interuss/uas_standards) just contains handy tools for working with standard interfaces -- none of the interfaces are actually defined there.  The majority of library content in that repository is code auto-generated from the source interface.

monitoring/mock_uss/msgsigning/routes_msgsigning.py
```diff
@@ -0,0 +1,31 @@
...
+    public_key_file_location = os.path.join(
+        webapp.config.get(config.KEY_CERT_BASE_PATH), db.value.public_key_name
+    )
+
+    logger.info("Retrieving public key file from {}".format(public_key_file_location))
+
+    return flask.send_file(public_key_file_location)
+
+
+@webapp.route("/mock/msgsigning/set-keypair", methods=["POST"])
```
BenjaminPelletier, line 22: We generally want to follow the [RESTful design philosophy](https://www.redhat.com/en/topics/api/what-is-a-rest-api) and make endpoints resource-based rather than action-based.  In this case, I would expect the resource to be the keypair mock_uss is using, so I would expect the endpoint corresponding to that resource to be something like `/mock/msgsigning/keypair`.  POST is the verb usually used to create a new item in a collection, so I don't think it fits here (there is no collection and we're not creating a new item).  PUT is the verb to upsert a particular, specific resource.  There is no ID on this resource and that's unusual, but I think appropriate since mock_uss can only have one active keypair, but the thing we want to do here is to directly modify a specific resource (the single active keypair resource), so it seems like PUT would be the appropriate verb.

```suggestion
@webapp.route("/mock/msgsigning/keypair", methods=["PUT"])
```

BenjaminPelletier (commented):
This looks really good; just one comment on the new content from the last round of revisions.

monitoring/mock_uss/msgsigning/routes_msgsigning.py
kpkeenan12: Thanks for the explanation! Are we good to keep those stub endpoints, then in the future PR come back and actually implement? I can take out for now and just add all at once later, let me know what you'd prefer

kpkeenan12: Thanks for the explanation!

BenjaminPelletier:
The CI test failure for monitoring tests is not caused by this PR; I'll add an issue

kpkeenan12:
Would you be able to re-run the monitoring tests CI job? @BenjaminPelletier  

Checking the failures and I saw this 

`  details: 'Expected: ''2022-12-14T19:36:22.672220Z'', received: ''2022-12-14T19:36:22.67222Z'''
`

I don't think my last commit (addressing your comment to use `PUT` over `POST` should have affected any tests

BenjaminPelletier:
Added #928, kicked off new run of "monitoring tests" job.

BenjaminPelletier (approved):
Looks good to me; just pending CI success

PR #929, opened 2022-12-14
[monitorlib] Add message signing module

kpkeenan12:
Duplicate of #893, but within the NASA fork. 

Adds a `messagesigning` module to `monitorlib`. Will integrate into the project based off of this review [comment](https://github.com/interuss/dss/pull/922#discussion_r1044903204) in the design PR.


kpkeenan12:
Hey @barroco @BenjaminPelletier, wanted to get your general thoughts on the PR so far. In the original [implementation](https://github.com/nasa/dss/blob/uss-qual-test-ext-msg-sign/monitoring/monitorlib/infrastructure.py#L62), we went with updating the parent `AuthAdapter` class to add the message signing headers to all outgoing requests. The message signing code receives an `object_to_sign` (either `PreparedRequest` if its a request being fired off by the `uss_qualifier`, or a flask response object if it is a response being signed by the `mock_uss` -- although the signing of responses is not being included in this PR. 

From this [comment](https://github.com/interuss/dss/pull/922#discussion_r1044903204) in the design, you mentioned some changes that would be needed for `SignedRequest`. To wrap my head around these changes I have some questions:

`...and probably SignedRequest would then be modified to let the message signing functionality of the base AuthAdapter replace the custom logic currently in SignedRequest such that the auth spec`


The logic in the parent `AuthAdapter` will sign any `PreparedRequest` object, not just the one to a FIMS server. Do you just mean that we can strip the code in `SignedRequest` that builds the headers, and keep the code that builds the payload to the FIMS server? if the UFT object in `AuthAdapter` is set, those headers will get built automatically, so really all `SignedRequest` needs to do at this point is build the `(payload = "&".join([k + "=" + v for k, v in query.items()]))` and make the POST request to the FIMS server.

Please let me know if my overall understanding of your recommendations are correct or if there's some gaps. Thanks!

BenjaminPelletier (commented):
> The logic in the parent AuthAdapter will sign any PreparedRequest object, not just the one to a FIMS server. Do you just mean that we can strip the code in SignedRequest that builds the headers, and keep the code that builds the payload to the FIMS server? if the UFT object in AuthAdapter is set, those headers will get built automatically, so really all SignedRequest needs to do at this point is build the (payload = "&".join([k + "=" + v for k, v in query.items()])) and make the POST request to the FIMS server.

Yes, exactly -- `SignedRequest` will become a plain POST with no separate authentication from the message signature authentication provided by the base AuthAdapter.

Other comments below.

monitoring/monitorlib/messagesigning/message_signer.py
```diff
@@ -0,0 +1,117 @@
...
+from Crypto.PublicKey import RSA
+from Crypto.Signature import pkcs1_15
+from Crypto.Hash import SHA256
+from loguru import logger
+import urllib.parse
+from monitoring.monitorlib.messagesigning.hasher import get_content_digest
+
+
+def get_x_utm_jws_header(cert_url):
+    return '"alg"="{}", "typ"="{}", "kid"="{}", "x5u"="{}"'.format(
```
kmorrisCAL, line 12: I couldn't track down an official statement from the RFC but is wrapping the dictionary key in double-quotes proper/allowed?
None of the examples I've seen have quoted dictionary keys, and it breaks the http-sfv parsing library I'm using...

Should this line be:
```
return 'alg="{}", typ="{}", kid="{}", x5u="{}"'.format(
```

monitoring/monitorlib/messagesigning/message_signer.py
```diff
@@ -0,0 +1,117 @@
...
+        }
+    else:
+        base_value_map = {
+            "@status": object_to_sign.status_code,
+            "content-type": headers["content-type"],
+            "content-digest": "sha-512=:{}:".format(content_digest),
+            "x-utm-jws-header": get_x_utm_jws_header(),
+        }
+    curr_time = str(int(time.time()))
+    signature_param_str = '"{}": ({});{}'.format(
```
kmorrisCAL, line 106: This line is missing the `created` keyword:
```
signature_param_str = '"{}": ({});created={}'.format(
```

BenjaminPelletier:
As a FYI, we're tentatively looking to [split out monitoring from this repo](https://lists.interussplatform.org/g/tsc/topic/splitting_out_monitoring/94745227) starting next Wednesday, December 28.  It would be great if we could merge this PR by then, or otherwise it may need to be migrated separately to the other repo.

monitoring/monitorlib/messagesigning/message_signer.py
kpkeenan12: Good catch! Will make that change asap

kpkeenan12:
> As a FYI, we're tentatively looking to [split out monitoring from this repo](https://lists.interussplatform.org/g/tsc/topic/splitting_out_monitoring/94745227) starting next Wednesday, December 28. It would be great if we could merge this PR by then, or otherwise it may need to be migrated separately to the other repo.

That seems reasonable! I should be able to get to all the requested changes by then

monitoring/uss_qualifier/run_locally.sh
```diff
@@ -25,7 +25,7 @@ monitoring/build.sh || exit 1
 CONFIG_NAME="${1:-configurations.dev.local_test}"
 CONFIG_FLAG="--config ${CONFIG_NAME}"
 
-AUTH_SPEC='DummyOAuth(http://host.docker.internal:8085/token,uss_qualifier)'
+AUTH_SPEC='DummyOAuth(http://host.docker.internal:8085/token,uss_qualifier, message_signing=UFT(signing_key_path= /var/test-certs/messagesigning/mock_faa_priv.pem, cert_url = http://host.docker.internal:8077/mock/msgsigning/.well-known/uas-traffic-management/pub.der))'
```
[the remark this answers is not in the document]
kpkeenan12: It should be both. Mock USS needs the auth spec so it signs it's responses/validate incoming signed http requests, and uss_qualifier needs it to sign it's outgoing requests

monitoring/monitorlib/messagesigning/message_signer.py
```diff
@@ -0,0 +1,117 @@
...
+    with open(private_key_path, "rb") as priv_key_file:
+        private_key = RSA.import_key(priv_key_file.read())
+    return (
+        base64.b64encode(pkcs1_15.new(private_key).sign(hash)).decode("utf-8"),
+        sig_input,
+    )
+
+
+def get_key_id():
+    return "mock_uss_priv_key"
```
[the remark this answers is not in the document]
kpkeenan12: Since there's no logic in this function, I'm putting the hardcoded string in the `get_x_utm_jws_header` function itself. Removing this function altogether

monitoring/monitorlib/messagesigning/message_signer.py
```diff
@@ -0,0 +1,117 @@
...
+import time
+from Crypto.PublicKey import RSA
+from Crypto.Signature import pkcs1_15
+from Crypto.Hash import SHA256
+from loguru import logger
+import urllib.parse
+from monitoring.monitorlib.messagesigning.hasher import get_content_digest
+
+
+def get_x_utm_jws_header(cert_url):
```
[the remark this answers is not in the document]
kpkeenan12: extracted this function into `uft.py` since its not defined in IETF. Okay to keep the name as is? Everything else in this file should be in IETF.

monitoring/uss_qualifier/run_locally.sh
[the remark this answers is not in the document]
BenjaminPelletier: Hmm, I think we're going to need some pretty articulate documentation about this.

Message signing architecture diagram

The architecture diagram doesn't show any connections from uss_qualifier that need message signing.  Logging control/retrieval with mock_uss is performed via a custom interface that is internal to InterUSS.  Instructions to inject operations is via the InterUSS flight planning/SCD automated testing interface which doesn't need message signing as it's not part of the UTM ecosystem under test.

Thinking harder about what this suite will do, however, I realize that part of the F3548 suite involves uss_qualifier acting as a USS to double-check the operational intent a USS claims to have created to serve a user flight intent requested through the flight planning/SCD automated testing interface.  This means uss_qualifier will call a USS's operational intent details endpoint directly, and that endpoint is within the ASTM standard, thus part of the UTM ecosystem under test, thus possibly in scope for message signing.  I think this introduces some issues.

Currently, uss_qualifier does not host any endpoints -- it makes outgoing calls only, on behalf of the USS running it.  By adding this line, we're effectively saying that the port-8077 mock USS is an essential component of the USS running uss_qualifier.  This is the first time we've made that assumption, I don't think it's documented anywhere, and it's surprising.  There are a number of implications: first, the port-8077 mock USS must have the same participant ID as the USS running uss_qualifier (because they're components of the same USS).  In a real deployment, I don't think this would be the case since a real USS running uss_qualifier would host their cert somewhere other than a mock USS instance, so we can't programmatically check that the participant IDs match.  Instead, we'd need to write some very clear documentation describing the situation so users can understand what is being assumed and what is required.  Second, we shouldn't even use the port-8077 mock_uss's cert_url as the cert_url for the USS running uss_qualifier even in local/dev tests because we're eventually going to [intentionally mangle](https://github.com/interuss/dss/tree/master/monitoring/uss_qualifier/suites/faa/uft/design#negative-tests--) the certs of the port-8077 mock_uss as part of the test.  The USS running uss_qualifier should always work properly throughout the test, so we shouldn't/can't mangle their certs if they need to use message signing.

This is a challenging problem.  I'd suggest that the simplest solution is probably to just not have uss_qualifier do message signing.  This will work with UFT because the message signing requirements have been relaxed to accept and log messages with no signatures to accommodate test tools, and uss_qualifier is indeed a test tool.  If that solution isn't acceptable, I think we'll need to update the design doc to think about this issue more carefully.  I suspect the solution will probably be that message signing will not be able to use the existing F3548 suite as currently envisioned.  The most obvious place where this won't work is in the negative cases -- behavior expectations are different between happy-path-F3548 and negative-case-message-signing.  When port-8077 mock_uss is set to have mangled message signing certs, the test should pass if the intent to create a flight fails (since message signing won't work properly) whereas in plain F3548, the test should pass if the intent to create a flight succeeds.  In that case, we'll simply need separate test scenarios for message signing, and in that case, there would be no need to check the validity of operational intent details since there's no message signing requirement that check would cause to fail, therefore uss_qualifier would not need to sign its messages.  I think this matches the original high-level vision for message signing testing which was that mock_uss is the tooling which manipulates and measures message signing capabilities (rather than uss_qualifier).

monitoring/monitorlib/messagesigning/message_signer.py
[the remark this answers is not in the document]
BenjaminPelletier: My personal preference would be to name it something like ietf.py to better indicate its content scope (to me, message_signer.py implies I should expect to find a class named MessageSigner in the module), but message_signer.py isn't unreasonable to contain generic/standard message signing tools.  I'd also personally prefer to name the action group rather than the actor as we don't have any other files I'm thinking of that name the actor (e.g., we have testing.py rather than tester.py, versioning.py rather than versioner.py or version_checker.py, auth.py rather than authorizer.py, fetch rather than fetcher, inspection.py rather than inspector.py, etc)

monitoring/monitorlib/messagesigning/message_signer.py
```diff
@@ -0,0 +1,104 @@
...
+    sig, sig_input = _get_signature(object_to_sign, signed_type, private_key_path)
+    content_digest = (
+        get_content_digest(object_to_sign.body)
+        if is_signing_request
+        else get_content_digest(object_to_sign.get_data())
+    )
+    signed_headers = {
+        "x-utm-message-signature-input": "utm-message-signature={}".format(sig_input),
+        "x-utm-message-signature": "utm-message-signature=:{}:".format(sig),
+        "x-utm-jws-header": get_x_utm_jws_header(cert_url),
```
BenjaminPelletier, line 23: Seems like this should be in uft.py since x-utm-jws-header isn't specified by IETF.

Basically, whenever I see "x-utm", I would expect that is a UFT-specific thing that shouldn't be in a generic message signing library -- the generic/IETF thing would accept some kind of parameter/argument, and the UFT-specific library would supply "x-utm*" for those arguments.

monitoring/monitorlib/auth.py
```diff
@@ -313,75 +313,6 @@ def issue_token(self, intended_audience: str, scopes: List[str]) -> str:
         }
         payload = "&".join([k + "=" + v for k, v in query.items()])
 
-        # Generate signature
```
BenjaminPelletier, line 316: It isn't valid to simply remove this block of code -- as currently proposed:

1. we're left with a number of SignedRequest class parameters that don't do anything and therefore also have incorrect documentation (e.g., signature_style)
2. message_signing isn't added as a parameter to the SignedRequest constructor which would pass it to the AuthAdapter parent class (or the existing parameters used to create an appropriate message_signing argument)
3. SignedRequest is no longer a valid class name for what this class does
4. dead signing code is not removed
5. there is no graceful deprecation of the now-lost capability to sign token requests in the UPP2 format, nor is there a comment in the PR that this capability is being removed

I'd recommend against trying combine SignedRequest signing removal into this PR as it's a non-trivial amount of effort and not integral to what the rest of the PR is trying to accomplish.

monitoring/monitorlib/messagesigning/message_signer.py
```diff
@@ -0,0 +1,104 @@
...
+    )
+    sig_base = ""
+    for component in covered_components:
+        sig_base += '"{}": {}\n'.format(component, base_value_map[component])
+    sig_base += signature_param_str
+    return sig_base
+
+
+def _wrap_components_in_quotes(components):
+    return " ".join(list(map(lambda comp: '"{}"'.format(comp), components)))
```
BenjaminPelletier, line 104: nit:

```suggestion
    return " ".join(f'"{comp}"' for comp in components)
```

monitoring/monitorlib/messagesigning/uft.py
```diff
@@ -0,0 +1,7 @@
+def get_x_utm_jws_header(cert_url):
+    return 'alg="{}", typ="{}", kid="{}", x5u="{}"'.format(
```
BenjaminPelletier, line 2: It seems odd to fill format string placeholders with constants; why not simply:

```suggestion
    return 'alg="RS256", typ="JOSE", kid="mock_uss_priv_key", x5u="{}"'.format(cert_url)
```

monitoring/monitorlib/messagesigning/uft.py
```diff
@@ -0,0 +1,7 @@
+def get_x_utm_jws_header(cert_url):
```
BenjaminPelletier, line 1: It seems like we know a USS will need to specify their `kid`; why not accept that as a parameter now?

monitoring/monitorlib/messagesigning/message_signer.py
```diff
@@ -0,0 +1,104 @@
...
+            "content-type": headers.get("content-type", ""),
+            "content-digest": "sha-512=:{}:".format(content_digest),
+            "x-utm-jws-header": get_x_utm_jws_header(cert_url),
+        }
+    else:
+        base_value_map = {
+            "@status": object_to_sign.status_code,
+            "content-type": headers["content-type"],
+            "content-digest": "sha-512=:{}:".format(content_digest),
+            "x-utm-jws-header": get_x_utm_jws_header(),
```
BenjaminPelletier, line 90: Missing cert_url

monitoring/monitorlib/messagesigning/message_signer.py
BenjaminPelletier, line 23: It's not wrong but seems a little weird to compute the x-utm-jws-header twice for each signature -- redundant calculations often introduce the possibility of inconsistency.  Why not compute the header once and pass it to the functions that need it (`_get_signature` here because it needs to provide it to `_get_signature_base`), as that would have the added benefit of making the low-level signature computation more agnostic of specific conventions chosen by UFT?

kpkeenan12:
Hey @BenjaminPelletier, due to some time constraints on another project I won't be able to finish this PR by the 12/28 deadline. I will still try and respond to  your review comments on this PR, but most likely will implement the changes on the new monitoring repo, when it becomes available. If you want, you can close this PR and we can use it as a reference point later on.

PR #931, opened 2022-12-19
[docs] Add reference to Python style guide

BenjaminPelletier:
In addition to adding the style guide, this PR also explicitly pins the version of the Black formatter for Python as initially this PR (which changes one line of a .md file) was failing Python formatting checks because the `latest_version` tag of `pyfound/black` was rev'd to a version that detects problems with the current repo content.  We should explicitly move between `pyfound/black` versions in the future.  Added #932 to track this.


barroco (approved):
👍

Issue #925, opened 2022-12-13
Add atproxy to local deployment + local uss_qualifier configuration


BenjaminPelletier:
Currently, atproxy does not benefit from being exercised in automated continuous integration tests.  We should:

1) Add a mock_uss capability named `atproxy_client` or similar that (when enabled) causes mock_uss to poll an atproxy server for requests, handle them, and POST back to the server with responses

2) Add another mock_uss container deployed locally with `make start-uss-mocks` that enables this functionality

3) Include atproxy as a USS under test in the configurations.dev.local_test uss_qualifier configuration

PR #933, opened 2022-12-21
[atproxy] Prepare atproxy for mock_uss backend implementation client

BenjaminPelletier:
As part of #925, atproxy needs to be restructured a bit to allow mock_uss to act as a backend implementation client.  This PR performs this restructuring by:

1. Separating atproxy library code from the actual webapp by putting the webapp in a subfolder; this allows the library code to be imported by another Flask application without mangling the Flask app singletons (boo singletons)
2. Moving data object definitions out of the handler modules so the data objects can be imported by a different Flask application without adding handler endpoints to that Flask application
3. Replacing magic strings indicating handler endpoints with an enum to avoid propagating magic strings in a backend implementation client


barroco (approved):
👍

PR #944, opened 2023-02-15
[tanka] Add support for private registry

barroco:
This PR updates the tanka manifests to support for private registry.


barroco:
Successfully tested with a personal private repository on docker hub.

build/deploy/core-service.libsonnet
```diff
@@ -100,6 +100,7 @@ local ingress(metadata) = base.Ingress(metadata, 'https-ingress') {
                 },
               },
             },
+            imagePullSecrets: if metadata.image_pull_secret != null then [{name: metadata.image_pull_secret}] else [],
```
marcadamsge, line 103: I might be a bit rusty with tanka, but shouldn't you put this line in the definitions in `base.libjsonnet`? The `base.libjsonnet` definitions also have access to the metadata. This way that configuration option would apply to all deployments in the project. Also what about other manifests that might need this setting, like `StatefulSet`s?

build/deploy/schema-manager.libsonnet
```diff
@@ -28,6 +28,7 @@ local schema_dir = '/db-schemas';
               },
               volumeMounts: volumes.mounts.caCert + volumes.mounts.clientCert,
             },
+            imagePullSecrets: if metadata.image_pull_secret != null then [{name: metadata.image_pull_secret}] else [],
```
marcadamsge, line 31: Same here, I think that if you put this line in `base.Job` it would apply to all jobs in the project

build/README.md
```diff
@@ -323,6 +327,20 @@ a PR to that effect would be greatly appreciated.
 
         -   Note that `VAR_DOCKER_IMAGE_NAME` is used in two places.
 
+    1.  `VAR_DOCKER_IMAGE_PULL_SECRET`: Secret name of the credentials to access
```
BenjaminPelletier, line 330: What should this value be if no credentials are needed to access the image registry (default situation)?  It looks like we need instructions regarding uncommenting the line or not and replacing the value or not.

build/deploy/core-service.libsonnet
barroco: Good point, the initial intention was to limit the use of credentials to the image built in this repository but it seems that it can be passed to all resources pulling images even if it is using a different registry. I will update it.

build/deploy/schema-manager.libsonnet
barroco: I will update the PodSpec which should apply to Job, Deployment and Statefulset resources.

build/README.md
barroco: Ben, please have a look to the latest change to see if it meets your expectation

barroco:
Successful deployment with 6412918 with `terraform-google-dss` with and without image_pull_secret set.

marcadamsge (approved):
LGTM

build/README.md
BenjaminPelletier: The baseline instruction is "replace all `VAR_*` instances with appropriate values", so I think we should say something like "If the image specified in `VAR_DOCKER_IMAGE_NAME` does not require authentication, then do not populate this instance and do not uncomment the line containing it"

BenjaminPelletier (approved):

PR #949, opened 2023-03-30
[tf-aws] Add missing image_pull_secret variable

barroco:
This PR adds a missing variable flag to `terraform-aws-dss`.
The variable flag should have been added as part of #944


BenjaminPelletier (approved):
Sorry, this fell off my todo list

Issue #874, opened 2022-10-28
Simplify onboarding new participants, including for more major cloud providers


barroco:
The instructions for bringing up a DSS instance are pretty actionable (complete, clear) currently, but they’re very long and require a fair amount of engineering expertise.  We have a tool under development called deployment_manager which should simplify this process substantially, and therefore make deployment of a DSS instance easier

Deployment instructions: https://github.com/interuss/dss/tree/master/build
Deployment tool: https://github.com/interuss/dss/tree/master/monitoring/deployment_manager

barroco:
Following discussions with @BenjaminPelletier @BradNicolle and @marcadamsge, the plan to update the DSS deployment approach to support other cloud providers and keep it manageable for InterUSS over time.

## Background
The deployment of the DSS is currently mostly documented on a README. Kubernetes (K8s) deployment instructions only cover GKE. Tanka is used to generate and configure kubernetes resources. In addition, the DSS codebase is being refactored to require only one container instead of two currently. Most of the complexity lies in getting a kubernetes cluster running for cockroach-db ready to be pooled and the pooling steps. We have undertaken the process of extracting self-contained modules to separate repositories. Finally, we are starting the work to support other cloud providers.

### Default DSS infrastructure
The DSS is composed of two different services to run (we assume the http-gateway and core-service application as one since refactoring is under way): the DSS API and the cockroach database. 
In addition, the current configuration proposes the following supporting services:
 - Istio: service mesh
 - Prometheus & Grafana

The requirements for the InterUSS standard deployment of the DSS in terms of infrastructure are:
- A Kubernetes cluster with 3 nodes
- 4 ingresses with individual static ips with DNS associated (1 for the gateway, 3 for cockroach db instances)

## Objectives and change plan overview

### 1. Infrastructure as code
Conceptually, the deployment will be broken down in three main categories:

**Infrastructure**: It is responsible for the cloud resources required to run the DSS services. It includes the kubernetes cluster creation, cluster nodes, load balancer and associated fixed IPs, etc. This stage is cloud provider specific. The objective is to support Amazon Web Services (EKS), Azure (AKS), Google (GKE). 
To manage multi-cloud resources, we propose to use terraform providers [C.1]. Using terraform providers will offer the following benefits of infrastructure as code:
Limit the number of untested command line steps in the READMEs. 
Allow users to keep track of the infrastructure lifecycle and run simple upgrades.
Practice multi-cloud deployment as part of the CI/CD.

[...]

PR #950, opened 2023-04-26
[helm] Initial Helm chart

barroco:
Progressing on the objective #874, this PR adds the ability to deploy the DSS using helm to a Kubernetes cluster.
It was tested on AWS and Google deployment using the terraform scripts to deploy the infrastructure.
A future PR will update the terraform-commons-dss to automatically deploy the DSS using this helm chart.

Few remarks:
1. Since this helm chart primary purpose is to be instrumented by terraform, documentation has been kept minimal. Users will be expected to get the instructions from there. (This may be reviewed based on user feedback)
2. initContainers have been added to the workloads to better handle the starting sequence of pods:
    1. cockroachdb
    2. schema-manager
    3. dss-core-service
3. Not all configuration flags have been added yet to the manifest compared to tanka. It will be added if necessary incrementally in upcoming PRs and based on the terraform configuration.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -1,7 +1,5 @@
 # DSS Deployment
 
-**Work in progress**
-
 This folder contains the increments toward the new deployment approach as described in [#874](https://github.com/interuss/dss/issues/874).
 
 The infrastructure folder contains the terraform modules to deploy the DSS to kubernetes clusters of various cloud providers:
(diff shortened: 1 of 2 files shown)


marcadamsge (commented):
Pretty advanced stuff 👍 I was surprised at first that your `values.schema.json` also includes values of the cockroachDB helm chart, but then I realized that you also use some of these values. I tried rendering that helm chart locally, I didn't try deploying it though, but it LGTM

BenjaminPelletier (approved):
Looks good to me, though I am probably not the most ideal technical reviewer.  I have some comments regarding the documentation, but your discretion whether they are addressed for this PR or in a separate PR.

PR #958, opened 2023-06-21
Use docker compose v2 instead of docker-compose v1

palyca:
These updates for the docker compose updates to the container naming convention changes in Docker V2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,7 +140,7 @@ probe-locally:
 
 .PHONY: collect-local-logs
 collect-local-logs:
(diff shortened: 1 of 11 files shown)


BenjaminPelletier (commented):
This looks good, but let's also update [standalone_instance.md](https://github.com/interuss/dss/blob/master/build/dev/standalone_instance.md) to reflect the use of docker compose v2 as well -- specifically, it shouldn't mention `docker-compose` (instead `docker compose`), and installing docker-compose separately is no longer a prereq (it comes as part of docker, I think)

BenjaminPelletier (approved):

PR #985, opened 2024-01-11
[scd] return proper http code for too big area search on the SCD subscriptions endpoint

Shastick:
While writing a scenario for the `uss_qualifier` I ran into a situation where the DSS does not use the proper HTTP code when the requested are is too large for subscription queries in the SCD context.

As per [specification](https://github.com/astm-utm/Protocol/blob/v1.0.0/utm.yaml#L2936), in this situation, we should receive a 413 code, and not the observed 500 one.

The location in the code was found through the following DSS logs:

```
2024-01-11T13:51:43.660Z	error	errors/errors.go:81	Error during unary server call	{"error_id": "E:7998749e-a7f4-4a2b-a39c-46d7dac2e222", "stacktrace": "Got an unexpected error\n --- at /app/pkg/scd/subscriptions_handler.go:443 (Server.QuerySubscriptions) ---\nCaused by: Error searching Subscriptions in repo\n --- at /app/pkg/scd/subscriptions_handler.go:414 (Server.QuerySubscriptions.func1) ---\nCaused by: Could not calculate spatial covering\n --- at /app/pkg/scd/store/cockroach/subscriptions.go:313 (repo.SearchSubscriptions) ---\nCaused by: Area is too large (510062190.739560km² > 2500.000000km²)\n --- at /app/pkg/geo/s2.go:177 (Covering) ---\nCaused by: Area too large\n --- at /app/pkg/geo/errors.go:31 (init) ---", "error": "Area too large", "code": 0}
```

The proposed fix solves the issue: when running against a local build of the DSS, I observed the proper HTTP code being returned when the requested search area was too big.

### DIFF
--- a/pkg/scd/subscriptions_handler.go
+++ b/pkg/scd/subscriptions_handler.go
@@ -439,8 +439,20 @@ func (a *Server) QuerySubscriptions(ctx context.Context, req *restapi.QuerySubsc
 
 	err = a.Store.Transact(ctx, action)
 	if err != nil {
-		return restapi.QuerySubscriptionsResponseSet{Response500: &api.InternalServerErrorBody{
-			ErrorMessage: *dsserr.Handle(ctx, stacktrace.Propagate(err, "Got an unexpected error"))}}
+
+		errResp := &restapi.ErrorResponse{Message: dsserr.Handle(ctx, err)}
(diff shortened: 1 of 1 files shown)


BenjaminPelletier (approved):
Awesome, thanks :)

PR #990, opened 2024-01-24
[infrastructure] Remove useless variables

barroco:
This PR removes two variables which should have been removed previously since they are not publicly exposed.

They are actually managed internally [here](https://github.com/interuss/dss/blob/master/deploy/infrastructure/dependencies/terraform-commons-dss/variables_internal.tf).


marcadamsge (approved):
LGTM from what I understood from the structure, I'm not super familiar though

BenjaminPelletier (approved):

PR #996, opened 2024-02-01
[apis] models.IDFromString strictly check for RFC 4122 variant of the V4  UUID format

Shastick:
The OpenAPI specification for the [SCD endpoints](https://github.com/astm-utm/Protocol/blob/v1.0.0/utm.yaml#L120) specifies that we should use a UUID V4 of the RFC4122 variant. It provides a regexp(*) for validating the UUIDs in an unambiguous way.

However, In the context of SCD, the current implementation allows the use of IDs that don't conform to the V4-RFC4122 variant, such as `00000179-e36d-40be-d38b-eca6ca350000`

### Problem

This might be problematic if other implementations chose to enforce this more strictly:
 - a client may successfully create an entity on this DSS implementation
 - a client will then expect other DSS's for the relevant area to reply to requests, which said DSS's might reject because the requested IDs are not valid
 
More generally, a request for creating an entity should either be valid for all DSSes or invalid for all DSSes (at least, that seems like a reasonable requirement)

### Approach

A simple solution is to add two additional checks for UUIDs: one for the version, and the other for the variant.

This is a simple fix, which is provided as a first version of this PR: this does not seem to break any existing test, incl. the `prober`.

**However:**

The OpenAPI specs for the [RID v2_1 requirement](https://github.com/uastech/standards/blob/ab6037442d97e868183ed60d35dab4954d9f15d0/remoteid/canonical.yaml#L1038) side of the DSS requires a UUID V4 without specifying a variant. Note that the given example happens to be a RFC4122 variant, but there is no mention of RFC4122 in that file.

## **Chose solution**

 - amend the RID specification to more explicitly mandate a UUID V4 of the RFC4122 variant and apply that requirement to all UUID's in the DSS.
 - ~~separate the validation logic for UUIDs depending on if they are used in the RID or SCD contexts~~
 
As indicated by @nasajoey, the requirement for a v4 UUID implicitly means that it needs to respect RFC4122, as no other RFC mentions such an UUID. It is thus acceptable to expect _all_ endpoints (both RID v1/v2 and SCD) to only accept such identifiers. 
 

(*): `^[0-9a-fA-F]{8}\\-[0-9a-fA-F]{4}\\-4[0-9a-fA-F]{3}\\-[8-b][0-9a-fA-F]{3}\\-[0-9a-fA-F]{12}$`

### DIFF
--- a/pkg/models/models.go
+++ b/pkg/models/models.go
@@ -63,11 +63,28 @@ func (manager Manager) String() string {
 	return string(manager)
 }
 
+// IDFromString converts a string to an ID, typically provided by an API client.
+// The OpenAPI spec for the SCD endpoints defines precisely the regex that UUIDs must conform to:
+//
+//	^[0-9a-fA-F]{8}\\-[0-9a-fA-F]{4}\\-4[0-9a-fA-F]{3}\\-[8-b][0-9a-fA-F]{3}\\-[0-9a-fA-F]{12}$
+//
+// This is the strictest possible interpretation for a "V4 UUID RFC 4122 Variant" (See RFC 4122, section-4.4) and means we must refuse
+// even UUIDs that have the correct format but are not technically V4 UUID as described RFC 4122 variant (notably the fixed bits)
+//
+// RFC 4122 is the only RFC that describes a "version 4" UUID: the format requirement using the above regex is
+// thus also extended to the RID v1 and v2 endpoints, which explicitly mandate that the UUIDs to be used are the V4 one.
+//
+// See https://github.com/astm-utm/Protocol/blob/cb7cf962d3a0c01b5ab12502f5f54789624977bf/utm.yaml#L128 for more details
 func IDFromString(s string) (ID, error) {
 	id, err := uuid.Parse(s)
 	if err != nil {
-		return ID(""), stacktrace.Propagate(err, "Error parsing ID in UUID format")
+		return ID(""), stacktrace.Propagate(err, "Error parsing ID in UUID V4 format: `%s`", s)
 	}
+
+	if id.Variant() != uuid.RFC4122 || id.Version() != 4 {
+		return ID(""), stacktrace.NewError("UUID must be V4 as per RFC4122, was: `%v`, id `%s`", id.Variant(), s)
+	}
+
 	return ID(id.String()), nil
 }
 
--- a/pkg/models/models_test.go
+++ b/pkg/models/models_test.go
@@ -1,0 +1,84 @@
+package models
+
+import "testing"
+
+func TestIDFromString(t *testing.T) {
+	type args struct {
+		s string
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    ID
+		wantErr bool
+	}{
+		{
+			name: "compliant 1",
+			args: args{
+				s: "00000179-e36d-40be-838b-eca6ca350000",
+			},
+			want:    ID("00000179-e36d-40be-838b-eca6ca350000"),
+			wantErr: false,
+		},
+		{
+			name: "compliant 2",
+			args: args{
+				s: "03e5572a-f733-49af-bc14-8a18bd53ee39",
+			},
+			want:    ID("03e5572a-f733-49af-bc14-8a18bd53ee39"),
+			wantErr: false,
+		},
+		{
+			name: "wrong version",
+			args: args{
+				s: "00000179-e36d-50be-838b-eca6ca350000",
+			},
+			want:    ID(""),
+			wantErr: true,
+		},
+		{
+			name: "wrong variant",
+			args: args{
+				s: "00000179-e36d-d0be-d38b-eca6ca350000",
+			},
+			want:    ID(""),
+			wantErr: true,
+		},
+		{
+			name: "random",
+			args: args{
+				s: "sdh2o4mxc912asdvt34123dsfg2123dcsasefvaqyb4bp963r3",
+			},
+			want:    ID(""),
+			wantErr: true,
+		},
+		{
+			name: "missing leading char",
+			args: args{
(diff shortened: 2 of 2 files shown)


pkg/scd/subscriptions_handler.go
```diff
@@ -305,7 +305,7 @@ func (a *Server) GetSubscription(ctx context.Context, req *restapi.GetSubscripti
 	id, err := dssmodels.IDFromString(string(req.Subscriptionid))
 	if err != nil {
 		return restapi.GetSubscriptionResponseSet{Response400: &restapi.ErrorResponse{
-			Message: dsserr.Handle(ctx, stacktrace.NewErrorWithCode(dsserr.BadRequest, "Invalid ID format: `%s`", req.Subscriptionid))}}
+			Message: dsserr.Handle(ctx, stacktrace.Propagate(err, "Invalid ID format: `%s`", req.Subscriptionid))}}
```
Shastick, line 308: Note: this change is not strictly required but I used it to get feedback on the cause of the failure to the caller: we may discuss if we should check and adapt other places that would have hidden the error.

nasajoey:
Thanks for detailing this issue!  One point to consider... the RID API that you link to has the following:

```
    UUIDv4:
      description: UUID v4.
      maxLength: 36
      minLength: 36
      type: string
      example: 03e5572a-f733-49af-bc14-8a18bd53ee39
```

It looks like this was thankfully taken from early USS specifications and at least included the "v4" information in naming.  "v4" means nothing if the variant (or type) of UUID is Microsoft. One might be able to argue that the requirement is the same between services (SCD and RID), it's just that the RID one doesn't have a regex.  If someone provides a Microsoft GUID variant UUID, then technically they are not providing a "UUIDv4". 

So I'd lean toward your first option, but re-emphasize (note the added **MORE**) that the requirement is already there:

> amend the RID specification to **MORE** explicitly mandate a UUID V4 of the RFC4122 variant and apply that requirement to all UUID's in the DSS.

I understand updating the RID API may be non-trivial in terms of process, but it feels like the right path.

BenjaminPelletier:
> amend the RID specification to more explicitly mandate a UUID V4 of the RFC4122 variant and apply that requirement to all UUID's in the DSS.

The RID specification is defined in F3411, so this kind of change would be done through the ASTM standards development process (outside InterUSS) and the average lead time for that would be roughly 1-3 years :)

> separate the validation logic for UUIDs depending on if they are used in the RID or SCD contexts

Let's do that :) (to the extent separate validation logic is necessary; see Joey's thread)

BenjaminPelletier:
I think I agree with Joey.  "UUID v4" presumably communicates some important information in the RID specification.  The only UUID standard I'm aware of that includes something related to "v4" is [RFC4122](https://datatracker.ietf.org/doc/html/rfc4122).  RFC4122's v4 only applies when the [variant](https://datatracker.ietf.org/doc/html/rfc4122#section-4.1.1) is 10x, so any variant other than 10x should be invalid.  Then, within that variant, the [version](https://datatracker.ietf.org/doc/html/rfc4122#section-4.1.3) must be 4.

pkg/models/models.go
```diff
@@ -63,10 +63,27 @@ func (manager Manager) String() string {
...
 	if err != nil {
-		return ID(""), stacktrace.Propagate(err, "Error parsing ID in UUID format")
+		return ID(""), stacktrace.Propagate(err, "Error parsing ID in UUID V4 format: `%s`", s)
+	}
+	// TODO check the other required fields: the UUID lib reports the value of the version field
+	// but does not validate that the other V4 specific bits are correct.
+	if id.Version() != 4 {
+		return ID(""), stacktrace.NewError("UUID must be a V4 version, was: `%v`, id: `%s`", id.Version(), s)
+	}
+	if id.Variant() != uuid.RFC4122 {
```
BenjaminPelletier, line 85: nit: Version only makes sense if the Variant is RFC4122, right?  If so, this check should probably be performed first.

pkg/scd/subscriptions_handler.go
BenjaminPelletier: Propagating is generally preferred to making a new error, but NewErrorWithCode is where we apply the error type (e.g., BadRequest).  That's not necessarily the best approach, but regarding this change in isolation, are we sure that all `err`s we may propagate here will originate with `dsserr.BadRequest`, or are we sure we will always indicate BadRequest further up the stack?

pkg/models/models.go
nasajoey: Agreed.  In RFC they note that "Variant" is really a "Type" and they discuss it prior to version for that reason... so if it isn't RFC4122, then basically bets are off in terms of the rest of the formatting.

pkg/scd/subscriptions_handler.go
Shastick: I'll revert this:
 - the main purpose was for troubleshooting and confirmation that the fix works as expected from the point of view of the `uss_qualifier`.
 - It would be hard to guarantee that the error type would be correct in all situations, or it would require more work that does not fall within the current scope and priorities.

Shastick:
PR has been updated:
- the parsing method now has the rationale for the strict requirement explained in a comment.
- error handling on the endpoint is reverted to its original

Please let me know if I need to update some documentation somewhere else. Note that the formatting requirement will soon be part of the `uss_qualifier`, so all implementations that don't respect the formatting will quickly be made aware of this.

BenjaminPelletier (approved):

BenjaminPelletier:
I don't think we need to update documentation anywhere -- the DSS intent remains to implement the requirements of the standards, including the APIs.  This PR was simply bringing us in line with that intent since the APIs do actually require the formatting that this PR ensures.

PR #999, opened 2024-02-07
[security] Update profiler to 0.4.0

BenjaminPelletier:
This is nominally a one-line PR that changes the profiler dependency from 0.2.0 to 0.4.0.  The other changes are due to running `go mod tiy -compat=1.17`.


BenjaminPelletier:
I'm not sure how to effectively test this change, so if reviewer(s) could consider carefully, that would be appreciated.

mickmis (approved):
(reviewing upon request of @barroco)

Given that we can't test automatically, we can look at the [changelog of the library](https://github.com/googleapis/google-cloud-go/blob/main/profiler/CHANGES.md). My impression is that:
- there is no breaking change
- there are a few bug fixes that don't look like could impact us
- mostly it is updating dependencies, the most notable one being an upgrade to protobuf-v2
Only the dependencies update could maybe have some red flags.
Now, IIUC with this library the only thing we do is that we spawn a routine to generate and upload profiles, with a behavior that seems to be determined by the server. Given that, as long as the server used supports the new version this upgrade should be OK. Which since this is a hosted cloud service it should.
And also if that change break something, that should be limited to this profiling feature and should in theory not have a broader impact. 

Only alternative would be to test manually, but it looks like this service is not enabled in any of the example deployment files:
https://github.com/interuss/dss/blob/7579299b9ae50f9ac505292d48ece2f40501aa84/deploy/services/helm-charts/dss/templates/dss-core-service.yaml#L47
https://github.com/interuss/dss/blob/7579299b9ae50f9ac505292d48ece2f40501aa84/build/deploy/metadata_base.libsonnet#L30

IMO it is low-risk enough to go ahead with the change.

BenjaminPelletier:
Per meeting yesterday, my understanding is that we believe the concurrency latency issue is separate and should not block this PR.

PR #1000, opened 2024-02-08
[terraform-google][tanka] Set SSL policy to enforce secure TLS connections

barroco:
This PR adds to tanka deployment the ability to set a SSL policy to the DSS Gateway ingress. Creation of the SSL Policy with terraform is addressed in #997.

When VAR_SSL_POLICY is provided, the following expected lines are added to the deployment compared to the original one:

```
>     networking.gke.io/v1beta1.FrontendConfig: ssl-frontend-config
...
> ---
> apiVersion: networking.gke.io/v1beta1
> kind: FrontendConfig
> metadata:
>   labels:
>     name: ssl-frontend-config
>   name: ssl-frontend-config
>   namespace: default
> spec:
>   sslPolicy: dss-ci-google-ue4-secure-ssl-policy
```

Tested on google cloud: https://dss.ci.google-interuss.uspace.dev/
Load balancer is correctly set with the provided ssl policy.

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -380,6 +380,12 @@ a PR to that effect would be greatly appreciated.
 
     -   If you are only turning up a single DSS instance for development, you
         may optionally change `single_cluster` to `true`.
+    
+    1.  `VAR_SSL_POLICY`: When deploying on Google Cloud, a [ssl policy](https://cloud.google.com/load-balancing/docs/ssl-policies-concepts) 
+        can be applied to the DSS Ingress. This can be used to secure the TLS connection.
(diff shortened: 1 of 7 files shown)


BenjaminPelletier (approved):
Looks good to me.  I've asked to see if @BradNicolle has a chance to take a look, but if not we can merge and make a follow-up PR to address any issues.

BradNicolle:
> Looks good to me. I've asked to see if @BradNicolle has a chance to take a look, but if not we can merge and make a follow-up PR to address any issues.

All looks good to me - I've patched in the Tanka changes and verified that the load balancer TLS version and ciphers match the policy.

Issue #1002, opened 2024-02-15
SCD Traffic concurrent prober check fails on cloud deployment


barroco:
**Describe the bug**
The prober fails when testing a cloud deployment. Failure in appended at the end of this report.

Problem has been identified when deploying the DSS on GCP using helm.

**To Reproduce**
- Deploy a Kubernetes cluster on GCP using terraform
- Deploy the DSS using Helm
- Run the oauth dummy server locally
- Run the prober using `interuss/monitoring:v0.3.0`

Note that the test has been conducted from a local machine (based in Switzerland) and the cluster is deployed in US east region.

**Stacktrace**
```
============================= test session starts ==============================
platform linux -- Python 3.11.7, pytest-6.2.4, py-1.11.0, pluggy-0.13.1
rootdir: /app/monitoring/prober
plugins: Faker-8.1.0, mock-3.6.1
collected 315 items

aux_/test_token_validation.py ...                                        [  0%]
aux_/test_version.py .                                                   [  1%]
monitorlib/test_geo.py .                                                 [  1%]
rid/v1/test_isa_expiry.py .......                                        [  3%]
rid/v1/test_isa_simple.py ...................                            [  9%]
rid/v1/test_isa_simple_heavy_traffic_concurrent.py .....                 [ 11%]
rid/v1/test_isa_validation.py .........                                  [ 14%]
rid/v1/test_subscription_isa_interactions.py ......                      [ 16%]
rid/v1/test_subscription_isa_slightly_overlapping.py .....               [ 17%]
rid/v1/test_subscription_simple.py ............                          [ 21%]
rid/v1/test_subscription_validation.py ........                          [ 24%]
rid/v1/test_token_validation.py ......                                   [ 26%]
rid/v2/test_isa_expiry.py .......                                        [ 28%]
rid/v2/test_isa_simple.py ...................                            [ 34%]
rid/v2/test_isa_validation.py .........                                  [ 37%]
rid/v2/test_subscription_isa_interactions.py ......                      [ 39%]
rid/v2/test_subscription_simple.py ............                          [ 42%]
rid/v2/test_subscription_validation.py ........                          [ 45%]
rid/v2/test_token_validation.py ......                                   [ 47%]
scd/test_constraint_simple.py ...................                        [ 53%]
scd/test_constraints_with_subscriptions.py ...........                   [ 56%]
scd/test_operation_references_error_cases.py .......................     [ 64%]
scd/test_operation_references_state_transition.py ......                 [ 66%]
[...]

Issue #742, opened 2022-03-16
Mitigate contention with dense, high-tempo operations


BenjaminPelletier:
When prober's `scd/test_operation_simple_heavy_traffic_concurrent.py` test runs on a real-world cross-data-center distributed CRDB cluster with 100 concurrent operations, often 1 operation mutation will fail (as many as 3 failures observed) with one of the [contention-type errors](https://www.cockroachlabs.com/docs/stable/transaction-retry-error-reference.html), usually ABORT_REASON_PUSHER_ABORTED.

Even when the number of concurrent operations is reduced to 40, it appears that a failure is sometimes observed even with 50 retries.  This is not yet verified as the failure observation was made before #740 was merged, and #740 is what confirms the 50 retries.

While even 10-20 concurrency support should serve all foreseeable medium-term deployments, we should understand this failure better and identify a mitigation to enable future scaling.  In the short term, we should reduce the number of concurrent operations to better align test acceptance criteria to current product needs (reducing concurrent operation count).  In the long term, we should mitigate the issue and restore the higher concurrency limit.

PR #1004, opened 2024-02-23
[dss] Reduce contention on OI creation/mutation endpoint

barroco:
This PR adds a lock at the beginning of the transactions creating or mutating operational intents.
Contention is mainly due to the ability of the transaction to increment the subscription's notification index.
The lock will ensure transactions with contention on subscriptions on the same cells to wait before starting preventing multiple retries.

Compared to the result reported in the issue #1002, we can identify the following overall test durations (single run) and number of transaction retries:

| test step | v0.11.0-rc1 | #1004 | 
| --- | --- | --- |
| test_create_ops_concurrent | 4s | 1.4 s | 
| test_mutate_ops_concurrent | 5.2s | 1.5 s |
| overall retries | 150 | 17 |
| overall CI run | ~ 24s | ~ 19s |

The following diagram shows the result of `scd/test_operation_simple_heavy_traffic_concurrent.py`. Test was run on a GCP deployment (3 core-services, 3 cockroach db nodes):

Screenshot 2024-02-23 at 18 26 20

This improvement should address #742.

Note that due to the sensitivity of the change, I tried to keep it as focused as possible. Some opportunities for refactoring will be addressed separately.

### DIFF
--- a/pkg/scd/operational_intents_handler.go
+++ b/pkg/scd/operational_intents_handler.go
@@ -430,6 +430,13 @@ func (a *Server) PutOperationalIntentReference(ctx context.Context, manager stri
 	action := func(ctx context.Context, r repos.Repository) (err error) {
 		var version int32 // Version of the Operational Intent (0 means creation requested).
 
+		// Lock subscriptions based on the cell to reduce the number of retries under concurrent load.
+		// See issue #1002 for details.
+		err = r.LockSubscriptionsOnCells(ctx, cells)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


BenjaminPelletier (approved):
It's great we can sufficiently alleviate the problem with a solution this simple -- thanks for doing this!

After this P0 is resolved, we can (at a more normal priority) take a look at those follow-up items like refactoring and additional improvements.  Thanks again!

PR #1005, opened 2024-02-27
[dss] allow accepted op-intent references to not have a subscription

Shastick:
It seems that the ASTM-F3548-v21 standard provides no ground to require the presence of a subscription when a client creates an Operation Intent Reference in the 'accepted' state.

This PR relaxes the requirement and allows such OIRs to not specify either a subscription ID, or the necessary information to create an implicit subscription.

Unfortunately, the current OpenAPI specification does not allow to return an Operational Intent reference that has no `subscription_id`, or that would have that field set to an empty string.

For this purpose we introduce an "empty UUID" placeholder that conforms with the UUIDV4 specification: `00000000-0000-4000-8000-000000000000`


Shastick:
Opening for a first round of review: a `prober` test is still failing as it requires an [adaptation of the expected return code](https://github.com/interuss/monitoring/pull/523/files)

Shastick:
@BenjaminPelletier with the latest prober version the CI is passing, this is ready for review

test/migrations/rid_db_post_migration_e2e.sh
```diff
@@ -82,7 +82,7 @@ docker run --link dummy-oauth-for-testing:oauth \
 	--link core-service-for-testing:core-service \
 	-v "${RESULTFILE}:/app/test_result" \
 	-w /app/monitoring/prober \
-	interuss/monitoring:v0.3.0 \
```
BenjaminPelletier, line 85: Can we jump straight to v0.5.0?

pkg/scd/operational_intents_handler.go
```diff
@@ -463,36 +476,36 @@ func (a *Server) PutOperationalIntentReference(ctx context.Context, manager stri
...
-			// Create implicit Subscription
-			if params.NewSubscription == nil || params.NewSubscription.UssBaseUrl == "" {
-				return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Missing new_subscription or uss_base_url in new_subscription")
-			}
-			if !a.EnableHTTP {
-				err := scdmodels.ValidateUSSBaseURL(string(params.NewSubscription.UssBaseUrl))
-				if err != nil {
-					return stacktrace.PropagateWithCode(err, dsserr.BadRequest, "Failed to validate USS base URL")
+			// Create implicit Subscription if the implicit subscription params are set:
+			// (for situations where these params are required, an error will have been returned earlier)
```
BenjaminPelletier, line 480: I'm not clear on what this comment means -- is it supposed to be "for situations where these params are required, an error will have been returned earlier *if the params were not provided*"?

BenjaminPelletier (approved):

test/migrations/rid_db_post_migration_e2e.sh
Shastick: My bad, indeed, we want that otherwise we break things.

pkg/scd/operational_intents_handler.go
Shastick: Rephrased and expanded the comment.

PR #1012, opened 2024-03-15
[scd] minimal implementation for the dss report handler

Shastick:
An implementation for the dss report handler that simply logs the report as a Json string.

(This is to have a version to develop a simple test scenario against, but the goal would be to have something that we can possibly ship)

### DIFF
--- a/pkg/scd/dss_report_handler.go
+++ b/pkg/scd/dss_report_handler.go
@@ -4,20 +4,28 @@ import (
 	"context"
 	"encoding/json"
 	"github.com/google/uuid"
+	"github.com/interuss/dss/pkg/api"
 	restapi "github.com/interuss/dss/pkg/api/scdv1"
+	dsserr "github.com/interuss/dss/pkg/errors"
 	"github.com/interuss/dss/pkg/logging"
 	"github.com/interuss/stacktrace"
 	"go.uber.org/zap"
 )
 
-// JSONLoggingDssReportHandler a DssReportHandler that simply logs the received report as JSON.
-type JSONLoggingDssReportHandler struct {
+// ReceivedReportHandler takes care of handling a DSS report received through the MakeDssReport REST handler.
+type ReceivedReportHandler interface {
+	// Handle a DSS report request. Returns the error report passed in 'req' after having set its identifier.
+	Handle(ctx context.Context, req *restapi.MakeDssReportRequest) (*restapi.ErrorReport, error)
+}
+
+// JSONLoggingReceivedReportHandler a DSSReportHandler that simply logs the received report as JSON.
+type JSONLoggingReceivedReportHandler struct {
 	// ReportLogger is the logger to which the received reports will be logged.
 	ReportLogger *zap.Logger
 }
 
 // HandleDssReport logs the received report as a JSON string to a logger.
-func (h JSONLoggingDssReportHandler) HandleDssReport(ctx context.Context, req *restapi.MakeDssReportRequest) (*restapi.ErrorReport, error) {
+func (h *JSONLoggingReceivedReportHandler) Handle(ctx context.Context, req *restapi.MakeDssReportRequest) (*restapi.ErrorReport, error) {
 	reportID, err := uuid.NewRandom()
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "Failed to generate report ID")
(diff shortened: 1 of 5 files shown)


pkg/scd/dss_report_handler.go
```diff
@@ -0,0 +1,36 @@
...
+		return nil, stacktrace.Propagate(err, "Failed to generate report ID")
+	}
+	rVal := req.Body
+	reportIDStr := reportID.String()
+	rVal.ReportId = &reportIDStr
+	// Serialize the report to a JSON string:
+	jsonReport, err := json.Marshal(req.Body)
+	if err != nil {
+		logging.WithValuesFromContext(ctx, logging.Logger).Error("Failed to serialize DSS Report", zap.Error(err))
+		return nil, stacktrace.Propagate(err, "Failed to serialize DSS Report")
```
mickmis, line 32: Failing to marshal would probably warrant returning code `dsserr.BadRequest`. (then check status and return 400 if so as suggested in other comment)

pkg/scd/dss_report_handler.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	h.ReportLogger.Info("Received DSS Report", zap.String("report", string(jsonReport)))
```
mickmis, line 34: I'd suggest passing as a field the report ID (that would make it easier to find the report in logs based on a report ID provided through another channel).

pkg/scd/dss_report_handler.go
```diff
@@ -0,0 +1,36 @@
...
+	"encoding/json"
+	"github.com/google/uuid"
+	restapi "github.com/interuss/dss/pkg/api/scdv1"
+	"github.com/interuss/dss/pkg/logging"
+	"github.com/interuss/stacktrace"
+	"go.uber.org/zap"
+)
+
+// JSONLoggingDssReportHandler a DssReportHandler that simply logs the received report as JSON.
+type JSONLoggingDssReportHandler struct {
```
mickmis, line 14: This is not the convention used in the other handlers. I'd suggest removing this struct and just embedding the code in `MakeDssReport` given it is that simple. We are just using the normal logger anyway, right?

pkg/scd/repos/repos.go
```diff
@@ -80,7 +81,14 @@ type Constraint interface {
...
 }
 
+// DssReport takes care of handling a DSS report.
+type DssReport interface {
+	// HandleDssReport handles a DSS report request. Returns the error report passed in 'req' after having set its identifier.
+	HandleDssReport(ctx context.Context, req *restapi.MakeDssReportRequest) (*restapi.ErrorReport, error)
+}
+
 // Repository aggregates all SCD-specific repo interfaces.
+// Note that 'DssReport' is not yet part of it, while we figure exactly how we want to handle DSS reports
```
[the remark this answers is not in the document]
Shastick: Makes sense. Will move it into the `dss_report_handler.go` file, then.

pkg/scd/server.go
```diff
@@ -35,9 +36,10 @@ func makeSubscribersToNotify(subscriptions []*scdmodels.Subscription) []restapi.
 
 // Server implements scdv1.Implementation.
 type Server struct {
-	Store      scdstore.Store
-	Timeout    time.Duration
-	EnableHTTP bool
+	Store            scdstore.Store
+	DssReportHandler repos.DssReport
```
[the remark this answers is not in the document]
Shastick: Strange that the linter does not catch it :o

pkg/scd/server.go
```diff
@@ -54,8 +56,13 @@ func (a *Server) MakeDssReport(ctx context.Context, req *restapi.MakeDssReportRe
 			Message: dsserr.Handle(ctx, stacktrace.PropagateWithCode(req.BodyParseError, dsserr.BadRequest, "Malformed params"))}}
 	}
 
-	return restapi.MakeDssReportResponseSet{Response400: &restapi.ErrorResponse{
-		Message: dsserr.Handle(ctx, stacktrace.NewErrorWithCode(dsserr.BadRequest, "Not yet implemented"))}}
+	resp, err := a.DssReportHandler.HandleDssReport(ctx, req)
+	if err != nil {
+		return restapi.MakeDssReportResponseSet{Response400: &restapi.ErrorResponse{
```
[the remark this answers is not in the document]
Shastick: Will do

pkg/scd/dss_report_handler.go
Shastick: The json report would contain it already, but it does not hurt to add it explicitly, I guess.

pkg/scd/dss_report_handler.go
mickmis: Fair :+1:

pkg/scd/dss_report_handler.go
```diff
@@ -0,0 +1,77 @@
...
+	rVal := req.Body
+	reportIDStr := reportID.String()
+	rVal.ReportId = &reportIDStr
+	// Serialize the report to a JSON string:
+	jsonReport, err := json.Marshal(req.Body)
+	if err != nil {
+		logging.WithValuesFromContext(ctx, logging.Logger).Error("Failed to serialize DSS Report", zap.Error(err))
+		return nil, stacktrace.PropagateWithCode(err, dsserr.BadRequest, "Failed to serialize DSS Report")
+	}
+	h.ReportLogger.Info("Received DSS Report", zap.String("reportID", reportIDStr), zap.String("report", string(jsonReport)))
```
mickmis, line 42: Using `zap.String` for the report will escape parts of the JSON making it not too readable. Here using `zap.Any` should produce better result.

mickmis (approved):
Please check the two comments I've left.
Also, have you tried invoking this endpoint locally to validate it works as expected? Just for safety before merging.
Otherwise LGTM

pkg/scd/dss_report_handler.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/interuss/dss/pkg/api"
+	restapi "github.com/interuss/dss/pkg/api/scdv1"
+	dsserr "github.com/interuss/dss/pkg/errors"
+	"github.com/interuss/dss/pkg/logging"
+	"github.com/interuss/stacktrace"
+	"go.uber.org/zap"
+)
+
+// DSSReporting takes care of handling a DSS report.
+type DSSReporting interface {
```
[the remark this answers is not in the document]
Shastick: I was considering this but avoided it, because we already have some files called `*_handlers` that seem to be specifically about handling REST calls?

This interfaces is about handling things _after_ they traversed the `MakeDssReport` function in `dss_report_handler` and it feels a bit weird to pass that to an interface called DSSReportHandler?

(Having things in a separate package would make it clearer: I'm open to suggestions)

Shastick:
> Also, have you tried invoking this endpoint locally to validate it works as expected? Just for safety before merging.
> Otherwise LGTM

Yes, I wrote a tiny test scenario. I'll run it again now to be sure

pkg/scd/dss_report_handler.go
mickmis: Actually I had in mind that the logs are necessarily JSON-encoded. While it's the default this is configurable so we can't assume that. In that case we could encode it in base64, but that may be going too far for now. So let's go ahead with this version.

Shastick: Yes, the logs are in JSON, and in the current setup (to the best of my understanding), we'll have one field of the overall JSON log be raw, escaped JSON. This should be easy to process.

pkg/scd/dss_report_handler.go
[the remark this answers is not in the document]
mickmis: True that. That's a reason why I'm not a fan of defining interfaces ahead of time without it being actually in use TBH ([c.f. as well this](https://go.dev/wiki/CodeReviewComments#interfaces)). But that's (still) nitpicking so I'm good merging it as it is. I leave it up to you.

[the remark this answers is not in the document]
Shastick: Thanks for the pointer. I'm still thinking too much in terms of Java interfaces -> i did some renaming to get closer to the spec.

PR #1013, opened 2024-03-26
[dss][scd] DSS0210,A2-7-2,4d: Consider expired subscriptions as deleted

barroco:
DSS0210,A2-7-2,4d requires that expired subscriptions should be automatically deleted. This PR adapts the querySubscription according to [tsc investigation (option 2)](https://github.com/interuss/tsc/pull/12) to filter out expired records. 
Note that following the model of RID, a future PR will need to add a garbage collection job. (Option 1)


mickmis (approved):
LGTM

mickmis (approved):
FYI I tested the changes locally, all tests pass fine with additional tests from interuss/monitoring#605.

PR #1014, opened 2024-04-11
[terraform-aws-kubernetes] Reduce the size of the default node disks

barroco:
The size of disks used by the EC2 node is not actually used by the workloads deployed in Kubernetes. Volumes created by Kubernetes are attached to that node and therefore provisioning 100G for the host do not make sense.
This PR reduces it at 10G.

Note that this change may require a recreation of the nodes, therefore, it is probably soon the right time to start versioning the terraform, tanka and helm manifests in a package manager, cf [C.4] of https://github.com/interuss/dss/issues/874.


BenjaminPelletier (approved):

BenjaminPelletier:
Until we have automatic versioning and migration, we should make sure to flag any breaking changes like this in the notes for the next release.

PR #1023, opened 2024-05-01
[dss][openapi-to-go-server] Move scopes to go constants in generated files

barroco:
This PR moves the repeated scope strings to a constant simplifying its reference in other parts of the code of the DSS.

The first commit changes the generation logic.
The second updates the example.
The third updates the interfaces of the DSS implementation.

### DIFF
--- a/pkg/api/ridv1/interface.gen.go
+++ b/pkg/api/ridv1/interface.gen.go
@@ -7,57 +7,59 @@ import (
 )
 
 var (
+	DssWriteIdentificationServiceAreasScope  = api.RequiredScope("dss.write.identification_service_areas")
+	DssReadIdentificationServiceAreasScope   = api.RequiredScope("dss.read.identification_service_areas")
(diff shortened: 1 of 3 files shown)


interfaces/openapi-to-go-server/apis.py
```diff
@@ -58,6 +59,19 @@ def filter_operations(self, tags: Set[str]):
         self.data_types = [dt for dt in self.data_types
                            if dt.name in required_data_types]
 
+    def security_scopes(self) -> Set[Tuple[str, operations.Scope]]:
```
Shastick, line 62: Might a `Dict` make more sense? 

A set of tuples is a lot like a Dict, though if you don't need to do lookups, the Set is fine.

Shastick (approved):
Looking good to me.

BenjaminPelletier (approved):

mickmis (approved):
Nice :+1:

PR #1032, opened 2024-05-15
Update dss-core-service.yaml

StevenPG:
Using the same values from the "if" branch to avoid duplicate configuration in values.yaml

When deploying as-is from helm, my configuration looks like the following example:

`
dss:
  jwksEndpoint: 'someEndpoint'
  jwksKeyIds: []
  conf:
     jwksEndpoint: 'someEndpoint'
     jwksKeyIds: []
     hostname: $host
`

It looks like the intention was to use the same value from the `if` block in the content. This PR should adjust that and match the same file that is provided and the user is instructed to copy-paste.

### DIFF
--- a/deploy/services/helm-charts/dss/templates/dss-core-service.yaml
+++ b/deploy/services/helm-charts/dss/templates/dss-core-service.yaml
@@ -46,10 +46,10 @@ spec:
             - --garbage_collector_spec=@every 30m
             - --gcp_prof_service_name=
             {{- if $dss.conf.jwksEndpoint }}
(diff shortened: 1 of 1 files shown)


StevenPG:
I'm not sure whether I should sign this CLA for a simple config change, can a contributor make this update on my behalf to correct the documentation and config?

BenjaminPelletier:
Unfortunately we [can't merge any changes without a signed CLA](https://github.com/interuss/tsc/blob/main/CHARTER.md#7-intellectual-property-policy) but @barroco can check on bugs

PR #1042, opened 2024-06-08
Remove redundant S2 cells queries for operational intents

BenjaminPelletier:
For what I believe are historical reasons (entity cells used to be in a separate table from the entities themselves rather than in a column of the entity), the S2 cells for an operational intent are currently retrieved separately and explicitly for each individual operational intent even though the cells had already been retrieved in the initial query for operational intent information.  This resulted in an enormous number of additional queries -- one for each operational intent result -- which would cause, e.g., an operational intent search to perform 200+ SQL queries if the search returned 200 operational intents.  All of these additional queries are unnecessary because each entity's S2 cells are already retrieved on the initial query.

This PR fixes that problem by simply deleting the redundant separate population of operational intent cells.  Thanks to @BradNicolle for the catch and suggestion.

### DIFF
--- a/pkg/scd/store/cockroach/operational_intents.go
+++ b/pkg/scd/store/cockroach/operational_intents.go
@@ -12,7 +12,6 @@ import (
 	dsssql "github.com/interuss/dss/pkg/sql"
 	"github.com/interuss/stacktrace"
 
-	"github.com/golang/geo/s2"
 	"github.com/jackc/pgx/v5"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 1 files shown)


BenjaminPelletier:
Note: the content being deleted cases int64 cell IDs to uint64 before casting to s2.CellIDs whereas [OperationalIntent.SetCells](https://github.com/interuss/dss/blob/1e01739d9d8370e8ff610fb9d019f9cc3fab3971/pkg/scd/models/operational_intents.go#L158) casts the int64s directly to s2.CellIDs.  We should verify that the intermediate uint64 cast is not necessary.

Shastick:
I ran the qualifier against a locally built version of this PR and everything passed.

(I still haven't gotten to running the qualifier as part of the CI as @barroco [suggested ](https://github.com/interuss/dss/pull/1039#pullrequestreview-2075945747), I'll create an issue and put it on my stack)

PR #1045, opened 2024-06-19
[ci] run the uss-qualifier as part of the CI

Shastick:
To increase test coverage, we add a step to the CI that will run the qualifier against the DSS version built from the current PR.

Notes:
 - this currently only runs "single instance tests", as, to the best of my understanding, only a single DSS instance is started in CI (therefore, only a single DSS is configured)
 - this does not run any CRDB-specific checks (they are skipped) 
 - the configuration was obtained by running the qualifier with the `--config-output` option, with the following modifications:
   - removal of the second DSS
   - add `has_private_address: true` to dss resource specifications (otherwise the qualifier fails on resource initialisation)
   - set `uss1: all_astm_dss_requirements` in the `participant_requirements` artifacts section
  
 
Test plan:
 This worked both locally when running the qualifier from a docker image built from a recent release of the `interuss/monitoring` image, and passes on the CI when running with the monitoring release `v0.8.0`
 
 See [this successful test run](https://github.com/interuss/dss/pull/1045/checks#step:11:205) for example

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,5 +63,7 @@ jobs:
       run: make start-locally
     - name: Probe local DSS instance
       run: make probe-locally
+    - name: Run Qualifier against local DSS instance
+      run: make qualify-locally
     - name: Bring down local DSS instance
       run: make down-locally
--- a/Makefile
+++ b/Makefile
@@ -145,6 +145,10 @@ start-locally:
 probe-locally:
 	build/dev/probe_locally.sh
 
+.PHONY: qualify-locally
+qualify-locally:
+	build/dev/qualify_locally.sh
+
 .PHONY: collect-local-logs
 collect-local-logs:
 	docker logs dss_sandbox-local-dss-core-service-1 2> core-service-for-testing.log
--- a/build/dev/dss_probing_qualifier_config.yaml
+++ b/build/dev/dss_probing_qualifier_config.yaml
@@ -1,0 +1,217 @@
+v1:
+    artifacts:
+        raw_report:
+            indent: null
+            redact_access_tokens: true
+        report_html: null
+        sequence_view:
+            redact_access_tokens: true
+            render_kml: true
+        templated_reports: null
+        tested_requirements:
+            - participant_requirements:
+                  uss1: all_astm_dss_requirements
+              report_name: requirements
+              requirement_collections:
+                  all_astm_dss_requirements:
+                      requirement_collections:
+                          - requirement_sets:
+                                - astm.f3411.v22a.dss_provider
+                                - astm.f3411.v19.dss_provider
+                                - astm.f3548.v21.dss_provider
+    test_run:
+        action:
+            on_failure: Continue
+            test_suite:
+                resources:
+                    f3411v19_dss_instances: netrid_dss_instances_v19
+                    f3411v22a_dss_instances: netrid_dss_instances_v22a
+                    f3548v21_dss_instances: scd_dss_instances
+                    flight_intents: che_non_conflicting_flights
+                    id_generator: id_generator
+                    planning_area: kentland_planning_area
+                    problematically_big_area: kentland_problematically_big_area
+                    second_utm_auth: second_utm_auth
+                    service_area: kentland_service_area
+                    test_exclusions: test_exclusions
+                    utm_client_identity: utm_client_identity
+                suite_type: suites.interuss.dss.all_tests
+        execution:
+            include_action_when: null
+            skip_action_when: null
+            stop_fast: true
+            stop_when_resource_not_created: false
+        non_baseline_inputs:
+            - v1.test_run.resources.resource_declarations.utm_auth
+            - v1.test_run.resources.resource_declarations.second_utm_auth
+            - v1.test_run.resources.resource_declarations.scd_dss_instances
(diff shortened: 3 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -65,3 +65,22 @@ jobs:
...
+        submodules: true
+        path: monitoring-repo
+        ref: main  # consider checking out a tag instead? Downside is that we'd need to update it with each release
+    - name: Run qualifier in monitoring repository
+      run: |
+        set -e
+        cd monitoring-repo
+        # Set the docker-compose file to use the locally built DSS instance
+        # Note that on OS X you'll need to use `sed -i ''` instead of `sed -i`
+        sed -i 's|interuss/dss:[^ ]*|interuss-local/dss:latest|g' build/dev/docker-compose.yaml
```
BenjaminPelletier, line 81: The ability to bring up a local DSS instance should certainly exist in the DSS repo -- we shouldn't need to rely on the content of another repository to do that.

Since uss_qualifier is designed to test systems however they're deployed, we should tell it how our DSS system is deployed when we want to test it.  That means we should have a test configuration for testing the DSS tracked in this repo.  The configuration will probably look a lot like [this one](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/configurations/dev/dss_probing.yaml), but we should remove the $refs because the targets exist in another repo and we don't want to be dependent on them for the configuration to be valid.  To do this, we can run uss_qualifier with [`--config-output`](https://github.com/interuss/monitoring/blob/8b760a6bb16d0f7fc90ed278ffd167958b5cb409/monitoring/uss_qualifier/main.py#L47).

Once we have a locally-deployed system to test and a configuration specifying how to test it, we should be able to just run uss_qualifier directly similar to how we [run `prober` directly](https://github.com/interuss/dss/blob/73c267551e3a047acb1fd53fc90c7d97efeba814/build/dev/probe_locally.sh#L38).  `monitoring`'s run_locally.sh is probably a good hint on what [that command to run uss_qualifier](https://github.com/interuss/monitoring/blob/8b760a6bb16d0f7fc90ed278ffd167958b5cb409/monitoring/uss_qualifier/run_locally.sh#L63) should look like, but we shouldn't need to run the script in the other repo since that's just an example and not [the generally-expected usage of uss-qualifier](https://github.com/interuss/monitoring/tree/main/monitoring/uss_qualifier#usage).

It is probably also a good idea to (separately) modernize [the local DSS deployment approach in this repo](https://github.com/interuss/dss/blob/master/build/dev/docker-compose_dss.yaml) to look more like the one in the monitoring repo, but the two repos should not be directly coupled.

.github/workflows/ci.yml
```diff
@@ -65,3 +65,22 @@ jobs:
       run: make probe-locally
     - name: Bring down local DSS instance
       run: make down-locally
+    - name: Checkout monitoring repository
```
BenjaminPelletier, line 68: We should be using uss_qualifier as a standalone tool that is released separately from the DSS.  When this is the case, we should simply be able to use the released `monitoring` docker image (like [the prober check](https://github.com/interuss/dss/blob/73c267551e3a047acb1fd53fc90c7d97efeba814/build/dev/probe_locally.sh#L43) does) and not check out the repo at all.  If we really need to use a specific commit of that repo that is not released, that probably indicates a problem in our release strategy since if this user needs that specific unreleased commit, it's likely others will want to use that specific unreleased commit as well, so we should make a `monitoring` release that includes that commit.

.github/workflows/ci.yml
```diff
@@ -65,3 +65,22 @@ jobs:
...
+        set -e
+        cd monitoring-repo
+        # Set the docker-compose file to use the locally built DSS instance
+        # Note that on OS X you'll need to use `sed -i ''` instead of `sed -i`
+        sed -i 's|interuss/dss:[^ ]*|interuss-local/dss:latest|g' build/dev/docker-compose.yaml
+        # Start all required containers to run the qualifier
+        make restart-all
+        cd monitoring/uss_qualifier/
+        # Run the qualifier
+        ./run_locally.sh configurations.dev.noop,configurations.dev.uspace,configurations.dev.geoawareness_cis,configurations.dev.generate_rid_test_data,configurations.dev.geospatial_comprehension,configurations.dev.general_flight_auth,configurations.dev.message_signing,configurations.dev.dss_probing,configurations.dev.f3548_self_contained,configurations.dev.netrid_v22a,configurations.dev.netrid_v19
```
BenjaminPelletier, line 86: Most of these configurations are not valid for testing the behavior of the DSS product itself.  A configuration very similar to [dss_probing](https://github.com/interuss/monitoring/tree/main/monitoring/uss_qualifier/configurations/dev#dss_probing) in the monitoring repo should be all we need to verify compliance of this DSS implementation to relevant requirements.  If [that test suite](https://github.com/interuss/monitoring/tree/main/monitoring/uss_qualifier/suites/interuss/dss) is missing tests to adequately verify compliance of a DSS implementation, we should add those missing tests to it rather than relying on other test suites not directly testing a DSS implementation to incidentally cover those tests.

Shastick:
The step is run and [succeeded](https://github.com/interuss/dss/pull/1045/checks#step:11:205).

BenjaminPelletier (approved):

PR #1052, opened 2024-07-08
[versioning] add new versioning endpoint

### LINKED ISSUE #1047: Implement versioning API
Since a DSS instance is merely a USS system component, versioning the entire USS system would be more straightforward if the InterUSS DSS implementation reported its software version according to a standard API.

Implement [the InterUSS versioning API](https://github.com/interuss/automated_testing_interfaces/tree/main/versioning) in the InterUSS DSS implementation.

The existing [aux API](https://github.com/interuss/dss/blob/ce9e97db6c256c2a6f1d589fcac8cf52d7f7b723/interfaces/aux_/aux_.yaml#L24) does have an existing version endpoint.  In the monitoring repo's [VersionProviderSpecification](https://github.com/interuss/monitoring/blob/69b0cbc76afa29754188d6d2793b8487a11c6a1e/monitoring/uss_qualifier/resources/versioning/client.py#L22), we could add another field named something like `interuss_dss` which would produce a new type of versioning client (perhaps `InterUSSDSSVersionClient`) which would query the existing aux version endpoint.  But, I think that would be backward-looking while transitioning to the standard API would be more forward-looking and encourage USSs to do the same.

Shastick:
Adds the versioning API to the DSS, according to https://github.com/interuss/automated_testing_interfaces/tree/main/versioning.

This is done by:
 - adding a submodule that points to the versioning endpoint spec [repository](https://github.com/interuss/automated_testing_interfaces)
 - running the generator, moving the generated files to a relevant place and checking them into git
 - serving the new route and returning the current binary version for any requested identity.
 
The endpoint was tested locally via the `build/dev/read_version.sh` test script.

resolves #1047

### DIFF
--- a/cmds/core-service/main.go
+++ b/cmds/core-service/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/interuss/dss/pkg/versioning"
 	"log"
 	"net/http"
 	"net/url"
@@ -20,6 +21,7 @@ import (
 	apiridv1 "github.com/interuss/dss/pkg/api/ridv1"
 	apiridv2 "github.com/interuss/dss/pkg/api/ridv2"
 	apiscdv1 "github.com/interuss/dss/pkg/api/scdv1"
+	apiversioningv1 "github.com/interuss/dss/pkg/api/versioningv1"
 	"github.com/interuss/dss/pkg/auth"
 	aux "github.com/interuss/dss/pkg/aux_"
 	"github.com/interuss/dss/pkg/build"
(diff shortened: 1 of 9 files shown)


cmds/core-service/main.go
```diff
@@ -44,6 +46,9 @@ var (
 	timeout    = flag.Duration("server timeout", 10*time.Second, "Default timeout for server calls")
 	locality   = flag.String("locality", "", "self-identification string used as CRDB table writer column")
 
+	systemIdentity = flag.String("system_identity", "", "System identity for which a version should be provided via the automated testing versioning interface")
+	systemVersion  = flag.String("system_version", "", "System version to provide via the automated testing versioning interface")
```
Shastick, line 50: I assume that, if the overall approach of the PR is correct, we will ultimately want this to return the same thing as the existing `aux/v1/version` endpoint. (obtained by a call to `version.Current()`.

(I'll update the PR once the open discussion points have been adressed)

BenjaminPelletier:
The system identity is the identifier of a system boundary -- basically, defining which subcomponents of a multi-component system are in scope for "the system" when we ask for the version of "the system".  InterUSS will not generally be part of the actual USS-regulator testing activities, so we won't (at least initially) be able to support including or excluding components of a DSS for versioning -- for us, the DSS is an atomic unit.  Since we don't support drawing a versioned system boundary through the atomic DSS unit, the InterUSS versioning endpoint should always return the same version (of the whole atomic DSS unit) regardless of what system boundary is requested.

The version returned can have any format and we already have an established version format in the aux API; we should just reuse that value here.  It wouldn't make sense to accept the version as a flag since the version is an intrinsic property of the software and can't be different for two different deployments/instances of the same binary/executable.

Shastick:
PR updated to reflect latest comment by @BenjaminPelletier, this is ready for review.

BenjaminPelletier (approved):

PR #1057, opened 2024-07-18
[scd] properly cleanup implicit sub on oir update

### LINKED ISSUE #1056: [scd] dangling implicit subscription when OIR with implicit subscription is mutated
**Describe the bug**

When updating an Operational Intent Reference (OIR) for which an implicit subscription was created, if the subscription id is not specified and the parameters for creating an implicit subscription are specified, the DSS will create a new implicit subscription without deleting the previous one.

This is [the region](https://github.com/interuss/dss/blob/8a6c91674868bb0f40043848a6b3e81aa5f5ebbe/pkg/scd/operational_intents_handler.go#L480) of the code where it happens.

**To Reproduce**

1. Create an OIR and request the creation of an implicit subscription
2. Mutate the OIR without specifying a subscription ID and specify some parameters for an implicit subscription
3. Observe that the returned OIR has a new implicit subscription ID
4. Query the previous implicit subscription: it still exists, even if the OIR is now attached to the newly created implicit subscription

**Expected behavior**

After the mutation of an OIR with an implicit subscription, under all circumstances, either the existing implicit subscription has been reused and properly updated, or a new implicit subscription has been created, and the previous one removed if it is no longer attached to any OIR.

**Additional questions/discussion**

Options to solve this:

- **A**) On mutation, when no subscription ID is specified and the parameters for an implicit subscription are specified, it is acceptable to reuse the existing implicit subscription and mutate it if necessary.
  - Important: with this approach, the DSS should make sure that the callback URL of the subscription is also updated if the one passed in the implicit parameters of the mutation differs from the original one. Or, if this is not allowable, it should return an error.

- **B**) On mutation, when no subscription ID is specified and the parameters for an implicit subscription are specified, delete the previous implicit subscription, if it exists and has no more OIRs attached, and create a new one.

I'm in favor of option B here, as I'm not sure that allowing to indirectly mutate the callback URL of an implicit subscription is a desirable feature: 
 - callers that wish to mutate the callback URL of a subscription may mutate it directly.
 - callers that desire that an implicit subscription be re-used can specify the relevant subscription ID in the mutation request

Side question: On mutation, when no subscription and no implicit subscription parameters are specified, should we:
 - delete the existing implicit subscription if we are allowed to (state is ACCEPTED), return an error if we are not allowed
 - mutate the existing implicit subscription, if one exists and if it is required
 
Another way to look at this question is to ask _how do we remove an implicit subscription from an OIR in the ACCEPTED state_ without deleting the OIR and re-creating it?
- Is this something we would wish to support?
- I would need to confirm this, but I believe that deleting the implicit subscription would not work (I expect that the logic checking whether a subscription is still covering an OIR to ignore the OIR's state)

Shastick:
This is a first stab at #1056, and covers two corner cases that currently result in dangling implicit subscriptions:

 - when an OIR previously linked to an implicit subscription is updated with a user-specified subscription
 - when an OIR previously linked to an implicit subscription is updated in such a way that a new implicit subscription is created.
 
The OIR upsert logic now check the following:
 - if a subscription was previously attached to the updated OIR
 - if the previous subscription was an implicit one
 - if the previous implicit subscription has no more dependent OIRs
 
We now delete subscriptions that meet all the conditions above.
 
## Notes

This PR chooses the approach to maintain the re-creation of an implicit subscription when, in a request to update an OIR, parameters for an implicit subscription are provided. 

Resolves #1056 

Test plan: with this PR, the DSS is passing the test scenario check through which I discovered the problem, but which has not yet been added to the qualifier. (It will be available once https://github.com/interuss/monitoring/pull/720 is merged)

### DIFF
--- a/pkg/scd/operational_intents_handler.go
+++ b/pkg/scd/operational_intents_handler.go
@@ -16,8 +16,8 @@ import (
 	"github.com/interuss/stacktrace"
 )
 
-// subscriptionIsOnlyAttachedToOIR will check if:
-// - the subscription exists and is implicit
+// subscriptionIsImplicitAndOnlyAttachedToOIR will check if:
+// - the subscription is defined and is implicit
 // - the subscription is attached to the specified operational intent
 // - the subscription is not attached to any other operational intent
 //
@@ -27,29 +27,22 @@ import (
 // NOTE: this should eventually be pushed down to CRDB as part of the queries being executed in the callers of this method.
 //
 //	See https://github.com/interuss/dss/issues/1059 for more details
-func subscriptionIsOnlyAttachedToOIR(ctx context.Context, r repos.Repository, oirID, subscriptionID *dssmodels.ID) (bool, error) {
-	// Get the Subscription supporting the OperationalIntent, if one is defined
-	if subscriptionID != nil {
-		sub, err := r.GetSubscription(ctx, *subscriptionID)
-		if err != nil {
-			return false, stacktrace.Propagate(err, "Unable to get OperationalIntent's Subscription from repo")
-		}
-		if sub == nil {
-			return false, stacktrace.NewError("OperationalIntent's Subscription missing from repo")
-		}
-
-		if sub.ImplicitSubscription {
-			// Get the Subscription's dependent OperationalIntents
-			dependentOps, err := r.GetDependentOperationalIntents(ctx, sub.ID)
-			if err != nil {
-				return false, stacktrace.Propagate(err, "Could not find dependent OperationalIntents")
-			}
-			if len(dependentOps) == 0 {
-				return false, stacktrace.NewError("An implicit Subscription had no dependent OperationalIntents")
-			} else if len(dependentOps) == 1 && dependentOps[0] == *oirID {
-				return true, nil
-			}
-		}
+func subscriptionIsImplicitAndOnlyAttachedToOIR(ctx context.Context, r repos.Repository, oirID *dssmodels.ID, subscription *scdmodels.Subscription) (bool, error) {
+	if subscription == nil {
+		return false, nil
+	}
+	if !subscription.ImplicitSubscription {
+		return false, nil
+	}
+	// Get the Subscription's dependent OperationalIntents
+	dependentOps, err := r.GetDependentOperationalIntents(ctx, subscription.ID)
+	if err != nil {
+		return false, stacktrace.Propagate(err, "Could not find dependent OperationalIntents")
+	}
+	if len(dependentOps) == 0 {
+		return false, stacktrace.NewError("An implicit Subscription had no dependent OperationalIntents")
+	} else if len(dependentOps) == 1 && dependentOps[0] == *oirID {
+		return true, nil
 	}
 	return false, nil
 }
@@ -101,16 +94,23 @@ func (a *Server) DeleteOperationalIntentReference(ctx context.Context, req *rest
 				"OperationalIntent owned by %s, but %s attempted to delete", old.Manager, *req.Auth.ClientID)
 		}
 
-<<<<<<< HEAD
-        if old.OVN != ovn {
-            return stacktrace.NewErrorWithCode(dsserr.VersionMismatch,
-                "Current version is %s but client specified version %s", old.OVN, ovn)
-        }
+		if old.OVN != ovn {
+			return stacktrace.NewErrorWithCode(dsserr.VersionMismatch,
+				"Current version is %s but client specified version %s", old.OVN, ovn)
+		}
 
-		removeImplicitSubscription, err := subscriptionCanBeRemoved(ctx, r, old.SubscriptionID)
-=======
-		removeImplicitSubscription, err := subscriptionIsOnlyAttachedToOIR(ctx, r, &id, old.SubscriptionID)
->>>>>>> fdd5e11 (rename and make subscriptionCanBeRemoved more specific)
+		var previousSubscription *scdmodels.Subscription
+		if old.SubscriptionID != nil {
+			previousSubscription, err = r.GetSubscription(ctx, *old.SubscriptionID)
+			if err != nil {
+				return stacktrace.Propagate(err, "Unable to get OperationalIntent's Subscription from repo")
+			}
+			if previousSubscription == nil {
+				return stacktrace.NewError("OperationalIntent's Subscription missing from repo")
+			}
+		}
+
+		removeImplicitSubscription, err := subscriptionIsImplicitAndOnlyAttachedToOIR(ctx, r, &id, previousSubscription)
 		if err != nil {
 			return stacktrace.Propagate(err, "Could not determine if Subscription can be removed")
 		}
@@ -458,9 +458,9 @@ func (a *Server) upsertOperationalIntentReference(ctx context.Context, authorize
 		return nil, nil, stacktrace.NewErrorWithCode(dsserr.BadRequest, "Invalid state for initial version: `%s`", params.State)
 	}
 
-	subscriptionID := dssmodels.ID("")
+	specifiedSubscriptionID := dssmodels.ID("")
 	if params.SubscriptionId != nil {
-		subscriptionID, err = dssmodels.IDFromOptionalString(string(*params.SubscriptionId))
+		specifiedSubscriptionID, err = dssmodels.IDFromOptionalString(string(*params.SubscriptionId))
 		if err != nil {
 			return nil, nil, stacktrace.NewErrorWithCode(dsserr.BadRequest, "Invalid ID format for Subscription ID: `%s`", *params.SubscriptionId)
 		}
@@ -469,7 +469,7 @@ func (a *Server) upsertOperationalIntentReference(ctx context.Context, authorize
 	// Check if a subscription is required for this request:
 	// OIRs in an accepted state do not need a subscription.
 	if state.RequiresSubscription() &&
-		subscriptionID.Empty() &&
+		specifiedSubscriptionID.Empty() &&
 		(params.NewSubscription == nil ||
 			params.NewSubscription.UssBaseUrl == "") {
 		return nil, nil, stacktrace.NewErrorWithCode(dsserr.BadRequest, "Provided Operational Intent Reference state `%s` requires either a subscription ID or information to create an implicit subscription", state)
@@ -493,7 +493,7 @@ func (a *Server) upsertOperationalIntentReference(ctx context.Context, authorize
 			return stacktrace.Propagate(err, "Could not get OperationalIntent from repo")
 		}
 
-		var previousSubscriptionID *dssmodels.ID
+		var previousSubscription *scdmodels.Subscription
 		if old != nil {
 			if old.Manager != manager {
 				return stacktrace.NewErrorWithCode(dsserr.PermissionDenied,
@@ -505,7 +505,13 @@ func (a *Server) upsertOperationalIntentReference(ctx context.Context, authorize
 			}
 
 			version = int32(old.Version)
-			previousSubscriptionID = old.SubscriptionID
+
+			if old.SubscriptionID != nil {
+				previousSubscription, err = r.GetSubscription(ctx, *old.SubscriptionID)
+				if err != nil {
+					return stacktrace.Propagate(err, "Unable to get OperationalIntent's Subscription from repo")
+				}
+			}
 		} else {
 			if ovn != "" {
 				return stacktrace.NewErrorWithCode(dsserr.NotFound, "OperationalIntent does not exist and therefore is not version %s", ovn)
@@ -514,24 +520,29 @@ func (a *Server) upsertOperationalIntentReference(ctx context.Context, authorize
 			version = 0
 		}
 
-		var sub *scdmodels.Subscription
+		// Determine if the previous subscription needs to be cleaned up
+		previousSubIsBeingReplaced := previousSubscription != nil && specifiedSubscriptionID != previousSubscription.ID
 		removePreviousImplicitSubscription := false
-		if subscriptionID.Empty() {
-			// Create an implicit subscription if the implicit subscription params are set:
-			// for situations where these params are required but have not been set,
-			// an error will have been returned earlier.
-			// If they are not set at this point, continue without creating an implicit subscription.
+		if previousSubIsBeingReplaced {
+			removePreviousImplicitSubscription, err = subscriptionIsImplicitAndOnlyAttachedToOIR(ctx, r, &id, previousSubscription)
+			if err != nil {
+				return stacktrace.Propagate(err, "Could not determine if previous Subscription can be removed")
+			}
+		}
+
+		// effectiveSubscription is the subscription that will end up being attached to the OIR
+		// it defaults to the previous subscription, but may be updated if required by the parameters
+		effectiveSubscription := previousSubscription
+		if specifiedSubscriptionID.Empty() {
+			// No subscription ID was provided:
+			// We will check if parameters to create a new implicit subscription were provided,
+			// otherwise we will do nothing.
 			if params.NewSubscription != nil && params.NewSubscription.UssBaseUrl != "" {
 				if !a.AllowHTTPBaseUrls {
 					err := scdmodels.ValidateUSSBaseURL(string(params.NewSubscription.UssBaseUrl))
 					if err != nil {
 						return stacktrace.PropagateWithCode(err, dsserr.BadRequest, "Failed to validate USS base URL")
 					}
-				}
-
-				removePreviousImplicitSubscription, err = subscriptionIsOnlyAttachedToOIR(ctx, r, &id, previousSubscriptionID)
-				if err != nil {
-					return stacktrace.Propagate(err, "Could not determine if previous Subscription can be removed")
 				}
 
 				// Note: parameters for a new implicit subscription have been passed, so we will create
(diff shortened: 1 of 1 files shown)


pkg/scd/operational_intents_handler.go
```diff
@@ -490,6 +500,17 @@ func (a *Server) upsertOperationalIntentReference(ctx context.Context, authorize
...
+				// If an implicit subscription creation is requested, we need to check if:
+				// - a previous subscription was attached,
+				// - if so, if it was an implicit subscription
+				// - if so, if we can remove it after creating the new implicit subscription
+				removePreviousImplicitSubscription, err = subscriptionCanBeRemoved(ctx, r, previousSubscriptionID)
+				if err != nil {
+					return stacktrace.Propagate(err, "Could not determine if previous Subscription can be removed")
+				}
+
+				// We _could_ reuse the previous subscription ID if it was an implicit subscription,
```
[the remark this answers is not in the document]
Shastick: The comment does not entirely reflect my intention, I will reword it.

in short: when an update is done and the parameters for the creation of an implicit subscription are passed, my understanding is that we actually _want_ to create a new implicit subscription.

Clients are allowed to change the subscription attached to an OIR, and from my understanding, I think it is valid to give the option to replace an existing subscription with an implicit one.

Clients that wish to reuse an existing subscription (whether or not it is an implicit one) should specify the existing subscription ID in the call parameters.

pkg/scd/operational_intents_handler.go
```diff
@@ -16,6 +16,33 @@ import (
 	"github.com/interuss/stacktrace"
 )
 
+func subscriptionCanBeRemoved(ctx context.Context, r repos.Repository, subscriptionID *dssmodels.ID) (bool, error) {
```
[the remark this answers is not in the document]
Shastick: I created https://github.com/interuss/dss/issues/1059 to keep track of this. Will also add in the comment.

pkg/scd/operational_intents_handler.go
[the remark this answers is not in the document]
Shastick: To be more precise: there was no design discussion about this – this is my interpretation of how the API should work.

If you think this needs discussing/clarifying we can either open an issue, discuss here or have a chat at the next weekly meeting.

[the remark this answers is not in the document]
BenjaminPelletier: > when an update is done and the parameters for the creation of an implicit subscription are passed, my understanding is that we actually want to create a new implicit subscription.

Yes, when [`new_subscription`](https://github.com/astm-utm/Protocol/blob/d26d3222b9816a5bc90d00287c9746c4255f8f0d/utm.yaml#L766) is specified, a new implicit subscription [should be created](https://github.com/astm-utm/Protocol/blob/d26d3222b9816a5bc90d00287c9746c4255f8f0d/utm.yaml#L771).

> I think it is valid to give the option to replace an existing subscription with an implicit one.

I don't know of an actual use case for this, but I agree this is what should happen if someone specifies `new_subscription` on an op intent with an existing subscription.  And, if that existing subscription was implicit and no longer has any dependent operational intents, then I would expect the existing subscription to be automatically deleted by the DSS since it was implicit (managed by DSS).

> Clients that wish to reuse an existing subscription (whether or not it is an implicit one) should specify the existing subscription ID in the call parameters.

Agreed.

> If you think this needs discussing/clarifying we can either open an issue, discuss here or have a chat at the next weekly meeting.

I don't think there is any ambiguity here so I don't think there's need for further discussion.  However, if someone thinks otherwise then we can certainly have that discussion.

pkg/scd/operational_intents_handler.go
```diff
@@ -16,6 +16,42 @@ import (
...
+		if sub.ImplicitSubscription {
+			// Get the Subscription's dependent OperationalIntents
+			dependentOps, err := r.GetDependentOperationalIntents(ctx, sub.ID)
+			if err != nil {
+				return false, stacktrace.Propagate(err, "Could not find dependent OperationalIntents")
+			}
+			if len(dependentOps) == 0 {
+				return false, stacktrace.NewError("An implicit Subscription had no dependent OperationalIntents")
+			} else if len(dependentOps) == 1 {
+				return true, nil
```
BenjaminPelletier, line 48: We're assuming the one dependent operational intent reference is the one the caller is interested in, but that assumption is neither documented nor enforced.  If we're just checking a particular subscription, then the answer shouldn't be "yes" or "no", it should be "only when removing this particular operational intent reference" (returning an optionally-nil pointer to an operational intent ID).  Or, if we're checking whether a particular subscription can be removed after removing the dependency of an operational intent reference on that subscription, then the ID of the operational intent should be passed into this function and validated that the single dependent operational intent is the operational intent passed.

The operational intent ID was not needed previously because the context in which the code used to be located had just identified this subscription as the one the operational intent was using.

pkg/scd/operational_intents_handler.go
```diff
@@ -16,6 +16,42 @@ import (
 	"github.com/interuss/stacktrace"
 )
 
+// subscriptionCanBeRemoved will check if:
+// - a previous subscription was attached,
```
BenjaminPelletier, line 20: Attached to what?  Neither this documentation nor the function signature mentions an operational intent.  I think we should add an op intent ID to the function signature (see other comment) and then this will make more sense (a previous subscription was attached to the specified operational intent).

pkg/scd/operational_intents_handler.go
```diff
@@ -490,6 +511,14 @@ func (a *Server) upsertOperationalIntentReference(ctx context.Context, authorize
 					}
 				}
 
+				removePreviousImplicitSubscription, err = subscriptionIsOnlyAttachedToOIR(ctx, r, &id, previousSubscriptionID)
```
mickmis, line 514: Any reason why not deleting the subscription in this block rather than setting a flag? That would simplify the control flow.

pkg/scd/operational_intents_handler.go
```diff
@@ -16,6 +16,44 @@ import (
...
+//
+//	See https://github.com/interuss/dss/issues/1059 for more details
+func subscriptionIsOnlyAttachedToOIR(ctx context.Context, r repos.Repository, oirID, subscriptionID *dssmodels.ID) (bool, error) {
+	// Get the Subscription supporting the OperationalIntent, if one is defined
+	if subscriptionID != nil {
+		sub, err := r.GetSubscription(ctx, *subscriptionID)
+		if err != nil {
+			return false, stacktrace.Propagate(err, "Unable to get OperationalIntent's Subscription from repo")
+		}
+		if sub == nil {
```
mickmis, line 37: We retrieve the subscription ID from the operational intent out of the store. IIUC it is then not realistic for the subscription to not exist because the store will guarantee that to not happen. 
IMO this also highlights that invoking `GetSubscription` here is not appropriatea: this should be done when `previousSubscriptionID` is retrieved, and then pass the subscription around and not just its ID.

pkg/scd/operational_intents_handler.go
```diff
@@ -16,6 +16,44 @@ import (
...
+		}
+
+		if sub.ImplicitSubscription {
+			// Get the Subscription's dependent OperationalIntents
+			dependentOps, err := r.GetDependentOperationalIntents(ctx, sub.ID)
+			if err != nil {
+				return false, stacktrace.Propagate(err, "Could not find dependent OperationalIntents")
+			}
+			if len(dependentOps) == 0 {
+				return false, stacktrace.NewError("An implicit Subscription had no dependent OperationalIntents")
```
mickmis, line 48: Not sure if that's OK: when this new logic will be introduced, there will exist in the store some dangling implicit subscriptions.

pkg/scd/operational_intents_handler.go
```diff
@@ -16,6 +16,44 @@ import (
...
+	if subscriptionID != nil {
+		sub, err := r.GetSubscription(ctx, *subscriptionID)
+		if err != nil {
+			return false, stacktrace.Propagate(err, "Unable to get OperationalIntent's Subscription from repo")
+		}
+		if sub == nil {
+			return false, stacktrace.NewError("OperationalIntent's Subscription missing from repo")
+		}
+
+		if sub.ImplicitSubscription {
```
mickmis, line 41: Better use an early return if the subscription is explicit.

pkg/scd/operational_intents_handler.go
Shastick: > We retrieve the subscription ID from the operational intent out of the store. IIUC it is then not realistic for the subscription to not exist because the store will guarantee that to not happen.

Operational intents are not guaranteed to have a subscription attached: OIR's in the 'ACCEPTED' state may well be created without a subscription.

pkg/scd/operational_intents_handler.go
Shastick: That is correct: although, I essentially kept the logic as it existed before being factored into this method.

I'd rather fix this in a separate PR, as there might be some discussion about whether or not we should opportunistically fix such inconsistencies when we encounter them vs. noisily failing so we become aware of them.

pkg/scd/operational_intents_handler.go
Shastick: At this point in the transaction the subscription cannot be deleted, because it is referenced by the operational intent. 

The previous implicit subscription can only be removed once the OIR has been updated to point to another subscription.

pkg/scd/operational_intents_handler.go
Shastick: > IMO this also highlights that invoking GetSubscription here is not appropriatea: this should be done when previousSubscriptionID is retrieved, and then pass the subscription around and not just its ID.

I disagree: there are circumstances where we won't need to fetch the subscription. For example, when updating an OIR, if the request specifies the same subscription ID as before, there is no need to get it from the store.

mickmis:
I tried to build an overview of what should happen wrt subscription in the upsert.

Is the subscription ID provided in the request parameters the same as the subscription ID attached to the existing operational intent? (in both cases the subscription ID may be none)
If yes: no need to do anything.
If no we have the following cases:
|                              | implicit old sub                                          | explicit old sub        | no old sub              |
|------------------------------|-----------------------------------------------------------|-------------------------|-------------------------|
| *implicit sub in params*     | delete old if no dependent OI                             | -                       | -                       |
| *explicit sub in params*     | delete old if no dependent OI                             | -                       | -                       |
| *no sub in params*           | delete old if no dependent OI                             | -                       | -                       | 
| *new implicit sub in params* | create new implicit sub<br/>delete old if no dependent OI | create new implicit sub | create new implicit sub | 

At the moment we have a bit of a complex control flow for handling subscriptions updates, based first on whether the subscription ID is empty, but if the above is correct, the control flow should be as simple as (and also extensible to the deletion of operational intents):
```
if params.sub_id != old_oi.sub_id:
    if old_sub.implicit:
        delete_if_no_dependent_ois()
    if params.new_sub:
        create_new_implicit_sub()
```

If I'm correct a small refactoring could be appropriate to factor away this logic IMO. WDYT?

Shastick:
Taking to draft as I need to handle another corner case.

Shastick:
> If I'm correct a small refactoring could be appropriate to factor away this logic IMO. WDYT?

Yes, it's actually possible to move the check out of the if block. The result does not yet feel entirely clean, but I'm not sure we can do better without doing a deeper refactoring

Shastick:
@mickmis where do we stand on the need to refactor the `UpdateOperationalIntentReference` method? If needed, the refactor should certainly happen in another PR: should we do it before or after merging this one? How urgent is this with respect to the current open issues that are caused by the dangling OIRs?

Once I'm done with handling the OVN check PRs and corresponding scenarios I can take a look at the refactor, but I'm wondering if it is aligned with our current priorities?

pkg/scd/operational_intents_handler.go
```diff
@@ -506,53 +549,60 @@ func (a *Server) upsertOperationalIntentReference(ctx context.Context, authorize
...
-			sub, err = r.GetSubscription(ctx, subscriptionID)
-			if err != nil {
-				return stacktrace.Propagate(err, "Unable to get Subscription")
-			}
-			if sub == nil {
-				return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Specified Subscription %s does not exist", subscriptionID)
+			// A subscription has been specified:
+			// If it is different from the previous subscription, we need to fetch it from the store
+			// in order to ensure it correctly covers the OIR.
+			if effectiveSubscription == nil || previousSubIsBeingReplaced {
```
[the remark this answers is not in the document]
mickmis: Actually now that I look at it again, I'm also not completely sure why this change here is needed? Was there previously an issue that is fixed by this change? Or is it just an optimization to avoid fetching twice the same subscription?

mickmis (commented):
> where do we stand on the need to refactor the UpdateOperationalIntentReference method? If needed, the refactor should certainly happen in another PR: should we do it before or after merging this one? How urgent is this with respect to the current open issues that are caused by the dangling OIRs?
> 
> Once I'm done with handling the OVN check PRs and corresponding scenarios I can take a look at the refactor, but I'm wondering if it is aligned with our current priorities?

After looking at it one more time, I do believe the change in this PR does what it is supposed to do. But I think it can and should be simplified more.

If not bound by any other constraint, IMO this function should be refactored before making such a change. Its control flow is already more complex than it should, and this change make that situation even worse. This makes it hard to identify tricky corner-cases, both at implementation and review. Also, the logic in this function is both one of the most complex and most central in the DSS.
Now, I'm not sure where we stand in terms of constraints and priorities. Do you have an opinion @BenjaminPelletier ?

pkg/scd/operational_intents_handler.go
```diff
@@ -692,6 +742,14 @@ func (a *Server) upsertOperationalIntentReference(ctx context.Context, authorize
 			return stacktrace.Propagate(err, "Failed to upsert OperationalIntent in repo")
 		}
 
+		// Check if the previously attached subscription should be removed
```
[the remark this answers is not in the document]
Shastick: My reply to a comment on a [previous revision](https://github.com/interuss/dss/pull/1057#discussion_r1728517254) still applies:

> At this point in the transaction the subscription cannot be deleted, because it is referenced by the operational intent.
> 
> The previous implicit subscription can only be removed once the OIR has been updated to point to another subscription.
>
> If you delete it before the r.UpsertOperationalIntent(ctx, op) statement at line 719, you get an error.

Moving the deletion to just before `r.UpsertOperationalIntent(ctx, op)` at line 740 will cause the scenario that checks implicit subscription handling to fail (I just checked again to make sure).

pkg/scd/operational_intents_handler.go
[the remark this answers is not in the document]
Shastick: > effectiveSubscription == nil will always be true here.

Thanks for catching. A previous revision was pre-setting the value and that ended up being even more confusing than it is now.

> Actually now that I look at it again, I'm also not completely sure why this change here is needed? Was there previously an issue that is fixed by this change? Or is it just an optimization to avoid fetching twice the same subscription?

This is to avoid re-fetching the same subscription already fetched at line 496 in the `if old.SubscriptionID != nil {` block if that subscription is not being replaced. In cases where it is being replaced, we need to fetch the subscription that will eventually be attached to the OIR to verify that it properly covers the OIR.

It indeed ensures that we don't do unnecessary roundtrips.

Shastick:
> After looking at it one more time, I do believe the change in this PR does what it is supposed to do. But I think it can and should be simplified more.

Do you mean that this PR can and should be simplified, or that the method should be simplified?

> If not bound by any other constraint, IMO this function should be refactored before making such a change. Its control flow is already more complex than it should, and this change make that situation even worse. This makes it hard to identify tricky corner-cases, both at implementation and review. Also, the logic in this function is both one of the most complex and most central in the DSS. Now, I'm not sure where we stand in terms of constraints and priorities. Do you have an opinion @BenjaminPelletier ?

I personally see this as two separate issues:

 - one of correctness: is the code doing what it is supposed to do? So far we have relied on tests to convince ourselves of that, and we should continue to do so here. Furthermore we know that the DSS is currently _not_ behaving as it should, and we have some evidence that this PR would address that. If we don't trust our tests enough to tell us if the DSS behaves as expected, we should probably start by adding tests until we reach that confidence.
 - one of maintainability: how long does it take to understand the code and to modify it, and how hard is it to add a feature or fix something without unintended consequences.
 
As such, I would see no issue in proceeding with adding something as long as it is _correct_ (when we apply our usual standard for correctness) even if it worsens _maintainability_ a little bit, if focusing on maintainability first means delaying the fix by an unknown amount of time.

I see three reasons for not waiting on a refactor to proceed with the fix:

 - I don't see a good path to simplifying things yet (we could have separate paths for OIR creation and update, maybe moving things to submethods will make the logic more tractable, maybe there's a major simplification that is possible and that we don't see, maybe we will disagree on the approach and need to align), so I expect that this might take a while
 - we plan to inline or push down some logic into the queries themselves rather than doing round-trips between the database and the Go logic. This in itself will deeply impact the relevant methods, and may cause us to remove or rework the simplifications we did to improve maintainability: any effort spent building the 'perfect' method now risks being thrown away if we don't combine it with pushing down the logic to the database.
 - longer term plans may involve moving away from an SQL/relational-based store, which will have an even deeper impact
 

I do however agree that the logic herein is pretty central to what the DSS does, and that it would deserve to be made as maintainable as possible as that will benefit us in the long term.

pkg/scd/operational_intents_handler.go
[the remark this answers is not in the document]
Shastick: > Thanks for catching. A previous revision was pre-setting the value and that ended up being even more confusing than it is now.

(Well, as a good illustration of how convoluted the control logic got, my 'simplification' was too much and needs a rework)

BenjaminPelletier:
I do think fixing the correctness problem from #1056 (which I think this PR does) is important; it has impact in a number of areas (e.g., [this table](https://github.com/interuss/dss/issues/1074#issuecomment-2305567576) from #1074 which probably means #1074 wouldn't have been filed if this issue had been fixed earlier).

I have not reviewed the full content of this PR in particular because Committer review time is limited and theoretically we should all have fairly similar overall thoughts (when we don't, that's a good candidate for additional documentation in the TSC repo).  I (and/or @barroco) can take a more holistic look to weigh in on the inherently-gray tradeoffs between correctness, maintainability, and timeliness, though I think in an ideal world (which indeed we don't live in, unfortunately) a good result could be worked out with the reviewer(s) already most familiar with the content.

I don't know if this general thought is applicable to this specific case, but often complexity can be managed by factoring more sections of a large function into their own functions with their own documentation, signatures, and encapsulated behavior (as has already been done once in this PR).  This lets the core of the complex function look more like brief pseudocode, and therefore be more understandable at a glance (like [@mickmis's pseudocode](https://github.com/interuss/dss/pull/1057#issuecomment-2307088229)).

We don't need this fix tomorrow, so we can still iterate on the most maintainable version, but I agree we don't want that to take forever.  If anyone thinks we could do a better job at picking tradeoffs, that would be a great topic for the Tuesday meeting, and probably worth giving a heads up on Slack 1-2 days beforehand that we'll want to discuss that so anyone who wants to can pre-read the background info (e.g., this PR) to be better prepared to discuss real time.

Shastick:
Given that 

> We don't need this fix tomorrow, so we can still iterate on the most maintainable version, but I agree we don't want that to take forever.

I'm happy to start with encapsulating things into submethods and see what kinds of results I can come up with today.

pkg/scd/operational_intents_handler.go
Shastick: Note to self: will need to confirm the above again, as I might have been confused by another error.

pkg/scd/operational_intents_handler.go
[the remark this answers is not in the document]
mickmis: > This is to avoid re-fetching the same subscription already fetched at line 496 in the if old.SubscriptionID != nil { block if that subscription is not being replaced. In cases where it is being replaced, we need to fetch the subscription that will eventually be attached to the OIR to verify that it properly covers the OIR.
> 
> It indeed ensures that we don't do unnecessary roundtrips.

Could you add a comment mentioning that? That clarifies that this is not actually a separate case, but an optimization. Ideally that could be hidden behind a function but not sure if that's doable or makes sense here.

mickmis:
> I personally see this as two separate issues:
> 
> * one of correctness: is the code doing what it is supposed to do? So far we have relied on tests to convince ourselves of that, and we should continue to do so here. Furthermore we know that the DSS is currently _not_ behaving as it should, and we have some evidence that this PR would address that. If we don't trust our tests enough to tell us if the DSS behaves as expected, we should probably start by adding tests until we reach that confidence.
> * one of maintainability: how long does it take to understand the code and to modify it, and how hard is it to add a feature or fix something without unintended consequences.
> 
> As such, I would see no issue in proceeding with adding something as long as it is _correct_ (when we apply our usual standard for correctness) even if it worsens _maintainability_ a little bit, if focusing on maintainability first means delaying the fix by an unknown amount of time.

You make great points here, thanks for the response, I'd just add that for correctness we need understanding in addition to testing (which is what I struggled to reach).

All in all I think that what you ended up doing @Shastick (i.e. timeboxing some enhancements) is a good compromise.

(thanks for the feedback @BenjaminPelletier)

pkg/scd/operational_intents_handler.go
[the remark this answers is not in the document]
mickmis: My bad, OK I see!

Shastick:
Closing this PR to open a new clean one, since so much of original base branch has changed.

Issue #1059, opened 2024-07-25
[scd] push down implicit subscription cleanup to Database, if possible


Shastick:
**Is your feature request related to a problem? Please describe.**
In general, we want to minimize locking time when interacting with Cockroach DB.

When removing and mutating Operational Intent References (OIR), we may do several round-trips to the database.

It is likely that these could be reduced (by eg moving some logic to the SQL and have it executed at the database level)

**Describe the solution you'd like**
The logic that has been factored into a separate method in #1057 could possibly be directly integrated into some CRDB queries, instead of the current back-and forth between the database and the backend.

**Describe alternatives you've considered**
An option is to keep things as-is: this is not urgent but is worth keeping track off.

Whether or not it is worth doing this right away depends on how easy we can push down the cleanup: I'll spend a little bit of time investigating this. If anyone things it is not worth the effort at the moment, please let me know.

**Additional context**
The opportunity for pushing down the cleanup logic became apparent while investigating some [dangling implicit subscription issue for the uss_qualifier](https://github.com/interuss/monitoring/issues/415) that led to #1057

Shastick:
While we may not be able to do the removal within an existing query, we might be able able to surface the fact that the subscription attached to an OIR can be removed via an existing query.

 - [DeleteOperationalIntent](https://github.com/interuss/dss/blob/4f9b508dbd447fdf0592abdd98e0ade5a9efc6b8/pkg/scd/operational_intents_handler.go#L129) could be modified to do the deletion directly at the database level, or at least to return a boolean indicating whether or not the removal can occur. The [implementation](https://github.com/interuss/dss/blob/ce9e97db6c256c2a6f1d589fcac8cf52d7f7b723/pkg/scd/store/cockroach/operational_intents.go#L146) is simple enough and a direct deletion seems possible.
 - Similarly [https://github.com/interuss/dss/blob/ce9e97db6c256c2a6f1d589fcac8cf52d7f7b723/pkg/scd/store/cockroach/operational_intents.go#L173], can likely be modified to at least return whether or not the cleanup can occur, and could possibly handle the deletion directly.
 
 Both cases seem possible, although I need to check if CRDB's SQL dialect would allow us to do so as I'm not entirely familiar with it.

PR #1060, opened 2024-07-29
[scd] inline cleanup of implicit subscription into CRDB call when removing OIR

Shastick:
This is the first part of #1059: 

It quite literally moves the Go logic into SQL, by means of nested queries and conditional DELETE statements.

(The PR sits on top of #1057) 

Test plan: 

this PR survives a run of every current uss_qualifier configuration when run locally and also passes the prober (both on CI and locally)

### DIFF
--- a/pkg/scd/operational_intents_handler.go
+++ b/pkg/scd/operational_intents_handler.go
@@ -16,6 +16,37 @@ import (
 	"github.com/interuss/stacktrace"
 )
 
+// subscriptionIsImplicitAndOnlyAttachedToOIR will check if:
+// - the subscription is defined and is implicit
+// - the subscription is attached to the specified operational intent
+// - the subscription is not attached to any other operational intent
+//
+// This is to be used in contexts where an implicit subscription may need to be cleaned up: if true is returned,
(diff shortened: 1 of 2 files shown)


mickmis:
I wanted to start the review of this PR but:
1. Is it appropriate to start relying more on CRDB given the current pending decisions? cc @BenjaminPelletier 
2. Is it reviewable in current state? It looks out of date from #1057 @Shastick

BenjaminPelletier:
I do think we want to continue to maintain our current products until we have a positive new direction.  We probably don't want to focus on optimizations targeted specifically for CRDB (though it's probably still valuable to complete outstanding work even if it did target CRDB specifically), but it seems like this improvement would be mostly applicable for most alternative SQL options.

Shastick:
> Is it reviewable in current state? It looks out of date from https://github.com/interuss/dss/pull/1057 @Shastick

This PR does not strictly need to be up to date with #1057 as it is going to replace the Go logic from #1057 with subqueries: I would definitely wait for the other to be merged before merging this one, but the approach using subqueries can already be discussed and reviewed.

Shastick:
Replaced with #1107

PR #1061, opened 2024-07-30
[core-service] Rename enable_http to allow_http_base_urls

brandoncorrea:
### Breaking Change
Clients of `core-service` will need to update the command-line parameter, `enable_http` to `allow_http_base_urls`, namely [Monitoring – core_service.sh L21/33](https://github.com/interuss/monitoring/blob/ad6483fbd8394eb20d2d4aa9c318e6a3701479cc/build/dev/startup/core_service.sh#L21-L33).

### DIFF
--- a/build/dev/haproxy_local_setup.sh
+++ b/build/dev/haproxy_local_setup.sh
@@ -163,7 +163,7 @@ docker run -d --name core-service-for-testing -p 8082:8082 \
 	-dump_requests \
 	-accepted_jwt_audiences core-service,localhost \
 	-enable_scd	\
-	-enable_http
+	-allow_http_base_urls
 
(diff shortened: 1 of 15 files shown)


brandoncorrea:
I'm in favor of supporting backwards compatibility. I'll update the PR accordingly.

brandoncorrea:
@barroco Both `enable_http` and `allow_http_base_urls` flags are now supported.

brandoncorrea:
A warning has been added when the old flag is set, preferring the new flag if both are set.

Looking further into how flags work, I realized my first backwards-compatibility code wasn't technically correct because flags aren't initialized at the head of the file (everything is default). I moved this to `main` where logging and flags have already been initialized.

brandoncorrea:
@barroco description has been updated!

Issue #1055, opened 2024-07-12
Update Helm chart values.yaml to latest stable version


StevenPG:
**Is your feature request related to a problem? Please describe.**

New users to the DSS project may choose to deploy their own instance using Helm. However, there is some potential for developer bewilderment because the hardcoded default version of the core-dss service is 0.7.0.

https://github.com/interuss/dss/blob/b8db6d73b799bba94531b6036b5a05eac507ed81/deploy/services/helm-charts/dss/values.yaml#L18

**Describe the solution you'd like**
This should be updated each time a new release is created, with the latest being 0.15.0 so a user of the github repository will be on the expected version when they pull the repository.

**Describe alternatives you've considered**

Removing the default version so that a user is required to either manually input, or set the version. With a note in the README about choosing the version to coincide from the release versions and the interuss Dockerhub organization.

StevenPG:
Awesome @barroco, the solution makes sense. I appreciate you resolving the issue!

PR #1063, opened 2024-08-05
[helm] Remove default dss image value to force user to set it manually

barroco:
As proposed in #1055 , this PR removes the default value to require users to set the DSS image manually.

An alternative would have been to set the `latest` tag as default but that could lead to unwanted upgrade of the image in case of rollout restart. A comment has been added in that sense.

(This change will be highlighted as breaking in the release note)

### DIFF
--- a/deploy/services/helm-charts/dss/README.md
+++ b/deploy/services/helm-charts/dss/README.md
@@ -8,8 +8,7 @@ Requirements and instructions to create a new Kubernetes cluster can be found [h
 3. Install [Helm](https://helm.sh/) version 3.11.3 or higher
 
 ## Usage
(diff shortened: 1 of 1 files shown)


marcadamsge (approved):
LGTM 👍

deploy/services/helm-charts/dss/values.example.yaml
```diff
@@ -1,5 +1,15 @@
 # See values.schema.json for schema description
 
+dss:
+  #  image: docker.io/interuss/dss:v0.15.0 # See https://hub.docker.com/r/interuss/dss/tags for official image releases.
```
mickmis, line 4: IMO in the example file it is OK to set the version (i.e. uncommenting this line).

mickmis (approved):

PR #1064, opened 2024-08-06
[dss] scd: oir upsert - inline cleanup of implicit subscription into CRDB

### LINKED ISSUE #1059: [scd] push down implicit subscription cleanup to Database, if possible

Shastick:
Test plan: both the prober and the qualifier (all scenarios) succeed when run against this PR, including the new [implicit cleanup](https://github.com/interuss/monitoring/pull/720) scenario when run locally.

Misc: this PR depends on #1107, please only consider its last commit.

closes #1059

### DIFF
--- a/pkg/scd/operational_intents_handler.go
+++ b/pkg/scd/operational_intents_handler.go
@@ -111,11 +111,6 @@ func (a *Server) DeleteOperationalIntentReference(ctx context.Context, req *rest
 			}
 		}
 
-		removeImplicitSubscription, err := subscriptionIsImplicitAndOnlyAttachedToOIR(ctx, r, id, previousSubscription)
-		if err != nil {
-			return stacktrace.Propagate(err, "Could not determine if Subscription can be removed")
-		}
-
 		// Gather the subscriptions that need to be notified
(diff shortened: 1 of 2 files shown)


Shastick:
Closing given that we're migrating to Yugabyte. Will reopen or create a new Yugabyte compatible version if/when inlining is relevant again.

PR #1071, opened 2024-08-21
[doc] Deployment documentation migration - architecture introduction

barroco:
As part of #874, the content related to deploying a DSS in the build/ folder should be migrated to /deploy.

This PR moves the architecture introduction to the new location.

### DIFF
--- a/deploy/architecture.md
+++ b/deploy/architecture.md
@@ -1,4 +1,4 @@
-# DSS deployment architecture
+# Deployment of a DSS Pool
 
 ## Introduction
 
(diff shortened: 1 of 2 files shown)


deploy/architecture.md
```diff
@@ -1,8 +1,20 @@
-# Kubernetes deployment
+# DSS deployment architecture
```
mickmis, line 1: Is 'architecture' the correct concept here? In my mind in the context of DSS pooling as described by the introduction, an architecture would be describing a specific arrangement of DSSs forming a pool. But here it seems like we are talking about how to actually define such an architecture and how to operate it.

PR #1080, opened 2024-08-27
[terraform] Explicitly set CRDB version in terraform examples and tooling

barroco:
Following #1077 and #1076, tanka manifests and helm chart values require the cockroachdb version to be explicitely input. This PR upgrades terraform tooling to explicitly set in examples (*.examples.tf) and CI ([deploy/operations/ci/aws-1/terraform.tfvars](https://github.com/interuss/dss/compare/master...Orbitalize:dss:tf-parametrize-cockroachdb-tag?expand=1#diff-457e94149888dc4db9ae6640c4141106cc1614f186640965833bb4e1f9c65231)) the cockcroach db version.

Users who have previously set a configuration following the instructions will have to add the image version to their configuration. A note will be provided in the release note.

### DIFF
--- a/deploy/infrastructure/dependencies/terraform-commons-dss/tanka.tf
+++ b/deploy/infrastructure/dependencies/terraform-commons-dss/tanka.tf
@@ -11,7 +11,7 @@ resource "local_file" "tanka_config_main" {
     VAR_CRDB_EXTERNAL_NODES      = join(",", [for a in var.crdb_external_nodes : "'${a}'"])
     VAR_STORAGE_CLASS            = var.kubernetes_storage_class
     VAR_DOCKER_IMAGE_NAME        = var.image
-    VAR_CRDB_DOCKER_IMAGE_NAME   = var.crdb_image_tag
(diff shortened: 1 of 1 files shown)


deploy/infrastructure/dependencies/terraform-commons-dss/tanka.tf
```diff
@@ -10,7 +10,8 @@ resource "local_file" "tanka_config_main" {
     VAR_INGRESS_NAME             = var.ip_gateway
     VAR_CRDB_EXTERNAL_NODES      = join(",", [for a in var.crdb_external_nodes : "'${a}'"])
     VAR_STORAGE_CLASS            = var.kubernetes_storage_class
-    VAR_DOCKER_IMAGE_NAME        = local.image
+    VAR_DOCKER_IMAGE_NAME        = var.image
```
BenjaminPelletier, line 13: I don't think I follow the difference between local and var here, but will assume this fix is good if the change was intentional (for non-CRDB image).

BenjaminPelletier (approved):

PR #1082, opened 2024-08-28
[scd] oir deletion now checks for OVN correctness

Shastick:
Adds the missing OVN check to the OIR deletion logic.

This is a first step towards fixing #1081 (it covers OIRs)

Test plan: a DSS version built from this PR passes the new [qualifier scenario](https://github.com/interuss/monitoring/pull/761) designed to detect this kind of issue.

PR #1083, opened 2024-08-28
[scd] cr deletion now checks for OVN correctness

### LINKED ISSUE #1081: [scd] deletion of entity ref. accepts any OVN for deletion.
**Describe the bug**
The [implementation](https://github.com/interuss/dss/blob/06a97569a0e8f38b23167e3e45f1d03f91db0525/pkg/scd/operational_intents_handler.go#L21) of the Operational Intent Reference (OIR) deletion endpoint ignores the provided OVN and will delete the OIR in any case.

The same issue seems to affect [Constraint Reference (CR) deletion](https://github.com/interuss/dss/blob/06a97569a0e8f38b23167e3e45f1d03f91db0525/pkg/scd/constraints_handler.go#L19).

(Subscriptions, on the other hand, are [correctly handled](https://github.com/interuss/dss/blob/06a97569a0e8f38b23167e3e45f1d03f91db0525/pkg/scd/subscriptions_handler.go#L508))

**Expected behavior**
The OIR deletion endpoint is expected to only delete the specified OIR if the passed OVN matches the one currently stored for that OIR.

If the OVNs don't match, we would expect a 409, as specified in the [astm-utm protocol's OpenAPI spec](https://github.com/astm-utm/Protocol/blob/63ba6c5ff9c9abb93c55fbdf2bffc0b089d10c4a/utm.yaml#L2475), and the OIR should not be deleted.

The CR deletion endpoint should behave in the same way.

In addition to the fix itself, we would expect the uss_qualifier to detect when a DSS implementation does not validate OVNs. Enriching the qualifier with the relevant checks is covered [by this issue](https://github.com/interuss/monitoring/issues/759).

**Additional context**
Originally described by @BenjaminPelletier in [this issue](https://github.com/interuss/dss/issues/1069#issuecomment-2313161276)

Shastick:
Add a check for the provided OVN before deleting constraint references.

Once #1082 is merged, this will close #1081 

Test plan: a DSS built from this PR was used as a target for [this scenario](https://github.com/interuss/monitoring/pull/762) and behaved properly.

### DIFF
--- a/pkg/scd/constraints_handler.go
+++ b/pkg/scd/constraints_handler.go
@@ -37,6 +37,13 @@ func (a *Server) DeleteConstraintReference(ctx context.Context, req *restapi.Del
 			Message: dsserr.Handle(ctx, stacktrace.NewErrorWithCode(dsserr.PermissionDenied, "Missing manager"))}}
 	}
 
+	// Retrieve OVN
+	ovn := scdmodels.OVN(req.Ovn)
+	if ovn == "" {
+		return restapi.DeleteConstraintReferenceResponseSet{Response400: &restapi.ErrorResponse{
+			Message: dsserr.Handle(ctx, stacktrace.NewErrorWithCode(dsserr.BadRequest, "Missing OVN for constraint to modify"))}}
+	}
+
 	var response *restapi.ChangeConstraintReferenceResponse
 	action := func(ctx context.Context, r repos.Repository) (err error) {
 		// Make sure deletion request is valid
@@ -49,6 +56,9 @@ func (a *Server) DeleteConstraintReference(ctx context.Context, req *restapi.Del
 		case old.Manager != dssmodels.Manager(*req.Auth.ClientID):
 			return stacktrace.NewErrorWithCode(dsserr.PermissionDenied,
 				"Constraint owned by %s, but %s attempted to delete", old.Manager, *req.Auth.ClientID)
+		case old.OVN != ovn:
+			return stacktrace.NewErrorWithCode(dsserr.VersionMismatch,
+				"Current version is %s but client specified version %s", old.OVN, ovn)
 		}
 
 		// Find Subscriptions that may overlap the Constraint's Volume4D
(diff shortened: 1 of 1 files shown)


Shastick:
@BenjaminPelletier, following our discussion at the last community meeting about maintaining and updating the qualifier configuration used in the CI for the DSS repo, should we maybe change the way we release scenarios that 'break' the current DSS implementation?

**So far we have been doing this:**
1. Add scenario (or fix existing one) that reveals an issue with the DSS
2. Fix the issue on the DSS side, manually test the PR against the new scenario
3. Release new DSS version
4. update DSS version on the monitoring repo, obtain a green CI
5. merge new scenario
6. release monitoring image
7. update monitoring image on the DSS repo
8. update CI qualifier configuration on the DSS repo

This works, but is a bit ad-hoc and requires trusting the manual step in 2) above. The main reason we do this is that we usually directly update the qualifier configurations when we add a scenario, which means CI runs will necessarily break if the DSS has an open issue.

**An alternative process could be:**
1. Add scenario that reveals issue with the DSS, but do not update the default configurations yet
2. Open PR, get it reviewed and merged
3. Release new version of the monitoring image
4. Within a PR on the DSS:
  - Fix the issue on the DSS,
  - update the qualifier config used in the DSS CI to use the new scenario, 
  - get a green CI that confirms the issue is fixed and merge
5. release the DSS image
9. Open a PR on the monitoring repo that:
  - updates to the latest released DSS
  - updates the configurations to test the newly fixed features
  
The second process is technically cleaner, albeit a little bit more involved, and introduces the risk to forget to add a relevant configuration to our defaults in step 6). On the other hand, it gets us closer to properly using the qualifier and forces us to go through the gymnastics of properly updating configurations on the DSS side. (Under the current process, we risk forgetting updating the monitoring image or the qualifier configuration on the DSS repo, which is guaranteed to be impossible with the alternative process)

I will try the second approach for the set of scenarios that check that OVNs are being verified to see if it is even practical: for entirely new scenarios it should be easy, but for updates to existing scenarios it will be less so.

Note that there will still be situations that require an ad-hoc approach.

Shastick:
@BenjaminPelletier this is ready for review: if we're interested in moving quickly, the two PRs fixing OVN handling can be merged and a release can be done so we may proceed with the qualifier scenarios https://github.com/interuss/monitoring/pull/761 and https://github.com/interuss/monitoring/pull/762

BenjaminPelletier:
I've added [this issue](https://github.com/interuss/tsc/issues/17) to track that discussion separate from this PR.

BenjaminPelletier (approved):

Issue #1088, opened 2024-08-30
[maintenance] scd improve readability of OIR upsertion logic and OIR subscription handling


Shastick:
The logic for [creating and updating Operational Intent References](https://github.com/interuss/dss/blob/0915eafcc6fba974797ea2e0aa86de795ce61875/pkg/scd/operational_intents_handler.go#L370) (OIR) is central to the DSS's role.

The [current work](https://github.com/interuss/dss/pull/1057) to address some open issues (#1056, #1074) revealed that the logic within this method is rather difficult to grasp, hindering its maintenance.

This issue's goals are:
 - provide a high level description of what this method needs to do
 - document how and where this method is tested
 - provide a place to discuss approaches to simplify and improve that method's maintainability

### Logic

The description below covers the logic of `upsertOperationalIntentReference` with respect to the handling of subscriptions and their cleanup when they are implicit.

Below, `cipasicr()` is a short-hand for _check if previously attached subscription is implicit and can be removed_

| _Subscription_     | Create OIR | Update OIR  |
|------------------|-----------------|---------|
| None             | 1. check state, <br/>2. create OIR                | 1. check state <br/>2. update OIR <br/>3. `cipasicr()`   |
| Provided         | 1. check sub covers OIR. If not and is implicit: adapt, otherwise fail. <br/>2 .create OIR | 1. check sub covers OIR. If not and is implicit: adapt, otherwise fail <br/>2. update OIR <br/>3. `cipasicr()` |
| Request Implicit | 1. create implicit sub<br/> 2. create OIR  | 1 .create implicit sub <br/>2. update OIR <br/>3. `cipasicr()` |

See also @mickmis's [comment](https://github.com/interuss/dss/pull/1057#issuecomment-2307088229) with a table that focuses purely on the subscription handling:

|                             | implicit old sub               | explicit old sub  | no old sub |
|-----------------------------|--------------------------------|-------------------|------------|
| implicit sub in params      | delete old if no dependent OI  | -                 | -          |
| explicit sub in params      | delete old if no dependent OI  | -                 | -          |
| no sub in params            | delete old if no dependent OI  | -                 | -          |
| request new implicit sub in params  | create new implicit sub, <br/> delete old if no dependent OI     |      create new implicit sub       |   create new implicit sub   |

### Approaches

In a first step, we'll want to factor pieces of `upsertOperationalIntentReference` to into submethods. This should make the upsertion logic's behavior more tractable.

[...]

PR #1089, opened 2024-08-30
[scd] factor out parameter validation for oir upsert method

Shastick:
A first small step towards #1088.

This takes care of moving the parameter validation to a submethod. It is not yet improving anything with respect to the core business logic, but allows readers to more directly focus on the code of what `upsertOperationalIntentReference` does.

Note about the usage of the `stacktrace` library and our dependency on error codes: as we use the error codes to decide what HTTP response to send back, it is important that we properly propagate them when we move logic one or several layers of method calls.

### DIFF
--- a/pkg/scd/operational_intents_handler.go
+++ b/pkg/scd/operational_intents_handler.go
@@ -364,7 +364,6 @@ func (a *Server) UpdateOperationalIntentReference(ctx context.Context, req *rest
 }
 
 type validOIRUpsertParams struct {
-	manager        dssmodels.Manager
 	id             dssmodels.ID
 	ovn            restapi.EntityOVN
 	state          scdmodels.OperationalIntentState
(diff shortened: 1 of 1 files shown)


pkg/scd/operational_intents_handler.go
```diff
@@ -363,95 +363,132 @@ func (a *Server) UpdateOperationalIntentReference(ctx context.Context, req *rest
...
+	id             dssmodels.ID
+	ovn            restapi.EntityOVN
+	state          scdmodels.OperationalIntentState
+	extents        []*dssmodels.Volume4D
+	uExtent        *dssmodels.Volume4D
+	cells          s2.CellUnion
+	subscriptionID dssmodels.ID
+}
+
+func (a *Server) validateAndReturnUpsertParams(
```
mickmis, line 377: Pass `AllowHTTPBaseUrls` as a param to avoid attaching this to `Server`?

pkg/scd/operational_intents_handler.go
```diff
@@ -363,95 +363,132 @@ func (a *Server) UpdateOperationalIntentReference(ctx context.Context, req *rest
...
+	// but this requires some changes in the caller that go beyond the immediate scope of #1088 and can be done later.
+	upsertParams, err := a.validateAndReturnUpsertParams(authorizedManager, entityid, ovn, params)
+	if err != nil {
+		// Important note:
+		//   - stacktrace.Propagate sets NoCode as the code for the error
+		//   - subsequently, stacktrace.GetCode(err) only looks into the first error in the stack and returns NoCode if the error has no code
+		//   - therefore, we need to explicitly use stacktrace.PropagateWithCode if we want the calling logic
+		//   to be able to retrieve the code of the error
+		// TODO we can consider fixing this in the stacktrace library
+		return nil, nil, stacktrace.PropagateWithCode(err, stacktrace.GetCode(err), "Failed to validate Operational Intent Reference upsert parameters")
```
mickmis, line 491: Here I suggest making `validateAndReturnUpsertParams` purely a validation function by:
- move out of `validateAndReturnUpsertParams` the two cases of errors `dsserr.PermissionDenied`
- never returning any code within `validateAndReturnUpsertParams`
- if `err != nil` set the code `dsserr.BadRequest`

That makes it clear this function validates the input to create the op intent and any error implies bad input. That should also enable passing the manager instead of the whole `authorizedManager`.

mickmis (commented):
Nice, thanks! That's indeed a good first step in cutting up this function. With this construct it will also be easy when it makes sense to start factoring away some logic into functions of  `validOIRUpsertParams`
I just have some minor comments.

mickmis (approved):
LGTM - thanks

PR #1090, opened 2024-08-30
[scd] factor out parameter validation against previous OIR

Shastick:
Moves the checks that compare the requesting manager to the one of the (possibly) existing OIR to a submethod.

Note the usage of `stacktrace.GetCode(err)` and `stacktrace.PropagateWithCode()`: I'm not sure we have a more elegant solution such as we had in #1089 if we want to avoid this.

### DIFF
--- a/pkg/scd/operational_intents_handler.go
+++ b/pkg/scd/operational_intents_handler.go
@@ -478,6 +478,36 @@ func checkUpsertPermissionsAndReturnManager(authorizedManager *api.Authorization
 	return dssmodels.Manager(*authorizedManager.ClientID), nil
 }
 
+// validateUpsertRequestAgainstPreviousOIRAndReturnVersion checks that the client requesting an OIR upsert has the necessary permissions and that the request is valid.
+// On success, the version of the OIR is returned:
+//   - upon initial creation (if no previous OIR exists), it is 0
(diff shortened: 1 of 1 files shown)


mickmis (approved):
LGTM - see the nit suggestion. LMK if you implement it or not so that I can merge or wait.

pkg/scd/operational_intents_handler.go
```diff
@@ -478,6 +478,36 @@ func checkUpsertPermissionsAndReturnManager(authorizedManager *api.Authorization
 	return dssmodels.Manager(*authorizedManager.ClientID), nil
 }
 
+// validateUpsertRequestAgainstPreviousOIRAndReturnVersion checks that the client requesting an OIR upsert has the necessary permissions and that the request is valid.
+// On success, the version of the OIR is returned:
+//   - upon initial creation (if no previous OIR exists), it is 0
+//   - otherwise, it is the version of the previous OIR
+func validateUpsertRequestAgainstPreviousOIRAndReturnVersion(
```
mickmis, line 485: nit: do not return version from this function and just perform validation.
Determining version can then just be:
```go
version := 0
if old != nil {
  version = int32(old.Version)
}
```

mickmis (approved):
This time including the comment I'm talking about.

PR #1092, opened 2024-08-30
[scd] oir upsert: factor out determination of subscriptions to notify

Shastick:
Another step towards #1088

### DIFF
--- a/pkg/scd/operational_intents_handler.go
+++ b/pkg/scd/operational_intents_handler.go
@@ -94,8 +94,8 @@ func (a *Server) DeleteOperationalIntentReference(ctx context.Context, req *rest
 			}
 		}
 
-		// Find Subscriptions that may overlap the OperationalIntent's Volume4D
-		allsubs, err := r.SearchSubscriptions(ctx, &dssmodels.Volume4D{
+		// Gather the subscriptions that need to be notified
+		notifyVolume := &dssmodels.Volume4D{
 			StartTime: old.StartTime,
 			EndTime:   old.EndTime,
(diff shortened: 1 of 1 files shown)


pkg/scd/operational_intents_handler.go
```diff
@@ -508,6 +508,56 @@ func validateUpsertRequestAgainstPreviousOIR(
 	return nil
 }
 
+func incrementIndicesAndGetRelevantSubscriptions(
```
BenjaminPelletier, line 511: nit: I think the primary thing this function is doing is getting relevant subscriptions, with a sub-behavior of incrementing notification indices

```suggestion
func getRelevantSubscriptionsAndIncrementIndices(
```

BenjaminPelletier (approved):

pkg/scd/operational_intents_handler.go
```diff
@@ -517,6 +517,56 @@ func validateUpsertRequestAgainstPreviousOIR(
 	return nil
 }
 
+func getRelevantSubscriptionsAndIncrementIndices(
```
mickmis, line 520: Drive-by comment: also use this in `DeleteOperationalIntentReference`? There seems to be some common logic (although maybe there are some small deviations)

Shastick: Good point. I'll quickly check: if the method can be used as-is I'll update to use it, otherwise I'll keep that for another day

Shastick: Ended up doing it as it was an easy fix.

mickmis (approved):

PR #1094, opened 2024-08-30
[scd] oir upsert: factor implicit subscription creation

Shastick:
Also moves the implicit subscription URL check to parameter validation

another step for #1088

### DIFF
--- a/pkg/scd/operational_intents_handler.go
+++ b/pkg/scd/operational_intents_handler.go
@@ -353,18 +353,20 @@ func (a *Server) UpdateOperationalIntentReference(ctx context.Context, req *rest
 }
 
 type validOIRParams struct {
-	id                                 dssmodels.ID
-	ovn                                restapi.EntityOVN
-	state                              scdmodels.OperationalIntentState
-	extents                            []*dssmodels.Volume4D
-	uExtent                            *dssmodels.Volume4D
-	cells                              s2.CellUnion
-	subscriptionID                     dssmodels.ID
-	ussBaseURL                         string
-	implicitSubscriptionRequested      bool
-	implicitSubscriptionBaseURL        string
-	implicitSubscriptionForConstraints bool
-	key                                map[scdmodels.OVN]bool
+	id                   dssmodels.ID
+	ovn                  restapi.EntityOVN
+	state                scdmodels.OperationalIntentState
+	extents              []*dssmodels.Volume4D
+	uExtent              *dssmodels.Volume4D
+	cells                s2.CellUnion
+	subscriptionID       dssmodels.ID
+	ussBaseURL           string
+	implicitSubscription struct {
+		requested      bool
+		baseURL        string
+		forConstraints bool
+	}
+	key map[scdmodels.OVN]bool
 }
 
 // validateAndReturnUpsertParams checks that the parameters for an Operational Intent Reference upsert are valid.
@@ -404,12 +406,16 @@ func validateAndReturnUpsertParams(
 		if params.NewSubscription.UssBaseUrl == "" {
 			return nil, stacktrace.NewError("Missing required USS base url for new subscription (in parameters for implicit subscription)")
 		}
+		// If an implicit subscription is requested, the Subscription ID cannot be present.
 		if params.SubscriptionId != nil {
 			return nil, stacktrace.NewError("Cannot provide both a Subscription ID and request an implicit subscription")
 		}
-		valid.implicitSubscriptionRequested = true
-		valid.implicitSubscriptionBaseURL = string(params.NewSubscription.UssBaseUrl)
-		valid.implicitSubscriptionForConstraints = *params.NewSubscription.NotifyForConstraints
+		valid.implicitSubscription.requested = true
+		valid.implicitSubscription.baseURL = string(params.NewSubscription.UssBaseUrl)
+		// notify for constraints defaults to false if not specified
+		if params.NewSubscription.NotifyForConstraints != nil {
+			valid.implicitSubscription.forConstraints = *params.NewSubscription.NotifyForConstraints
+		}
 	}
 
 	if !allowHTTPBaseUrls {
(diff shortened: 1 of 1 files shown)


pkg/scd/operational_intents_handler.go
```diff
@@ -364,14 +364,18 @@ func (a *Server) UpdateOperationalIntentReference(ctx context.Context, req *rest
...
+	ovn                                restapi.EntityOVN
+	state                              scdmodels.OperationalIntentState
+	extents                            []*dssmodels.Volume4D
+	uExtent                            *dssmodels.Volume4D
+	cells                              s2.CellUnion
+	subscriptionID                     dssmodels.ID
+	ussBaseURL                         string
+	implicitSubscriptionRequested      bool
+	implicitSubscriptionBaseURL        string
+	implicitSubscriptionForConstraints bool
```
mickmis, line 377: nit: use an inner anonymous struct

pkg/scd/operational_intents_handler.go
```diff
@@ -641,40 +689,15 @@ func (a *Server) upsertOperationalIntentReference(ctx context.Context, authorize
...
-					subToUpsert.NotifyForConstraints = *params.NewSubscription.NotifyForConstraints
-				}
-
-				sub, err = r.UpsertSubscription(ctx, &subToUpsert)
-				if err != nil {
+			// Create an implicit subscription if one has been requested.
+			// Requesting neither an explicit nor an implicit subscription is allowed for ACCEPTED states:
+			// for other states, an error will have been returned earlier.
+			// if no implicit subscription is requested and we reached this point, we will proceed without subscription
+			if validParams.implicitSubscriptionRequested {
```
mickmis, line 696: Actually now we don't need anymore to check for `validParams.subscriptionID.Empty()` above: we can directly check for `validParams.implicitSubscriptionRequested`.

pkg/scd/operational_intents_handler.go
```diff
@@ -396,11 +400,42 @@ func validateAndReturnUpsertParams(
...
+		}
+	}
+
+	if params.NewSubscription != nil {
+		// The spec states that NewSubscription.UssBaseUrl is required and an empty value
+		// makes no sense, so we will fail if an implicit subscription is requested but the base url is empty
+		if params.NewSubscription.UssBaseUrl == "" {
+			return nil, stacktrace.NewError("Missing required USS base url for new subscription (in parameters for implicit subscription)")
+		}
+		if params.SubscriptionId != nil {
```
mickmis, line 418: That is checked below (although only a subscription is required). Adding this check here cover more cases but is also duplicate with below. If this behavior change is desired it should be in a separate PR. But I'm not sure that's really necessary, so I'd propose to just remove this check here.

pkg/scd/operational_intents_handler.go
```diff
@@ -396,11 +400,42 @@ func validateAndReturnUpsertParams(
...
+			return nil, stacktrace.NewError("Cannot provide both a Subscription ID and request an implicit subscription")
+		}
+		valid.implicitSubscriptionRequested = true
+		valid.implicitSubscriptionBaseURL = string(params.NewSubscription.UssBaseUrl)
+		valid.implicitSubscriptionForConstraints = *params.NewSubscription.NotifyForConstraints
```
mickmis, line 423: `NotifyForConstraints` may be nil: that was checked before but not anymore: it should be added back.

Shastick: Good catch!

pkg/scd/operational_intents_handler.go
Shastick: Unfortunately not:

`validParams.subscriptionID.Empty()` is allowed to be empty while `validParams.implicitSubscriptionRequested` is false when the OIR is in the `ACCEPTED` state.

When we enter the `validParams.subscriptionID.Empty()` branch, we only want to do something if `validParams.implicitSubscriptionRequested` is true.

If we do `validParams.implicitSubscriptionRequested` in the place where we currently check `validParams.subscriptionID.Empty()`, we will go through the `// Use existing Subscription` branch.

There's probably a better way to express this (possibly with a switch), but I'll be keeping that for later.

pkg/scd/operational_intents_handler.go
Shastick: > That is checked below (although only a subscription is required)

Did you mean that this is checked above? (It used to be checked below but I figured I'd move the format check before this check)

I figured that I'd rather fail early here where we are actually validating the parameters: this might indeed be changing the behavior of the DSS and I can split that to a separate PR for clarity.

Shastick: Done ine https://github.com/interuss/dss/pull/1097

pkg/scd/operational_intents_handler.go
mickmis: Right! Indeed the different is subtle.

pkg/scd/operational_intents_handler.go
```diff
@@ -396,11 +402,44 @@ func validateAndReturnUpsertParams(
...
+			return nil, stacktrace.NewError("Invalid ID format for Subscription ID: `%s`", *params.SubscriptionId)
+		}
+	}
+
+	if params.NewSubscription != nil {
+		// The spec states that NewSubscription.UssBaseUrl is required and an empty value
+		// makes no sense, so we will fail if an implicit subscription is requested but the base url is empty
+		if params.NewSubscription.UssBaseUrl == "" {
+			return nil, stacktrace.NewError("Missing required USS base url for new subscription (in parameters for implicit subscription)")
+		}
```
mickmis, line 419: That's also a slight behavior change, however this would just error out differently later on, so OK to keep in this PR IMO.

pkg/scd/operational_intents_handler.go
mickmis: Thanks for #1097, but was not removed from this one?

mickmis (approved):
LGTM - just however what was split away in #1097 was not removed from this PR.

pkg/scd/operational_intents_handler.go
Shastick: I was expecting to get #1097 merged first, then keep the reordering of the checks in here (and have not yet rebased)

mickmis (approved):

Issue #1078, opened 2024-08-23
[SCD] Enable USS to propose OVN to increase parallelization


LeslieW:
# Implementation tasks
- [x] Update CRDB schemas #1095
- [x] Store past OVNs #1096
- [x] Create fork of astm-utm/Protocol https://github.com/interuss/astm-utm-protocol
- [x] Update OpenAPI definition of interuss/astm-utm-protocol for DSS + USS interuss/astm-utm-protocol#3
- [x] Use interuss/astm-utm-protocol as source for OpenAPI
- [x] Implement new DSS endpoint for requested OVN
- [x] Create USS qualifier test scenario to test new optional behavior for both DSS and USS 

# Original issue
**Is your feature request related to a problem? Please describe.**
Recent load testing for SCD revealed a limitation: a USS may need their OVN in internal storage before it is received and processed from the DSS. The following scenarios come to mind:

- The USS may need to plan many flights in close succession and having the OVNs up front would reduce retries due to a missing OVN.
- Another USS may request op intent details before the DSS response has been received/processed by the owning USS.

**Describe the solution you'd like**
It would be helpful if each USS could optionally provide an OVN to the DSS. Perhaps something similar to below:

Each USS would have a current and proposed version of an operational intent. By default the 
{operational-intent-id}/ endpoint would return the current version. A proposed version becomes the current version when the response is received from the DSS, or if another USS specifically requests that version (meaning the DSS response is still pending or was lost, but the DSS accepted the request). Each USS would additionally have a {operational-intent-id}/{version} endpoint that would return either the current or proposed version. A USS is not required to store previous versions beyond current and proposed.

Visual resources
![ovn_proposal](https://github.com/user-attachments/assets/0f1aa610-379e-4976-8b75-9dd37d804264)

**Describe alternatives you've considered**
Alternatives seem limited as this is a distributed system -- without this distributed transaction a USS can only do a best effort reconciliation which increases load on the DSS or latency when responding to other USSs.  

**Additional context**
@BenjaminPelletier and @Callumdmay may have additional context (much of the above was pulled from a conversation we had).

BenjaminPelletier:
It seems like there are two features mentioned here:

1. Allow a USS to request a specific OVN (such that the USS knows what the OVN is likely to be and can begin effectively serving operational intent details before a response is received from the DSS)
2. Enable USSs to get the details for the correct operational intent version even in the presence of a DSS request that is pending from the perspective of the serving USS

[...]

PR #1095, opened 2024-09-02
[deploy][scd/crdb] Add columns uss_requested_ovn and past_ovns to scd_operations table; Bump schema_version to v3.2.0

mickmis:
First step of #1078: this adds to the cockroach DB `scd_operations` table the `ovn` and `past_ovns` columns.
This change only concerns the DB and does not introduce any behavior change.
Note the `README.md` updates: removal of `*_bootstrapper.sh` (`latest` version is specified there) and addition of `schema_manager.yaml`.

# CRDB schema
A `NULL` value in the `ovn` column signifies that the operational intent uses the existing DSS-generated OVN, which is derived from the timestamp of the last update. If it has a non-`NULL` value, it may not be an empty string, and it signifies that the OVN in force was proposed by the USS.

The `past_ovns` column is intended to contain all past OVNs of the operational intent, both DSS-generated and USS-proposed. This mechanism will be implemented in the next PR.

# Testing
Do note that I've only tested this change by deploying locally the DSS and by validating that the CI passes. I've not tested deployment to the cloud: @barroco should this be tested manually before merging this PR or it this not necessary?


pkg/scd/store/cockroach/operational_intents.go
```diff
@@ -199,6 +222,20 @@ func (s *repo) UpsertOperationalIntent(ctx context.Context, operation *scdmodels
...
 		return nil, stacktrace.Propagate(err, "Failed to convert id to PgUUID")
 	}
+
+	var ovn pgtype.Text // if the OVN is to be generated by the DSS, it must be NULL in DB, not just an empty string
+	if operation.OVN != "" {
+		ovn = pgtype.Text{
+			String: operation.OVN.String(),
+			Valid:  true,
+		}
+	}
```
Shastick, line 232: Side question: if we are going to set the `ovn` field only when it was specified by a client, should we possibly name it `client_specified_ovn`?

Alternatively, could we set it in any case and compute the time-based OVN once at insertion time instead of when it is retrieved from the database? Or is it important to be able to distinguish between client-provided and server-generated OVN, and to do this via a null `ovn` field?

build/db_schemas/scd/upto-v3.2.0-add_ovn_columns.sql
```diff
@@ -0,0 +1,14 @@
+ALTER TABLE scd_operations
+    ADD COLUMN IF NOT EXISTS ovn       STRING
+        CHECK (ovn != ''), -- ovn must be NULL if unspecified, not an empty string
+    ADD COLUMN IF NOT EXISTS past_ovns STRING[] NOT NULL
+        DEFAULT ARRAY []::STRING[]
+        CHECK (
+            array_position(past_ovns, NULL) IS NULL AND
+            array_position(past_ovns, '') IS NULL AND
+            array_position(past_ovns, ovn) IS NULL
+            ); -- past_ovns must not contain NULL elements, empty strings or current ovn
```
Shastick, line 10: Side thought: how big do we expect these arrays to grow? 

I don't believe that enforcing the constraint will ever become a problem, but this still makes me wonder how many modifications of an OIR (which is what will cause this array to grow) we should expect in the worst case.

mickmis: I do not know. But given that we don't use it to query anything, I don't expect its size to have any significant impact.

pkg/scd/store/cockroach/operational_intents.go
mickmis: > Alternatively, could we set it in any case and compute the time-based OVN once at insertion time instead of when it is retrieved from the database? Or is it important to be able to distinguish between client-provided and server-generated OVN, and to do this via a null ovn field?

I did consider that initially, and at the time of making this PR the assumption was that if the provided suffix is not valid, the DSS just ignores it and do as usual.
However that assumption changed so maybe that's to be reconsidered.

mickmis: I've looked into it, but unfortunately there isn't a simple way (i.e. w/o data migration) of computing the OVN at insertion time instead of when it is retrieved. The risk is that with the DSS update the previous OVNs cease to be valid which is of course not acceptable. 
I will update the column however to make it more explicit that this is not the OVN in all cases however.

build/deploy/examples/minimum/main.jsonnet
```diff
@@ -31,7 +31,7 @@ local metadata = metadataBase {
   schema_manager+: {
     image: 'VAR_DOCKER_IMAGE_NAME',
     desired_rid_db_version: '4.0.0',
-    desired_scd_db_version: '3.1.0',
+    desired_scd_db_version: '3.2.0',
```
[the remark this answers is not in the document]
mickmis: Done!

build/db_schemas/scd/upto-v3.2.0-add_ovn_columns.sql
```diff
@@ -0,0 +1,14 @@
+ALTER TABLE scd_operations
+    ADD COLUMN IF NOT EXISTS uss_requested_ovn STRING
+        CHECK (uss_requested_ovn != ''), -- uss_requested_ovn must be NULL if unspecified, not an empty string
+    ADD COLUMN IF NOT EXISTS past_ovns         STRING[] NOT NULL
+        DEFAULT ARRAY []::STRING[]
+        CHECK (
+            array_position(past_ovns, NULL) IS NULL AND
+            array_position(past_ovns, '') IS NULL AND
+            array_position(past_ovns, uss_requested_ovn) IS NULL
+            ); -- past_ovns must not contain NULL elements, empty strings or current uss_requested_ovn
```
mickmis, line 10: @BenjaminPelletier @barroco FTR: the only modification this data migration is doing is to add columns. Given that the queries are only ever referencing explicitly columns (as opposed to a wildcard selection), I think it is safe to conclude that previous versions of the DSS runtime are compatible with this newer version of the schema.

(follow-up to the discussion last tuesday about upgrades during the InterUSS weekly)

PR #1096, opened 2024-09-02
[scd] Store past OVNs on operational intents upsert

mickmis:
PR stacked on top of #1095 and #1100: please only consider beb2042b3308c689bc7b27d572fb23dc81de4dbe for review.
Continuation of #1078 

This enables storing all past OVNs in the DB.

### DIFF
--- a/pkg/scd/operational_intents_handler.go
+++ b/pkg/scd/operational_intents_handler.go
@@ -391,7 +391,7 @@ type validOIRParams struct {
 	key map[scdmodels.OVN]bool
 }
 
-func (vp *validOIRParams) toOIR(manager dssmodels.Manager, attachedSub *scdmodels.Subscription, version scdmodels.VersionNumber) *scdmodels.OperationalIntent {
+func (vp *validOIRParams) toOIR(manager dssmodels.Manager, attachedSub *scdmodels.Subscription, version scdmodels.VersionNumber, pastOVNs []scdmodels.OVN) *scdmodels.OperationalIntent {
 	// For OIR's in the accepted state, we may not have a attachedSub available,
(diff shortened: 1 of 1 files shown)


pkg/scd/operational_intents_handler.go
```diff
@@ -365,7 +365,7 @@ func (a *Server) UpdateOperationalIntentReference(ctx context.Context, req *rest
 
 type validOIRParams struct {
 	id             dssmodels.ID
-	ovn            restapi.EntityOVN
+	ovn            scdmodels.OVN
```
Shastick, line 368: Thanks, that's a convenient change

Shastick: How do we/will we differentiate between a client specifying an OVN as part of a 'normal' mutation (where the current OVN needs to be provided)?

(I see that #1078 mentions a new `requested_ovn_suffix` parameter which I assume will be added in upcoming PRs?)

Shastick (approved):
PR looks good to me (review was limited to commit `4a3db8c`)

Shastick:
Side thought: if the parent PR still needs a bit of work, the current refactoring effort (#1088) would benefit from the change of `ovn` to `scdmodels.OVN`. Could you consider splitting that to a separate PR?

(Otherwise I could also do so)

pkg/scd/operational_intents_handler.go
mickmis: > (I see that https://github.com/interuss/dss/issues/1078 mentions a new requested_ovn_suffix parameter which I assume will be added in upcoming PRs?)

Exactly

PR #1097, opened 2024-09-02
[scd] oir upsert: check uss base URL and if both subscription ID and implicit request are specified

Shastick:
These checks are now done explicitly and early.

This change of behavior was split off from #1094 

It is part of the refactors for #1088


mickmis (approved):
LGTM

PR #1098, opened 2024-09-03
[scd] oir upsert: factor out extension and validation of existing subscription

Shastick:
Another step towards #1088

### DIFF
--- a/pkg/scd/operational_intents_handler.go
+++ b/pkg/scd/operational_intents_handler.go
@@ -701,6 +701,81 @@ func validateKeyAndProvideConflictResponse(
 	return nil, nil
 }
 
+// dependOnExistingSubscriptionIfAllowedAndCovered ensures that:
+//   - the subscription exists
+//   - the subscription is owned by the manager
+//   - the subscription covers the requested geo-temporal extent
+//   - if the subscription is implicit, it is extended if required to cover the requested parameters
+//   - if the subscription is explicit and it is not covering the requested parameters, an error is returned
+//
+// IMPORTANT: this method currently only EXTENDS implicit subscriptions:
+//
+//	It WILL NOT shrink an implicit subscription if that subscription covers more than the area of its dependant OIRs.
+//	This is relevant for situations where clients manually specify implicit subscriptions that were previously
+//	created when creating or updating their OIRs: they may get notifications for areas they are not interested in.
+func dependOnExistingSubscriptionIfAllowedAndCovered(ctx context.Context, r repos.Repository, manager dssmodels.Manager, params *validOIRParams) (*scdmodels.Subscription, error) {
+	if params.subscriptionID.Empty() {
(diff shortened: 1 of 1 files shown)


pkg/scd/operational_intents_handler.go
```diff
@@ -739,49 +814,15 @@ func (a *Server) upsertOperationalIntentReference(ctx context.Context, authorize
...
-				} else {
-					return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Subscription does not cover entire spatial area of the OperationalIntent")
-				}
-			}
-			if updateSub {
-				sub, err = r.UpsertSubscription(ctx, sub)
-				if err != nil {
-					return stacktrace.Propagate(err, "Failed to update existing Subscription")
-				}
+				// Propagating the code here is important for properly handling all failure cases in the caller
```
mickmis, line 820: Doesn't the simple `Propagate` does that in any case? I thought the error code would be unwrapped at the top of the stack.

pkg/scd/operational_intents_handler.go
```diff
@@ -660,6 +660,81 @@ func validateKeyAndProvideConflictResponse(
...
+//	It WILL NOT shrink an implicit subscription if that subscription covers more than the area of its dependant OIRs.
+//	This is relevant for situations where clients manually specify implicit subscriptions that were previously
+//	created when creating or updating their OIRs: they may get notifications for areas they are not interested in.
+func dependOnExistingSubscriptionIfAllowedAndCovered(ctx context.Context, r repos.Repository, manager dssmodels.Manager, params *validOIRParams) (*scdmodels.Subscription, error) {
+	if params.subscriptionID.Empty() {
+		// This situation should have been caught earlier but we check again to be safe
+		return nil, stacktrace.NewError("Cannot depend on a non-specified Subscription")
+	}
+
+	sub, err := r.GetSubscription(context.Background(), params.subscriptionID)
```
mickmis, line 681: Context of function must be used here.

mickmis (commented):
TBH I'm a bit less convinced by factoring away this function, its abstraction isn't very clear, its name and arguments are an indication of that.
I'm also not sure it is worth spending the time refactoring it more.
Now I don't think it makes anything worse than it is currently either.
So I'm OK merging the PR (after addressing the comment), but also OK not merging it, up to you.

pkg/scd/operational_intents_handler.go
Shastick: You're right: if the passed cause has a code, it will be re-used.

(But, for the record, if it has none, there will be no attempt to go look a level deeper for a code)

Shastick:
@mickmis the PR now contains a proposal that hides away less logic

mickmis (commented):
Looks like the case for the start time got lost along the way.
Except from that LGTM indeed it is better that way IMO.

mickmis (approved):

PR #1100, opened 2024-09-06
[scd] Use scdmodels for OVN+VersionNumber and simplify 'version' variable flow

mickmis:
See for context https://github.com/interuss/dss/pull/1096#issuecomment-2326345564
cc/ @Shastick


Shastick (approved):
Thanks!

PR #1107, opened 2024-09-09
[dss] scd: inline oir deletion implicit subscription cleanup into CRDB calls

Shastick:
First part of #1059 

This pushes down the Go logic added in #1104 into CRDB my means of sub-requests.

The new qualifier scenario [checking for implicit sub cleanups](https://github.com/interuss/monitoring/pull/720) succeeds on this PR when ran locally.

### DIFF
--- a/pkg/scd/operational_intents_handler.go
+++ b/pkg/scd/operational_intents_handler.go
@@ -111,11 +111,6 @@ func (a *Server) DeleteOperationalIntentReference(ctx context.Context, req *rest
 			}
 		}
 
-		removeImplicitSubscription, err := subscriptionIsImplicitAndOnlyAttachedToOIR(ctx, r, id, previousSubscription)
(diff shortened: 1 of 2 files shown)


mickmis:
Doesn't this qualify as relying more on CRDB? Is this acceptable? Also for #1064
cc/ @barroco @BenjaminPelletier

Shastick:
> Doesn't this qualify as relying more on CRDB? Is this acceptable? Also for #1064 cc/ @barroco @BenjaminPelletier

Any backend that offers an SQL interface and supports transactions should work with this: subqueries are a feature of the SQL standard and not limited to CRDB.

Shastick:
Closing given that we're migrating to Yugabyte. Will reopen or create a new Yugabyte compatible version if/when inlining is relevant again.

Issue #814, opened 2022-09-21
Reduce likelihood of contention in transactions


BenjaminPelletier:
Currently, there are circumstances when many transactions may fundamentally contend to change the same data fields.  Specifically, if there are many flights being updated in the same area at the same time, the notification indices for each flight's associated subscription will need to be updated for each change to a flight in the area.  We can reduce the likelihood of a deadlock or poor database behavior in two ways:

1. Always sort ID parameters before [passing them in a SQL command](https://github.com/interuss/dss/blob/c399b901c89c1761425dda664cfce3d712a5f11b/pkg/rid/store/cockroach/subscriptions.go#L280) since they will be evaluated in series rather than parallel
2. Add a "for update" flag when reading database entities to use a `SELECT FOR UPDATE` SQL command where appropriate

Shastick:
@BenjaminPelletier, before I go one with adding optional "FOR UPDATE" to various requests, may I ask you to check https://github.com/interuss/dss/pull/1117 and confirm that this is the approach you were thinking off?

PR #1112, opened 2024-09-12
[dss] sort cell ids in queries

Shastick:
Sorts cell identifiers (usually stored in the `cells` column of various entities) in CRDB queries.

This is expected to reduce contention in two ways:
- similar queries will be guaranteed to access data in a similar order, helping with distributed locking
- sorted arrays may be easier to compare with the `&&` operator

All queries that contain a condition of the form `cells && <some_int64_array>` depend on `sql.CellUnionToCellIds()` or `sql.CellUnionToCellIdsWithValidation()`, so this PR simply added `slices.Sort()` step to both of these methods.

First step of #814


mickmis:
What's the status of this PR? @Shastick

pkg/sql/utils.go
```diff
@@ -24,5 +26,7 @@ func CellUnionToCellIdsWithValidation(cu s2.CellUnion) ([]int64, error) {
 		}
 		pgCids[i] = int64(cell)
 	}
+	// Sort the cell IDs for optimisation purpose (see https://github.com/interuss/dss/issues/814)
+	slices.Sort(pgCids)
```
mickmis: Call `CellUnionToCellIds` here instead of duplicating logic?

Shastick: I'm not sure I understand the remark?

`CellUnionToCellIdsWithValidation` is doing the same thing as `CellUnionToCellIds`, but validates each individual cell before adding the cell to `pgCids`.

Calling `CellUnionToCellIds` would mean doing two iterations on the `s2.CellUnion` and provide no benefit in exchange?

Do you mean you'd rather do an iteration for validating (without allocating `pgCids) and then calling `CellUnionToCellIds`?

I don't know how many cells a `s2.CellUnion` will contain and I guess it needs to be a lot for this to have an impact on performance, but given that:
 - it's a PR targeted at performance
 - in the long run `CellUnionToCellIds` should possibly be removed because it does not validation
 
 I went with having some minimal duplication.

mickmis: > Calling CellUnionToCellIds would mean doing two iterations on the s2.CellUnion and provide no benefit in exchange?

Avoid code duplication?

> Do you mean you'd rather do an iteration for validating (without allocating pgCids) and then calling CellUnionToCellIds`?

Yep

> Do you mean you'd rather do an iteration for validating (without allocating pgCids) and then calling CellUnionToCellIds`?

In terms of performance the complexity of the sorting will anyway be an order of magnitude higher than iterating on the slice.

But in any case we can easily do that then:
```go

func cellUnionToCellIds(cu s2.CellUnion, doValidation bool) ([]int64, error) {
	pgCids := make([]int64, len(cu))
	for i, cell := range cu {
		if doValidation {
			if err := geo.ValidateCell(cell); err != nil {
				return nil, stacktrace.Propagate(err, "Error validating cell")
			}
		}
		pgCids[i] = int64(cell)
	}
	// Sort the cell IDs for optimisation purpose (see https://github.com/interuss/dss/issues/814)
	slices.Sort(pgCids)
	return pgCids, nil
}

func CellUnionToCellIds(cu s2.CellUnion) []int64 {
    cIDs, _ := cellUnionToCellIds(cu, false) 
    return cIDs
}

func CellUnionToCellIdsWithValidation(cu s2.CellUnion) ([]int64, error) {
    return cellUnionToCellIds(cu, true)
}
```

mickmis (approved):
LGTM

PR #1113, opened 2024-09-12
[dss/SCD] early subscription lock for OIR deletions

Shastick:
The early locking already happens for insertions/mutations, there is no reason not to do it for deletions.

Part of #814

### DIFF
--- a/pkg/scd/operational_intents_handler.go
+++ b/pkg/scd/operational_intents_handler.go
@@ -99,6 +99,13 @@ func (a *Server) DeleteOperationalIntentReference(ctx context.Context, req *rest
 				"Current version is %s but client specified version %s", old.OVN, ovn)
 		}
 
+		// Early lock on the subscriptions covering the cells relevant to the OIR
+		// See issue #1002 for details.
+		err = r.LockSubscriptionsOnCells(ctx, old.Cells)
+		if err != nil {
+			return stacktrace.Propagate(err, "Unable to acquire lock")
+		}
+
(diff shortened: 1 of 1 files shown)


pkg/scd/operational_intents_handler.go
```diff
@@ -99,6 +99,13 @@ func (a *Server) DeleteOperationalIntentReference(ctx context.Context, req *rest
 				"Current version is %s but client specified version %s", old.OVN, ovn)
 		}
 
+		// Early lock on the subscriptions covering the cells relevant to the OIR
+		// See issue #1002 for details.
+		err = r.LockSubscriptionsOnCells(ctx, old.Cells)
```
BenjaminPelletier, line 104: I would expect this lock to happen as basically the first thing in the transaction (similar to upsert); is there a reason we are leaving lines 82-100 out of the lock?

Shastick: Before having obtained the operational being deleted, we don't actually know which cells to lock.

We could call `r.LockSubscriptionsOnCells` at line 84 already, but not earlier. 

I willing let the checks before the locking to avoid (relatively) expensive locking operations in situations where the OIR is not owned by the caller or if the OVN is wrong.

Note that in the OIR upsert logic, the lock is not perfect, as we use the cells specified in the request to lock subscriptions, but not the ones specified in the possibly existing OIR. (This is likely not dramatic)

BenjaminPelletier (approved):

PR #1119, opened 2024-09-13
[scd] Enable USSs to request an OVN for operational intents

mickmis:
Final step towards addressing #1078.
Still pending testing with upcoming uss_qualifier test scenario, but review welcomed in the meantime in any case.


mickmis:
In interuss/monitoring#816 I could test that this works as expected. Merging this PR.

PR #1122, opened 2024-09-20
[cmd] Enable db-manager CLI to support subcommands

mickmis:
This PR refactors the `db-manager` command to support subcommands (notably for #1116).
- The existing `db-manager` logic is now defined as the `migration` subcommand. Various updates of scripts, deployment files, and documentations are made in that regard.
- Introduce use of spf13/cobra and spf13/pflag libraries to support this.
- Note that not all flags are migrated to using spf13/pflag, however the existing ones that use `flag` are still supported (notably for the cockroachdb flags).
- `go mod tidy` was ran.
- The diff is unfortunately not very readable, however here is a change overview that should help review:
  - `cmds/db-manager/main.go` is a new file
  - `cmds/db-manager/main.go` was moved to `cmds/db-manager/migration/migration.go` and adapted:
    - panics are replaced by returning errors
    - flags are defined with spf13/pflag
    - context is passed from parent
    - check for parameter `schemas_dir` is replaced by use of function `MarkFlagRequired`

### DIFF
--- a/cmds/core-service/README.md
+++ b/cmds/core-service/README.md
@@ -40,11 +40,11 @@ docker container run -p 26257:26257 -p 8080:8080 --rm cockroachdb/cockroach:v24.
 Once an initialized CockroachDB cluster is available, the necessary databases within the CRDB cluster must be created/configured properly.  This can be accomplished with [migrate_local_db.sh](../../build/dev/migrate_local_db.sh), as documented in the [standalone instance documentation](../../build/dev/standalone_instance.md), when using the standard standalone development DSS instance, or it can be accomplished manually with commands similar to those below starting from the repo root folder:
 
 ```bash
-go run ./cmds/db-manager \
+go run ./cmds/db-manager migrate \
   --schemas_dir ./build/db_schemas/rid \
   --db_version latest \
   --cockroach_host localhost
-go run ./cmds/db-manager \
+go run ./cmds/db-manager migrate \
   --schemas_dir ./build/db_schemas/scd \
   --db_version latest \
   --cockroach_host localhost
--- a/cmds/db-manager/main.go
+++ b/cmds/db-manager/main.go
@@ -1,0 +1,29 @@
+package main
+
+import (
+	"flag"
+	"log"
+	"os"
+
+	"github.com/interuss/dss/cmds/db-manager/migration"
+	"github.com/spf13/cobra"
+)
+
+var (
+	DBManagerCmd = &cobra.Command{
+		Use:   "db-manager",
+		Short: "DSS database management utility",
+	}
+)
+
+func init() {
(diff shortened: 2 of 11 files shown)


cmds/db-manager/main.go
```diff
@@ -1,250 +1,29 @@
...
-	semver.Sort(versions)
-
-	// Render sorted step list
-	result := make([]MigrationStep, len(versions)+1)
-	result[0].version = *semver.New("0.0.0")
-	for i := 0; i < len(versions); i++ {
-		result[i+1] = steps[*versions[i]]
+	DBManagerCmd = &cobra.Command{
+		Use:   "db-manager",
+		Short: "DSS database management",
```
BenjaminPelletier, line 15:
```suggestion
		Short: "DSS database management utility",
```

(nit: I think this is a better answer to "what is db-manager?")

cmds/core-service/README.md
```diff
@@ -40,11 +40,11 @@ docker container run -p 26257:26257 -p 8080:8080 --rm cockroachdb/cockroach:v24.
 Once an initialized CockroachDB cluster is available, the necessary databases within the CRDB cluster must be created/configured properly.  This can be accomplished with [migrate_local_db.sh](../../build/dev/migrate_local_db.sh), as documented in the [standalone instance documentation](../../build/dev/standalone_instance.md), when using the standard standalone development DSS instance, or it can be accomplished manually with commands similar to those below starting from the repo root folder:
 
 ```bash
-go run ./cmds/db-manager \
+go run ./cmds/db-manager migration \
```
BenjaminPelletier, line 43: This all seems great except I think the action is better suited to a verb ("migrate") rather than a noun ("migration") -- `db-manager migrate <options>` seems more intuitive to me than `db-manager migration <options>` since I think the latter can really only be implying one thing ("perform the 'migrate' action") whereas the latter could potentially be implying one of multiple things ("validate the ability to perform a migration", "create a new migration", "determine the status of a migration", "abort a migration").

I think the package name of `migration` is fine since all its content relates to migration, but it seems like the command should be a verb.

BenjaminPelletier (approved):
It is annoying when git doesn't recognize a move + edit.  When that is known to be coming, the move can be done in the first commit (updating references, etc as necessary) and then the changes can be done in a second commit so reviewers can look at the changes separately from the move.  But not worth doing for this PR, I think.

cmds/core-service/README.md
mickmis: Agreed

mickmis:
> It is annoying when git doesn't recognize a move + edit. When that is known to be coming, the move can be done in the first commit (updating references, etc as necessary) and then the changes can be done in a second commit so reviewers can look at the changes separately from the move. But not worth doing for this PR, I think.

Indeed, I've not thought about it, thanks for the tip. Ended up doing it in this PR it was fast enough.

PR #1133, opened 2024-11-19
[crdb] Refactor cockroach package to prepare for Yugabyte compatible migration

barroco:
In preparation to switching to yugabyte. The objective is to keep both implementations side by side before switching it completely once a decision has been made on the long term plan. 
This PR renames the package cockroach and reorganize it to make it more generic in order to accomodate future Yugabyte queries. An upcoming PR will introduce the Yugabyte compatibility changes.
No behaviour change is expected.
Note that this PR introduces the ability to detect the type of database.

Commits have been prepared to simplify review.

### DIFF
--- a/pkg/datastore/datastore.go
+++ b/pkg/datastore/datastore.go
@@ -10,15 +10,8 @@ import (
 )
 
 type Datastore struct {
-	Database
 	Version *Version
 	Pool    *pgxpool.Pool
-}
-
-type Database interface {
-	GetSchemaVersion(ctx context.Context, dbName string) (*semver.Version, error)
-	DatabaseExists(ctx context.Context, name string) (bool, error)
-	CreateDatabase(ctx context.Context, name string) error
 }
 
 var UnknownVersion = &semver.Version{}
--- a/pkg/datastore/version.go
+++ b/pkg/datastore/version.go
@@ -1,7 +1,6 @@
 package datastore
 
 import (
-	"fmt"
 	"github.com/coreos/go-semver/semver"
 	"github.com/interuss/stacktrace"
 	"go.uber.org/multierr"
@@ -43,8 +42,4 @@ func NewVersion(fullVersion string) (*Version, error) {
 	}
 
 	return nil, stacktrace.Propagate(multierr.Combine(err, err2), "Unable to extract datastore type and version")
-}
-
-func (v *Version) String() string {
-	return fmt.Sprintf("%s@%s", v.Type, v.SemVer.String())
 }
--- a/cmds/db-manager/migration/migrate.go
+++ b/cmds/db-manager/migration/migrate.go
@@ -85,7 +85,7 @@ func migrate(cmd *cobra.Command, _ []string) error {
 		ds.Pool.Close()
 	}()
 
-	log.Printf("CRDB server version: %s", ds.Version.String())
+	log.Printf("CRDB server version: %s", ds.Version.SemVer.String())
 
 	// Make sure specified database exists
 	exists, err := ds.DatabaseExists(ctx, dbName)


pkg/datastore/version.go
```diff
@@ -0,0 +1,64 @@
...
+type Version struct {
+	version *semver.Version
+	dsName  string
+}
+
+var COCKROACHDB = &VersionRegex{"cockroach", `v((0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*))`}
+var YUGABYTE = &VersionRegex{"yugabyte", `-YB-((0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*))`}
+
+func (dt *VersionRegex) ParseVersion(fullVersion string) (*semver.Version, error) {
+	re := regexp.MustCompile(dt.VersionRegex)
```
mickmis, line 24: Store this as a variable instead of the regex string: since it is hardcoded better panic early rather than when attempting to use it.

pkg/datastore/version.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (m *Version) IsCockroachDB() bool {
+	return m.dsName == COCKROACHDB.Name
+}
+
+func (m *Version) IsYugabyte() bool {
+	return m.dsName == YUGABYTE.Name
+}
+
+func (m *Version) String() string {
```
mickmis, line 62: This doesn't seem to produce the same version than is accepted by `ParseVersion`, is this expected?

pkg/datastore/version.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (m *Version) Version() *semver.Version {
+	return m.version
+}
+
+func (m *Version) DbType() string {
+	return m.dsName
+}
+
+func VersionFromString(fullVersion string) (*Version, error) {
```
mickmis, line 40: All in all I suggest removing altogether `VersionRegex`, `COCKROACHDB` and `YUGABYTE`; and inlining `ParseVersion` in here to avoid additional abstractions. `IsCockroachDB` and `IsYugabyte` could just use an internal flag.

pkg/datastore/connectParameters_test.go
```diff
@@ -1,10 +1,29 @@
-package cockroach
+package datastore
 
 import (
 	"github.com/stretchr/testify/require"
 	"testing"
 )
 
+// ConnectParametersFromMap constructs a ConnectParameters instance from m.
+func connectParametersFromMap(m map[string]string) ConnectParameters {
```
mickmis, line 9: :+1:

pkg/datastore/datastore.go
```diff
@@ -0,0 +1,150 @@
...
+func initDatastore(ctx context.Context, pool *pgxpool.Pool) (*Datastore, error) {
+	version, err := fetchVersion(ctx, pool)
+	if err != nil {
+		return nil, err
+	}
+
+	if version.IsCockroachDB() {
+		return &Datastore{Version: version, Pool: pool}, nil
+	}
+	return nil, fmt.Errorf("%s is not implemented yet", version.dsName)
```
mickmis, line 66: Use stacktrace.Propagate?

pkg/datastore/datastore.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, stacktrace.Propagate(err, "Error querying datastore version")
+	}
+
+	return VersionFromString(fullVersion)
+}
+
+func (ds *Datastore) CreateDatabase(ctx context.Context, dbName string) error {
+	createDB := fmt.Sprintf("CREATE DATABASE %s", dbName)
+	if _, err := ds.Pool.Exec(ctx, createDB); err != nil {
+		return fmt.Errorf("failed to create new database %s: %v", dbName, err)
```
mickmis, line 85: Use stacktrace.Propagate?

pkg/datastore/datastore.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (ds *Datastore) DatabaseExists(ctx context.Context, dbName string) (bool, error) {
+	const checkDbQuery = `
+		SELECT EXISTS (
+			SELECT * FROM pg_database WHERE datname = $1
+		)`
+
+	var exists bool
+	if err := ds.Pool.QueryRow(ctx, checkDbQuery, dbName).Scan(&exists); err != nil {
+		return false, err
```
mickmis, line 98: Use stacktrace.Propagate?

pkg/datastore/datastore.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	return exists, nil
+}
+
+// GetSchemaVersion returns the Schema Version of the requested DB Name
+func (ds *Datastore) GetSchemaVersion(ctx context.Context, dbName string) (*semver.Version, error) {
```
mickmis, line 105: This is a copied from `(db *DB) GetVersion`. So remove the other one?

mickmis (commented):
All comments are minor, otherwise LGTM

pkg/datastore/datastore.go
mickmis: Irrelevant

pkg/datastore/connectParameters.go
```diff
@@ -0,0 +1,109 @@
...
+
+	dsnMap["pool_max_conns"] = fmt.Sprintf("%d", cp.MaxOpenConns)
+
+	if sslMode == "disable" {
+		return formatDSN(dsnMap), nil
+	}
+
+	dir := cp.SSL.Dir
+	if dir == "" {
+		return "", stacktrace.NewError("Missing crdb ssl_dir")
```
mickmis, line 102: All `stacktrace.NewError` mention 'crdb': should be adapted.

pkg/datastore/datastore.go
```diff
@@ -0,0 +1,150 @@
...
+	"time"
+)
+
+type Datastore struct {
+	Database
+	Version *Version
+	Pool    *pgxpool.Pool
+}
+
+type Database interface {
```
mickmis, line 18: This interface actually does not seem required? It is never used.

pkg/datastore/datastore.go
```diff
@@ -0,0 +1,150 @@
...
+	"context"
+	"fmt"
+	"github.com/coreos/go-semver/semver"
+	"github.com/interuss/stacktrace"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"time"
+)
+
+type Datastore struct {
+	Database
```
mickmis, line 13: Any reason why embedding that interface in the stuct?

pkg/datastore/version.go
```diff
@@ -0,0 +1,50 @@
...
+
+	v, err2 := parseVersion(fullVersion, yugabyteRegex)
+	if err2 == nil {
+		return &Version{v, Yugabyte}, nil
+	}
+
+	return nil, stacktrace.Propagate(multierr.Combine(err, err2), "Unable to extract datastore type and version")
+}
+
+func (v *Version) String() string {
```
mickmis, line 48: It looks surprising that this marshalls to a string that is different from the one accepted by `NewVersion`. Is that expected?

mickmis (approved):
Few minor comments, otherwise LGTM

PR #1134, opened 2024-11-25
Scd subscriptions ttl

callumdmay:
There has been prior effort in addressing dangling subscriptions in #1109 and #1088, however in our load testing efforts on v0.18.0 we have observed that some expired subscriptions are not being cleaned up. The root cause of this should be identified eventually, but as a mitigation we can add a row-level ttl to the table to prevent unbounded table growth.


callumdmay:
Hey @barroco, what's the advantage of using the command over enforcing a ttl? That would do the cleanup automatically. Happy to adjust the ttl window as well if we would like to keep around expired subscriptions for longer.

mickmis:
What's the status of this PR? Shall we close it given the current ongoing effort of migrating to Yugabyte? @barroco

PR #1138, opened 2024-12-02
[yugabyte] SQL Migrations

barroco:
This PR adds migration scripts for yugabyte and adapts the migration CLI accordingly.

## Migration
Yugabyte does not support the exact same data types and declaration syntax as CockroachDB, therefore the following adaptations have been made:

Data Types ([CockroachDB](https://www.cockroachlabs.com/docs/stable/data-types) -> [Yugabyte](https://docs.yugabyte.com/preview/api/ysql/datatypes/)):
1. INT64 -> BIGINT
2. STRING -> TEXT

Indexes:
1. Inverted indexes are implemented using [ybgin](https://docs.yugabyte.com/preview/explore/ysql-language-features/indexes-constraints/gin/)

Syntax required changes:
1. INDEX can't be declared in the CREATE TABLE block.
2. Keyword STORING is replaced by INCLUDE.
3. Index names must be unique.

## Additional notes
- It was agreed to squash the migration to reflect rid v4.0.0 and scd v3.2.0 of the crdb schema.

## CI and Tests
- A job in the CI has been added to cover changes dedicated to Yugabyte migration.
- Migrations up and down have been successfully tested locally.

### DIFF
--- a/cmds/db-manager/migration/migrate.go
+++ b/cmds/db-manager/migration/migrate.go
@@ -115,13 +115,16 @@ func migrate(cmd *cobra.Command, _ []string) error {
 	}
 
 	// Reconnect to target database
-	ds, err = connectTo(ctx, dbName)
+	ds2, err := connectTo(ctx, dbName)
 	if err != nil {
 		return fmt.Errorf("failed to reconnect to database %s: %w", dbName, err)
 	}
+	defer func() {
+		ds2.Pool.Close()
+	}()
 
 	// Read current schema version of database
-	currentVersion, err := ds.GetSchemaVersion(ctx, dbName)
+	currentVersion, err := ds2.GetSchemaVersion(ctx, dbName)
 	if err != nil {
 		return fmt.Errorf("failed to get current database version for %s: %w", dbName, err)
 	}
@@ -168,7 +171,7 @@ func migrate(cmd *cobra.Command, _ []string) error {
 		if isCockroach {
 			// Ensure SQL session has implicit transactions disabled for CRDB versions 22.2+
 			sessionConfigurationSQL := ""
-			if ds.Version.SemVer.Compare(*semver.New("22.2.0")) >= 0 {
+			if ds2.Version.SemVer.Compare(*semver.New("22.2.0")) >= 0 {
 				sessionConfigurationSQL = "SET enable_implicit_transaction_for_batch_statements = false;\n"
 			}
 
@@ -180,7 +183,7 @@ func migrate(cmd *cobra.Command, _ []string) error {
 		}
 
 		// Execute migration step
-		if _, err := ds.Pool.Exec(ctx, migrationSQL); err != nil {
+		if _, err := ds2.Pool.Exec(ctx, migrationSQL); err != nil {
 			return fmt.Errorf("failed to execute %s migration step %s: %w", dbName, fullFilePath, err)
 		}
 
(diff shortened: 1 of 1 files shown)


cmds/db-manager/migration/migrate.go
```diff
@@ -157,25 +173,37 @@ func migrate(cmd *cobra.Command, _ []string) error {
 
 		// Ensure SQL session has implicit transactions disabled for CRDB versions 22.2+
 		sessionConfigurationSQL := ""
-		if ds.Version.SemVer.Compare(*semver.New("22.2.0")) >= 0 {
+		if isCockroach && ds.Version.SemVer.Compare(*semver.New("22.2.0")) >= 0 {
 			sessionConfigurationSQL = "SET enable_implicit_transaction_for_batch_statements = false;\n"
```
mickmis, line 177: Put this inside the below block `if isCockroach` out of clarity?

cmds/db-manager/migration/migrate.go
```diff
@@ -157,25 +173,37 @@ func migrate(cmd *cobra.Command, _ []string) error {
...
+		migrationSQL := ""
+		if isCockroach {
+			migrationSQL = sessionConfigurationSQL + fmt.Sprintf("USE %s;\n", dbName) + string(rawMigrationSQL)
+		} else {
+			ds, err = datastore.Dial(ctx, connectParameters)
+			if err != nil {
+				return fmt.Errorf("failed to reconnect to database %s: %w", dbName, err)
+			}
+			migrationSQL = sessionConfigurationSQL + string(rawMigrationSQL)
+		}
```
mickmis, line 189: While functionally equivalent, shouldn't this be inside a `if isYugabyte` block to be conceptually correct? IIUC this is required for yugabyte to reconnect to a different DB, this is not required because 'the DB is not cockroach'.

build/db_schemas/yugabyte/rid/upto-v1.0.0-crdb_v4.0.0.sql
```diff
@@ -0,0 +1,45 @@
...
+CREATE INDEX subs_by_time_with_owner ON subscriptions (ends_at) INCLUDE (owner);
+
+CREATE TABLE identification_service_areas (
+    id UUID PRIMARY KEY,
+    owner TEXT NOT NULL,
+    url TEXT NOT NULL,
+    starts_at TIMESTAMPTZ,
+    ends_at TIMESTAMPTZ,
+    updated_at TIMESTAMPTZ NOT NULL,
+    cells BIGINT[],
```
mickmis, line 29: This is missing `NOT NULL` constraint,  while the `cells` on `subscriptions` has it. 
I'm not sure I understand that part of the crdb migrations:
https://github.com/interuss/dss/blob/ed31e91c617fe23b9edeab3a2ca2e885ba7d34f4/build/db_schemas/rid/upto-v2.0.0-add_inverted_indices.sql#L38-L43

If at the end we are supposed to have or not a `NOT NULL` constraint, but in any case it looks like it should be consistent between `subscriptions` and `operations`.

mickmis (approved):
Full review, LGTM, just minor comments/questions to address before merging.

BenjaminPelletier (approved):

BenjaminPelletier:
Just to document my understanding of the reasoning behind "It was agreed to squash the migration to reflect rid v4.0.0 and scd v3.2.0 of the crdb schema.":

We have no intent to support the ability to users to switch between database backends at any time, and we further expect any potential migration of data between database backends to be a one-time occurrence.  Therefore, the process of switching to YugabyteDB from an earlier schema version simply includes the step of migrating to rid v4.0.0 and scd v3.2.0 as the first part of the process.  YugabyteDB schemas will be on an entirely separate track from CRDB schemas and therefore there is no need to synchronize schema versions.  Instead, any particular DSS release will have a minimum and maximum CRDB schema version supported (for the duration that CRDB is still supported) and a separate minimum and maximum YugabyteDB schema version supported.

PR #1142, opened 2025-01-02
[cleanup] Remove rid database v3 backward compatibility (RID schema version 4 is required after this commit)

barroco:
Back in 2021, support for backward compatiblity was kept to keep support for schema v3 and allow production users to migrate to v4 at their pace. 3 years later and due to the opportunity to simplify the switch to yugabyte, this PR removes the support for rid database v3 schemas.

### DIFF
--- a/pkg/rid/store/cockroach/store.go
+++ b/pkg/rid/store/cockroach/store.go
@@ -100,10 +100,11 @@ func (s *Store) CheckCurrentMajorSchemaVersion(ctx context.Context) error {
 
 // Interact implements store.Interactor interface.
 func (s *Store) Interact(ctx context.Context) (repos.Repository, error) {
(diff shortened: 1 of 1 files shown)


pkg/rid/store/cockroach/store.go
```diff
@@ -100,15 +100,10 @@ func (s *Store) CheckCurrentMajorSchemaVersion(ctx context.Context) error {
 
 // Interact implements store.Interactor interface.
 func (s *Store) Interact(ctx context.Context) (repos.Repository, error) {
-	logger := logging.WithValuesFromContext(ctx, s.logger)
```
mickmis, line 103: This change looses the values from `ctx`.

mickmis (approved):
LGTM with a minor issue to be fixed before merging, see comment.

PR #1143, opened 2025-01-02
[yugabyte] Yugabyte compatible implementation

barroco:
This PR introduces the ability to run core-service with a yugabyte datastore.

Key changes:
1. ybgin indexes on cells have been removed since they don't support to be scanned with other indexes (which is always the case).
2. SQL query syntax has been changed to be supported by CRDB as well as Yugabyte. No behaviour change is expected.

Note that it was identified that the current implementation may show some failing test runs of the prober with Yugabyte. Current assumption is that the problem is due to default transaction isolation which is not the same in CRDB (Serializable) and [Yugabyte (Snapshot)](https://docs.yugabyte.com/preview/explore/transactions/isolation-levels/). The later may lead to serialization anomalies for single statement queries. Adapting the code to support this specific use case with Yugabyte may require extensive refactoring which may better be addressed in a separate PR.
Prober tests are catching this issue unreliably. It has been kept disabled for Yugabyte until the implementation is stable.

Commits have been organised in order to facilitate the review.

### DIFF
--- a/build/db_schemas/README.md
+++ b/build/db_schemas/README.md
@@ -21,3 +21,10 @@ places:
 * /pkg/{rid|scd}/store/cockroach/store.go
 * /deploy/infrastructure/dependencies/terraform-commons-dss/default_latest.tf
 * /deploy/services/helm-charts/dss/templates/schema-manager.yaml
+
+## Yugabyte schema versions
+
+Versions 1.0.0 of the schemas reflect the latest versions of the crdb schemas. If 
+some adaptations are required during the development phase until the first release,
+changes should be done using version 1.0.1. This paragraph may be removed after the
+first release.


build/dev/docker-compose_dss.yaml
```diff
@@ -105,13 +110,20 @@ services:
...
+        condition: service_completed_successfully
+        required: false
     networks:
       - dss_sandbox_default_network
     healthcheck:
       test: wget -O - 'http://localhost/healthy' || exit 1
       interval: 3m
       start_period: 30s
       start_interval: 5s
+    profiles: ["", "with-yugabyte"]
```
mickmis, line 127: Why the empty string? Is this supposed to represent 'no profile'?

pkg/rid/store/cockroach/subscriptions.go
```diff
@@ -87,7 +87,7 @@ func (r *repo) MaxSubscriptionCountInCellsByOwner(ctx context.Context, cells s2.
 	// strict we could keep this count in memory, (or in some other storage).
 	var query = `
```
mickmis, line 88: `q1` and `q2` are added but never used. Are they really necessary for yagabyte?

pkg/scd/store/cockroach/constraints.go
```diff
@@ -127,13 +127,38 @@ func (c *repo) GetConstraint(ctx context.Context, id dssmodels.ID) (*scdmodels.C
...
+			%s = $7,
+			%s = $8,
+			%s = $9,
+			%s = transaction_timestamp()
+		WHERE scd_constraints.%s = $1
 		RETURNING
-			%s`, constraintFieldsWithoutPrefix, constraintFieldsWithPrefix)
+			%s`,
+			constraintFieldsWithoutPrefix,
+			constraintFieldsWithIndices[0],
```
mickmis, line 149: If those are referenced individually, might as well directly set them in the query, this would be more readable.

pkg/scd/store/cockroach/constraints.go
```diff
@@ -127,13 +127,38 @@ func (c *repo) GetConstraint(ctx context.Context, id dssmodels.ID) (*scdmodels.C
...
+		SET %s = $2,
+			%s = $3,
+			%s = $4,
+			%s = $5,
+			%s = $6,
+			%s = $7,
+			%s = $8,
+			%s = $9,
+			%s = transaction_timestamp()
+		WHERE scd_constraints.%s = $1
```
mickmis, line 145: The `WHERE` should not be required?
(it is not specified for availabilities)

pkg/scd/store/cockroach/constraints.go
```diff
@@ -203,7 +228,8 @@ func (c *repo) SearchConstraints(ctx context.Context, v4d *dssmodels.Volume4D) (
 				COALESCE(starts_at <= $3, true)
 			AND
 				COALESCE(ends_at >= $2, true)
-			LIMIT $4`, constraintFieldsWithoutPrefix)
+			LIMIT $4;
+			`, constraintFieldsWithoutPrefix)
```
mickmis, line 232: Is setting a final `;` required for yugabyte? If so, why is it not the case of query on scd_constraints and availabilities above? And if not, why adding it here?

pkg/scd/store/cockroach/operational_intents.go
```diff
@@ -196,13 +196,44 @@ func (s *repo) DeleteOperationalIntent(ctx context.Context, id dssmodels.ID) err
...
 		upsertOperationsQuery = fmt.Sprintf(`
-			UPSERT INTO
+			INSERT INTO
 				scd_operations
 				(%s)
 			VALUES
 				($1, $2, $3, $4, $5, $6, $7, $8, $9, transaction_timestamp(), $10, $11, $12, $13)
-			RETURNING
-				%s`, operationFieldsWithoutPrefix, operationFieldsWithPrefix)
+			ON CONFLICT (%s) DO UPDATE
```
mickmis, line 204: Same as above: better specify directly the field names if they are referenced individually.

pkg/scd/store/cockroach/subscriptions.go
```diff
@@ -185,15 +185,42 @@ func (c *repo) pushSubscription(ctx context.Context, q dsssql.Queryable, s *scdm
...
+				%s = $7,
+				%s = $8::bool,
+				%s = $9,
+				%s = $10,
+				%s = $11,
+				%s = transaction_timestamp()
 		RETURNING
-			%s`, subscriptionFieldsWithoutPrefix, subscriptionFieldsWithPrefix)
+			%s`,
+			subscriptionFieldsWithoutPrefix,
```
mickmis, line 209: Same as above: better specify directly the field names if they are referenced individually.

pkg/scd/store/cockroach/subscriptions.go
```diff
@@ -185,15 +185,42 @@ func (c *repo) pushSubscription(ctx context.Context, q dsssql.Queryable, s *scdm
...
 		VALUES
 			($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, transaction_timestamp())
+		ON CONFLICT (%s) DO UPDATE
+			SET %s = $2,
+				%s = $3::int,
+				%s = $4,
+				%s = $5::int,
+				%s = $6,
+				%s = $7,
+				%s = $8::bool,
```
mickmis, line 202: Is setting the types here required for yugabyte? If so, why is it not required as well for operational intents?

build/db_schemas/yugabyte/rid/downfrom-v1.0.1-drop-ybgin-indexes.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX s_cell_idx ON subscriptions USING ybgin (cells);
```
mickmis, line 1: If yugabyte is not deployed anywhere, might as well update v1.0.0 IMO. And it probably makes sense keeping doing so until the v1.0.0 is actually a true working v1 that is released.

mickmis (commented):
Only minor comments, LGTM otherwise.

mickmis (approved):
LGTM

PR #1147, opened 2025-02-05
Fix for code scanning alert no. 2: Incorrect conversion between integer types

BenjaminPelletier:
Potential fix for [https://github.com/interuss/dss/security/code-scanning/2](https://github.com/interuss/dss/security/code-scanning/2)

To fix the problem, we need to ensure that the `uint64` value parsed from the string does not exceed the maximum value that can be held by an `int64` before performing the conversion. This can be done by adding a bounds check before the conversion. If the value exceeds `math.MaxInt64`, we should handle it appropriately, such as returning an error.

The best way to fix this issue is to add a check to ensure that the `uint64` value is within the range of an `int64` before performing the conversion. If the value is out of range, we should return an error.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/pkg/models/models.go
+++ b/pkg/models/models.go
@@ -7,7 +7,6 @@ import (
 	"github.com/google/uuid"
 	"github.com/interuss/stacktrace"
 	"github.com/jackc/pgx/v5/pgtype"
-	"math"
 )
 
 type (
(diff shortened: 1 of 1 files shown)


mickmis (approved):
LGTM as is, just a nit.

PR #1167, opened 2025-04-09
[terraform-aws] Switch to IRSA and upgrade aws-load-balancer-controller to 2.12

barroco:
In preparation of upgrading the deployment infrastructure to support Kubernetes up to 1.32 (#1155), this PR switches to the recommended IAM permissions configuration using [IAM role for service account](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.12/deploy/installation/#option-a-recommended-iam-roles-for-service-accounts-irsa) and removes the policy attached to nodes. 

The AWS load balancer controller helm is upgraded to the latest version 1.12.0.

This was tested on a EKS cluster 1.29 and 1.32.

PR #1169, opened 2025-04-09
[terraform] Support Kubernetes up to 1.32

### LINKED ISSUE #1155: Update Terraform deployment to use latest kubernetes_version
Currently published Terraform terraform.tfvars files uses a kubernetes_version no longer supported by GCP.  Changing to latest version fails a check in variables.gen.tf

Requesting update to support kubernetes_version 1.31 or above

barroco:
This PR adds support to Kubernetes up to 1.32 and document the migration steps.

> [!NOTE] 
> It depends on #1166, #1167 and #1168.

This should close #1155 .

### DIFF
--- a/deploy/infrastructure/dependencies/terraform-aws-kubernetes/variables.gen.tf
+++ b/deploy/infrastructure/dependencies/terraform-aws-kubernetes/variables.gen.tf
@@ -95,29 +95,12 @@ variable "kubernetes_version" {
   description = <<-EOT
   Desired version of the Kubernetes cluster control plane and nodes.
 
-<<<<<<< HEAD
   Supported versions: 1.24 to 1.32
-=======
-    Supported versions:
-      - 1.24
-      - 1.25
-      - 1.26
-      - 1.27
-      - 1.28
-      - 1.29
-      - 1.30
-      - 1.31
-      - 1.32
->>>>>>> 96cc6475 (Update templates and modules)
   EOT
 
   validation {
     condition     = contains(["1.24", "1.25", "1.26", "1.27", "1.28", "1.29", "1.30", "1.31", "1.32"], var.kubernetes_version)
-<<<<<<< HEAD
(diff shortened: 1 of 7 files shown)


deploy/MIGRATION.md
```diff
@@ -99,6 +99,16 @@ CockroachDB requires to upgrade one minor version at a time, therefore the follo
 
 Migrations of GKE clusters are managed using terraform.
 
+#### 1.28 to 1.32
```
the-glu, line 102: You can also regroup / refactor all others upgrade since they are the same instructions no ?

Something like:

---

#### Generic instructions

For each intermediate version up to the target version (eg. if you upgrade from `1.27` to `1.30`, apply thoses instructions for `1.28`, `1.29`, `1.30`), do:

1. Change your `terraform.tfvars` to use `<New version>` by adding or updating the `kubernetes_version` variable:
   ```terraform
   kubernetes_version = <New version>
   ```
1. Run `terraform apply`. This operation may take more than 30min.
1. Monitor the upgrade of the nodes in the Google Cloud console.

deploy/MIGRATION.md
```diff
@@ -143,6 +153,18 @@ Before proceeding, always check on the cluster page the *Upgrade Insights* tab w
 availability of Kubernetes resources in each version. The following sections omit this check if no resource is
 expected to be reported in the context of a standard deployment performed with the tools in this repository.
 
+#### 1.28 to 1.32
```
the-glu, line 156: And here same logic for generic instructions + specific ones for `1.29`, and `1.25` ?

the-glu (approved):
LGTM, just a nit about regrouping upgrade instructions

I assume that not the case but just to be sure, https://github.com/interuss/dss/blob/3eff6d77546a75f43491c596fa49ef216c6f0efb/deploy/operations/Dockerfile#L24 shouldn't be updated to the latest version, correct?

Issue #1171, opened 2025-04-10
Health check for shared configuration correctness


callumdmay:
We operate a DSS pool shared amongst multiple different organizations,  and a recurring challenge we face is effectively verifying the correct configuration of newly deployed DSS instances, particularly ensuring they are connected to the same underlying data source as existing instance.

The current InterUSS DSS implementation offers a valuable /healthy endpoint for general service availability. Building on this, we were thinking it would be really helpful to have a similar endpoint that reads a unique, hard-to-guess value (like a UUID) from a special health check table in the database. This would give us a simple way to confirm that all DSS instances are pointing to the same data source by checking if they all return the same value.

Ideally, for our use case, we'd love to not only verify new setups but also continuously monitor the health of our pool. Perhaps this could be extended to allow the unique value to change periodically (either through another endpoint or based on a configurable time interval). This would provide a simple and effective way to ensure all DSS instances remain correctly configured over time.

Happy to answer questions or provide additional clarification!

BenjaminPelletier:
/healthy is nice because it's extremely lightweight, but I agree it does not provide a great deal of information -- it doesn't even verify connectivity to the underlying database, let alone successful pooling with other DSS instances.  It seems like there should be a large amount of value in a DSS operator being able to answer the question "Is the DSS pool configured and working the way I expect it to be?" and it seems like this would essentially do that.  Especially with increasing levels of operationalization, it seems like this kind of feature is becoming more and more important and urgent.

## Outline 1
Brainstorming, what do you think of this implementation outline?
1. Add a pool_participants table in the [database](https://github.com/interuss/dss/tree/master/build/db_schemas) with "DSS instance" and "Last heartbeat" columns
2. At startup, a DSS instance upserts its DSS instance ID and the current time into that table
3. DSS exposes an [aux endpoint](https://github.com/interuss/dss/tree/master/interfaces/aux_) to query pool participants, which basically returns the contents of the pool_participants table
4. A USS verifies successful initial pooling by verifying that the expected participants are present in that response
5. DSS also exposes another aux endpoint/method that updates its "Last heartbeat" value to the current timestamp, and the DSS instance operator has its automation call that endpoint as frequently as desired/agreed by other pool participants
6. A USS verifies ongoing pool integrity by verifying that all expected participants have a "Last heartbeat"

## Variants to Outline 1
There are many more specific decisions if that overall outline seems good to everyone, but I think we would want to settle on a high-level outline like Outline 1 before diving into specifics like these:
[...]

Issue #1140, opened 2024-12-05
Host configuration resources for pooled DSS instances


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
To execute pooling actions, our [setup](https://github.com/interuss/dss/tree/master/build#deploying-a-dss-instance-via-kubernetes) and [pooling procedures]() have instructions like "Reach out to existing operators to request their public cert," "share your CRDB node addresses with the operators of the existing DSS instances," and "The joining USS will also need a list of node addresses to which they should connect."  These steps require DSS instance operators to exchange information out-of-band and the security and authenticity of this information exchange depends on the nature of that out-of-bad communication.  The out-of-band communication is also usually manual, and therefore subject to substantial delays.

**Describe the solution you'd like**
Enable exchange of this information automatically by having a DSS instance publish its configuration information in a well-known location.  After this is implemented, the only piece of information that would need to be exchange out-of-band would be the base URL of each DSS instance (existing and new, when pooling).  This would enable the following automated equivalents when existing operators indicate their DSS instance base URLs as https://dss.uss1.example.com and https://dss.uss2.example.com, and a new USS joining the pool indicates their DSS instance base URL as https://dss.uss3.example.com:

"Reach out to existing operators to request their public cert"
--> GET https://dss.uss1.example.com/configuration/individual_ca.crt (and also for https://dss.uss2.example.com/configuration/individual_ca.crt)

"share your CRDB node addresses with the operators of the existing DSS instances"
--> GET https://dss.uss3.example.com/configuration/interface.json
```json
{
  "crdb_node_addresses": [
    "0.db.dss.uss3.example.com",
    "1.db.dss.uss3.example.com",
    "2.db.dss.uss3.example.com"
  ],
  ...
}
```

"The joining USS will also need a list of node addresses to which they should connect."
--> GET https://dss.uss1.example.com/configuration/interface.json (and also for https://dss.uss2.example.com/configuration/interface.json)

The configuration endpoints would likely be protected by access tokens similar to existing access tokens, but with a scope of (for instance) `interuss.dss_host`.

[...]

PR #1174, opened 2025-04-15
[aux] Add empty DSS instances endpoint

BenjaminPelletier:
This PR takes a first step toward addressing #1171 and #1140 (via database entries rather than just endpoints per instance) by defining an aux-API `pool/dss_instances` endpoint that can be read to (in the future) retrieve information about DSS instances in the pool -- especially heartbeat information to verify integrity of the pool on an ongoing basis.

Future PRs can build on this by:
1. Creating a place in the database to store this DSS instance information
2. Reading from the database to report DSS instance information
3. Upserting an entry for the current DSS instance on startup
4. Adding a PUT `pool/dss_instances/{instance_id}/heartbeat` endpoint to register a heartbeat via client (but this request would be refused if the DSS instance the request was addressed to was not {instance_id})

This PR was tested with the added `read_dss_instances.sh` script.

### DIFF
--- a/interfaces/aux_/aux_.yaml
+++ b/interfaces/aux_/aux_.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.2
 info:
   title: DSS Auxiliary API
-  version: 1.0.0
+  version: 1.1.0
 
 components:
   schemas:
@@ -19,6 +19,59 @@ components:
         message:
           description: Human-readable message indicating what error occurred and/or why.
           type: string
+
+    DSSInstancesResponse:
+      type: object
+      properties:
+        dss_instances:
+          type: array
+          items:
+            $ref: '#/components/schemas/DSSInstance'
+          default: []
+    DSSInstance:
+      type: object
+      properties:
+        instance_id:
+          description: Identity of this DSS instance participating in the pool (locality).
+          type: string
+        most_recent_heartbeat:
+          description: Most recent heartbeat registered for this DSS instance.
+          anyOf:
+          - $ref: '#/components/schemas/Heartbeat'
+      required:
+      - instance_id
+    Heartbeat:
+      type: object
+      properties:
+        timestamp:
+          description: Time at which heartbeat was registered.
+          type: string
+          format: date-time
+          example: '1985-04-12T23:20:50.52Z'
+        reporter:
+          description: Identity (via access token `sub` claim) of client reporting the heartbeat, or omitted if no client reported the heartbeat.
+          type: string
+          example: 'uss1'
+          default: ''
+        source:
+          description: Source/trigger of this heartbeat.
+          type: string
+          example: Startup
+        index:
+          description: Index of this heartbeat within the set of all heartbeats for this pool participant.
+          type: integer
+          format: int64
+          minimum: 0
+          example: 1
+          default: 0
+        unique_value:
+          description: If specified, a sufficiently unique value to verify synchronization between DSS instances with additional certainty.  A UUIDv4 is recommended.
+          type: string
+          example: 9d531807-f3fa-435a-ad54-4bca01a5df43
+          default: ''
+      required:
+      - timestamp
+      - source
 
 paths:
   /aux/v1/version:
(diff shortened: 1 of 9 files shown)


pkg/aux_/pool_participants.go
```diff
@@ -0,0 +1,19 @@
...
+
+func (a *Server) GetDSSInstances(ctx context.Context, req *restapi.GetDSSInstancesRequest) restapi.GetDSSInstancesResponseSet {
+	if req.Auth.Error != nil {
+		resp := restapi.GetDSSInstancesResponseSet{}
+		setAuthError(ctx, stacktrace.Propagate(req.Auth.Error, "Auth failed"), &resp.Response401, &resp.Response403, &resp.Response500)
+		return resp
+	}
+
+	// Pool participant storage is not yet implemented.
+	return restapi.GetDSSInstancesResponseSet{Response200: &restapi.DSSInstancesResponse{DssInstances: &[]restapi.DSSInstance{}}}
```
mickmis, line 18: Probably not for this PR, but IMO it would be nice to have the code generator always generate a response 501 to be able to return a 501 for any call in such situations.

interfaces/aux_/aux_.yaml
```diff
@@ -20,6 +20,59 @@ components:
...
+        reporter:
+          description: Identity (via access token `sub` claim) of client reporting the heartbeat, or omitted if no client reported the heartbeat.
+          type: string
+          example: 'uss1'
+          default: ''
+        source:
+          description: Source/trigger of this heartbeat.
+          type: string
+          example: Startup
+        index:
```
mickmis, line 60: What would that be used for? It does seem to overlap partially with the timestamp.

interfaces/aux_/aux_.yaml
```diff
@@ -20,6 +20,59 @@ components:
...
+          description: Index of this heartbeat within the set of all heartbeats for this pool participant.
+          type: integer
+          format: int64
+          minimum: 0
+          example: 1
+          default: 0
+        unique_value:
```
mickmis, line 67: My understanding of a heartbeat is that it is a signal of liveliness from the instance. With that understanding, I'm not sure this unique value should actually be part of the heartbeat?
In addition, how would that be used? IIUC comparing the last timestamps of heartbeats would be equivalent.

interfaces/aux_/aux_.yaml
```diff
@@ -20,6 +20,59 @@ components:
...
+      properties:
+        dss_instances:
+          type: array
+          items:
+            $ref: '#/components/schemas/DSSInstance'
+          default: []
+    DSSInstance:
+      type: object
+      properties:
+        instance_id:
```
mickmis, line 34: Since we are already in `DSSInstance`:
```suggestion
        id:
```

interfaces/aux_/aux_.yaml
BenjaminPelletier: My intent is to use it for resolving update race conditions, with a side benefit of keeping a bit of historical information (for instance, allowing a client to determine the approximate rate at which heartbeats were being made).  In the later PUT, I would expect the heartbeat payload to specify an `index` of {current `index` + 1} and any other values would be rejected 409.

interfaces/aux_/aux_.yaml
BenjaminPelletier: My intent was to include the option of transmitting a "hard-to-guess value" as stronger evidence of synchronization than just transmitting a timestamp, but it seems like the timestamp would be sufficiently hard to guess as to make this field essentially redundant -- I'll remove.

pkg/aux_/pool_participants.go
BenjaminPelletier: I think that's in scope for this PR -- 501 is less misleading than success with an empty list.  Will update.

interfaces/aux_/aux_.yaml
mickmis: Alright, makes sense indeed!

mickmis (approved):
LGTM

Issue #1175, opened 2025-04-15
Align naming for CockroachDB and YugabyteDB


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
Even though we are nominally code-complete on Yugabyte support for core-service, we still have a number of folders and documentation referring to Cockroach rather than something more generic; perhaps "Postgres" ([example folder](https://github.com/interuss/dss/tree/master/pkg/rid/store/cockroach) that [includes Yugabyte support](https://github.com/interuss/dss/blob/e42942c37f6e256144de8bd0b9ec0bcddc2639a4/pkg/rid/store/cockroach/store.go#L24)).

**Describe the solution you'd like**
We should clean up folder and file naming along with documentation to ensure we are not misrepresenting the structure and capabilities of the various modules/subsystems.  It seems like many things currently labeled "cockroach" would probably better be labeled "postgres" as they support multiple PostgresQL database interfaces.

**Describe alternatives you've considered**
N/A

**Additional context**
When adding database capabilities for #1171 and/or #1140, it will be important to understand where the database-agnostic logic lives and where database-specific logic needs to be provided.

PR #1176, opened 2025-04-16
Rename cockroach* flags to datastore*

BenjaminPelletier:
Per #1175, this PR furthers the naming adjustment by renaming all cockroach* flags to datastore*.

This changes the names of flags to the core-service executable, some of which are specified in deployment configurations (tanka + helm charts).  It seems like a tk-apply will update flags appropriately as long as the current codebase of tanka files is used in the new rollout -- I would expect that note to be added in the release notes of the next release that incorporates this PR (expecting v0.20.0).  I'm not as sure about Helm; if @barroco could consider the implications of this PR on Helm deployments, that would be great.


BenjaminPelletier:
To align priorities (spending time in other places in the short term), I intend to keep this PR in draft until the beginning of May.

BenjaminPelletier:
Closing as obsolete

PR #1177, opened 2025-04-16
[aux] Add access to DAR ID via aux API

BenjaminPelletier:
A key piece of information about the pool of DSS instances to which a particular DSS instance is connected is the identity of the DSS Airspace Representation data store.  The technical embodiment of this concept is easily accessible in CockroachDB as the Cockroach cluster ID -- if two DSS instances' CRDB nodes have the same cluster ID, then those DSS instances are nearly certain to be part of the same pool/cluster.  This concept is also present in Yugabyte, though somewhat harder to access.

Providing access to this information is likely to be very useful as an easy verification that two DSS instances are part of the same pool, and can partially address #1171.

This PR was tested with read_pool_info.sh using the standard local CRDB-based deployment, as well as the local Yugabyte-based deployment with `COMPOSE_PROFILES=with-yugabyte`.

### DIFF
--- a/build/dev/read_pool_info.sh
+++ b/build/dev/read_pool_info.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+# Retrieve token from dummy OAuth server
+ACCESS_TOKEN=$(curl --silent \
+    "http://localhost:8085/token?grant_type=client_credentials&scope=interuss.pool_status.read&intended_audience=localhost&issuer=localhost&sub=manual_tester" \
+| python extract_json_field.py 'access_token')
+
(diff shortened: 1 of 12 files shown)


BenjaminPelletier:
> Note that this doesn't fully confirm the pooling, both DSS instances could be initialized without being pooled and return the same value.

I definitely agree this doesn't fully confirm pooling, but I don't think both DSS instances could be initialized without being pooled and return the same value if I understand what is going on properly.  I believe initializing a CRDB cluster involves assigning a random UUID as the cluster ID, so I would expect the two unpooled instances not to return the same value since the probability of two random UUIDs colliding is essentially nil.  We should definitely follow up if this understanding isn't accurate.

PR #1178, opened 2025-04-29
Upgrade go version to 1.24; Upgrade golint-ci to 2.1.5; Fix linting issues; Generate new dummy keys for DummyOAuth

mickmis:
This PR upgrades the dss Go version to 1.24(.2) and golint-ci to 2.1.5.
The following changes were necessary:
- Key pair in `auth2.(pem|key)` had a public key with size 1023 bits which upgraded `crypto/rsa` rejects by default now (rejects all keys under 1024 bits). Now replaced with new keys with size 2048 bits.
- Updated generation code to print the public key.
- Many code changes due to upgraded linter version
- Switched golint-ci to use yml file instead of CLI parameters. Syntax of arguments have changed with upgrade to v2.


pkg/rid/application/isa.go
```diff
@@ -34,7 +34,7 @@ type ISAApp interface {
 }
 
 func (a *app) GetISA(ctx context.Context, id dssmodels.ID) (*ridmodels.IdentificationServiceArea, error) {
-	repo, err := a.Store.Interact(ctx)
+	repo, err := a.Interact(ctx)
```
BenjaminPelletier, line 37: I'm hesitant on this one -- I wouldn't view [`app`](https://github.com/interuss/dss/blob/716fb896c72340b318595b6e72f83f6aaf0d453b/pkg/rid/application/application.go#L13) as inheriting from [`Store`](https://github.com/interuss/dss/blob/716fb896c72340b318595b6e72f83f6aaf0d453b/pkg/rid/store/store.go#L13), but rather having a `Store` member, and this change seems like a regression relative to that concept.  In other words, I think not providing an explicit name is intended to achieve merely an anonymous field rather than a promoted field.  If the linter is disallowing explicit (via anonymous field name) access to members of promoted fields, could you link to the rationale why that is necessarily bad style?  It seems like the other way to fix such a lint finding would be to give the anonymous field an actual name.

BenjaminPelletier (commented):
Everything looks good to me except the one comment below.

pkg/rid/application/isa.go
mickmis: The linter reported the following:
>  QF1008: could remove embedded field "Store" from selector (staticcheck)

Which IMO makes sense for an embedded field, which BTW is a bit different than an anonymous one, the aim being more towards composition rather than reducing verbosity.
And I also agree with you that in that case the change I originally did is a regression.
Actually *Store* should not be an embedded field in the first place, which is confirmed by this TODO I just discovered:
https://github.com/interuss/dss/blob/5168bd2d3ea4c8a64c9aa165ca0018f92a5337cb/pkg/rid/application/application.go#L14-L17

> It seems like the other way to fix such a lint finding would be to give the anonymous field an actual name.

Yep that is the proper fix IMO, I will do that.

BenjaminPelletier (approved):

PR #1188, opened 2025-05-23
Add an overrideable way to name the load balancers so that multiple DSS can be deployed in one VPC

StevenPG:
I will raise an issue to discuss. This is my proposed solution to the issue I've run into.

While deploying the DSS into an AWS VPC, the networking components are named in such a way that there can only be one running instance. Subsequent deployments result in 400 errors on the Service objects in Kubernetes with a "duplicate name" error. 

This can be confusing for a developer, since the terraform is designed in such a way that two environments can be deployed.

### DIFF
--- a/deploy/services/helm-charts/dss/templates/cockroachdb-loadbalancers.yaml
+++ b/deploy/services/helm-charts/dss/templates/cockroachdb-loadbalancers.yaml
@@ -1,4 +1,5 @@
 {{- $cloudProvider := $.Values.global.cloudProvider}}
+{{- $cockroachDbExternalNodeSuffix := default "" $.Values.cockroachdb.conf.crdbNodeSuffix }}
 
 {{- if $.Values.cockroachdb.enabled }}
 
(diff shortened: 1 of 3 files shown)


StevenPG:
Testing `helm template` without new variables in https://github.com/interuss/dss/commit/34d494b43f0b0b34556d43aa8918a742691637b2

```
    service.beta.kubernetes.io/aws-load-balancer-name: dss-gateway-external
    service.beta.kubernetes.io/aws-load-balancer-name: cockroach-db-external-node
```

with new variables

```
    service.beta.kubernetes.io/aws-load-balancer-name: myGatewayIdentifier-703ecd1d-b1bf-4f35-a80d-c7654e41437d
    service.beta.kubernetes.io/aws-load-balancer-name: cockroach-db-node-bfb5d477-1f44
```

```
dss:
  image: docker.io/interuss/dss:v0.18.0 # See https://hub.docker.com/r/interuss/dss/tags for official image releases.
  # When running local images in minikube, uncomment the following line
  # imagePullPolicy: Never
  gatewayId: myGatewayIdentifier-703ecd1d-b1bf-4f35-a80d-c7654e41437d
...
cockroachdb:
  conf:
    join: []
    crdbNodeSuffix: cockroach-db-node-bfb5d477-1f44-48b6-a711-0db21845b1b4
```

StevenPG:
@barroco Updated in 3341b214e4b68f3c401fe4c945fdb6f9bd47f641

StevenPG:
@barroco I created a companion issue here! https://github.com/interuss/dss/issues/1189

I'll link this thread and make an initial comment there

PR #1190, opened 2025-05-27
Yugabyte: Working GCP implementation

the-glu:
This PR introduce changes in the teraform files to support yugabyte in GCP.

It follow #1186 . You probably want to review the last commit only.

It needs the following PRs to make it working even with yugabyte not working correctly with server_broadcast_address ( https://github.com/yugabyte/yugabyte-db/issues/27367 ), we do 'cheat' by injecting into the host file the public host we want, making yugabyte broadcasting the public hostname as private rpc address.

- https://github.com/yugabyte/charts/pull/207
- https://github.com/yugabyte/charts/pull/208

It does the following:

- Added a new terraform option datastore_type, and needed yugabyte options
- Fixed and added a public name option in certificate generation for yugabyte nodes
  - The previous PR doesn't include the extension file, generating certificate only valid for the 'main' host
- Add correct load balances for yugabyte, exposing all ports
  - Some port are exposed even if not valid to avoid a UI issue doing queries on unavailable port and waiting for timeout (exposing them reject directly the tcp connection)

It doesn't have, for now (but the PR is quite big):

- Documentation
- Some variables are shared with cochroachdb and haven't been renamed to generic ones.
- AWS implementation is broken dues to changes in the common part, but next PR will fix that :)
- Default values

It has been tested with two independent GCP k8s clusters that successfully formed a cluster:

image
image

PR #1191, opened 2025-05-27
Yugabyte: Working AWS implementation

the-glu:
This PR does follows #1190

It improve terraform files on the aws side to support yugabyte.

It's mostly similar from the previous PR, with the following changes:

- Added missing k8s storage parameter
- Disabled ulimit checks

The same limitation applies (e.g. limited documentation ;) )

It has been validated with an aws cluster and a aws-gcp cluster:

image


deploy/infrastructure/dependencies/terraform-aws-kubernetes/network_lb.tf
```diff
@@ -82,3 +82,27 @@ resource "aws_eip" "ip_crdb" {
     ExpectedDNS = format("%s.%s", count.index, var.crdb_hostname_suffix)
   }
 }
+
+# Public Elastic IPs for the yubagybte master instances
+resource "aws_eip" "ip_yugabyte_masters" {
```
barroco, line 87: Could you please open an issue to optimize the number of public ip addresses and investigate having masters and tserver on the same node / ip ?

deploy/infrastructure/dependencies/terraform-commons-dss/helm.tf
```diff
@@ -102,6 +102,19 @@ resource "local_file" "helm_chart_values" {
...
+        master = {
+          storageClass = var.kubernetes_storage_class
+        }
+        tserver = {
+          storageClass = var.kubernetes_storage_class
+        }
+      }
+
+      preflight = {
+        skipUlimit = true
```
barroco, line 115: Could you please confirm this is for AWS and GCP as well?

the-glu: Yes, I removed in all cases just in case

deploy/infrastructure/dependencies/terraform-aws-kubernetes/network_lb.tf
the-glu: Done via #1199

PR #1194, opened 2025-06-04
[aux] Add CAs endpoint

the-glu:
Contributing to #1140, this PR add an endpoint listing current CAs in the pool, and the CA(s) of the instance.

Idea would be to use that list of certificates to join a pool.

It doesn't include security, assuming thoses are public.
I only updated the yugabyte script to expose the CA of the instance, it won't works for cockroachdb.

### DIFF
--- a/deploy/services/helm-charts/dss/templates/_volumes.tpl
+++ b/deploy/services/helm-charts/dss/templates/_volumes.tpl
@@ -13,6 +13,9 @@
 - mountPath: /opt/yugabyte-certs/ca.crt
   name: ca-certs
   subPath: root.crt
+- mountPath: /opt/yugabyte-certs/ca-instance.crt
+  name: ca-certs
+  subPath: ca-instance.crt
 {{- end -}}
 {{- end -}}
 {{- define "client-certs:volume" -}}
--- a/interfaces/aux_/aux_.yaml
+++ b/interfaces/aux_/aux_.yaml
@@ -84,6 +84,15 @@ components:
       - timestamp
       - source
 
+    CAsResponse:
+      type: object
+      properties:
+        CAs:
+          description: A list of certificates, each in PEM format.
+          type: array
+          items:
+            type: string
+
 paths:
   /aux/v1/version:
     get:
@@ -201,6 +210,44 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+        '501':
+          description: >-
+            The server has not implemented this operation.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+  /aux/v1/configuration/accepted_ca_certs:
+    get:
+      summary: Current certificates of certificate authorities (CAs) that this DSS instance accepts as legitimate signers of node certificates for the pool of DSS instances constituting the DSS Airspace Representation.
+      operationId: getAcceptedCAs
+      tags: [ dss ]
+      responses:
+        '200':
+          description: The information is successfully returned.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CAsResponse'
+        '501':
+          description: >-
+            The server has not implemented this operation.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+  /aux/v1/configuration/ca_certs:
+    get:
+      summary: Current certificates of certificate authorities (CAs) that signed the node certificates for this DSS instance. May return more that one certificate (e.g. for rotations).  Other DSS instances in the pool should accept node certificates signed by these CAs.
+      operationId: getInstanceCAs
+      tags: [ dss ]
+      responses:
+        '200':
+          description: The information is successfully returned.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CAsResponse'
         '501':
           description: >-
             The server has not implemented this operation.
--- a/deploy/operations/certificates-management/ca_pool.py
+++ b/deploy/operations/certificates-management/ca_pool.py
@@ -74,6 +74,7 @@ def regenerate_ca_files(cluster):
         f.write("\n\n".join(CAs))
 
     shutil.copy(cluster.ca_pool_ca, cluster.client_ca)
+    shutil.copy(cluster.ca_cert_file, cluster.client_instance_ca)
 
     for node_type in ["master", "tserver"]:
         shutil.copy(cluster.ca_pool_ca, getattr(cluster, f"{node_type}_ca"))
--- a/deploy/operations/certificates-management/cluster.py
+++ b/deploy/operations/certificates-management/cluster.py
@@ -52,6 +52,10 @@ class Cluster(object):
         return os.path.join(self.client_certs_dir, "root.crt")
 
     @property
+    def client_instance_ca(self):
+        return os.path.join(self.client_certs_dir, "ca-instance.crt")
+
+    @property
     def master_certs_dir(self):
         return os.path.join(self.directory, "masters")
 
--- a/pkg/api/auxv1/interface.gen.go
+++ b/pkg/api/auxv1/interface.gen.go
@@ -7,9 +7,9 @@ import (
 )
 
 var (
+	InterussPoolStatusReadScope             = api.RequiredScope("interuss.pool_status.read")
 	DssWriteIdentificationServiceAreasScope = api.RequiredScope("dss.write.identification_service_areas")
 	DssReadIdentificationServiceAreasScope  = api.RequiredScope("dss.read.identification_service_areas")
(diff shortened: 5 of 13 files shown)


interfaces/aux_/aux_.yaml
```diff
@@ -208,6 +217,44 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/CAsResponse'
+        '501':
+          description: >-
+            The server has not implemented this operation.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+  /aux/v1/pool/ca:
```
BenjaminPelletier: Since the CA certificate(s) are for this DSS instance, it seems like this should be a RESTful resource of the DSS instance being queried:

```suggestion
  /aux/v1/ca_certs:
```

(`ca_certs` makes it clearer what the resource contains IMO)

Or, if we wanted to have a conceptual configuration RESTful resource for which this is a subresource (probably my slight preference), then perhaps something like `/aux/v1/configuration/ca_certs`.  But, importantly, the CA certificate(s) returned by this endpoint are not characteristics or subresources of the pool as a whole.

interfaces/aux_/aux_.yaml
```diff
@@ -208,6 +217,44 @@ paths:
...
+    get:
+      summary: Queries the current CA certificates of the DSS instance. May return more that one CA (e.g. for rotations).
```
BenjaminPelletier: To be extra clear (I don't think this changes the meaning of what you had IIUC):

```suggestion
      summary: Current certificates of certificate authorities (CAs) that signed the node certificates for this DSS instance. May return more that one certificate (e.g. for rotations).  Other DSS instances in the pool should accept node certificates signed by these CAs.
```

interfaces/aux_/aux_.yaml
```diff
@@ -208,6 +217,44 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+  /aux/v1/pool/cas:
```
BenjaminPelletier: While I agree this resource relates to and/or describes this DSS instance's beliefs about the pool, I don't think it's actually a characteristic of the pool itself -- instead, it's a characteristic of the DSS instance (describing the DSS instance's beliefs about the pool).  So, I think it should probably be either:

```suggestion
  /aux/v1/accepted_ca_certs:
```

...or something like `/aux/v1/configuration/accepted_ca_certs`

interfaces/aux_/aux_.yaml
```diff
@@ -84,6 +84,15 @@ components:
       - timestamp
       - source
 
+    CAsResponse:
+      type: object
+      properties:
+        CAs:
+          description: A list of certificate, in PEM format.
```
BenjaminPelletier: nit:

```suggestion
          description: A list of certificates, each in PEM format.
```

interfaces/aux_/aux_.yaml
```diff
@@ -208,6 +217,44 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+  /aux/v1/pool/cas:
+    get:
+      summary: Queries the current CA certificates of the pool of DSS instances constituting the DSS Airspace Representation.
```
BenjaminPelletier:
```suggestion
      summary: Current certificates of certificate authorities (CAs) that this DSS instance accepts as legitimate signers of node certificates for the pool of DSS instances constituting the DSS Airspace Representation.
```

interfaces/aux_/aux_.yaml
```diff
@@ -208,6 +217,44 @@ paths:
...
+      operationId: getPoolCAs
```
BenjaminPelletier:
```suggestion
      operationId: getAcceptedCAs
```

interfaces/aux_/aux_.yaml
```diff
@@ -208,6 +217,44 @@ paths:
...
+          description: >-
+            The server has not implemented this operation.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+  /aux/v1/pool/ca:
+    get:
+      summary: Queries the current CA certificates of the DSS instance. May return more that one CA (e.g. for rotations).
+      operationId: getCAs
```
BenjaminPelletier:
```suggestion
      operationId: getInstanceCAs
```

pkg/aux_/ca.go
```diff
@@ -0,0 +1,42 @@
...
+	CAFile := connectParameters.GetInstanceCAFile()
+
+	if CAFile == "" {
+		return restapi.GetCAsResponseSet{Response200: &restapi.CAsResponse{}}
+	}
+
+	data, err := os.ReadFile(CAFile)
+
+	if err != nil {
+		msg := "Unable to read data"
```
BenjaminPelletier: Especially since this error is likely to pop up in normal circumstances (e.g., developer has a stand-alone instance deployed without any cert files, or the deployment doesn't yet include ca-instance.crt as noted above), let's make sure the issue is clear:

```suggestion
		msg := "Unable to read CA certificate file for this DSS instance"
```

It also seems like we would want to enable as efficient/effective debugging for this error as practical; if someone encountered this error, what would their next debugging step be?  I would expect they would want to know what `err` actually said, and likely the value of `CAFile` as well.  Can we make sure this information is relatively easily available, probably by printing it to the log when this error occurs?

pkg/aux_/pool_ca.go
```diff
@@ -0,0 +1,42 @@
...
+	CAFile := connectParameters.GetCAFile()
+
+	if CAFile == "" {
+		return restapi.GetPoolCAsResponseSet{Response200: &restapi.CAsResponse{}}
+	}
+
+	data, err := os.ReadFile(CAFile)
+
+	if err != nil {
+		msg := "Unable to read data"
```
BenjaminPelletier: Similar comment as above

```suggestion
		msg := "Unable to read certificate file for accepted CAs"
```

deploy/services/helm-charts/dss/templates/_volumes.tpl
```diff
@@ -13,6 +13,9 @@
 - mountPath: /opt/yugabyte-certs/ca.crt
   name: ca-certs
   subPath: root.crt
+- mountPath: /opt/yugabyte-certs/ca-instance.crt
```
BenjaminPelletier, line 16: IIUC, this is what makes ca-instance.crt available for service deployments using Helm charts.  The other currently-supported ways to deploy services are CRDB [via tanka](https://github.com/interuss/dss/tree/master/build#deploying-a-dss-instance-via-kubernetes) and CRDB via Helm charts (I assume the folder /opt/yugabyte-certs isn't used in CRDB deployments) and I think we'd want to make ca-instance.crt available when using those deployment approaches too(e.g., [CRDB-tanka](https://github.com/interuss/dss/blob/12e25383f1d76e875f9b77241621936dba81afc6/deploy/services/tanka/volumes.libsonnet#L77)).

Note that this could be done in a follow-up PR as long as the open action is captured somewhere.

If I read correctly, it seems like ca-instance.crt needs to exist before executing the deployment.  If that's correct, it seems like we also need to add documentation and/or tooling (likely again in a separate PR) that will cause this file to exist at the point it's needed.

interfaces/aux_/aux_.yaml
the-glu: I do agree, I will move it to `configuration/` it make more sense to have everything there.

(Edited: I mixed endpoints ^^')

deploy/services/helm-charts/dss/templates/_volumes.tpl
the-glu: Yes, right now the goal I had in mind is to improve the yugabyte certificate tool, so I only updated the yugabyte part (who is indeed only not present in tanka). 

This won't even work for CRDB via helm, I didn't updated the bash scripts / states to expose it ^^'

I did assume that since yugabyte part hasn't been released yet, there are no need to document the need to make the file existing, but that of course not true if we do add CRDB support for that part :)

I will create an issue listing missing features on the CRDB side.

pkg/aux_/ca.go
the-glu: I did add the error / path directly in the message, without extra logging. Would that be ok? I don't think the file name / error is going to be sensitive :)

deploy/services/helm-charts/dss/templates/_volumes.tpl
the-glu: Issue is there: https://github.com/interuss/dss/issues/1196

the-glu:
@BenjaminPelletier Ready for a second review pass, I let some conversion open that you can close if you're ok with then :)

pkg/aux_/ca.go
BenjaminPelletier: I agree I can't think of why any of that would be sensitive, and putting it in the message makes it more readily accessible than having to go back to the logs.

PR #1195, opened 2025-06-04
[aux] Add DAR ID on yugabyte deployments

the-glu:
This PR add the DAR ID for yugabyte deployments.

There is a cluster id notion on yugabyte, however it's not available via the postgres api.

We _could_ cheats by using the yugabyte api directly, but it mean communicating with the masters (not the tservers), on a new port, witch may introduce complexity.

As a solution, this PR propose to use the uuid of tserver instances (contacted to a predictable hash).

This id is not fix and can change with time as nodes are added/removed. However at any point in time, it should be unique across a cluster.

This mean https://github.com/interuss/dss/blob/master/deploy/operations/monitoring.md#procedure-option-2 is not an acceptable option
This mean https://github.com/interuss/dss/blob/master/deploy/operations/monitoring.md#procedure-option-1 may fail if queries are done at the exact same time of an addition/removal of a server in the pool. However, a simple bouncing strategy would work around that (probably exceptional) case.


BenjaminPelletier:
> This id is not fix and can change with time as nodes are added/removed.

I think this is probably a red flag for some of the purposes of DAR ID.  Under this definition, the ID of the entire DAR would change when a new USS joined the pool, and that would render confusing statements like "USS Q is joining the DAR with ID XYZ" since the DAR USS Q joins will not be XYZ once it has joined.  Instead, from an ID perspective, the XYZ DAR will cease to exist and a new JKL DAR will be formed.  I think this will cause problems, including the https://github.com/interuss/dss/blob/master/deploy/operations/monitoring.md#procedure-option-2 identified in the description.

IIUC, the Yugabyte cluster ID is immutable for the entire lifetime of the database cluster (very similar to the CRDB cluster ID), so I prefer some means of acquiring and using this actual value that has all the desirable characteristics even though doing so will be more difficult than with CRDB.  As an example brainstorm I don't necessarily think is the best approach: we could "fork" the Yugabyte image by creating our own Dockerfile based on their image and add a startup script that runs a query for the cluster ID locally and then writes the result into the database if it doesn't yet exist.

the-glu:
> IIUC, the Yugabyte cluster ID is immutable for the entire lifetime of the database cluster (very similar to the CRDB cluster ID), so I prefer some means of acquiring and using this actual value that has all the desirable characteristics even though doing so will be more difficult than with CRDB. As an example brainstorm I don't necessarily think is the best approach: we could "fork" the Yugabyte image by creating our own Dockerfile based on their image and add a startup script that runs a query for the cluster ID locally and then writes the result into the database if it doesn't yet exist.

Then we could require access to the non-sql port of the yugabyte cluster as well if that doesn't fill the purpose of the DAR ID.

It does seems that the cluster uuid (not the universe one) is available on the tserver via the http api, I just need to confirm it stable with addition / removal, we could use that.

Would that be ok?

the-glu:
> My first thought is that providing core-service access to the non-SQL port of the Yugabyte node(s) it's talking to shouldn't be too concerning (but maybe I'm overlooking something). Exposing any Yugabyte ports to the open Internet beyond the ones necessary to form and maintain a cluster would be undesirable though. I wouldn't expect a core-service job from DSS Instance 1 to be able to query the cluster ID from a Yugabyte node from DSS Instance 2 -- instead, DSS Instance 1's core-service job would query a Yugabyte node within DSS Instance 1.

I do agree and that at least true for the purpose there so we're probably fine.

More generally but so you know, to have functional yugabyte web interface tools working, opening those port between all dss instance of a cluster seems to be a requirement however.

> Sounds reasonable, though I don't have an informed opinion. I assume we'll eventually update [this documentation](https://github.com/interuss/dss/blob/master/deploy/operations/pooling.md#the-cockroachdb-cluster-is-initialized) to be agnostic between CRDB and Yugabyte, but I would hope to keep the concept unchanged since Yugabyte does have a cluster ID that behaves in the same way described there IIUC.

Yes, the universe uuid fill the same purpose and seems to be accessible so we're fine. Notice there cloud be multiple universe uuid in a xcluster setup, but we can probably exclude that setup from supported configuration.

However I didn't find a 'stable' way or getting any stable uuid on the `tserver` alone (testing in a cluster failled).

But during exploration, I found that latest version of yugabyte include that uuid in yb_servers() sql function, released 3 weeks ago (I was on previous version), that make everything else not an issue anymore ;)

I updated the code and added a warning / explanation that yugabyte > 2.25.2.0 is needed, I think we can assume that people are going to deploy new clusters at this point :)

BenjaminPelletier (approved):

PR #1203, opened 2025-06-18
[yugabyte] Rename crdb_hostname_suffix to db_hostname_suffix

the-glu:
This PR rename the `crdb_hostname_suffix` to `db_hostname_suffix` since it's also used for yugabyte.

It's not backward compatible, a release note is needed for the change :)


deploy/infrastructure/utils/definitions/crdb_hostname_suffix.tf
```diff
@@ -1,10 +0,0 @@
-variable "crdb_hostname_suffix" {
```
barroco, line 1: Do you think we could set a condition on this variable which error with an error inviting the user to rename the variable ?

the-glu: Done :)

PR #1211, opened 2025-07-01
[helm] Improve public endpoint setting with backward compatiblity

the-glu:
Improve management of the new public_endpoint setting in helm chart by using an explicit setting, that can be set to "" to run older version.

This still enforce usage of the future latest image on the terraform states.

### DIFF
--- a/deploy/services/helm-charts/dss/values.schema.json
+++ b/deploy/services/helm-charts/dss/values.schema.json
@@ -271,6 +271,10 @@
             "hostname": {
               "type": "string",
               "description": "Public hostname of the dss. Example: dss.example.com"
+            },
+            "publicEndpoint": {
+              "type": "string",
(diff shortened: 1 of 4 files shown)


deploy/services/helm-charts/dss/values.schema.json
```diff
@@ -271,6 +271,10 @@
             "hostname": {
               "type": "string",
               "description": "Public hostname of the dss. Example: dss.example.com"
+            },
+            "publicEndpoint": {
+              "type": "string",
+              "description": "Public endpoint of the dss. Example: https://dss.example.com. Only for dss version >= 0.20.0, otherwise left this value empty."
```
barroco:
```suggestion
              "description": "Public endpoint of the dss. Example: https://dss.example.com. Only for dss version >= 0.21.0, otherwise left this value empty."
```

PR #1212, opened 2025-07-01
 [deploy] Move locality to a common variable and make it mandatory 

the-glu:
`locality` is now a mandatory variable.

This PR move it out of the `crdb` context and make it mandatory everywhere and pass it in helm charts.

It need an explicit release note about setting it to the new location.

This PR follow #1211 , you may want to read only the last commit.


deploy/infrastructure/utils/definitions/locality.tf
barroco, line 1: Similarly to `crdb_hostname_suffix`, keep a crdb_locality with a failing condition if value is provided to guide users.

PR #1213, opened 2025-07-01
[aux] Add heartbeat functionnality

the-glu:
This PR add basic heartbeats functionality.

There are a few questions in comments about details :)

### DIFF
--- a/interfaces/aux_/aux_.yaml
+++ b/interfaces/aux_/aux_.yaml
@@ -220,6 +220,72 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+  /aux/v1/pool/dss_instances/heartbeat:
+    put:
+      summary: Record a new heartbeat from the DSS instance
+      operationId: putDSSInstancesHeartbeat
+      tags: [ dss ]
+      security:
+      - Auth:
+        - interuss.pool_status.heartbeat.write
+      parameters:
+      - name: source
+        in: query
+        required: true
+        description: The source of the timestamp
+        schema:
+          type: string
+      - name: timestamp
+        in: query
+        required: false
+        description: Override the timestamp value of the heartbeat. If not set, will use the current time. RFC 3339 format.
+        schema:
+          type: string
+          format: date-time
+      - name: next_heartbeat_expected_before
+        in: query
+        required: false
+        description: Set the time before the next heartbeat is expected. RFC 3339 format.
+        schema:
+          type: string
+          format: date-time
+      responses:
+        '201':
+          description: The heartbeat have been recorded. The known DSS instances participating in the pool are successfully returned.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/DSSInstancesResponse'
+        '400':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+          description: >-
+            The request was not properly formed or the parameters are invalid
+        '401':
+          description: >-
+            Bearer access token was not provided in Authorization header, token
+            could not be decoded, or token was invalid.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+        '403':
+          description: >-
+            The access token was decoded successfully but did not include a
+            scope appropriate to this endpoint.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+        '501':
+          description: >-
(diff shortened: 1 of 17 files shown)


interfaces/aux_/aux_.yaml
```diff
@@ -220,6 +220,72 @@ paths:
...
+        schema:
+          type: string
+          format: date-time
+      responses:
+        '201':
+          description: The heartbeat have been recorded. The known DSS instances participating in the pool are successfully returned.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/DSSInstancesResponse'
```
the-glu, line 258: Based on doc description, I added the same reply from the `/aux/v1/pool/dss_instances`, but we may want something else, like a redirect

interfaces/aux_/aux_.yaml
```diff
@@ -263,6 +329,7 @@ security:
       - dss.read.identification_service_areas
       - dss.write.identification_service_areas
       - interuss.pool_status.read
+      - interuss.pool_status.write
```
the-glu: Do we want a specific heartbeats scope?

pkg/aux_/pool_participants.go
```diff
@@ -46,10 +48,84 @@ func (a *Server) GetDSSInstances(ctx context.Context, req *restapi.GetDSSInstanc
...
+	repo, err := a.Store.Interact(ctx)
+	if err != nil {
+		resp.Response500 = &api.InternalServerErrorBody{ErrorMessage: *dsserr.Handle(ctx, stacktrace.Propagate(err, "Unable to interact with the store"))}
+		return resp
+	}
+
+	heartbeat := models.Heartbeat{
+		Source:   *req.Source,
+		Reporter: *req.Auth.ClientID,
+		Locality: a.Locality,
```
the-glu, line 92: I forced the locally to be the one of the DSS instance processing the call, but we may want to allow any value?

pkg/aux_/server.go
```diff
@@ -14,7 +14,8 @@ import (
 
 // Server implements auxv1.Implementation.
 type Server struct {
-	Store auxstore.Store
+	Store    auxstore.Store
+	Locality string
```
the-glu, line 18: I passed the locality for the enforced (or default) value, but it is the correct way ?

pkg/aux_/store/datastore/dss.go
```diff
@@ -26,14 +27,19 @@ func (r *repo) SaveOwnMetadata(ctx context.Context, locality string, publicEndpo
 	}
 
 	if exists {
+		q, err := r.Query(ctx, "UPDATE pool_participants SET public_endpoint = $2, updated_at = transaction_timestamp() WHERE locality = $1", locality, publicEndpoint)
+		q.Close()
```
the-glu, line 31: I side fixed a non-closed query that wasn't causing issue, but did when copy-pasting for heartbeats.

interfaces/aux_/aux_.yaml
BenjaminPelletier: I could be convinced of alternatives (like the redirect to .../dss_instances), but I don't see why this would cause any problems.

interfaces/aux_/aux_.yaml
BenjaminPelletier: That's a good question.  I think probably "yes" gives us the most future-proofing since very likely the periodic heartbeat process shouldn't necessarily get carte blanche to write any part of the status.  So, `interuss.pool_status.heartbeat.write` seems like a good scope (other variations ok too).

pkg/aux_/pool_participants.go
BenjaminPelletier: I think forcing it to this instance's locality is good.  If we allowed "any" value, I think we would want to make sure to validate that the specified value actually is a proper locality, and I think that would be challenging.

pkg/aux_/server.go
BenjaminPelletier: Seems reasonable to me (though I'm sure there are other reasonable solutions also)

pkg/aux_/store/datastore/dss.go
BenjaminPelletier: Thanks!

BenjaminPelletier (approved):

interfaces/aux_/aux_.yaml
the-glu: Switched to `interuss.pool_status.heartbeat.write` :)

PR #1235, opened 2025-07-30
[yugabyte] Update resources for yugabyte services

the-glu:
Update resources for yugabyte servers with more sensitive defaults after running full test suites:

- Lights resources increased added on yugabyte tservers nodes
- Lowered requests on all cased to allow deployments on small cluster.

Both on tanka and helm values.

(There are differences between tanka and helm deployments, eg. we don't set limits on helm, hence the differences in the diff)

### DIFF
--- a/deploy/services/tanka/yugabyte.libsonnet
+++ b/deploy/services/tanka/yugabyte.libsonnet
@@ -420,7 +420,7 @@ local volumes = import 'volumes.libsonnet';
                   },
                   requests: {
                     cpu: if metadata.yugabyte.light_resources then 0.75 else 1,
-                    memory: if metadata.yugabyte.light_resources then "2Gi" else "2Gi",
+                    memory: "2Gi",
                   },
                 },
                 ports: [{


deploy/services/tanka/yugabyte.libsonnet
```diff
@@ -415,12 +415,12 @@ local volumes = import 'volumes.libsonnet';
...
-                    cpu: if metadata.yugabyte.light_resources then 0.1 else 2,
-                    memory: if metadata.yugabyte.light_resources then "0.5Gi" else "4Gi",
+                    cpu: if metadata.yugabyte.light_resources then 1.5 else 2,
+                    memory: if metadata.yugabyte.light_resources then "2Gi" else "4Gi",
                   },
                   requests: {
-                    cpu: if metadata.yugabyte.light_resources then 0.1 else 2,
-                    memory: if metadata.yugabyte.light_resources then "0.5Gi" else "4Gi",
+                    cpu: if metadata.yugabyte.light_resources then 0.75 else 1,
+                    memory: if metadata.yugabyte.light_resources then "2Gi" else "2Gi",
```
barroco:
```suggestion
                    memory: "2Gi",
```

PR #1237, opened 2025-07-30
[terraform/tanka] Move generated tanka file into tanka project

the-glu:
The tanka file generated by terraform states is not directly usable anymore, because of the introduction of [tanka folder hierarchy](https://tanka.dev/directory-structure/).

This PR fix it by using a (similarly named to existing folders) workspace folder into the tanka file.


the-glu:
> I am expecting some updates to the documentation of the following modules:

True, fixed :)

PR #1243, opened 2025-08-15
[docs] Migrate relevant documentation to mkdocs-generated website

mickmis:
This migrates all documentation I did find relevant to be migrated, i.e. end-user documentation, not code/dev documentation. I used my best judgement here, but some might be wrong. All old pages now have a link to https://interuss.github.io/dss instead.

The content is left as is, save for links that were fixes. More work will be required to make the documentation better.
Do note that some links to documentation refers to documentation that remained with the code, in those cases I put the full web GitHub link on master branch.

Do note that I've moved `assets` into `docs/assets` in order for those files to be included in the documentation.

I've pushed manually the built documentation on 'dev' version, it can be seen here: https://interuss.github.io/dss/dev/


mickmis:
> It seems like there is just a placeholder at that URL; does it maybe get overwritten whenever a PR is merged? If so, let's ad-hoc freeze PR merges, have you manually push again, then I'll review, approve, and merge this PR.

Indeed! Sorry about that. I've pushed it manually to a dedicated version that won't get overridden by the CI on master: https://interuss.github.io/dss/pr-1243/
I will delete it after merging.

PR #1244, opened 2025-08-18
[dss/scd] Add rest converter for UssAvailabilityStatus

RustedBones:
Add `ToRest` conversion helper to `UssAvailabilityStatus` to factorize some code in the `uss_availability_handler`

### DIFF
--- a/pkg/scd/models/availability.go
+++ b/pkg/scd/models/availability.go
@@ -49,3 +49,10 @@ func UssAvailabilityStateFromString(s string) (UssAvailabilityState, error) {
 func UssAvailabilityStateFromRest(s restapi.UssAvailabilityState) (UssAvailabilityState, error) {
 	return UssAvailabilityStateFromString(string(s))
 }
+
+func (u UssAvailabilityStatus) ToRest() *restapi.UssAvailabilityStatus {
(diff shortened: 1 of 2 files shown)


pkg/scd/models/availability.go
```diff
@@ -49,3 +49,10 @@ func UssAvailabilityStateFromString(s string) (UssAvailabilityState, error) {
 func UssAvailabilityStateFromRest(s restapi.UssAvailabilityState) (UssAvailabilityState, error) {
 	return UssAvailabilityStateFromString(string(s))
 }
+
+func (u UssAvailabilityStatus) ToRest() *restapi.UssAvailabilityStatus {
```
mickmis, line 53: Drop the pointer? `UssAvailabilityState.ToRest` does so and all usages dereference the pointer.

RustedBones: The other `ToRest` apis not returning a pointer are for `string` type aliases.
The ones returning a struct use a pointer:
- `func (o *OperationalIntent) ToRest() *restapi.OperationalIntentReference`
- `func (c *Constraint) ToRest() *restapi.ConstraintReference`

mickmis: Alright makes sense :+1:

PR #1245, opened 2025-08-18
[dss/scd] update uss availability only if old version matches

### LINKED ISSUE #1170: utml.yaml: PUT /dss/v1/uss_availability/{uss_id}: old_version check is not implemented
**Describe the bug**
According to the description of the field `old_version` of the `PUT /dss/v1/uss_availability/{uss_id}` endpoint of `utm.yaml`, the endpoint is supposed to support a form of compare-and-swap operation:
> Version of USS's availability to change, for consistent read-modify-write operations and consistent retry behavior.

However this is not currently implemented in: https://github.com/interuss/dss/blob/2c0ff28842fe6b106b81c98a262827f2ee7d4372/pkg/scd/uss_availability_handler.go#L75

**Expected behavior**
The endpoint implements the behavior described by the OpenAPI specifications.

RustedBones:
Fix #1170


BenjaminPelletier (approved):
We should also make a TODO to adjust the autogeneration code to make the order of elements deterministic (presumably `sort_keys=True`) so PRs like this won't be nearly as big

PR #1254, opened 2025-08-27
[rid] Disable fallback on 'defaultdb'

### LINKED ISSUE #1251: RID server may incorrectly fall back to 'defaultdb' database
**Describe the bug**
Initially reported here: https://github.com/interuss/dss/issues/1241#issuecomment-3223443136

In v4.0.0 of DB schemas the RID database was renamed from `defaultdb` to `rid`. Out of retrocompatibility, at startup, when setting up connection to the database for RID, `createRIDServers` may fall back to using `defaultdb` instead of `rid`:
https://github.com/interuss/dss/blob/e56fefbc2842aa06aa51547d0fbbc028cf543720/cmds/core-service/main.go#L158-L176

This is done when `NewStore` fails. However no specific check of the error is done: if it fails e.g. because of transient network issues, the connection will fall back to `defaultdb`. After a migration to v4 schema, this DB still exists however it is empty.

The impact is probably low: the service should be stopped at that point since in theory do data should be present in `defaultdb`. A restart would likely solve the transient issue.

**Expected behavior**
Fall back to `defaultdb` only when the error is about a non-existent DB.

Or remove this altogether since it actually might have never been required: no retrocompatibility of schema is expected between major versions, since the the store will return an error at init if the schema is of a different major version: 
https://github.com/interuss/dss/blob/acee6db8b2777cf3f5f011242978b2c3d30184cf/pkg/rid/store/cockroach/store.go#L86-L104
And the database rename bumped up from 3 to 4 the major version.

the-glu:
Fix #1251

Remove backward compatibility to 'defaultdb' since the store will return an error at init if the schema is of a different major version


BenjaminPelletier (approved):
I think this is fine because we have [already made schema v4 required](https://github.com/interuss/dss/pull/1142).

PR #1261, opened 2025-09-02
[db-manager] Move rid cleanup cronjob to db-manager evict

the-glu:
Base version for implementation of #1250 , with some open questions :)

Doesn't include infrastructure changes.

PR #1262, opened 2025-09-02
[rid] Fix Garbage Collector to use the current repository version

### LINKED ISSUE #1252: RID garbage collector may be using a different repo version
**Describe the bug**
Initially reported here: https://github.com/interuss/dss/issues/1241#issuecomment-3223463556

In `createRIDServers` we create a `Repository` out of the `Store`. When creating the `Repository` it is determined by `Interact` or `Transact` which version of the repository to use. Most of those are done dynamically for each call. However for the garbage collector this is done at startup and this never changes during runtime:
https://github.com/interuss/dss/blob/e56fefbc2842aa06aa51547d0fbbc028cf543720/cmds/core-service/main.go#L178-L182

As such, if there is a migration applied during the lifetime of the instance, the garbage collector may remain stuck with an old version of the repository while the rest of the code does not.

The impact is likely low at the moment since we have a single version of the repository present (recently the v3 was deleted).

**Expected behavior**
Garbage collector determines repository version at each invokation. 
Passing the store instead of the repository could be an easy fix.

MariemBaccari:
- Fixes #1252
- Passes the Store instead of the Repository to the Garbage Collector creation
- Recreates the repository inside the Garbage Collector methods 
- Updates unit tests to reflect changes

### DIFF
--- a/cmds/core-service/main.go
+++ b/cmds/core-service/main.go
@@ -175,11 +175,7 @@ func createRIDServers(ctx context.Context, locality string, logger *zap.Logger) 
 		}
 	}
 
-	repo, err := ridStore.Interact(ctx)
-	if err != nil {
-		return nil, nil, stacktrace.Propagate(err, "Unable to interact with store")
-	}
-	gc := ridc.NewGarbageCollector(repo, locality)
+	gc := ridc.NewGarbageCollector(ridStore, locality)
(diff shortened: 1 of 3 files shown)


MariemBaccari:
Already fixed by #1261

PR #1263, opened 2025-09-02
[logging] Add Subject field to log output 

### LINKED ISSUE #1051: Include `sub` field in log output lines
When examining DSS logs I would like to quickly filter requests from a specific authorized USS client.

Include the `sub` field of the request's decoded JWT in a field in the JSON log entry for each request. `req_sub` would look parallel next to `req_dump` and similar fields.

- Process logs and decode JWT headers then re-remit the log with the `sub` value extracted.
- Extract the `uss_base_url` from certain operations and filter based on them. Not all requests have this, however.

MariemBaccari:
- Fixes #1051
- Adds an authentication middleware to decode tokens and pass the Subject claim to the context. 
- Retrieves the value from the context during logging.

When provided with a valid token, the log outputs the following:
`2025-09-03T06:21:28.402Z	info	logging/http.go:95	GET / HTTP/1.1	{"address": ":8082", "req_dump": "", "req_sub": "uss1", "resp_dump": "404 page not found\n", "req_headers": {"Accept":["application/json"],"Authorization":["Bearer ...`

When the token decoding fails, the log outputs the following: (edited following the nit fix)
`info	logging/http.go:95	GET / HTTP/1.1	{"address": ":8082", "req_dump": "", "resp_sub_err": "Access token validation failed: token is expired by 29h16m20s", "resp_dump": "404 page not found\n", "req_headers": {"Accept":["application/json"],"Authorization":["Bearer`

Currently, we decode the token twice (in Authorize and then in the new ExtractClaims call in the authentication middleware). A PR removing the second decoding will follow soon.


barroco (approved):
nit: I think the stack trace in the error message doesn't bring any value. I would recommend to only keep the error message.

PR #1266, opened 2025-09-09
[auth] Decode token once 

MariemBaccari:
This PR is a follow-up to #1263 and removes the duplicate decoding of the authentication token. 
It decodes the token once in the authentication middleware and passes the result in the context.


cmds/core-service/main.go
```diff
@@ -385,15 +386,29 @@ func healthyEndpointMiddleware(logger *zap.Logger, next http.Handler) http.Handl
...
-
 		claims, err := authorizer.ExtractClaims(r)
+		if err == nil {
+			if !authorizer.AcceptedAudiences[claims.Audience] {
+				err = stacktrace.NewErrorWithCode(dsserr.Unauthenticated, "Invalid access token audience: %v", claims.Audience)
+			}
+		}
+
+		ctx = context.WithValue(r.Context(), auth.CtxAuthKey{}, auth.CtxAuthValue{
+			Claims: claims,
```
barroco: Instead of using the context to pass the subject to the logger, we may want to zap it in the logger directly like the httpmiddleware is [doing it](https://github.com/interuss/dss/pull/1266/files#diff-5858646a5e039a83d56e3fd01dd2b5c0d086e6709edf93fa1fbcd56b52aaad95R50). Could you please evaluate this approach and propose a new version ?

barroco (dismissed):

cmds/core-service/main.go
MariemBaccari: In [9cbfe67](https://github.com/interuss/dss/pull/1266/commits/9cbfe6fa07d438b26f60b348f5359a07131c7595) I added the logger as an argument to the auth middleware to write the values there and pass the new (enriched) logger to the httpMiddleware. What do you think ?

pkg/auth/auth.go
```diff
@@ -182,44 +183,46 @@ func (a *Authorizer) setKeys(keys []interface{}) {
 	a.keyGuard.Unlock()
 }
 
+type CtxAuthKey struct{}
+type CtxAuthValue struct {
```
mickmis: nit: you are in package `auth`, so you may drop the `Auth` part from the variable names.

cmds/core-service/main.go
```diff
@@ -381,19 +380,33 @@ func healthyEndpointMiddleware(logger *zap.Logger, next http.Handler) http.Handl
...
 			//remove the stacktrace using the formatting specifier "%#s"
-			ctx = context.WithValue(r.Context(), logging.CtxAuthError{}, fmt.Sprintf("%#s", err))
-		} else {
-			ctx = context.WithValue(r.Context(), logging.CtxAuthSubject{}, claims.Subject)
+			errMsg = fmt.Sprintf("%#s", err)
 		}
 
+		ctx = context.WithValue(ctx, logging.CtxAuthKey{}, logging.CtxAuthValue{
+			Subject: claims.Subject,
+			ErrMsg:  errMsg,
```
mickmis: Why is propagating the error necessary? Remove if not. 
Then the subject may be propagated with a single `string` 

Also, nit, but for context key you may do:
```go
type CtxKey string

ctx = context.WithValue(ctx, CtxKey("sub"), parsedSub)
```

mickmis (commented):
I see this PR intends to optimize the runtime of the DSS in avoiding decoding the token twice by passing it through the context.
The main challenge here is dealing with the different approaches of how the logging and the authorization are performed. The logging is made through an http middleware, while the authorization is by invoking the authorizer close to the endpoint handler.

This PR deals with that by moving part of the authorization in a middleware. This is fine as long as:
- the split makes sense, here on one side we validate the token validity w.r.t. the instance (signature, audience) and on the other side w.r.t. the request (scopes)
- those related logics are not scattered away in the codebase

As such I suggest here to:
- move `authMiddleware` to being a method of `Authorizer` in `auth` package and rename it to reflect better its function, e.g. `TokenMiddleware`:
```go
func (a *Authorizer) TokenMiddleware(handler http.Handler) http.Handler

// in main.go:
``` 
- clarify that `Authorize` authorizes the request knowing that the token was already validated before by the middleware 
- unexpose unecessary exposed functions and structs (this is BTW a hint towards related logic being scattered) such as `ExtractClaims` or `Claims` or `AcceptedAudiences`

cmds/core-service/main.go
MariemBaccari: Propagating the error allows us to log it when the sub can't be found:

When provided with a valid token, the log outputs the following:
`2025-09-03T06:21:28.402Z	info	logging/http.go:95	GET / HTTP/1.1	{"address": ":8082", "req_dump": "", "req_sub": "uss1", "resp_dump": "404 page not found\n", "req_headers": {"Accept":["application/json"],"Authorization":["Bearer ...`

When the token decoding fails, the log outputs the following: 
`info	logging/http.go:95	GET / HTTP/1.1	{"address": ":8082", "req_dump": "", "resp_sub_err": "Access token validation failed: token is expired by 29h16m20s", "resp_dump": "404 page not found\n", "req_headers": {"Accept":["application/json"],"Authorization":["Bearer`

Is this aligned with the expected behavior ? If not, I can remove it.

pkg/auth/auth.go
```diff
@@ -120,10 +120,18 @@ func (r *JWKSResolver) ResolveKeys(ctx context.Context) ([]interface{}, error) {
...
-	logger            *zap.Logger
-	keys              []interface{}
-	keyGuard          sync.RWMutex
+	logger   *zap.Logger
+	keys     []interface{}
+	keyGuard sync.RWMutex
+
 	acceptedAudiences map[string]bool
+
+	decodedClaims *middlewareResult
```
mickmis: Not OK to pass this through the `Authorizer` struct, we will have race conditions: `decodedClaims` is scoped to the request, while the authorizer is scoped to the runtime. Pass this through the context instead, like for the `sub`. The context is made for stuff scoped to the request.

pkg/logging/http.go
```diff
@@ -39,8 +39,11 @@ func (w *tracingResponseWriter) WriteHeader(statusCode int) {
 	w.next.WriteHeader(statusCode)
 }
 
-type CtxAuthError struct{}
-type CtxAuthSubject struct{}
+type CtxKey string
+type CtxAuthValue struct {
```
mickmis: nit: rather than a custom struct, I'd suggest to flatten this and just use two different keys with a string value type `CtxKey("sub")` and `CtxKey("sub_err_msg")`.

pkg/logging/http.go
```diff
@@ -72,12 +75,11 @@ func HTTPMiddleware(logger *zap.Logger, dump bool, handler http.Handler) http.Ha
 			}
 		}
 
-		subject, ok := r.Context().Value(CtxAuthSubject{}).(string)
-		if !ok {
-			authErrorMsg := r.Context().Value(CtxAuthError{}).(string)
-			logger = logger.With(zap.String("resp_sub_err", authErrorMsg))
+		authResults := r.Context().Value(CtxKey("sub")).(CtxAuthValue)
```
mickmis: This removes the type assertion (with the `value, ok := var.(Type)` pattern), meaning that a wrong type will cause this line to panic (which may also happen when there is no value at all).
This is acceptable when it is obvious that a wrong type is a programming error. Which in this current version of the PR is the case (`sub` should always be added by the token middleware), however if you do implement my previous nit about flattening the `CtxAuthValue`, this might not be the case anymore (where you would have either `sub` or `sub_err_msg`).

pkg/auth/auth_test.go
```diff
@@ -115,7 +115,12 @@ func TestRSAAuthInterceptor(t *testing.T) {
 
 	for i, test := range authTests {
 		t.Run(strconv.Itoa(i), func(t *testing.T) {
-			res := a.Authorize(nil, test.req, []api.AuthorizationOption{})
+			claims, err := a.extractClaims(test.req)
+			a.decodedClaims = &middlewareResult{
+				claims: claims,
+				err:    err,
+			}
+			res := a.Authorize(nil, test.req.WithContext(context.Background()), []api.AuthorizationOption{})
```
mickmis: nit: use `t.Context()` instead

pkg/auth/auth.go
```diff
@@ -182,30 +190,68 @@ func (a *Authorizer) setKeys(keys []interface{}) {
...
+		handler.ServeHTTP(w, r.WithContext(ctx))
+	})
+}
+
+// Authorize extracts and verifies bearer tokens from a http.Request after it was validated by the TokenMiddleware.
 func (a *Authorizer) Authorize(_ http.ResponseWriter, r *http.Request, authOptions []api.AuthorizationOption) api.AuthorizationResult {
-	keyClaims, err := a.ExtractClaims(r)
-	if err != nil {
-		return api.AuthorizationResult{Error: stacktrace.PropagateWithCode(err, dsserr.Unauthenticated, "Failed to extract claims from access token")}
+	if a.decodedClaims == nil {
```
mickmis: This should come from the context, and not form the `Authorizer`, c.f. other comment.

pkg/auth/auth.go
```diff
@@ -182,30 +190,68 @@ func (a *Authorizer) setKeys(keys []interface{}) {
...
+		return api.AuthorizationResult{Error: stacktrace.NewErrorWithCode(dsserr.Unauthenticated, "Access token not found")}
 	}
 
-	if !a.acceptedAudiences[keyClaims.Audience] {
```
mickmis, line 192: Actually, why not keeping the audience check here if we have the claims?

pkg/auth/auth.go
```diff
@@ -240,15 +286,6 @@ func (a *Authorizer) ExtractClaims(r *http.Request) (claims, error) {
 	return keyClaims, nil
 }
 
-func HasScope(scopes []string, requiredScope api.RequiredScope) bool {
```
mickmis, line 243: Any reason why this function has moved? This makes the diff harder to read.

mickmis (changes requested):
Some comments, most importantly the one about `decodedClaims`. But we are getting there :)

barroco:
@MariemBaccari thanks again for the initial contribution.
I am closing it according to the tsc guidelines: https://github.com/interuss/tsc/blob/main/repo_contributions.md#mark-pr-ready-for-review
Feel free to re-open the PR whenever some progress has been made.

pkg/auth/auth.go
MariemBaccari: I moved it back, sorry for the mixup.

pkg/logging/http.go
```diff
@@ -72,12 +76,14 @@ func HTTPMiddleware(logger *zap.Logger, dump bool, handler http.Handler) http.Ha
...
-		subject, ok := r.Context().Value(CtxAuthSubject{}).(string)
-		if !ok {
-			authErrorMsg := r.Context().Value(CtxAuthError{}).(string)
-			logger = logger.With(zap.String("resp_sub_err", authErrorMsg))
-		} else {
-			logger = logger.With(zap.String("req_sub", subject))
+		if errMsgValue := r.Context().Value(CtxKeyErrMsg); errMsgValue != nil {
+			if errMsg, ok := errMsgValue.(string); ok && errMsg != "" {
+				logger = logger.With(zap.String("resp_sub_err", errMsg))
+			}
```
mickmis: I do not think it is really necessary within the logging middleware to get the JWT subject decoding error, if any. In case of error let's just not put anything in the context. The error will be handled by the authorizer anyway. And propagating an error within the context is already not a very much common practice, so let's not do it twice. 

So I suggest to remove this and `CtxKeyErrMsg`.

pkg/auth/auth.go
```diff
@@ -182,11 +182,49 @@ func (a *Authorizer) setKeys(keys []interface{}) {
...
+)
+
+// TokenMiddleware decodes the authentication token and passes the claims to the authorizer and to the context for logging.
+func (a *Authorizer) TokenMiddleware(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		claims, err := a.extractClaims(r)
+		if err != nil {
+			// remove the stacktrace using the formatting specifier "%#s"
+			ctx = context.WithValue(ctx, logging.CtxKeyErrMsg, fmt.Sprintf("%#s", err))
```
mickmis: To be removed (see other comment).

pkg/auth/auth.go
```diff
@@ -182,11 +182,49 @@ func (a *Authorizer) setKeys(keys []interface{}) {
...
+			ctx = context.WithValue(ctx, ctxKeyError, err)
+		} else {
+			ctx = context.WithValue(ctx, logging.CtxKeySub, claims.Subject)
```
mickmis: We don't actually need to inject the subject here, do we? Since we are already injecting the claims. I suggest:
- remove injection of the subject by itself
- use the injected `claims` in the logging middleware to retrieve the subject
- you will probably encounter a dependency loop between packages `auth` and `logging`: factor away `claims` (and `ScopeSet`) in a sub package (e.g. `pkg/auth/claims`) so that it can be used both by `auth` and `logging` without a dependency loop

Then, I suggest using the following pattern to set / retrieve the claims in / from the context:
```go
// file `pkg/auth/claims/claims.go`

type ctxKey string
var key = ctxKey("claims")

func NewContext(ctx context.Context, claims Claims) context.Context {
	return context.WithValue(ctx, key, claims)
}

func FromContext(ctx context.Context) (claims Claims, error) {
	claims, ok := ctx.Value(key).(Claims)
	if !ok {
		return "", fmt.Errorf(...)
	}

	return claims, nil
}

type ScopeSet ...

type Claims struct ...

```

That way from other packages we can just invoke `ctx := claims.NewContext(ctx, claims)` and `claims := claims.FromContext(ctx)`.

mickmis (commented):
One last pass of simplification and we are good to go.

pkg/logging/http.go
```diff
@@ -72,12 +70,9 @@ func HTTPMiddleware(logger *zap.Logger, dump bool, handler http.Handler) http.Ha
...
 
-		subject, ok := r.Context().Value(CtxAuthSubject{}).(string)
-		if !ok {
-			authErrorMsg := r.Context().Value(CtxAuthError{}).(string)
-			logger = logger.With(zap.String("resp_sub_err", authErrorMsg))
-		} else {
-			logger = logger.With(zap.String("req_sub", subject))
+		claimsValue, _ := claims.FromContext(r.Context())
+		if claimsValue.Subject != "" {
+			logger = logger.With(zap.String("req_sub", claimsValue.Subject))
```
mickmis: nit (limit scope of variable `claimsValue`):
```suggestion
		if claimsValue, _ := claims.FromContext(r.Context()); claimsValue.Subject != "" {
			logger = logger.With(zap.String("req_sub", claimsValue.Subject))
```

pkg/auth/claims/claims.go
```diff
@@ -18,6 +19,34 @@ var (
...
 )
 
+type ctxKey string
+
+var (
+	claimsKey = ctxKey("claims")
+	errKey    = ctxKey("error")
+)
+
+func NewContext(ctx context.Context, claims Claims, err error) context.Context {
```
mickmis: This abstracts error checking, making for an odd control flow. I suggest doing the error check at call site and split `NewContext` into a separate function like `NewContextFromError`.

pkg/auth/auth.go
mickmis: Audience check is still removed?

mickmis (approved):
LGTM modulo minor comments

pkg/auth/auth.go
MariemBaccari: I had previously missed it, I put it back now.

PR #1288, opened 2025-10-21
[client-certificates] Add generate-clients step and fix default client

barroco:
This PR cleans up client certificate generation.
The first commit adjusts the certificate manager to use the expected default yugabyte name in the CN of the certificate and adds a new command to regenerate the client if the cluster has already been initialized.
The second and third adjust tanka and helm volume mounts to the new file name.

### DIFF
--- a/deploy/operations/certificates-management/init.py
+++ b/deploy/operations/certificates-management/init.py
@@ -240,7 +240,7 @@ def do_init(cluster):
     )
 
 def do_generate_clients(cluster):
-    """Gemerate clients certificates"""
+    """Generate clients certificates"""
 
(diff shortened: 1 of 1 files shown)


deploy/operations/certificates-management/init.py
```diff
@@ -238,3 +238,12 @@ def do_init(cluster):
     l.info(
         "The new cluster certificates are ready! Don't forget to 'apply' the configuration."
     )
+
+def do_generate_clients(cluster):
+    """Gemerate clients certificates"""
```
mickmis: typo

mickmis (approved):
LGTM

PR #1290, opened 2025-10-21
[security] Require client certificate to connect to yugabyte sql interface

barroco:
This PR enforces ssl and authentication of clients on ysql connections using client certificate.

### DIFF
--- a/deploy/services/helm-charts/dss/values.yaml
+++ b/deploy/services/helm-charts/dss/values.yaml
@@ -64,6 +64,8 @@ yugabyte:
       placement_cloud: "cloud-1"
       placement_region: "uss-1"
       placement_zone: "zone-1"
+      use_client_to_server_encryption: true
+      ysql_hba_conf_csv: 'hostssl all all 0.0.0.0/0 cert'
 
 monitoring:
   enabled: false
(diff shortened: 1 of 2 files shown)


deploy/services/tanka/yugabyte-auxiliary.libsonnet
mickmis, line 31: `use_client_to_server_encryption` is not present in the master gflags (it is for the tserver though), while it is set for the helm charts.

deploy/services/helm-charts/dss/values.yaml
```diff
@@ -52,3 +52,12 @@ yugabyte:
     clientToServer: true
     insecure: false
     provided: true
+
+  gflags:
+    master:
+      use_client_to_server_encryption: true
```
the-glu: It's not needed to set this flag explicitly, it's handled in the state file via line 52 above (19d9d1ce3665e9d62d9a2538267116eb6a90644b9275b38ab30aeac127076243R52)

deploy/services/tanka/yugabyte-auxiliary.libsonnet
the-glu: Yes, it don't need to be set for helm charts ^^'

PR #1306, opened 2025-11-25
[doc] Set of small fixes

the-glu:
* Move minikuke at same level of others infrastructure modules
* Improve service index file
* Improve service titles
* Mark tanka documentation as needing more work
* Fix warnings about links
* Improve wording for (.md | folder | folder) that don't make sense on a webpage


mickmis (approved):
LGTM

PR #1312, opened 2025-12-08
[yugabyte] Instruction to build interuss yugabyte image

the-glu:
Instruction to build interuss yugabyte's image with mtls fix until the fix is present in an official yugabyte release.

### DIFF
--- a/README_DSS.md
+++ b/README_DSS.md
@@ -43,6 +43,9 @@ The current implementation relied historically on CockroachDB for data storage a
 between DSS participants but is migrating to Yugabyte with both version supported. See [implementation details](implementation_details.md) for
 more information.
 
+The DSS temporarily requires a patched version of Yugabyte, the reader may find [here](build/yugabyte_image_building.md) instructions on how to build it.
+
+
 ### DB Manager
 
 Component responsible for performing orderly database schema migrations, including database bootstrapping. Invoked automatically on first run along with the CRDB cluster-init job, or when requested manually by a user.
--- a/build/yugabyte_image_building.md
+++ b/build/yugabyte_image_building.md
@@ -1,0 +1,34 @@
+Custom yugabyte image building
+==============================
+
+To support mtls authentification, we need to build a custom image until [a fix](https://github.com/yugabyte/yugabyte-db/commit/89685fa888daca54eb3164a8c301e3bda8cf41b0) is relased into an official image.
+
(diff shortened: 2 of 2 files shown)


yugabyte_image_building.md
```diff
@@ -0,0 +1,34 @@
+Custom yugabyte image building
+==============================
+
+To support mtls authentification, we need to build a custom image until [a fix](https://github.com/yugabyte/yugabyte-db/commit/89685fa888daca54eb3164a8c301e3bda8cf41b0) is relased into an official image.
```
mickmis, line 4:
```suggestion
To support mtls authentification, we need to build a custom image until [a fix](https://github.com/yugabyte/yugabyte-db/commit/89685fa888daca54eb3164a8c301e3bda8cf41b0) is released into an official image.
```

yugabyte_image_building.md
```diff
@@ -0,0 +1,34 @@
...
+
+You will need a [working Yugabyte building environment](https://docs.yugabyte.com/stable/contribute/core-database/build-from-src-almalinux/). It's recommanded to install a fresh AlmaLinux 8.
```
mickmis, line 6:
```suggestion
You will need a [working Yugabyte building environment](https://docs.yugabyte.com/stable/contribute/core-database/build-from-src-almalinux/). It's recommended to install a fresh AlmaLinux 8.
```

the-glu:
> The doc file LGTM (modulo typos), but I'm unsure about where it is located and whether or not it should be linked to. Notably, should it be in the documentation website?
> When would we need to build this image? That would probably give a hint as to where it should be located and from where it should be linked to.

I think it's mostly internal, e.g. targeting people with release-access levels, so I think keeping in the repo is fine.

It will be needed if we need a new version of yugabyte and need to reapply the patch.

mickmis:
> I think it's mostly internal, e.g. targeting people with release-access levels, so I think keeping in the repo is fine.

OK makes sense

> It will be needed if we need a new version of yugabyte and need to reapply the patch.

OK then it might make sense to put it under `build/`, WDYT?

the-glu:
> OK then it might make sense to put it under `build/`, WDYT?

I would have think that `build/` is for the DSS, not "external" dependencies, but I'm not sure it change much, I moved it ^^'

the-glu:
> Looking at the .md, it might makes sense to link that from https://github.com/interuss/dss/blob/master/README_DSS.md. Maybe under https://github.com/interuss/dss/blob/master/README_DSS.md#yugabyte something like:
> 
> > The DSS temporarily requires a patched version of Yugabyte, the reader may find _here_ instructions on how to build it.

Done :)

PR #1315, opened 2025-12-08
[ci] Reduce infrastructure information exposure

barroco:
In order to reduce the information exposure of the infrastructure used by the CI job `Deploy DSS`, terraform variables containing references to actual AWS resources or domain names have been moved to environment secrets.

There are still some URL references in the [USS qualifier specific configuration](https://github.com/interuss/dss/blob/master/deploy/operations/ci/aws-1/test-resources.yaml) which will be addressed separately if needed.

### DIFF
--- a/.github/workflows/dss-deploy.yml
+++ b/.github/workflows/dss-deploy.yml
@@ -42,6 +42,7 @@ jobs:
           COMPOSE_PROFILES: aws-1
           AWS_REGION: us-east-1
           TF_VAR_app_hostname: ${{ secrets.APP_HOSTNAME }}
+          TF_VAR_authorization: ${{ secrets.AUTHORIZATION }}
           TF_VAR_aws_iam_administrator_role: ${{ secrets.AWS_IAM_ADMINISTRATOR_ROLE }}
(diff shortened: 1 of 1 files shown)


.github/workflows/dss-deploy.yml
```diff
@@ -20,30 +21,32 @@ jobs:
...
-          aws sts get-caller-identity
-
-      - name: Test Deployment Scenario AWS-1
+      - name: "Test Deployment Scenario: AWS-1"
         shell: bash
         working-directory: ./deploy/operations/
         env:
           COMPOSE_PROFILES: aws-1
           AWS_REGION: us-east-1
+          TF_VAR_app_hostname: ${{ secrets.APP_HOSTNAME }}
```
mickmis, line 44: `TF_VAR_authorization` is missing

mickmis (approved):
Was this tested to work? If so, LGTM modulo adding missing var

.github/workflows/dss-deploy.yml
barroco: Tests are still in progress [here](https://github.com/Orbitalize/dss/actions).

PR #1330, opened 2025-12-30
[helm] Add support for external monitoring option

the-glu:
Add support for the [external monitoring option](https://interuss.github.io/dss/latest/operations/monitoring/#externally-exposing-prometheus) on helm chart as well.

Implementation is better than the tanka one, as it's support https, a following up PR will bring it up to date.

### DIFF
--- a/deploy/services/helm-charts/dss/templates/prometheus-loadbalancers.yaml
+++ b/deploy/services/helm-charts/dss/templates/prometheus-loadbalancers.yaml
@@ -22,7 +22,7 @@ metadata:
     app: {{$.Release.Name}}-prometheus
     name: {{$.Release.Name}}-prometheus-external
   annotations:
-    cloud.google.com/backend-config: '{"default": "promtheus-external"}'
+    cloud.google.com/backend-config: '{"default": "prometheus-external"}'
   name: {{$.Release.Name}}-prometheus-external
 spec:
   ports:
@@ -33,7 +33,7 @@ spec:
   selector:
      app.kubernetes.io/instance: "{{$.Release.Name}}"
      app.kubernetes.io/name: "prometheus"
-  type: NodePort
+  type: ClusterIP
 
 ---
 apiVersion: networking.k8s.io/v1
--- a/deploy/services/helm-charts/dss/values.yaml
+++ b/deploy/services/helm-charts/dss/values.yaml
@@ -72,11 +72,8 @@ monitoring:
     ip: ""
     subnet: ""
     hostname: ""
-    # AWS
-    allowedIPs:
-      - 1.2.3.4/32
-    # GCP
-    allowedIPsPolicy: "cloud-armor-policy-prometheus"
+    allowedIPs: []
+    allowedIPsPolicy: ""
 
 prometheus:
   server:
@@ -279,4 +276,3 @@ grafana:
   persistence:
     type: pvc
     enabled: true
->>>>>>> f3220540 ([helm] Add support for monitoring stack)
--- a/deploy/services/helm-charts/dss/values.example.yaml
+++ b/deploy/services/helm-charts/dss/values.example.yaml
@@ -80,6 +80,15 @@ global:
 
 monitoring:  # Set to true to enable monitoring stack
   enabled: true
+  externalService:
+    enabled: false  # Set this to true to expose prometheus on a public endpoint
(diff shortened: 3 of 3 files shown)


deploy/services/helm-charts/dss/values.schema.json
```diff
@@ -349,6 +349,38 @@
...
+            "subnet": {
+              "description": "Optional: Subnet of the public ip for AWS only",
+              "type": "string"
+            },
+            "hostname": {
+              "description": "Optional: Hostname of the public ip for GCP only",
+              "type": "string"
+            },
+            "allowedIPs": {
+              "description": "List of IP allowed to connect to the service, for AWS only",
```
barroco:
```suggestion
              "description": "List of IP ranges allowed to connect to the service, for AWS only",
```

deploy/services/helm-charts/dss/templates/prometheus-loadbalancers.yaml
```diff
@@ -0,0 +1,115 @@
...
+     app.kubernetes.io/instance: "{{$.Release.Name}}"
+     app.kubernetes.io/name: "prometheus"
+  type: NodePort
+
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  annotations:
+    {{- include (printf "%s-ingress-dss-gateway-annotations" $cloudProvider)
```
barroco:
```suggestion
    {{- include (printf "%s-ingress-prometheus-annotations" $cloudProvider)
```

deploy/services/helm-charts/dss/templates/prometheus-loadbalancers.yaml
```diff
@@ -0,0 +1,115 @@
...
+spec:
+  ports:
+    - name: prometheus
+      port: 9090
+      targetPort: 9090
+  publishNotReadyAddresses: true
+  selector:
+     app.kubernetes.io/instance: "{{$.Release.Name}}"
+     app.kubernetes.io/name: "prometheus"
+  type: NodePort
```
barroco: Why using a Nodeport here instead of a LoadBalancer ?

deploy/services/helm-charts/dss/templates/prometheus-loadbalancers.yaml
```diff
@@ -0,0 +1,115 @@
...
+{{- if $.Values.monitoring.enabled }}
+{{- if $.Values.monitoring.externalService.enabled }}
+
+{{- if eq $cloudProvider "google" }}
+
+---
+apiVersion: cloud.google.com/v1
+kind: BackendConfig
+metadata:
+  name: promtheus-external
```
barroco:
```suggestion
  name: prometheus-external
```

deploy/services/helm-charts/dss/templates/prometheus-loadbalancers.yaml
```diff
@@ -0,0 +1,115 @@
...
+
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: {{$.Release.Name}}-prometheus
+    name: {{$.Release.Name}}-prometheus-external
+  annotations:
+    cloud.google.com/backend-config: '{"default": "promtheus-external"}'
```
barroco:
```suggestion
    cloud.google.com/backend-config: '{"default": "prometheus-external"}'
```

deploy/services/helm-charts/dss/values.yaml
```diff
@@ -67,6 +67,16 @@ yugabyte:
 
 monitoring:
   enabled: false
+  externalService:
```
barroco, line 70: I don't think we want any example values in this file since it is default values taken into account by the chart.
Please move them to `values.example.yaml`.

deploy/services/helm-charts/dss/templates/prometheus-loadbalancers.yaml
the-glu: That a mistake/leftover from tanka, but I switched to a `ClusterIP` (to be consistent with the DSS service that use the same system with the certificate)

deploy/services/helm-charts/dss/values.yaml
the-glu: Fixed (and added quick doc in `values.example.yaml` as well)

PR #1350, opened 2026-01-19
[prometheus] Add anotations on deployment to ensure redeployment on config changes

the-glu:
Fix issue with prometheus not being redeployed on config changes in helm and tanka by adding annotations.

On helm we have limited capability over the helm chart, so annotation is manual.
On tanka that automatic.

### DIFF
--- a/deploy/services/helm-charts/dss/values.yaml
+++ b/deploy/services/helm-charts/dss/values.yaml
@@ -85,7 +85,7 @@ prometheus:
       evaluation_interval: 5s
 
     podAnnotations:
-      serverFilesVersion: "1"  # Increment this value when changing serverFiles to force redeployment
+      serverFilesVersion: "1"  # Since the prometheus helm chart does not detect changes in server files, this value is used to force the redeployment following the update of the serverFiles defined below. 
 
     service:
       annotations:


deploy/services/helm-charts/dss/values.yaml
```diff
@@ -84,6 +84,9 @@ prometheus:
       scrape_timeout: 5s
       evaluation_interval: 5s
 
+    podAnnotations:
+      serverFilesVersion: "1"  # Increment this value when changing serverFiles to force redeployment
```
barroco:
```suggestion
      serverFilesVersion: "1"  # Since the prometheus helm chart does not detect changes in server files, this value is used to force the redeployment following the update of the serverFiles defined below. 
```

deploy/services/helm-charts/dss/values.yaml
```diff
@@ -101,7 +104,7 @@ prometheus:
   alertmanager:
     enabled: false
 
-  serverFiles:
+  serverFiles:  # Caution: When changing this, please update the podAnnotations above
```
barroco:
```suggestion
  serverFiles:  # Caution: Since the prometheus helm chart does not detect changes in server files, update the `prometheus.server.podAnnotations.serverFilesVersion` to force the redeployment of prometheus. 
```

barroco (approved):

PR #1354, opened 2026-01-19
[certificates] Add monitoring certificates 

the-glu:
Follow #1353 

Add monitoring certificates generation.

Goal is to use that for grafana <-> prometheus communication, and prometheus's federation.

### DIFF
--- a/deploy/operations/certificates-management/cluster.py
+++ b/deploy/operations/certificates-management/cluster.py
@@ -104,7 +104,7 @@ class Cluster(object):
     def clients(self):
         return [
             "yugabyte",
-            "client.grafana",  # Grafana expect the 'client.'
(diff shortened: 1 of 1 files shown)


deploy/operations/certificates-management/cluster.py
```diff
@@ -94,7 +102,10 @@ def is_ready(self):
 
     @property
     def clients(self):
-        return ["yugabyte"]  # TODO: Do we need more, like a specifc one for the DSS?
+        return [
+            "yugabyte",
+            "client.grafana",  # Grafana expect the 'client.'
```
barroco:
```suggestion
            "client.grafana",  # Grafana expects the 'client.'
```

barroco (approved):

Issue #1362, opened 2026-01-21
[doc] Performance boundaries


barroco:
Document the performance of the DSS under specific conditions (max. latency between instances, number of instances, ...) and define the expected boundaries in terms of query per seconds (QPS) (typical QPS, max QPS under peaks, ...).

mickmis:
Please find below a proposal for a load-testing scenario intended to be realistic that we can simulate in order to determine the performance boundaries of the DSS and hopefully replicate issues that are reported from users. Feedback is important and appreciated, notably on the 'realistic' part, and the bolded questions.
Tagging for feedback: @BenjaminPelletier @wing-utm-sharing-airspace @BradNicolle @davidmurphy1152
 
# Deployment
12 (CRDB+DSS) nodes per USS total ([setup from dss#1311](https://github.com/interuss/dss/issues/1311)):
- 4 USSs
- 3 nodes per USS

## Simulation of deployment over internet
Used [GCP data](https://docs.cloud.google.com/network-intelligence-center/docs/performance-dashboard/how-to/view-google-cloud-latency?hl=fr:), [AWS data](https://www.cloudping.co/), and [this article](https://www.bigbitbus.com/2018/05/07/What-Is-Your-Ping-AWS-And-Google-Cloud/) for guess-timates of P50 and P90 values for latency and packet loss.

- intra-USS (assumed in the same cloud region): 
  - latency: P50: 0.5ms, P90: 5ms
  - packet loss: P50: 0%, P90: 0.02%
- inter-USS intra-cloud (assumed east-west US coasts):
  - latency: P50: 40ms, P90: 65ms
  - packet loss: P50: 0.05% P90: 0.15%
- inter-USS inter-cloud (assumed east-west US coasts):
  - latency: P50: 65ms, P90: 85ms
  - packet loss: P50: 0.1% P90: 0.20%

# Test scenario
## Flights specifications
- extents:
  - time: current day, H-5m to H+25m (i.e. currently active)
  - altitude: 0-120m
  - geometry: 
    - narrow rectangle spanning 2 or 3 km (i.e. intersecting 3 or 4 cells at least, given that DSS cells are ~1km2)
    - centered on a location L0-7
- state: Accepted/Activated
- uss_base_url: USS A/B/C/D
- subscription: explicit if USS A or B

## Target state
- 64 concurrently active flights
  - **Is this a correct order of magnitude for an upper bound of the number of concurrently active flights?**
[...]

Issue #1311, opened 2025-12-05
High Latency for SCD DSS Requests


wing-utm-sharing-airspace:
**Describe the bug**
We've noticed timeouts on various DSS SCD API calls (exceeding 10s) in several of our DSS pools. Looking at the CRDB console for one instance of this issue, I'm seeing a lot of transaction restarts and very high tail latency for SCD Subscription queries in particular:

<img width="998" height="584" alt="Image" src="https://github.com/user-attachments/assets/c7902a85-b7a2-41d7-8a0d-22ad8f243cec" />

<img width="999" height="775" alt="Image" src="https://github.com/user-attachments/assets/7e695357-12d8-41bf-9e80-f8f80756db44" />

<img width="2007" height="443" alt="Image" src="https://github.com/user-attachments/assets/99c72246-745f-40b1-816a-4a6b17f1dfe4" />

This looks somewhat similar to https://github.com/interuss/dss/issues/1241.

**To Reproduce**
Unsure the best way to reproduce but will provide some context about the deployment and traffic levels where we've seen this issue:

Below screenshot shows QPS overlayed with latency spike:
<img width="670" height="564" alt="Image" src="https://github.com/user-attachments/assets/d9da4f67-7d7f-4215-8de4-d1a1467bfc51" />

This is a pooled deployment where 3 providers each have 3 CRDB nodes.
We're running DSS v0.20.0 and CRDB v24.1.3

**Expected behavior**
This is instance is receiving fairly spikey, high load compared with what we need to support our current operations but would appreciate investigation into the restarts and high latency to determine system limitations and available performance improvements.

**Additional context**
Sample of logs from around the time of the latency spike:
[downloaded-logs-20251205-124443.json](https://github.com/user-attachments/files/23966839/downloaded-logs-20251205-124443.json)

the-glu:
Hello,

What the approximate size of the `scd_subscriptions` table?

Do you use the command to clean the table? If I'm correct, in v0.20.0 it's via https://github.com/interuss/dss/tree/interuss/dss/v0.20.0/cmds/db-manager/cleanup and is not done automatically.

wing-utm-sharing-airspace:
Sorry for the late response - we ran the cleanup tool a couple of weeks ago but I don't believe we run it super regularly so possible there's a ton of subscriptions building up in those tables. I'll look into that as well.

[...]

PR #1363, opened 2026-01-21
[scd] Change the subcription locking mechanism to a simple, shared lock.

the-glu:
This PR proposes a new method of locking subscriptions, by having a shared lock used for all subscriptions, instead of locking all rows.

See  https://github.com/interuss/dss/issues/1311#issuecomment-3778873578 for graphs and details about the behavior change.

I added documentation in operations/performances, expecting #1362 to reuse it as well.


build/db_schemas/yugabyte/scd/upto-v1.1.0-add_locks.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS scd_locks (
```
mickmis, line 1: For both schema migration files: this is missing the down script.

build/db_schemas/scd/upto-v3.3.0-add_locks.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS scd_locks (
+  id INT64 PRIMARY KEY
+);
+
+INSERT INTO scd_locks (id) VALUES (0);
+
+UPDATE schema_versions
+SET schema_version = 'v3.3.0'
```
mickmis, line 8: I'm wondering whether this should be a major version bump: an older schema with a newer image will appear to work, however the locking won't be effective on the DSS instances with the older images. 

Do I understand correctly that this will have an impact only on performances? Or will it have an impact on the correctness of the results? If the former: probably not needed to major version bump. But not if the latter.

build/db_schemas/scd/upto-v3.3.0-add_locks.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS scd_locks (
+  id INT64 PRIMARY KEY
+);
+
+INSERT INTO scd_locks (id) VALUES (0);
```
mickmis: Why not use the same pattern as for the `schema_versions` table? Use-case is similar. 
Also the column `id` is IMO misleading by its name.

build/db_schemas/scd/upto-v3.3.0-add_locks.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS scd_locks (
```
mickmis, line 1: Would it make sense to add a column specifying the instance that created the lock? That might facilitate debugging if required.

docs/operations/performances.md
```diff
@@ -0,0 +1,28 @@
+# Performances
+
+## The SCD global lock option
+
+!!! danger
+     All DSS instances in a DSS pool must use the same value for this option. Mixing will result in dramatically lower performance.
```
mickmis, line 6: IMO there isn't enough safeguards or mitigation that this does not happen. At least IMO we want some visibility, e.g. expose this configuration flag on the `_aux` interface? 
cc @BenjaminPelletier

pkg/scd/store/datastore/subscriptions.go
```diff
@@ -380,26 +380,37 @@ func (c *repo) IncrementNotificationIndices(ctx context.Context, subscriptionIds
...
 
-	_, err := c.q.Exec(ctx, query, dsssql.CellUnionToCellIds(cells), ids)
-	if err != nil {
-		return stacktrace.Propagate(err, "Error in query: %s", query)
+		_, err := c.q.Exec(ctx, query)
+		if err != nil {
+			return stacktrace.Propagate(err, "Error in query: %s", query)
+		}
+
+	} else {
```
mickmis: Readability: outdent this `else` and early return in the `if`.

build/db_schemas/scd/upto-v3.3.0-add_locks.sql
the-glu: It won't work if the option is used, but only in that case, so I'm not sure about the answer.

I would say that as it's only an experimental one, that safe to keep it as minor, as it shouldn't be required for normal usage.

build/db_schemas/scd/upto-v3.3.0-add_locks.sql
the-glu: Not really, as we just do a select for update. We could add a column and update it, but that would mean more 'work' in the transaction (we now have to commit something).

And as it would be in the transaction, we would only see the value when it's committed (so too late).

build/db_schemas/scd/upto-v3.3.0-add_locks.sql
the-glu: Idea is to potentially having multiple locks (one per cell) if needed / with more works to have smarter lock, so the pattern is different.

For `id`, I will rename it to `key` (similar to pg's lock function)

docs/operations/performances.md
mickmis: From InterUSS weekly: 
- solution to this should not take a large effort
- failing fast is OK
- addressing this through doc only is OK
- exposing through aux interface is OK

the-glu: I added the info in the aux interface

build/db_schemas/scd/upto-v3.3.0-add_locks.sql
mickmis: > Idea is to potentially having multiple locks (one per cell) if needed / with more works to have smarter lock, so the pattern is different.

I don't think this is reflected anywhere in the PR, but that's pretty relevant info, could you add it somewhere? (in doc, in code as comment, in sql as comment, etc.)

build/db_schemas/scd/upto-v3.3.0-add_locks.sql
mickmis: OK makes sense. I wrote this comment not understanding fully the mechanism.

docs/operations/performances.md
mickmis: Then, since it is exposed, maybe it would makes sense adding some test case in monitoring ensuring the values are consistent across the instances?

build/db_schemas/scd.libsonnet
mickmis, line 7: Missing the downfrom

build/db_schemas/scd/upto-v3.3.0-add_locks.sql
mickmis: OK so putting more thoughts into this, as is, migrating first the schema is OK without having the DSS instances images being updated (since we are only adding a table). That's probably the most important. So we should not need a major upgrade indeed.

docs/operations/performances.md
the-glu: Mmm I'm not sure it's worth the hassle of having a full validation for that, as we may have better solutions in the future (e.g. with crdb support dropped) and that will only affect performances, in 'high load' cases.

build/db_schemas/scd/upto-v3.3.0-add_locks.sql
the-glu: Added in the code as comment

the-glu:
NB: CI failure should be fixed with #1374

Issue #1257, opened 2025-08-27
Improve observability of DSS instances


mickmis:
**Is your feature request related to a problem? Please describe.**
Feedback from real-life DSS deployments at least twice reported performance issues caused (possibly partially) by accumulation of stale data in the DB, leading to latency and contention resulting in degraded overall performances.

Those accumulations could probably have been detected before they had any performance impact. E.g. if the number of subscriptions was exposed as a metric and an alert was setup if it reaches a certain threshold.

**Describe the solution you'd like**
Solution would need to be discussed, but should cover:
- exposing some metrics (Prometheus, REST endpoint, in heartbeat, ...)
- alerting with some sane defaults
- operational documentation to configure those correctly, as values would probably change between deployments

**Describe alternatives you've considered**
N/A

**Additional context**
#1241
#1074 
Discussed during InterUSS weekly 2025-08-26. Issue #1181 was mentioned as possibly relevant.

PR #1365, opened 2026-01-29
[monitoring] Add OpenTelemetry option

the-glu:
This is a set of 3 PRs, proposition to add [OpenTelemetry](https://opentelemetry.io/) to generate optional traces.
You can check the [base PR](https://github.com/interuss/dss/pull/1365), the one for [handlers](https://github.com/interuss/dss/pull/1366) and the one for [datastores](https://github.com/interuss/dss/pull/1367).

The idea is to contribute to #1257 by providing an option to enable traces.

This first PR provide base support, including HTTP & SQL traces.

Traces looks like that (in that PR): 
image_2026-01-29_12-39-43

As OpenTelemetry is a framework, this allow processing traces with various compatibles services. Self-hostable examples include [jaeger](https://www.jaegertracing.io/), [openobserve](https://github.com/openobserve/openobserve), [Grafana Tempo](https://grafana.com/docs/tempo/latest/), [SigNoz](https://github.com/SigNoz/signoz). Multiples SaaS solutions exists (some previous tools provide them). As long as there is OpenTelemetry support  it can be used, and that also mean it could be integrated into the existing stack of an organization.

As an example, same traces (of last PR) in SigNoz:

image_2026-01-29_13-58-33.

Idea behind thoses PR is to use it only for tracing, not metrics nor logging, however it also possible to do so in the future.

Goal is to have it optional behind a flag, and only enabled if needed, mostly for debugging. Configuration is done via OpenTelemetry's own flags.

Documentation is not done yet, but this can be tested with a local jaeger (port 16686) with the following changes:

```patch

diff --git a/build/dev/docker-compose_dss.yaml b/build/dev/docker-compose_dss.yaml
index e66d2ba4..d204104b 100644
--- a/build/dev/docker-compose_dss.yaml
+++ b/build/dev/docker-compose_dss.yaml
@@ -133,6 +133,7 @@ services:
       COMPOSE_PROFILES: ${COMPOSE_PROFILES}
       # Note: requires the Dockerfile to have been built with "-cover" in the EXTRA_GO_INSTALL_FLAGS var
       GOCOVERDIR: "/startup/coverdata"
+      OTEL_EXPORTER_OTLP_ENDPOINT: "http://jaeger:4317"
     command: /startup/core_service.sh ${DEBUG_ON:-0}
     ports:
       - "4000:4000"
@@ -178,6 +179,16 @@ services:
       start_period: 30s
       start_interval: 5s
 
+  jaeger:
+    image: jaegertracing/jaeger:2.14.0
+    ports:
+      - "16686:16686"
+      - "4317:4317"
+      - "4318:4318"
+      - "5778:5778"
+      - "9411:9411"
+    networks:
+      - dss_sandbox_default_network
 networks:
   dss_sandbox_default_network:
     name: dss_sandbox-default
diff --git a/build/dev/startup/core_service.sh b/build/dev/startup/core_service.sh
index f09bda59..8c96cf3e 100755
--- a/build/dev/startup/core_service.sh
+++ b/build/dev/startup/core_service.sh
@@ -44,5 +44,6 @@ else
   -enable_scd \
   -allow_http_base_urls \
   -locality local_dev \
-  -public_endpoint http://127.0.0.1:8082
+  -public_endpoint http://127.0.0.1:8082 \
+  -enable_opentelemetry
 fi
```


cmds/core-service/main.go
```diff
@@ -400,6 +411,18 @@ func main() {
...
 
+	// Set up OpenTelemetry.
+	if *enableOpenTelemetry {
+		otelShutdown, err := setupOTelSDK(ctx)
+		if err != nil {
+			logger.Panic("Failed to initialize OpenTelemetry", zap.Error(err))
+		}
+		// Handle shutdown properly so nothing leaks.
+		defer func() {
+			err = errors.Join(err, otelShutdown(context.Background()))
```
mickmis: What's the intent with that `defer` call? In any case I suspect there is something wrong since it is called after the previous potential panic, yet it tries to join the `err` (which is overriden with `:=` so will always be nil here).

cmds/core-service/otel.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// shutdown calls cleanup functions registered via shutdownFuncs.
+	// The errors from the calls are joined.
+	// Each registered cleanup will be invoked once.
+	shutdown := func(ctx context.Context) error {
+		var err error
+		for _, fn := range shutdownFuncs {
+			err = errors.Join(err, fn(ctx))
+		}
+		shutdownFuncs = nil
```
mickmis: Why is this necessary?

cmds/core-service/otel.go
```diff
@@ -0,0 +1,83 @@
...
+		return err
+	}
+
+	// handleErr calls shutdown for cleanup and makes sure that all errors are returned.
+	handleErr := func(inErr error) {
```
mickmis: Why is this necessary? Why not inline it? It is called once.

cmds/core-service/otel.go
```diff
@@ -0,0 +1,83 @@
...
+	// Set up trace provider.
+	tracerProvider, err := newTracerProvider(ctx)
+	if err != nil {
+		handleErr(err)
+		return shutdown, err
+	}
+	shutdownFuncs = append(shutdownFuncs, tracerProvider.Shutdown)
+	otel.SetTracerProvider(tracerProvider)
+
+	return shutdown, err
```
mickmis:
```suggestion
	return shutdown, nil
```

cmds/core-service/otel.go
```diff
@@ -0,0 +1,83 @@
...
+		propagation.TraceContext{},
+		propagation.Baggage{},
+	)
+}
+
+func newTracerProvider(ctx context.Context) (*trace.TracerProvider, error) {
+
+	traceExporter, err := otlptracegrpc.New(ctx)
+	if err != nil {
+		panic(err)
```
mickmis: Don't panic, return error, also below.

cmds/core-service/main.go
the-glu: That coming from the example but indeed that seems strange, I fixed it with a simple defer, we don't need / care about potential error on shutdown anyways.

cmds/core-service/otel.go
the-glu: By default `setupOTelSDK` is plan to have multiple shutdown function, but indeed since we only use traces everything else is useless, I simplified everything

cmds/core-service/main.go
```diff
@@ -314,6 +316,14 @@ func RunHTTPServer(ctx context.Context, ctxCanceler func(), address, locality st
 	handler = logging.HTTPMiddleware(logger, *dumpRequests, handler)
 	handler = authorizer.TokenMiddleware(handler)
 
+	if *enableOpenTelemetry {
+		httpSpanName := func(operation string, req *http.Request) string {
+			return fmt.Sprintf("%s %s", req.Method, req.URL.Path)
```
mickmis, line 321: I see you changed that, and indeed for the better, however checking the [semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name), the second part should be the route template and not the actual path (e.g. `/uss/v1/operational_intents/{entityid}/telemetry` and not `/uss/v1/operational_intents/4654654/telemetry`).

cmds/core-service/otel.go
```diff
@@ -0,0 +1,61 @@
...
+			semconv.ServiceNameKey.String("dss"),
+		),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	tracerProvider := trace.NewTracerProvider(
+		trace.WithBatcher(traceExporter,
+			trace.WithBatchTimeout(time.Second)),
```
mickmis: How was this value selected?

mickmis (approved):
LGTM modulo comments

cmds/core-service/main.go
the-glu: Yes I wanted to do that in the `openapi-server` generator (since here we don't have it), but I started some cleanup in it first.

I suggest we merge it like this for know and improve it in a follow up PRs. (And ev. switch back to HTTP here if that too problematic).

the-glu: (See #1391)

cmds/core-service/otel.go
the-glu: From documentation, but removed to use defaults.

PR #1369, opened 2026-02-11
[tooling] Enforce openapi scopes generator order

the-glu:
The `openapi-to-go-server` return a unordered set of scopes, meaning that we get random unrelated changes everytime we build the API.

(Example: https://github.com/interuss/dss/pull/1363/changes/de42be09c471319408a3dff39b5d78b50a52cab0#diff-bb21552cc2d287b0949b37cae5b6dcf993eae2b0ca31a1e0e63f0dca8fbcd606L10 )

This PR enforce sorting to ensure stable generation of interface files.

### DIFF
--- a/interfaces/openapi-to-go-server/apis.py
+++ b/interfaces/openapi-to-go-server/apis.py
@@ -59,7 +59,7 @@ class API:
         self.data_types = [dt for dt in self.data_types
                            if dt.name in required_data_types]
 
-    def security_scopes(self) -> Set[Tuple[str, operations.Scope]]:
+    def security_scopes(self) -> List[Tuple[str, operations.Scope]]:
         """Returns a set of unique security scopes used by this API.
 
         :return: set of unique tuples containing the authorization scheme and scope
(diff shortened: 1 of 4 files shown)


interfaces/openapi-to-go-server/apis.py
```diff
@@ -59,7 +59,7 @@ def filter_operations(self, tags: Set[str]):
         self.data_types = [dt for dt in self.data_types
                            if dt.name in required_data_types]
 
-    def security_scopes(self) -> Set[Tuple[str, operations.Scope]]:
+    def security_scopes(self) -> List[Tuple[str, operations.Scope]]:
         """Returns a set of unique security scopes used by this API.
```
the-glu, line 63: nb: it's still a set of scopes, just represented as a list, so I didn't changed the description, but we may want to.

barroco: We are losing the deduplication in case the scope is defined multiple times, no ?
Wouldn't it be better to sort on the rendering side ?

the-glu: We're not, the function itself is still using a set to do deduplication.

For the location of the sort, I would think it's better there, as everyone using the function is taking advantage of the sort. (But right now it's only used once)

barroco (approved):

PR #1380, opened 2026-02-25
[scd] Validate Volume4D when parsing

RustedBones:
Factorize validation for `Volume3D` and `Volume4D` during parsing:
- Ensure `startTime` < `endTime` when defined
- Ensure `altLo` < `altHi` when defined

Introduce parsing options to tune validation:
- Ensure `altLo` and `altHi` set when `RequireAltitudeBounds` enabled
- Ensure `startTime` ans `endTime` set when `RequireTimeBounds` enabled

Configuration options for SCD entities mutation:
- op-intent:  `RequireTimeBounds: true, RequireAltitudeBounds: true`
- constraints: `RequireTimeBounds: true, RequireAltitudeBounds: false`
- subscriptions: `RequireTimeBounds: false, RequireAltitudeBounds: false`

For SCD queries: `RequireTimeBounds: false, RequireAltitudeBounds: false`

### DIFF
--- a/pkg/models/scd_conversions.go
+++ b/pkg/models/scd_conversions.go
@@ -7,38 +7,63 @@ import (
 	"github.com/interuss/stacktrace"
 )
 
+type Volume4DOpts struct {
+	RequireAltitudeBounds bool
+	RequireTimeBounds     bool
+}
+
+func (opts *Volume4DOpts) to3DOpts() Volume3DOpts {
+	return Volume3DOpts{
+		RequireAltitudeBounds: opts.RequireAltitudeBounds,
+	}
+}
+
 // Volume4DFromSCDRest converts vol4 SCD v1 REST model to a Volume4D
-func Volume4DFromSCDRest(vol4 *restapi.Volume4D) (*Volume4D, error) {
-	vol3, err := Volume3DFromSCDRest(&vol4.Volume)
+func Volume4DFromSCDRest(vol4 *restapi.Volume4D, opts Volume4DOpts) (*Volume4D, error) {
+	vol3, err := Volume3DFromSCDRest(&vol4.Volume, opts.to3DOpts())
 	if err != nil {
 		return nil, err // No need to Propagate this error as this stack layer does not add useful information
 	}
 
-	result := &Volume4D{
-		SpatialVolume: vol3,
-	}
-
+	var startTime *time.Time
 	if vol4.TimeStart != nil {
 		ts, err := time.Parse(time.RFC3339Nano, vol4.TimeStart.Value)
 		if err != nil {
 			return nil, stacktrace.Propagate(err, "Error converting start time")
 		}
-		result.StartTime = &ts
+		startTime = &ts
+	} else if opts.RequireTimeBounds {
+		return nil, stacktrace.NewError("Missing start time")
 	}
 
+	var endTime *time.Time
 	if vol4.TimeEnd != nil {
 		ts, err := time.Parse(time.RFC3339Nano, vol4.TimeEnd.Value)
 		if err != nil {
 			return nil, stacktrace.Propagate(err, "Error converting end time")
 		}
-		result.EndTime = &ts
+		endTime = &ts
+	} else if opts.RequireTimeBounds {
+		return nil, stacktrace.NewError("Missing end time")
 	}
 
-	return result, nil
+	if startTime != nil && endTime != nil && startTime.After(*endTime) {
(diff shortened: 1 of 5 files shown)


pkg/models/scd_conversions.go
```diff
@@ -7,38 +7,63 @@ import (
...
+func (opts *Volume4DOpts) to3DOpts() Volume3DOpts {
+	return Volume3DOpts{
+		RequireAltitudeBounds: opts.RequireAltitudeBounds,
+	}
+}
+
 // Volume4DFromSCDRest converts vol4 SCD v1 REST model to a Volume4D
-func Volume4DFromSCDRest(vol4 *restapi.Volume4D) (*Volume4D, error) {
-	vol3, err := Volume3DFromSCDRest(&vol4.Volume)
+func Volume4DFromSCDRest(vol4 *restapi.Volume4D, opts Volume4DOpts) (*Volume4D, error) {
```
mickmis: Using something inspired from the options/builder patterns could provide a cleaner and extensible interface here, i.e. sth like:
```go
type Volume4DValidator func(*Volume4D) error

func WithRequireTimeBounds() Volume4DValidator {
    return func(v *Volume4D) {
        if v.StartTime == nil {
            return fmt.Errorf("no start time")
        }
        if v.EndTime == nil {
            return fmt.Errorf("no end time")
        }
    }
}

func Volume4DFromSCDRest(vol4 *restapi.Volume4D, validators ...Volume4DValidator) (*Volume4D, error) {
    ...
    for _, validator := range validators {
        if err := validator(svc); err != nil {
            ...
        }
    }
    ...
}
```

pkg/models/scd_conversions.go
```diff
@@ -7,38 +7,63 @@ import (
...
+	}
+
+	return &Volume4D{
+		SpatialVolume: vol3,
+		StartTime:     startTime,
+		EndTime:       endTime,
+	}, nil
+}
+
+type Volume3DOpts struct {
```
mickmis: Do we actually ever need to validate those directly on a `Volume3D`? If not consider doing the validation only from `Volume4DFromSCDRest`.

mickmis (commented):
Intent SGTM, I do have suggestions regarding implementation, please LMK what you think.

pkg/scd/operational_intents_handler.go
```diff
@@ -518,21 +518,10 @@ func validateAndReturnOIRUpsertParams(
...
 	}
 
-	if valid.uExtent.StartTime == nil {
-		return nil, stacktrace.NewError("Missing time_start from extents")
-	}
-	if valid.uExtent.EndTime == nil {
-		return nil, stacktrace.NewError("Missing time_end from extents")
-	}
-
 	if now.After(*valid.uExtent.EndTime) {
```
mickmis, line 521: nit, not for this PR: pushing it further this could be a validator `WithRequireEndTimeBefore(now)`

pkg/models/scd_conversions.go
```diff
@@ -7,34 +7,78 @@ import (
...
+		if v.EndTime == nil {
+			return stacktrace.NewError("Missing end time")
+		}
+		return nil
+	}
+}
+
+func WithRequireAltitudeBounds() Volume4DValidator {
+	return func(v *Volume4D) error {
+		if v.SpatialVolume == nil {
```
mickmis: I understand this is to guard against a panic, however this is never valid, is it? Return an error if this is nil?

mickmis (approved):
LGTM modulo comment on `SpatialVolume`

pkg/scd/operational_intents_handler.go
RustedBones: Just to be sure we're alligned on this one, the doc states:

> Spacetime extents that bound this operational intent. Start and end times, as well as lower and upper altitudes, are required for each volume. The end time may not be in the past.

I think to move the validation on the unioned extent so we can accept composite extens with some expired volumes but not all. is that correct ?

pkg/models/scd_conversions.go
RustedBones: Actually this cannot happen. `SpatialVolume` is always set by the conversion.

pkg/scd/operational_intents_handler.go
RustedBones: See https://github.com/interuss/dss/pull/1381 for union validation

mickmis: Missed that comment - my bad.

PR #1381, opened 2026-02-25
[scd] Factorize volume4D union from rest

RustedBones:
Fix TODO from the code:
> TODO to factor out logic below into common multi-vol4d parse

Remove unused `extents` field when unioning
Add validator for expired 4D extent

### DIFF
--- a/pkg/scd/constraints_handler.go
+++ b/pkg/scd/constraints_handler.go
@@ -424,10 +424,6 @@ func validateAndReturnConstraintUpsertParams(
 		return nil, stacktrace.Propagate(err, "Invalid extents")
 	}
 
-	if now.After(*valid.uExtent.EndTime) {
-		return nil, stacktrace.NewError("Constraint may not end in the past")
-	}
-
 	valid.cells, err = valid.uExtent.CalculateSpatialCovering()
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "Invalid area")


pkg/models/scd_conversions.go
```diff
@@ -77,6 +86,30 @@ func Volume4DFromSCDRest(vol4 *restapi.Volume4D, validators ...Volume4DValidator
 	return volume, nil
 }
 
+// UnionVolume4DFromSCDRest converts a slice of vol4 SCD v1 REST model to a single bounding Volume4D
+func UnionVolume4DFromSCDRest(vol4s []restapi.Volume4D, validators ...Volume4DValidator) (*Volume4D, error) {
```
mickmis: Naming consistency:
```suggestion
func UnionVolumes4DFromSCDRest(vol4s []restapi.Volume4D, validators ...Volume4DValidator) (*Volume4D, error) {
```

pkg/models/scd_conversions.go
```diff
@@ -77,6 +86,30 @@ func Volume4DFromSCDRest(vol4 *restapi.Volume4D, validators ...Volume4DValidator
...
+	union, err := UnionVolumes4D(volumes...)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "Failed to union volumes")
+	}
+
+	for _, validator := range validators {
+		if err := validator(union); err != nil {
+			return nil, err
+		}
+	}
```
mickmis: Possible behavior change: before this PR the validators are invoked on each individual volume, and after it is on the aggregated one. Could we face a scenario where some individually invalid volumes are unionized into a valid one? Or the opposite? 
Out of caution and since this is used as well as validation of parameters, I would actually do the validation on both individual and unionized volumes. And in any case, make clear in the function doc what is the expected behavior. WDYT?
 
Also, factor away this loop? Duplicated with function above.

pkg/scd/constraints_handler.go
mickmis: Remove this?

pkg/models/scd_conversions.go
RustedBones: > Could we face a scenario where some individually invalid volumes are unionized into a valid one? Or the opposite?

Yes, that was my concern in https://github.com/interuss/dss/pull/1380#discussion_r3046280072. I think extent with some expired volumes should not be rejected by the DSS, that's why I moved the validation on the unioned volume. Is that fine ?

I'll add some doc and factor the loop.

mickmis: > Just to be sure we're alligned on this one, the doc states:
> > Spacetime extents that bound this operational intent. Start and end times, as well as lower and upper altitudes, are required for each volume. The end time may not be in the past.
> 
> I think to move the validation on the unioned extent so we can accept composite extens with some expired volumes but not all. is that correct ?

Hum that's an interesting case. IMO with a strict interpretation of the wording, the API would require all individual volumes to not end in the past. Now, implementation-wise, I do agree we should accept it as long as the aggregated volume is not in the past. And that's indeed what the current implementation is doing. So, agree with you for the 'end time not in the past' check.

As far as the other validations are concerned (time and altitude bounds), if they fail on an individual volume, they will fail for the aggregated one. So there will be no behavior change with the changes in this PR.

All in all LGTM with the clarification in the function doc of the behavior.

PR #1384, opened 2026-03-07
[scd] Time-bound subscription locks and add lock diagnostics

### LINKED ISSUE #1311: High Latency for SCD DSS Requests

ashishjsharda:
## Problem
Issue #1311 reports high latency and timeouts for SCD operations under overlapping traffic patterns. Lock acquisition in `LockSubscriptionsOnCells` becomes expensive when many subscriptions share S2 cells, regardless of temporal relevance.

## Solution
Narrow lock scope to subscriptions that overlap both the relevant S2 cells and the operation time window, while preserving correctness by always locking explicitly referenced subscription IDs. Add low-noise diagnostics to improve production observability of costly lock queries.

### 1. Time-bounded lock scope
Updated `LockSubscriptionsOnCells` to accept `startTime` and `endTime`, locking rows matching:
- `cells && $1`
- `COALESCE(starts_at <= endTime, true)`
- `COALESCE(ends_at >= startTime, true)`
- `OR id = ANY(explicitSubscriptionIDs)`

### 2. Call-site updates
Updated OIR mutation paths to pass the relevant time window:
- **Delete OIR:** uses `old.StartTime` / `old.EndTime`
- **Upsert OIR:** uses `validParams.uExtent.StartTime` / `validParams.uExtent.EndTime`

### 3. Lock diagnostics (thresholded, low-noise)
Warnings emitted when:
- `duration >= 500ms`, or
- `matched_rows >= 1000`

Fields logged:
- `global_lock`
- `duration`
- `matched_rows`
- `cell_count`
- `explicit_subscription_id_count`

Failure warnings:
- `SCD global lock query failed`
- `SCD subscription lock query failed`

### 4. Documentation
Extended `docs/operations/performances.md` with diagnostics behavior, fields, and thresholds.

## Validation
- `go test ./pkg/scd/store/datastore` — passed
- `go test ./pkg/scd/...` — passed locally

## Risk / Compatibility
- `LockSubscriptionsOnCells` interface signature updated; all in-repo call sites patched.
- Cell-based lock scope intentionally narrowed by time overlap; explicit subscription ID locking preserved.
- Diagnostics are passive and thresholded, minimizing steady-state log noise.

## Follow-up
- [ ] Validate against existing SCD load-test scenario from #1311 (p95/p99 latency, timeout/error rates)
- [ ] Tune thresholds based on production observations if needed

Closes #1311

### DIFF
--- a/pkg/scd/store/datastore/subscriptions.go
+++ b/pkg/scd/store/datastore/subscriptions.go
@@ -23,7 +23,7 @@ var (
 )
 
 const (
-	// These thresholds keep lock diagnostics low-noise in production while still surfacing likely bottlenecks.
+	// This threshold keep lock diagnostics low-noise in production while still surfacing likely bottlenecks.
 	lockQuerySlowThreshold = 4 * time.Second
 )
 


docs/operations/performances.md
```diff
@@ -28,3 +28,24 @@ All graphs have been generated with the [loadtest present in the monitoring repo
...
 ![](../assets/perfs_scd_lock_notoverlapping.png)
 *Non-overlapping requests. Notice the reduction of performance on the right, with a single lock.*
+
+## SCD lock diagnostics logs
+
+To help diagnose latency spikes (for example as discussed in [#1311](https://github.com/interuss/dss/issues/1311)), the SCD subscription lock path emits targeted warning logs when a lock query looks expensive.
+
+The warning is emitted only when one of the following thresholds is met:
+* Lock query duration is greater than or equal to `500ms`
+* Number of matched/locked rows is greater than or equal to `1000`
```
barroco: How those values were selected ?

docs/operations/performances.md
```diff
@@ -28,3 +28,24 @@ All graphs have been generated with the [loadtest present in the monitoring repo
...
+
+The log message is `SCD lock query diagnostics` and includes:
```
barroco: nit: I would suggest a log message stating more explicitly the reason of the log line such as `Expensive SCD lock detected`

pkg/scd/store/datastore/subscriptions.go
```diff
@@ -20,6 +22,12 @@ var (
 	subscriptionFieldsWithoutPrefix string
 )
 
+const (
+	// These thresholds keep lock diagnostics low-noise in production while still surfacing likely bottlenecks.
+	lockQuerySlowThreshold = 500 * time.Millisecond
+	lockQueryLargeRowCount = 1000
```
barroco: nit: It may be useful to have those values configurable so it can be adjusted per use cases. This can be done in a follow up PR.

barroco (commented):
Thank you very much @ashishjsharda for this contribution.

Please see comments inline.

docs/operations/performances.md
ashishjsharda: Great question. They were chosen as pragmatic initial guardrails to keep logs low-noise while still surfacing clear outliers:

- `500ms` was selected as a “this is unexpectedly slow for lock acquisition” threshold relative to our normal successful lock path.
- `1000` matched rows was selected as a “potentially broad lock scope” signal (large enough to avoid routine noise, small enough to catch problematic cases).

These are intentionally heuristic and meant for operational triage, not hard SLO limits. I can add a short note in the doc clarifying that these are initial defaults and may be tuned based on production observations.

docs/operations/performances.md
ashishjsharda: Good suggestion, agreed. I’ll update the diagnostic log message to make the reason explicit (for example `Expensive SCD lock detected`) while keeping the same structured fields (`duration`, `matched_rows`, `cell_count`, etc.) so existing analysis remains straightforward.

pkg/scd/store/datastore/subscriptions.go
ashishjsharda: Agreed. For this PR I kept constants to minimize scope and risk, but making these thresholds configurable is a sensible follow-up. I can open a follow-up issue/PR to expose them via config (with the current values as defaults) so deployments can tune per workload profile.

docs/operations/performances.md
barroco: Ok, thanks. Those values may generate log lines for every requests, looking at the graphs on the original issue.
While there is no requirement related to DSS response time for Operational Intent in SCD in F3548-21, the standard seems to assume that 5 seconds is a reasonable upper bound time for DSS responses in general. Therefore, 4 seconds may be more appropriate based on the experience of the deployed DSS pool referred in the issue.

Regarding the matched rows, it may possibly be removed, see my comments below.

pkg/scd/store/datastore/subscriptions.go
```diff
@@ -391,11 +400,28 @@ func (c *repo) LockSubscriptionsOnCells(ctx context.Context, cells s2.CellUnion,
...
+		duration := time.Since(start)
 		if err != nil {
+			logger.Warn("SCD global lock query failed",
+				zap.Duration("duration", duration),
+				zap.Error(err),
+			)
 			return stacktrace.Propagate(err, "Error in query: %s", query)
 		}
 
+		matchedRows := int(result.RowsAffected())
```
barroco: According to the [documentation](https://pkg.go.dev/database/sql#Result), it seems to always return 0. Do you know if the package pgxpool is implemented differently ? In this case, there is only one row in this table so we can remove that part of the check.

pkg/scd/store/datastore/subscriptions.go
```diff
@@ -417,11 +449,30 @@ func (c *repo) LockSubscriptionsOnCells(ctx context.Context, cells s2.CellUnion,
...
+		logger.Warn("SCD subscription lock query failed",
+			zap.Duration("duration", duration),
+			zap.Int("cell_count", len(cells)),
+			zap.Int("explicit_subscription_id_count", len(subscriptionIds)),
+			zap.Error(err),
+		)
 		return stacktrace.Propagate(err, "Error in query: %s", query)
 	}
 
+	matchedRows := int(result.RowsAffected())
```
barroco: Similarly, this seems to always return 0. Is RowsAffected of pgxpool behaving differently than the [sql definition](https://pkg.go.dev/database/sql#Result) ?
If it is not the case, I suggest to remove the check on row count. Counting with a query would be too expensive and may make the problem of performance worse.

pkg/scd/store/datastore/subscriptions.go
```diff
@@ -406,8 +432,14 @@ func (c *repo) LockSubscriptionsOnCells(ctx context.Context, cells s2.CellUnion,
         FROM
             scd_subscriptions
         WHERE
-            cells && $1
-        OR
+			(
```
barroco, line 432: @the-glu can you please evaluate the impact of this change with the load tests ?

pkg/scd/store/datastore/subscriptions.go
ashishjsharda: Thanks for flagging this. I double-checked the pgx/pgconn behavior: Exec returns a CommandTag, and for PostgreSQL SELECT tags include a row count (SELECT <rows>), so RowsAffected() is not universally 0 for this query type.

That said, your broader point about usefulness/noise is valid. To keep diagnostics robust and simple, I’m fine removing matched_rows from the warning criteria and relying on duration-based alerting.

pkg/scd/store/datastore/subscriptions.go
ashishjsharda: Agreed. I haven’t posted comparative load-test numbers yet. I’ll run the SCD loadtest scenarios (overlapping and non-overlapping) on this branch and share before/after results focused on p95/p99 latency and warning volume.

the-glu: This seems to improve the situation with default tests:

image

However, this is highly dependent on the flights duration. Loadtest use tests of 10s, switching to 90s show no improvements (but that expected).

I noticed however that lock may be wrong (but that was already the case for cells): `notifyVolume` can be different from `validParams`, explaining some transactions restarts.

This may increase those transactions restarts as the lock is now more specific (and may be more 'wrong'), but that only in case of updates.

docs/operations/performances.md
ashishjsharda: Good suggestion, addressed. I renamed the diagnostic warning message to `Expensive SCD lock detected` while keeping the structured context fields.

docs/operations/performances.md
ashishjsharda: Thanks, agreed on both points.

I pushed an update in `590c4fe`:
- switched diagnostics to a duration-only trigger at `4s`
- removed `matched_rows` from warning criteria,
- and updated docs accordingly.

This should keep the signal focused on truly expensive lock calls and avoid per-request noise in normal traffic.

pkg/scd/store/datastore/subscriptions.go
ashishjsharda: Thanks, agreed. I addressed this in `590c4fe8`:

- moved to a duration-only warning threshold (`4s`),
- removed `matched_rows` from warning criteria entirely.

So this no longer depends on `RowsAffected()` behavior for `SELECT ... FOR UPDATE`, and avoids extra/noisy row-count-based triggering.

pkg/scd/store/datastore/subscriptions.go
ashishjsharda: Thanks a lot for running these tests and sharing the comparison, this is very helpful.

The “improves in default 10s flights, no improvement at 90s” result makes sense, and matches the expectation that benefits are workload-pattern dependent.

On the lock-scope concern (`notifyVolume` potentially differing from `validParams`): agreed this is important. To keep this PR scoped and low-risk, I’d like to handle that as a dedicated follow-up investigation/PR focused specifically on update locking correctness and restart behavior. I’ll open a follow-up item and link it here so we can track it explicitly.

For this PR, with the current changes plus the diagnostics tuning updates (`590c4fe`), are you comfortable with merge?

pkg/scd/store/datastore/subscriptions.go
```diff
@@ -20,6 +22,11 @@ var (
 	subscriptionFieldsWithoutPrefix string
 )
 
+const (
+	// These thresholds keep lock diagnostics low-noise in production while still surfacing likely bottlenecks.
```
barroco: nit: This comment is now inaccurate since there is now only one threshold.

ashishjsharda: Good catch, thanks. Updated the comment to singular since only one threshold remains

PR #1391, opened 2026-03-11
 [monitoring] Add telemetry in router 

the-glu:
Add telemetry directly in router, providing clean tracenames.

Follow #1390 and #1365 

Example server tested with `./run_example.sh` and still working as expected.

Name has been removed from default handler, meaning unknown route or `/healthy` will just be named 'http', but that probably ok.

image_2026-03-11_10-40-38

Router set parent trace name to the path + start a span for actual processing.
Example traces:

image_2026-03-11_10-37-40
image_2026-03-11_10-38-11

### DIFF
--- a/interfaces/openapi-to-go-server/templates/server.go.template
+++ b/interfaces/openapi-to-go-server/templates/server.go.template
@@ -8,10 +8,24 @@ type APIRouter struct {
     Authorizer <API_PACKAGE>.Authorizer
 }
 
+var tracer = otel.Tracer("<PACKAGE>.<API_PACKAGE>")
+
 // *<PACKAGE>.APIRouter (type defined above) implements the <API_PACKAGE>.PartialRouter interface
 func (s *APIRouter) Handle(w http.ResponseWriter, r *http.Request) bool {
     for _, route := range s.Routes {
         if route.Method == r.Method && route.Pattern.MatchString(r.URL.Path) {
+
+            // We retrieve the current span from the otelhttp handler to set its name property.
+            span := trace.SpanFromContext(r.Context())
+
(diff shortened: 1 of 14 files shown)


interfaces/openapi-to-go-server/templates/server.go.template
```diff
@@ -8,10 +8,24 @@ type APIRouter struct {
...
+            span := trace.SpanFromContext(r.Context())
+
+            if span.IsRecording() {
+                // Current span is the one from the otelhttp handler
+                span.SetName(fmt.Sprintf("%s %s", r.Method, route.Path))
+            }
+
+            ctx, span := tracer.Start(r.Context(), route.Name)
+            defer span.End()
+            r = r.WithContext(ctx)
```
mickmis, line 27: I don't fully understand what this is doing:
- shouldn't we prevent calling `tracer.Start` if tracing is disabled? no check is done here
- except if that is the intent with `IsRecording`? but I don't understand what this returns exactly, what does it mean to be recording?  [this](https://opentelemetry.io/docs/specs/otel/trace/api/#isrecording) seems to hint at the fact this this is not exactly equivalent to 'is tracing enabled'

the-glu: * No need for that, the api take care of handling it in the back.
* `IsRecording` check if we can edit the current span name. Should no handler be attached, a dummy span will be returned and we cannot edit it name (or for some reason, the span form otelhttp is not editable). But basically, it come down to "can we set it name".

mickmis: Can you clarify through a comment that the intent of the first few lines is to set the name of the span that was created up the stack; and do the same up the stack (if not already there) specifying that the proper name will be set down the stack?

the-glu: Done!

PR #1398, opened 2026-03-18
[auth] Remove token signatures in logs

the-glu:
I noticed that the DSS logs are recording full tokens, which is a security concern.

This PR redacts the signature of JWT tokens, making them unusable while still logging the core payload to be debuggable.

### DIFF
--- a/pkg/logging/http.go
+++ b/pkg/logging/http.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/interuss/dss/pkg/auth/claims"
@@ -48,8 +49,10 @@ func redactHeaders(headers http.Header) http.Header {
 	newHeaders := headers.Clone()
 
 	for key, values := range newHeaders {
-		for i, val := range values {
-			newHeaders[key][i] = tokenRegex.ReplaceAllString(val, "$1[REDACTED]")
+		if strings.ToLower(key) == "authorization" {
+			for i, val := range values {
+				newHeaders[key][i] = tokenRegex.ReplaceAllString(val, "$1[REDACTED]")
+			}
 		}
 	}
 


pkg/logging/http.go
```diff
@@ -40,6 +41,21 @@ func (w *tracingResponseWriter) WriteHeader(statusCode int) {
 	w.next.WriteHeader(statusCode)
 }
 
+var tokenRegex = regexp.MustCompile(`(?i)(Bearer\s+[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.)([A-Za-z0-9-_]+)`)
+
+func redactHeaders(headers http.Header) http.Header {
+
+	newHeaders := headers.Clone()
+
+	for key, values := range newHeaders {
```
barroco, line 51: Perhaps it would be more efficient to directly target the authorization header instead of looping over all headers.

the-glu: Yes, fixed (but we still need to do a loop, as it's key insensitive).

pkg/logging/http.go
```diff
@@ -40,6 +42,23 @@ func (w *tracingResponseWriter) WriteHeader(statusCode int) {
...
+		if strings.ToLower(key) == "authorization" {
+			for i, val := range values {
+				newHeaders[key][i] = tokenRegex.ReplaceAllString(val, "$1[REDACTED]")
```
mickmis, line 54: If we are targeting the authorization header, might as well redact the whole value and not try to match the token itself, no?

mickmis (approved):
LGTM modulo comment

pkg/logging/http.go
the-glu: Not sure I understand, idea was to keep the base of the token for debugging, but make it useless without the signature.

barroco: I think there is value is keeping the content of the token safely by redacting the signature.
Since this is already an improvement compared to the current state, I propose to move forward and @mickmis feel free to propose another PR or open issue when you come back if there is still any concern.

barroco (approved):

pkg/logging/http.go
mickmis: Ah! I missed in the regex the last capture group and misread the title: I thought the intent was to redact the whole token. SGTM then. 

BTW, looking at it again, it might be worth adding unit tests covering different scenarios for this function.

PR #1403, opened 2026-03-31
Refactor store/datastore with common interfaces and helpers

the-glu:
This PR refactor the store / datastore system into common interfaces and helpers.

(Badly) duplicated code has been removed and everything is now normalized.

There are a few remarks / side effects, noted directly in comments.

Quick summary of changes:

* Store interface (`[db]/store/store.go`) moved to a generic one, old one just use the common.
* Base store in `pkg/datastore/store.go` that handle all generic part of old stores (e.g creation, `Interact/Transact`). What is left in `[db]/datatore/store.go` is special elements (eg. the config flag for scd, Cleanup) and call to BaseStore
* New helper in `pkg/datastoreutils/dial.go` that create data stores in a generic way for all DB. Replace all store creation that where in `core-service/db-manager` (and different in both cases).

The rest is unchanged, this don't redo the whole system but improve the current one.

### DIFF
--- a/cmds/core-service/main.go
+++ b/cmds/core-service/main.go
@@ -8,7 +8,6 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
-	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -24,8 +23,7 @@ import (
 	aux "github.com/interuss/dss/pkg/aux_"
 	auxc "github.com/interuss/dss/pkg/aux_/store/datastore"
 	"github.com/interuss/dss/pkg/build"
-	"github.com/interuss/dss/pkg/datastore"
-	"github.com/interuss/dss/pkg/datastore/flags" // Force command line flag registration
+	"github.com/interuss/dss/pkg/datastoreutils"
 	"github.com/interuss/dss/pkg/logging"
 	"github.com/interuss/dss/pkg/rid/application"
 	rid_v1 "github.com/interuss/dss/pkg/rid/server/v1"
@@ -36,7 +34,6 @@ import (
 	"github.com/interuss/dss/pkg/version"
 	"github.com/interuss/dss/pkg/versioning"
 	"github.com/interuss/stacktrace"
-	"github.com/robfig/cron/v3"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 	"go.uber.org/zap"
 )
@@ -65,30 +62,6 @@ var (
 	scdGlobalLock = flag.Bool("enable_scd_global_lock", false, "Experimental: Use a global lock when working with SCD subscriptions. Reduce global throughput but improve throughput with lot of subscriptions in the same areas.")
 )
 
-const (
-	codeRetryable = stacktrace.ErrorCode(1)
-)
-
-func getDBStats(ctx context.Context, db *datastore.Datastore, databaseName string) {
-	logger := logging.WithValuesFromContext(ctx, logging.Logger)
-	statsPtr := db.Pool.Stat()
-	stats := make(map[string]string)
-	stats["DBName"] = databaseName
-	stats["AcquireCount"] = strconv.Itoa(int(statsPtr.AcquireCount()))
-	stats["AcquiredConns"] = strconv.Itoa(int(statsPtr.AcquiredConns()))
-	stats["CanceledAcquireCount"] = strconv.Itoa(int(statsPtr.CanceledAcquireCount()))
-	stats["ConstructingConns"] = strconv.Itoa(int(statsPtr.ConstructingConns()))
-	stats["EmptyAcquireCount"] = strconv.Itoa(int(statsPtr.EmptyAcquireCount()))
-	stats["IdleConns"] = strconv.Itoa(int(statsPtr.IdleConns()))
-	stats["MaxConns"] = strconv.Itoa(int(statsPtr.MaxConns()))
-	stats["TotalConns"] = strconv.Itoa(int(statsPtr.TotalConns()))
-	if stats["TotalConns"] == "0" {
-		logger.Warn("Failed periodic DB Ping (TotalConns=0)", zap.String("Database", databaseName))
-	} else {
-		logger.Info("Successful periodic DB Ping ", zap.String("Database", databaseName))
-	}
-}
-
 func createKeyResolver() (auth.KeyResolver, error) {
 	switch {
 	case *pkFile != "":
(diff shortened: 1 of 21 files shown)


cmds/db-manager/cleanup/evict.go
```diff
@@ -54,12 +53,16 @@ func evict(cmd *cobra.Command, _ []string) error {
 	ctx, cancel := context.WithTimeout(ctx, *timeout)
 	defer cancel()
 
-	scdStore, err := getSCDStore(ctx)
+	logger := logging.WithValuesFromContext(ctx, logging.Logger)
+
+	datastoreutils.ApplicationName = "db-manager"
```
the-glu, line 58: This is probably not the best solution, but I wanted to avoid the parameters in Dial just for that :/

cmds/db-manager/cleanup/evict.go
```diff
@@ -169,45 +172,6 @@ func evict(cmd *cobra.Command, _ []string) error {
...
 
-func getSCDStore(ctx context.Context) (*scdc.Store, error) {
-	connectParameters := datastoreflags.ConnectParameters()
-	connectParameters.ApplicationName = "db-manager"
-	connectParameters.DBName = scdc.DatabaseName
-	datastore, err := datastore.Dial(ctx, connectParameters)
-	if err != nil {
-		logParams := connectParameters
-		logParams.Credentials.Password = "[REDACTED]"
-		return nil, fmt.Errorf("failed to connect to SCD database with %+v: %w", logParams, err)
```
the-glu, line 180: Errors message are now standarized, and therefor different, as I took ones from `core-services`. If really needed, we could also add logs of parameters globally, but user already set thoses options so..

pkg/datastore/flags/flags.go
```diff
@@ -17,7 +17,6 @@ func ConnectParameters() datastore.ConnectParameters {
 
 func init() {
 	flag.StringVar(&connectParameters.ApplicationName, "datastore_application_name", "dss", "application name for tagging the connection to the database")
-	flag.StringVar(&connectParameters.DBName, "datastore_db_name", "", "database name to connect to")
```
the-glu, line 20: This was never used, expect in some tests. All normal functions overwrote this value, making the parameter useless.

Now it's located in `<module>/store/datastore/store.go`, in one location + manually added in tests.

pkg/aux_/store/datastore/store.go
```diff
@@ -24,146 +19,47 @@ const (
...
-}
+	s := &Store{}
 
-// CheckCurrentMajorSchemaVersion checks that store supports the current major schema version.
-func (s *Store) CheckCurrentMajorSchemaVersion(ctx context.Context) error {
-	vs, err := s.GetVersion(ctx)
+	base, err := datastore.NewBaseStore(ctx, db, flags.ConnectParameters().MaxRetries, func(q dssql.Queryable) repos.Repository {
+		return &repo{
+			Queryable: q,
+			clock:     s.Clock,
```
the-glu, line 40: NB: Tests want to overwrite that, so we create the store, use a pointer there and add the basestore later..

pkg/datastoreutils/dial.go
```diff
@@ -0,0 +1,76 @@
+package datastoreutils
```
the-glu, line 1: NB: This live there instead of others packages because we need the flags to build connections parameters, and there are lot of import loops.

pkg/datastoreutils/dial.go
```diff
@@ -0,0 +1,76 @@
...
+	s, err := newStore(db)
+	if err != nil {
+		db.Pool.Close()
+		if strings.Contains(err.Error(), "connect: connection refused") || strings.Contains(err.Error(), fmt.Sprintf("database \"%s\" does not exist", dbName)) || strings.Contains(err.Error(), "database has not been bootstrapped with Schema Manager") {
+			return zero, stacktrace.PropagateWithCode(err, CodeRetryable, "Failed to create %s store", dbName)
+		}
+		return zero, stacktrace.Propagate(err, "Failed to create %s store", dbName)
+	}
+
+	c := cron.New()
```
the-glu, line 68: This is now added to every connections, instead of just `core-service`.

I don't think this has negative aspect: worst case if db migration / cleanup takes more than 1 minutes it will show states.

(We could also have a parameter)

pkg/datastoreutils/dial.go
```diff
@@ -0,0 +1,76 @@
...
+	stats["DBName"] = databaseName
+	stats["AcquireCount"] = strconv.Itoa(int(statsPtr.AcquireCount()))
+	stats["AcquiredConns"] = strconv.Itoa(int(statsPtr.AcquiredConns()))
+	stats["CanceledAcquireCount"] = strconv.Itoa(int(statsPtr.CanceledAcquireCount()))
+	stats["ConstructingConns"] = strconv.Itoa(int(statsPtr.ConstructingConns()))
+	stats["EmptyAcquireCount"] = strconv.Itoa(int(statsPtr.EmptyAcquireCount()))
+	stats["IdleConns"] = strconv.Itoa(int(statsPtr.IdleConns()))
+	stats["MaxConns"] = strconv.Itoa(int(statsPtr.MaxConns()))
+	stats["TotalConns"] = strconv.Itoa(int(statsPtr.TotalConns()))
+	if stats["TotalConns"] == "0" {
```
the-glu, line 36: I'm not 100% sure if there is something hidden, but why do we fill up all values just to use that one??

cmds/core-service/main.go
```diff
@@ -299,8 +212,6 @@ func RunHTTPServer(ctx context.Context, ctxCanceler func(), address, locality st
 	if *enableSCD {
 		scdV1Server, err = createSCDServer(ctx, logger)
 		if err != nil {
-			ridV1Server.Cron.Stop()
-			ridV2Server.Cron.Stop()
```
the-glu, line 303: 'Hidden' changes: crons that print stats are not stopped anymore if scd fails, but I didn't find a real purpose for that (since it's going to stop anyways)

pkg/datastoreutils/dial.go
```diff
@@ -0,0 +1,76 @@
...
+
+	cp := flags.ConnectParameters()
+	cp.DBName = dbName
+	cp.ApplicationName = ApplicationName
+
+	db, err := datastore.Dial(ctx, cp)
+
+	if err != nil {
+		if strings.Contains(err.Error(), "connect: connection refused") {
+			return zero, stacktrace.PropagateWithCode(err, CodeRetryable, "Failed to connect to datastore server for %s", dbName)
```
the-glu, line 55: Before message included a nice human name of the DB, that not the case any more, but it can be guess from db name.

I can be readded if needed with an extra parameter

cmds/db-manager/cleanup/evict.go
the-glu: -> ok

pkg/datastoreutils/dial.go
the-glu: -> to clenaup

pkg/datastoreutils/dial.go
the-glu: -> ok

pkg/datastoreutils/dial.go
the-glu: -> add a parameter

cmds/db-manager/cleanup/evict.go
the-glu: -> remove it

PR #1408, opened 2026-03-31
[datastore] Use a generic datastore

the-glu:
Move datastores to a generic one, with common function in a `BaseStore`.

Pre-cleanup DBName with harcoded strings, but final cleanup (with total removal on cmd side) will come in a later PR.

### DIFF
--- a/pkg/aux_/store/datastore/store.go
+++ b/pkg/aux_/store/datastore/store.go
@@ -3,17 +3,11 @@ package datastore
 import (
 	"context"
 
-	"github.com/cockroachdb/cockroach-go/v2/crdb"
-	"github.com/interuss/dss/pkg/datastore/flags"
-	dssql "github.com/interuss/dss/pkg/sql"
-
-	crdbpgx "github.com/cockroachdb/cockroach-go/v2/crdb/crdbpgxv5"
-	"github.com/coreos/go-semver/semver"
 	"github.com/interuss/dss/pkg/aux_/repos"
 	"github.com/interuss/dss/pkg/datastore"
+	"github.com/interuss/dss/pkg/datastore/flags"
 	"github.com/interuss/dss/pkg/logging"
-	"github.com/interuss/stacktrace"
-	"github.com/jackc/pgx/v5"
+	dssql "github.com/interuss/dss/pkg/sql"
 	"github.com/jonboulle/clockwork"
 	"go.uber.org/zap"
 )
@@ -24,11 +18,6 @@ const (
 	currentYugabyteMajorSchemaVersion = 1
 )
 
-var (
-	// DefaultClock is what is used as the Store's clock, returned from Dial.
-	DefaultClock = clockwork.NewRealClock()
-)
-
 type repo struct {
 	dssql.Queryable
 	clock   clockwork.Clock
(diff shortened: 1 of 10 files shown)


pkg/rid/store/datastore/store.go
```diff
@@ -24,144 +19,38 @@ const (
...
-	// TODO: consider what tx opts we want to support.
-	// TODO: we really need to remove the upper cockroach package, and have one
-	// "store" for everything
-
-	ctx = crdb.WithMaxRetries(ctx, flags.ConnectParameters().MaxRetries)
-
-	return crdbpgx.ExecuteTx(ctx, s.db.Pool, pgx.TxOptions{
-		IsoLevel: pgx.Serializable,
-	}, func(tx pgx.Tx) error {
-		// Is this recover still necessary?
```
the-glu, line 120: NB: Since it was not present anymore in SCD; I did assume it's not needed.

the-glu: Also: Similar code seems to be present in https://github.com/cockroachdb/cockroach-go/blob/master/crdb/common.go#L41 (called by https://github.com/cockroachdb/cockroach-go/blob/master/crdb/tx.go#L299 )

pkg/datastore/store.go
```diff
@@ -0,0 +1,84 @@
...
+	dssql "github.com/interuss/dss/pkg/sql"
+	"github.com/interuss/stacktrace"
+	"github.com/jackc/pgx/v5"
+	"github.com/jonboulle/clockwork"
+)
+
+type BaseStore[R any] struct {
+	DB           *Datastore
+	Clock        clockwork.Clock
+	databaseName string
```
mickmis: Duplicated with db.Pool.Config().ConnConfig.Database? Remove, or add TODO if plan is to remove later.

pkg/datastore/store.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/cockroachdb/cockroach-go/v2/crdb"
+	crdbpgx "github.com/cockroachdb/cockroach-go/v2/crdb/crdbpgxv5"
+	"github.com/coreos/go-semver/semver"
+	dssql "github.com/interuss/dss/pkg/sql"
+	"github.com/interuss/stacktrace"
+	"github.com/jackc/pgx/v5"
+	"github.com/jonboulle/clockwork"
+)
+
+type BaseStore[R any] struct {
```
mickmis: Rename to `Store`? 
The other `Store` structs are in other packages, and should disappear anyway IMO.

pkg/datastore/store.go
```diff
@@ -0,0 +1,84 @@
...
+
+func NewBaseStore[R any](ctx context.Context, db *Datastore, maxRetries int, newRepo func(dssql.Queryable) R) (BaseStore[R], error) {
+
+	dbName := db.Pool.Config().ConnConfig.Database
+
+	vs, err := db.GetSchemaVersion(ctx, dbName)
+	if err != nil {
+		return BaseStore[R]{}, stacktrace.Propagate(err, "Failed to get schema version for %s", dbName)
+	}
+	return BaseStore[R]{
```
mickmis: Call here `CheckCurrentMajorSchemaVersion` and unexpose it?

pkg/aux_/store/datastore/store.go
```diff
@@ -24,146 +18,38 @@ const (
...
 	logger  *zap.Logger
 	version *datastore.Version
 }
 
-// Store is an implementation of store.Store using Cockroach DB or yugabyte as its backend
-// store.
-//
-// TODO: Add the RID/SCD interfaces here, and collapse this store with the
-// outer pkg/cockroach
 type Store struct {
```
mickmis, line 28: Delete those `Store` structs and push the cleanup methods into the repository in a follow-up PR? (put a TODO)

pkg/aux_/store/datastore/store.go
```diff
@@ -24,146 +18,38 @@ const (
...
-
-	return crdbpgx.ExecuteTx(ctx, s.db.Pool, pgx.TxOptions{
-		IsoLevel: pgx.Serializable,
-	}, func(tx pgx.Tx) error {
-		// Is this recover still necessary?
-		defer recoverRollbackRepanic(ctx, tx)
-		return f(ctx, &repo{
-			Queryable: tx,
-			clock:     s.clock,
-			logger:    logger,
```
mickmis, line 127: FTR: this was not setting the version, and now is. But should be OK.

pkg/scd/store/datastore/store.go
```diff
@@ -3,16 +3,14 @@ package datastore
 import (
 	"context"
 
-	"github.com/cockroachdb/cockroach-go/v2/crdb"
-	crdbpgx "github.com/cockroachdb/cockroach-go/v2/crdb/crdbpgxv5"
-	"github.com/coreos/go-semver/semver"
-	"github.com/interuss/dss/pkg/datastore"
 	"github.com/interuss/dss/pkg/datastore/flags"
+	dssql "github.com/interuss/dss/pkg/sql"
```
mickmis, line 7:
```suggestion
	dsssql "github.com/interuss/dss/pkg/sql"
```

pkg/datastore/store.go
the-glu: Yes, they are gone in #1411

pkg/datastore/store.go
the-glu: We need schema versions, so that not a simple fix, will do in a follow up PR

pkg/aux_/store/datastore/store.go
the-glu: Done in #1413

PR #1409, opened 2026-03-31
[datastore] Normalize datastore initialization across the codebase with a generic dial function

the-glu:
Follow #1408 , extracted from #1403

Move to a generic `dial` function, normalizing `datastore.Dial` calls, checks of errors and (optional) cron creation.

### DIFF
--- a/cmds/core-service/main.go
+++ b/cmds/core-service/main.go
@@ -23,8 +23,7 @@ import (
 	aux "github.com/interuss/dss/pkg/aux_"
 	auxc "github.com/interuss/dss/pkg/aux_/store/datastore"
 	"github.com/interuss/dss/pkg/build"
-	"github.com/interuss/dss/pkg/datastore"
-	"github.com/interuss/dss/pkg/datastore/flags" // Force command line flag registration
+	"github.com/interuss/dss/pkg/datastoreutils"
 	"github.com/interuss/dss/pkg/logging"
 	"github.com/interuss/dss/pkg/rid/application"
 	rid_v1 "github.com/interuss/dss/pkg/rid/server/v1"
@@ -35,7 +34,6 @@ import (
 	"github.com/interuss/dss/pkg/version"
 	"github.com/interuss/dss/pkg/versioning"
 	"github.com/interuss/stacktrace"
-	"github.com/robfig/cron/v3"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 	"go.uber.org/zap"
 )
@@ -64,20 +62,6 @@ var (
 	scdGlobalLock = flag.Bool("enable_scd_global_lock", false, "Experimental: Use a global lock when working with SCD subscriptions. Reduce global throughput but improve throughput with lot of subscriptions in the same areas.")
 )
 
-const (
-	codeRetryable = stacktrace.ErrorCode(1)
-)
-
-func checkDatabase(ctx context.Context, db *datastore.Datastore, databaseName string) {
-	logger := logging.WithValuesFromContext(ctx, logging.Logger)
-	statsPtr := db.Pool.Stat()
-	if int(statsPtr.TotalConns()) == 0 {
-		logger.Warn("Failed periodic DB Ping (TotalConns=0)", zap.String("Database", databaseName))
-	} else {
-		logger.Info("Successful periodic DB Ping", zap.String("Database", databaseName))
-	}
-}
-
 func createKeyResolver() (auth.KeyResolver, error) {
 	switch {
 	case *pkFile != "":
@@ -100,24 +84,9 @@ func createKeyResolver() (auth.KeyResolver, error) {
 }
 
 func createAuxServer(ctx context.Context, locality string, publicEndpoint string, scdGlobalLock bool, logger *zap.Logger) (*aux.Server, error) {
-	connectParameters := flags.ConnectParameters()
-	connectParameters.DBName = "aux"
-	datastore, err := datastore.Dial(ctx, connectParameters)
+	auxStore, err := auxc.Dial(ctx, logger, true)
 	if err != nil {
-		if strings.Contains(err.Error(), "connect: connection refused") {
-			return nil, stacktrace.PropagateWithCode(err, codeRetryable, "Failed to connect to database for pool information store")
-		}
-		return nil, stacktrace.Propagate(err, "Failed to connect to pool information database; verify your database configuration is current with https://github.com/interuss/dss/tree/master/build#upgrading-database-schemas")
-	}
-
-	auxStore, err := auxc.NewStore(ctx, datastore, logger)
-	if err != nil {
-		// TODO: More robustly detect failure to create SCD server is due to a problem that may be temporary
-		if strings.Contains(err.Error(), "connect: connection refused") || strings.Contains(err.Error(), "database \"aux\" does not exist") {
-			datastore.Pool.Close()
-			return nil, stacktrace.PropagateWithCode(err, codeRetryable, "Failed to connect to datastore server for strategic conflict detection store")
-		}
-		return nil, stacktrace.Propagate(err, "Failed to create strategic conflict detection store")
+		return nil, err
 	}
 
 	repo, err := auxStore.Interact(ctx)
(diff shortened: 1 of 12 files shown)


cmds/core-service/main.go
```diff
@@ -286,8 +212,6 @@ func RunHTTPServer(ctx context.Context, ctxCanceler func(), address, locality st
 	if *enableSCD {
 		scdV1Server, err = createSCDServer(ctx, logger)
 		if err != nil {
-			ridV1Server.Cron.Stop()
-			ridV2Server.Cron.Stop()
```
the-glu, line 290: This will be fixed in a follow up PR. It's needed, but not implemented correctly (e.g. missing in scd)

pkg/datastoreutils/dial.go
```diff
@@ -0,0 +1,64 @@
+package datastoreutils
```
mickmis, line 1: Why not in `datastore` package directly?

pkg/aux_/store/datastore/store.go
```diff
@@ -53,3 +54,12 @@ func (s *Store) CleanUp(ctx context.Context) error {
 	_, err := s.DB.Pool.Exec(ctx, query)
 	return err
 }
+
+func Dial(ctx context.Context, logger *zap.Logger, withCheckCron bool) (*Store, error) {
```
mickmis, line 58: Inline those `Dial` functions?

pkg/datastoreutils/dial.go
```diff
@@ -0,0 +1,64 @@
...
+	logger := logging.WithValuesFromContext(ctx, logging.Logger)
+	statsPtr := db.Pool.Stat()
+	if int(statsPtr.TotalConns()) == 0 {
+		logger.Warn("Failed periodic DB Ping (TotalConns=0)", zap.String("Database", databaseName))
+	} else {
+		logger.Info("Successful periodic DB Ping", zap.String("Database", databaseName))
+	}
+}
+
+func DialStore[S any](ctx context.Context, dbName string, withCheckCron bool, newStore func(*datastore.Datastore) (S, error)) (S, error) {
```
mickmis, line 32: In the end, do we ever create stores without dialing them? If not, collapse `DialStore` and `NewStore`? (at least for what is exposed outside the package)

pkg/datastoreutils/dial.go
```diff
@@ -0,0 +1,64 @@
...
+
+	if err != nil {
+		if strings.Contains(err.Error(), "connect: connection refused") {
+			return zero, stacktrace.PropagateWithCode(err, CodeRetryable, "Failed to connect to datastore server for %s", dbName)
+		}
+		return zero, stacktrace.Propagate(err, "Failed to connect to %s database", dbName)
+	}
+	s, err := newStore(db)
+	if err != nil {
+		db.Pool.Close()
```
mickmis, line 49: Are you sure this is safe to call here? (before it was within the next if block)
-> fix in a separate PR if it is

cmds/core-service/main.go
```diff
@@ -100,24 +84,9 @@ func createKeyResolver() (auth.KeyResolver, error) {
 }
 
 func createAuxServer(ctx context.Context, locality string, publicEndpoint string, scdGlobalLock bool, logger *zap.Logger) (*aux.Server, error) {
-	connectParameters := flags.ConnectParameters()
-	connectParameters.DBName = "aux"
-	datastore, err := datastore.Dial(ctx, connectParameters)
+	auxStore, err := auxc.Dial(ctx, logger, true)
```
mickmis, line 87: The `withCronCheck` should be false here

pkg/datastoreutils/dial.go
```diff
@@ -0,0 +1,64 @@
...
+		if strings.Contains(err.Error(), "connect: connection refused") || strings.Contains(err.Error(), fmt.Sprintf("database \"%s\" does not exist", dbName)) || strings.Contains(err.Error(), "database has not been bootstrapped with Schema Manager") {
+			return zero, stacktrace.PropagateWithCode(err, CodeRetryable, "Failed to create %s store", dbName)
+		}
+		return zero, stacktrace.Propagate(err, "Failed to create %s store", dbName)
+	}
+
+	if withCheckCron {
+		c := cron.New()
+		if _, err := c.AddFunc("@every 1m", func() { checkDatabase(ctx, db, dbName) }); err != nil {
+			db.Pool.Close()
```
mickmis, line 59: This was not here before, is this safe to call?
-> fix in a separate PR if it is

cmds/core-service/main.go
mickmis: Mention this in a TODO?

the-glu: Already ready in #1410

pkg/datastoreutils/dial.go
the-glu: Tests does, for now, as they want to override the clock.

We could add it as parameters, but that would be future PR no?

cmds/core-service/main.go
the-glu: Are you sure? I would normalize scd/rid/aux there, no? Why aux wouldn't be checked like the others?

pkg/aux_/store/datastore/store.go
the-glu: I'm not sure how and why, what do you have in mind?

pkg/datastoreutils/dial.go
the-glu: Because that would create a loop: flags refrence datastore, and we need flags there :/

pkg/datastoreutils/dial.go
the-glu: Before it was only called in one return, so I normalized it. Since the db is going to be unused anymore, I do think it's ok and make sense to properly clean up in all cases.

pkg/datastoreutils/dial.go
mickmis: Looking at it, it would make a lot of sense to me to:
- rename the package `flags` into `params` (or sth similar)
- move `ConnectParameters` and dependent structs into this package
- move the content of this file into `datastore` which should not have an import loop anymore

Usually the subpackages don't depend on the parent ones. And having a package 'utils' is usually to be avoided.

pkg/datastoreutils/dial.go
mickmis: > Tests does, for now, as they want to override the clock.

If it is for tests only it should not be exported? Although that may be required since that is declared in a separate package. But maybe now with the previous suggestion of solving the import loop that may be done?

cmds/core-service/main.go
mickmis: I was talking about behavior change. 
About normalizing yes makes sense but in a separate PR then? If this behavior change causes any type of issue or interrogation, a commit message with 'generic dial function' won't help identifying the change, even more so when obfuscated by a refactor like here.

pkg/datastoreutils/dial.go
the-glu: Okay, would it be ok to do it in a follow up pr to limit changes ?

cmds/core-service/main.go
the-glu: Ok, but the whole PR is about normalizing initializations, that not the only behavior change and this one is really minor, do you really see the need of moving it to a separate PR?

pkg/datastoreutils/dial.go
the-glu: It's doable for scd/aux, but not for rid, as `application_test` is in another package.

I switched them as private when it's possible for now, but I think we probably want a parameter in a future PR.

pkg/datastoreutils/dial.go
mickmis: yep - add a TODO here if PR not already open though?

pkg/aux_/store/datastore/store.go
mickmis: Hum actually no, please disregard. I just wanted to eliminate a layer of indirection that I thought unneeded. But this is passing `aux` and `newStore` to the function so that probably makes sense as is. This might change though in later PR with the merging of  `datastoreutils`.

cmds/core-service/main.go
mickmis: > that not the only behavior change

Which ones did I miss?

> this one is really minor

I disagree, it is a pretty visible one, one more DB is going to have pings every minutes. That deserves to be mentioned in the changelog IMO, thus should be included in the git commit message. If it is too much overhead to open a separate PR, a mention of the change in the title of this PR could also work.

pkg/datastoreutils/dial.go
mickmis: But is it safe to be called? 

> Since the db is going to be unused anymore

Are you sure about that? Do we actually want to to close the DB connection pool at the first failed ping? What about transactions currently running? And this does not seem to be properly terminating the service? Looks like the routine will just die in the background before closing the connection pool.

pkg/datastoreutils/dial.go
mickmis: Discussed offline: OK

pkg/datastoreutils/dial.go
mickmis: Discussed offline: OK

pkg/datastoreutils/dial.go
the-glu: Yup, done

pkg/datastoreutils/dial.go
the-glu: After discussion, updated title + add remarks in `NEXT_RELEASE_NOTES`

cmds/core-service/main.go
the-glu: After discussions, updated PR title + added remark in changelog.

PR #1415, opened 2026-04-07
[datastore] Make CheckCurrentMajorSchemaVersion internal, move parameters to NewStore

the-glu:


pkg/datastore/store.go
```diff
@@ -62,19 +69,15 @@ func (s *Store[R]) GetVersion(ctx context.Context) (*semver.Version, error) {
 	return s.version, nil
 }
 
-func (s *Store[R]) CheckMajorSchemaVersion(ctx context.Context, crdbExpected, ybExpected int64, module string) error { // TODO: Make it internal, pass parameters to newStore
-	vs, err := s.GetVersion(ctx)
```
mickmis, line 66: Delete this method? Looks now unused.

the-glu: Yes correct, I was confused by 'Get**Schema**Version'

PR #1429, opened 2026-04-13
[aux] Fill nil on empty body on PutDSSInstancesHeartbeat

the-glu:
I noticed during some auth test that the endpoint crash instead of returning a 500 with an empty body, dues to a nil pointer being used. This PR fixes it.

### DIFF
--- a/pkg/aux_/pool_participants.go
+++ b/pkg/aux_/pool_participants.go
@@ -86,6 +86,11 @@ func (a *Server) PutDSSInstancesHeartbeat(ctx context.Context, req *restapi.PutD
 		return resp
 	}
 
+	if req.Source == nil {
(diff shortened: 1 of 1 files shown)


pkg/aux_/pool_participants.go
```diff
@@ -86,6 +86,11 @@ func (a *Server) PutDSSInstancesHeartbeat(ctx context.Context, req *restapi.PutD
 		return resp
 	}
 
+	if req.Source == nil {
```
BenjaminPelletier, line 89: The interface specifies this field as [required](https://github.com/interuss/dss/blob/f829c07902da3a01b14098decc808d1ba7f63ece/interfaces/aux_/aux_.yaml#L241), so if it isn't present, that is a bad request (400) rather than an internal server error (500)

BenjaminPelletier (approved):

PR #1430, opened 2026-04-13
 [tooling] Add authentification tests 

the-glu:
This PR follows #1429 

The PR add authentication tests. Endpoints list is based on generated `server.gen.go`: idea is to look at endpoints actually in the code and not interfaces that may be out-of-date (or added without changed the script there).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,7 @@ cleanup-test-go-units-crdb:
 build-dss:
 	build/dev/run_locally.sh build
 
-test-e2e: down-locally build-dss start-locally probe-locally evict-locally collect-local-logs down-locally
+test-e2e: down-locally build-dss start-locally probe-locally evict-locally security-locally collect-local-logs down-locally
 
 tag:
 	scripts/tag.sh $(UPSTREAM_OWNER)/dss/v$(VERSION)
@@ -176,6 +176,10 @@ probe-locally:
 .PHONY: evict-locally
 evict-locally:
 	build/dev/evict_locally.sh
+
+.PHONY: security-locally
+security-locally:
+	build/dev/security_locally.sh
 
 .PHONY: qualify-locally
 qualify-locally:
--- a/build/dev/security_locally.sh
+++ b/build/dev/security_locally.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+set -x
+
+# Find and change to repo root directory
+OS=$(uname)
+if [[ "$OS" == "Darwin" ]]; then
+	# OSX uses BSD readlink
+	BASEDIR="$(dirname "$0")"
+else
+	BASEDIR=$(readlink -e "$(dirname "$0")")
+fi
+cd "${BASEDIR}/../.." || exit 1
+
+CORE_SERVICE_CONTAINER="dss_sandbox-local-dss-core-service-1"
+OAUTH_CONTAINER="dss_sandbox-local-dss-dummy-oauth-1"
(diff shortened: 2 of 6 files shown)


test/evict/query_helper.py
```diff
@@ -25,7 +25,7 @@ def get_token(self) -> str:
         ).read()
         data = json.loads(r)
 
-        if not "access_token":
+        if "access_token" not in data:
```
the-glu, line 28: I noticed and fixed that one when copy-pasting it, probably not worth a dedicated PR ^^'

build/dev/security_locally.sh
```diff
@@ -0,0 +1,48 @@
...
+    echo "Activating yugabyte options"
+    export DB_HOSTNAME=local-dss-ybdb
+    export DB_PORT=5433
+    export DB_USERNAME=yugabyte
+else
+    echo "Staying with cockroachdb defaults options"
+fi
+
+if ! python test/security/test.py; then
+    echo "Evict tests did not succeed."
```
BenjaminPelletier:
```suggestion
    echo "Security tests did not succeed."
```

build/dev/security_locally.sh
```diff
@@ -0,0 +1,48 @@
...
+    exit 1
+else
+    echo "Evict tests succeeded."
```
BenjaminPelletier:
```suggestion
    echo "Security tests succeeded."
```

Makefile
```diff
@@ -177,6 +177,10 @@ probe-locally:
 evict-locally:
 	build/dev/evict_locally.sh
 
+.PHONY: security-locally
```
BenjaminPelletier, line 180: A user should never be surprised by a PR check failing on GitHub if they had already tested it locally according to documentation.  Let's add this action to a broad target like `make dss-tests` or else update documentation to list it as something a user needs to run (`make security-locally`) as one of their verification checks.

build/dev/security_locally.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# If yugabyte container is running, assume that we're running in yugabyte mode
+if [ "$( docker container inspect -f '{{.State.Status}}' "dss_sandbox-local-dss-ybdb-1" )" == "running" ]; then
+    echo "Activating yugabyte options"
+    export DB_HOSTNAME=local-dss-ybdb
+    export DB_PORT=5433
+    export DB_USERNAME=yugabyte
+else
+    echo "Staying with cockroachdb defaults options"
+fi
```
BenjaminPelletier: Is any of this necessary?

pkg/aux_/pool_participants.go
```diff
@@ -86,6 +86,11 @@ func (a *Server) PutDSSInstancesHeartbeat(ctx context.Context, req *restapi.PutD
 		return resp
 	}
 
+	if req.Source == nil {
+		resp.Response500 = &api.InternalServerErrorBody{ErrorMessage: *dsserr.Handle(ctx, stacktrace.Propagate(err, "Source not set"))}
```
BenjaminPelletier: This is a field in the request; if it needs to be set and isn't, then the problem is a bad request not an internal server error.

build/dev/security_locally.sh
the-glu: I copy pasted too much :/

pkg/aux_/pool_participants.go
the-glu: Yes, since it's a stacked PR I close that one, it will be adressed in #1429

BenjaminPelletier (approved):

PR #1436, opened 2026-04-15
[rid/datastore/ListExpired(Subscriptions|ISAs)] Properly escape writer; Fix removal of expired entities with NULL writer (empty locality)

the-glu:
I noticed that `List*` directly inject the `writer` parameter into the SQL query.

That not great at it could lead to SQL injections (and won't works with special values). Right now it's only used by `evict` and is coming from a flag, not from http queries, so it's not a security issue. However, it seems safer to escape it, for best practices and avoid potential issue when used in the future in other contexts.

This also fix the case with "", the `= NULL` condition won't match NULL values.

### DIFF
--- a/NEXT_RELEASE_NOTES.md
+++ b/NEXT_RELEASE_NOTES.md
@@ -48,6 +48,8 @@ The release notes should contain at least the following sections:
 
 ## Important information
 
+* Fixed a bug where the `evict` command ignored entries without a locality. If your DSS instance does not have a locality set, the next `evict` run may be slow while it processes the backlog of old entries.
+
 ## Minimal database schema version
 
 | Schema  | CockroachDB | Yugabyte |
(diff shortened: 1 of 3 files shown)


mickmis (approved):
Code change LGTM. I updated the title to reflect more accuratly the change. Could you just add a note in the next release notes about the fix on the null writer? This could have a visible impact on users, as it may clean up expired entities that were not cleaned up previously. That's the case if the DSS instance was not specifying any 'locality'.

the-glu:
> Code change LGTM. I updated the title to reflect more accuratly the change. Could you just add a note in the next release notes about the fix on the null writer? This could have a visible impact on users, as it may clean up expired entities that were not cleaned up previously. That's the case if the DSS instance was not specifying any 'locality'.

Done, I let you check the wording ;)

Issue #1418, opened 2026-04-08
[datastore] Improve store/datastore implementation


the-glu:
The current datastore/store implementation is quite complex with some 'loops' and database specific elements in generic classes.

Here is the current schema of interfaces/packages:

<img width="1982" height="1602" alt="Image" src="https://github.com/user-attachments/assets/daacfd81-7410-4107-bb94-01e5f703272f" />

Other packages work with instances of the only type in the schema.

I propose change it it like this:

<img width="2104" height="1582" alt="Image" src="https://github.com/user-attachments/assets/aef5588b-c994-45ab-be64-a9b39ccfc293" />

- The common store.Store takes care of initialization in a generic way, without knowing about specific databases.
- No intermediate stores in specific packages.
- Specific database implementations moved to dedicated structs (with explicit naming), including initialization, and are embedded as needed in the package's repository implementations. Common struct used for function needed during initialization by store.Store.
- Should a new type of storage be used, only the implementation of the generic store + the implementation of the repository would be added + a way to initialize the correct "sub-store." This is shown with low opacity on the schema.
- Other packages still use the only type in the schema.

the-glu:
Following discussions at yesterday weekly, I rechecked concept maps and they seem correct. As discussed I would suggest to "merge" datastore.Store with datastore.Datastore & raft.Store with raft.Group, but that would be the only change on the concepts side.

PR #1437, opened 2026-04-15
[docs] Improve data storage documentation

BenjaminPelletier:
This PR attempts to better-document storage architecture and elements in the codebase as they currently exist, per #1418.  I expect changes to some of this content soon after this PR per #1418, but hopefully the definitions in this PR will serve as a clearer baseline from which to better understand those additional changes.

I request that reviewers attempt to identify any descriptions which may be inaccurate or unclear so we can make sure there is a universally-shared baseline understanding of the current approach.


the-glu (approved):
LGTM

mickmis (approved):
LGTM

PR #1445, opened 2026-04-22
[store] Revork init structure; use Store interfaces instead of datastore.Store interfaces

### LINKED ISSUE #1418: [datastore] Improve store/datastore implementation

the-glu:

#1448 : Add 'store_type' flag

---

This PR is the 'core' of the proposal.

It reworks init functions to be cleaner, with a config object, helping specific stores (rid/*) tell the generic init function what to do.

`Dial` functions and `datastore.Store` structs in specific stores have been removed. Initialization is now done in `store.Store` functions and uses only `(rid/*).Store` interfaces. Usages have been updated (but are just a type change).

### DIFF
--- a/pkg/datastore/store.go
+++ b/pkg/datastore/store.go
@@ -160,7 +160,10 @@ func Init[R any](ctx context.Context, cfg Config[R], withCheckCron bool) (*Store
 			return nil, stacktrace.Propagate(err, "Failed to schedule db check for %s", cfg.DBName)
 		}
 		c.Start()
-		go func() { <-ctx.Done(); c.Stop() }()
+		go func() {
+		   <-ctx.Done()
+		   c.Stop()
+		}()
 	}
 
 	return db, nil
--- a/pkg/datastore/store.go
+++ b/pkg/datastore/store.go
@@ -38,38 +38,13 @@ type Store[R any] struct {
 	maxRetries    int
 }
 
-func NewStore[R any](ctx context.Context, db *Store[R], maxRetries int, crdbExpected int64, ybExpected int64, newRepo func(dsssql.Queryable) R) (Store[R], error) {
-
-	dbName := db.Pool.Config().ConnConfig.Database
-
-	vs, err := db.GetSchemaVersion(ctx, dbName)
-	if err != nil {
-		return Store[R]{}, stacktrace.Propagate(err, "Failed to get schema version for %s", dbName)
-	}
-
-	if err := checkMajorSchemaVersion(ctx, db, vs, crdbExpected, ybExpected); err != nil {
-		return Store[R]{}, err
-	}
-
-	return Store[R]{
-		Pool:          db.Pool,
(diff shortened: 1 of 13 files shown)


pkg/datastore/store.go
```diff
@@ -189,6 +118,70 @@ func Dial[R any](ctx context.Context, connParams params.ConnectParameters) (*Sto
...
+	if err == nil {
+		err = checkMajorSchemaVersion(ctx, db, vs, cfg.CrdbMajorSchemaVersion, cfg.YbMajorSchemaVersion)
+	}
+	if err != nil {
+		db.Pool.Close()
+		if strings.Contains(err.Error(), "connect: connection refused") || strings.Contains(err.Error(), fmt.Sprintf("database \"%s\" does not exist", cfg.DBName)) || strings.Contains(err.Error(), "database has not been bootstrapped with Schema Manager") {
+			return nil, stacktrace.PropagateWithCode(err, CodeRetryable, "Failed to create %s store", cfg.DBName)
+		}
+		return nil, stacktrace.Propagate(err, "Failed to create %s store", cfg.DBName)
+	}
```
mickmis, line 147: Not needed to call checkMajorSchemaVersion:
```suggestion
	vs, err := db.GetSchemaVersion(ctx, cfg.DBName)
	if err != nil {
		db.Pool.Close()
		if strings.Contains(err.Error(), "connect: connection refused") || strings.Contains(err.Error(), fmt.Sprintf("database \"%s\" does not exist", cfg.DBName)) || strings.Contains(err.Error(), "database has not been bootstrapped with Schema Manager") {
			return nil, stacktrace.PropagateWithCode(err, CodeRetryable, "Failed to create %s store", cfg.DBName)
		}
		return nil, stacktrace.Propagate(err, "Failed to create %s store", cfg.DBName)
	}
	
	if err := err = checkMajorSchemaVersion(ctx, db, vs, cfg.CrdbMajorSchemaVersion, cfg.YbMajorSchemaVersion); err != nil {
		db.Pool.Close()
		return something
	}
```

pkg/datastore/store.go
```diff
@@ -189,6 +118,70 @@ func Dial[R any](ctx context.Context, connParams params.ConnectParameters) (*Sto
...
+
+	if withCheckCron {
+		c := cron.New()
+		if _, err := c.AddFunc("@every 1m", func() { checkDatabase(ctx, db, cfg.DBName) }); err != nil {
+			db.Pool.Close()
+			return nil, stacktrace.Propagate(err, "Failed to schedule db check for %s", cfg.DBName)
+		}
+		c.Start()
+		go func() { <-ctx.Done(); c.Stop() }()
```
mickmis:
```suggestion
		go func() {
		   <-ctx.Done()
		   c.Stop()
		}()
```

mickmis (approved):
LGTM modulo minor comments

PR #1447, opened 2026-04-22
[store] Move CodeRetryable to generic store

### LINKED ISSUE #1418: [datastore] Improve store/datastore implementation

the-glu:

#1448 : Add 'store_type' flag

---

Very small PR that moves `CodeRetryable` out into the generic part of the store system, as it can be used in every store type case.


mickmis (approved):
reviewed ab128c4e6eb0db1279156f752eee801dc767f1f6
LGTM

PR #1451, opened 2026-04-29
[Raft] Add Raft store stubs

MariemBaccari:
This PR adds Raft store SCD stubs to prepare for the incoming implementation. 
RID and AUX will be added in a follow-up PR.

Implements #1700

### DIFF
--- a/pkg/raftstore/store.go
+++ b/pkg/raftstore/store.go
@@ -1,0 +1,30 @@
+package raftstore
+
+import (
+	"context"
+)
+
+type Store[R any] struct {
+	newRepo func() R
+}
+
+func Init[R any](newRepo func() R) *Store[R] {
+	return &Store[R]{newRepo: newRepo}
+}
+
+// Transact proposes the entry to Raft and blocks until it is committed and applied.
+func (s *Store[R]) Transact(ctx context.Context, f func(context.Context, R) error) error {
+	// TODO: implement
+	return nil
+}
+
+// Interact returns a repository that can be used to query the store without proposing a Raft entry.
+func (s *Store[R]) Interact(_ context.Context) (R, error) {
+	return s.newRepo(), nil
+}
+
+// Close shuts down the consensus instance.
+func (s *Store[R]) Close() error {
+	// TODO: implement
+	return nil
+}
--- a/pkg/scd/store/raftstore/store.go
+++ b/pkg/scd/store/raftstore/store.go
@@ -1,0 +1,13 @@
+package raftstore
+
+import (
+	"github.com/interuss/dss/pkg/raftstore"
+	"github.com/interuss/dss/pkg/scd/repos"
+)
+
+// repo is a full implementation of scd.repos.Repository for Raft-based storage.
+type repo struct{}
+
+func Init() (*raftstore.Store[repos.Repository], error) {
+	return raftstore.Init[repos.Repository](func() repos.Repository { return &repo{} }), nil
+}
--- a/pkg/scd/store/raftstore/subscriptions.go
+++ b/pkg/scd/store/raftstore/subscriptions.go
@@ -5,41 +5,36 @@ import (
 	"time"
 
 	"github.com/golang/geo/s2"
+	dsserr "github.com/interuss/dss/pkg/errors"
 	dssmodels "github.com/interuss/dss/pkg/models"
 	scdmodels "github.com/interuss/dss/pkg/scd/models"
+	"github.com/interuss/stacktrace"
 )
 
 func (r *repo) SearchSubscriptions(_ context.Context, v4d *dssmodels.Volume4D) ([]*scdmodels.Subscription, error) {
-	// TODO: implement
-	return nil, nil
+	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "SearchSubscriptions not implemented for raftstore")
 }
 
 func (r *repo) GetSubscription(_ context.Context, id dssmodels.ID) (*scdmodels.Subscription, error) {
-	// TODO: implement
-	return nil, nil
+	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "GetSubscription not implemented for raftstore")
 }
 
 func (r *repo) UpsertSubscription(_ context.Context, sub *scdmodels.Subscription) (*scdmodels.Subscription, error) {
-	// TODO: implement
-	return nil, nil
+	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "UpsertSubscription not implemented for raftstore")
 }
 
 func (r *repo) DeleteSubscription(_ context.Context, id dssmodels.ID) error {
-	// TODO: implement
-	return nil
+	return stacktrace.NewErrorWithCode(dsserr.NotImplemented, "DeleteSubscription not implemented for raftstore")
 }
 
 func (r *repo) IncrementNotificationIndices(_ context.Context, subscriptionIds []dssmodels.ID) ([]int, error) {
-	// TODO: implement
-	return nil, nil
+	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "IncrementNotificationIndices not implemented for raftstore")
 }
 
 func (r *repo) LockSubscriptionsOnCells(_ context.Context, cells s2.CellUnion, subscriptionIds []dssmodels.ID, startTime *time.Time, endTime *time.Time) error {
-	// TODO: implement
-	return nil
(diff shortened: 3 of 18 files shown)


cmds/core-service/main.go
```diff
@@ -167,16 +168,16 @@ func RunHTTPServer(ctx context.Context, ctxCanceler func(), address, locality st
...
-	if err != nil {
-		return stacktrace.Propagate(err, "Failed to create aux server")
-	}
-
-	// Initialize remote ID
-	ridV1Server, ridV2Server, err = createRIDServers(ctx, locality, logger)
-	if err != nil {
-		return stacktrace.Propagate(err, "Failed to create remote ID server")
+	// Initialize aux and remote ID if implemented by the store
+	if params.GetStoreParameters().StoreType != params.RaftStoreType {
```
the-glu: I would not have explicit references to store type here: the code in that file should not initialize aux/rid/scd server iff it didn't get a store, but it's the responsibility of the store interface behind to use the params and to return (or not) a store based on them.

pkg/store/params/params.go
```diff
@@ -4,6 +4,8 @@ import (
 	"flag"
 )
 
+const RaftStoreType = "raft"
```
the-glu: For consistency, SQL should also be a constant no?

pkg/store/params/params.go
the-glu: Description need an update (and use constants)

pkg/scd/store/raftstore/store.go
```diff
@@ -0,0 +1,35 @@
...
+func Init(logger *zap.Logger) (dssstore.Store[repos.Repository], error) {
+	return &Store{logger: logger}, nil
+}
+
+// Store implements store.Store[repos.Repository] for Raft-based in-memory storage.
+type Store struct {
+	logger *zap.Logger
+}
+
+func (s *Store) Close() error {
```
the-glu: Thoses one should probably go in a shared, generic raftstore package no? So when rid/aux are implemented, shared function are already out of the specifc scd one.

the-glu (commented):
Thanks for the contribution, a few suggestions based on the store architecture.

cmds/core-service/main.go
MariemBaccari: I addressed this by using a stacktrace error code to detect unsupported store types instead. Does that seem reasonable ?

pkg/raftstore/store.go
```diff
@@ -0,0 +1,26 @@
...
+func Init[R any]() *Store[R] {
+	return &Store[R]{}
+}
+
+// Transact proposes the entry to Raft and blocks until it is committed and applied.
+func (s *Store[R]) Transact(ctx context.Context, f func(context.Context, R) error) error {
+	panic("not implemented")
+}
+
+// Interact returns a read-only repository that can be used to query the store without proposing a Raft entry.
```
the-glu: Just for info:  there are a few cases when interact() is used in RW mode, but that in RID&aux

cmds/core-service/main.go
```diff
@@ -199,17 +203,20 @@ func RunHTTPServer(ctx context.Context, ctxCanceler func(), address, locality st
...
+
+	if ridV1Server != nil {
+		ridV1Router := apiridv1.MakeAPIRouter(ridV1Server, authorizer)
+		ridV2Router := apiridv2.MakeAPIRouter(ridV2Server, authorizer)
+		multiRouter.Routers = append(multiRouter.Routers, &ridV1Router, &ridV2Router)
+	}
+	if auxV1Server != nil {
+		auxV1Router := apiauxv1.MakeAPIRouter(auxV1Server, authorizer)
+		multiRouter.Routers = append(multiRouter.Routers, &auxV1Router)
+	}
```
the-glu: I would made the SCD's server logic consistant: CodeUnsupported check (nit: + creation there after tests)

the-glu (approved):
LGTM, Just a small fix for scd's CodeUnsupported normalization.

cmds/core-service/main.go
the-glu: Yes thanks, just added a comment about normalization.

pkg/raftstore/store.go
MariemBaccari: I fixed the comment, thanks

cmds/core-service/main.go
```diff
@@ -169,16 +169,30 @@ func RunHTTPServer(ctx context.Context, ctxCanceler func(), address, locality st
 
 	// Initialize aux
 	auxV1Server, err = createAuxServer(ctx, locality, *publicEndpoint, *scdGlobalLock, logger)
-	if err != nil {
+	if stacktrace.GetCode(err) == store.CodeUnsupported {
```
mickmis: No special handling should be required here. I think it is fine to let the error bubble up and return a 500 (or better, 501[1]) if they are invoked.
Alternatively, if we want to not register some endpoints, this should be made possible explictly through some config flags (just like the one for scd).

[1] IMO the tool `openapi-to-go-server` should be adding auto-generated 501 responses in the generated models. Then we could wire the handlers so that a `dsserr.NotImplemented` maps to a HTTP 501. But that's not within the scope of this PR.

pkg/store/store.go
```diff
@@ -20,5 +20,6 @@ type Store[R any] interface {
 }
 
 const (
-	CodeRetryable = stacktrace.ErrorCode(1)
+	CodeRetryable   = stacktrace.ErrorCode(1)
+	CodeUnsupported = stacktrace.ErrorCode(2)
```
mickmis: Rather than defining this new error, use `dsserr.NotImplemented`.

pkg/scd/store/raftstore/subscriptions.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+	"time"
+
+	"github.com/golang/geo/s2"
+	dssmodels "github.com/interuss/dss/pkg/models"
+	scdmodels "github.com/interuss/dss/pkg/scd/models"
+)
+
+func (r *repo) SearchSubscriptions(_ context.Context, v4d *dssmodels.Volume4D) ([]*scdmodels.Subscription, error) {
+	panic("SearchSubscriptions not yet implemented in raft store")
```
mickmis: Don't panic: return a code `dsserr.NotImplemented`. Then it's fine if it bubbles up.

cmds/core-service/main.go
MariemBaccari: Since rid and aux won't be implemented in the first stages of Raft store integration, removing the special handling here will prevent the core-service from starting with just scd. If we add stub rid and aux interfaces, the initialization will fail due to `Interact` and `SaveOwnMetadata` not being implemented. 

On the other hand, implementing new config flags is invasive considering that rid and aux will be implemented eventually so we will have a temporary config change. 

Could you specify which alternative you had in mind ? I'm not sure how to go about this.

mickmis: > If we add stub rid and aux interfaces, the initialization will fail due to Interact and SaveOwnMetadata not being implemented.

How about having the stubs of the `Store` functions (Interact, SaveOwnMEtadata, ...) not return any error then?

cmds/core-service/main.go
```diff
@@ -199,25 +214,22 @@ func RunHTTPServer(ctx context.Context, ctxCanceler func(), address, locality st
...
-		}
+	if ridV1Server != nil {
+		ridV1Router := apiridv1.MakeAPIRouter(ridV1Server, authorizer)
+		ridV2Router := apiridv2.MakeAPIRouter(ridV2Server, authorizer)
+		multiRouter.Routers = append(multiRouter.Routers, &ridV1Router, &ridV2Router)
+	}
+	if auxV1Server != nil {
+		auxV1Router := apiauxv1.MakeAPIRouter(auxV1Server, authorizer)
+		multiRouter.Routers = append(multiRouter.Routers, &auxV1Router)
+	}
```
mickmis: This should not be needed anymore with the suggested change (Interact, saveownmetadata stubs).

pkg/raftstore/store.go
```diff
@@ -0,0 +1,30 @@
...
+
+type Store[R any] struct{}
+
+func Init[R any]() *Store[R] {
+	return &Store[R]{}
+}
+
+// Transact proposes the entry to Raft and blocks until it is committed and applied.
+func (s *Store[R]) Transact(ctx context.Context, f func(context.Context, R) error) error {
+	return stacktrace.NewErrorWithCode(dsserr.NotImplemented, "Transact not yet implemented for Raft store")
```
mickmis: Do put a TODO with no error returned.

mickmis (approved):
LGTM modulo suggested change.

PR #1454, opened 2026-05-06
[monitoring] Add and expose opentelemetry metrics

the-glu:
This PR add metrics in already existing telemetry base, with expose via a prometheus compatible endpoint.

This don't include specifc metric, but there are already informatives ones with out-of-the-box feature of `otelhttp` and `otelpgx`.
 
 Both of existing one have been enriched: name has been added to otelhttp, and stats recording have been added to `otelpgx`.

Documentation / infra changes & correct dashboard will come in future PRs.

Contrib to #1257 

Quick-llm-generated dashboard to have a glance at existing metrics:

image_2026-05-06_14-21-55
image_2026-05-06_14-22-41


pkg/api/ridv1/server.gen.go
```diff
@@ -8,7 +8,9 @@ import (
 	"github.com/interuss/dss/pkg/api"
 	dsserr "github.com/interuss/dss/pkg/errors"
 	"github.com/interuss/stacktrace"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 	"go.opentelemetry.io/otel"
+	semconv "go.opentelemetry.io/otel/semconv/v1.27.0"
```
mickmis: Why v1.27.0? The latest seems to be v1.40.0:  https://pkg.go.dev/go.opentelemetry.io/otel/semconv

cmds/core-service/main.go
```diff
@@ -52,6 +52,7 @@ var (
 	dumpRequests        = flag.Bool("dump_requests", false, "Log full HTTP request and response (note: will dump sensitive information to logs; intended only for debugging and/or development)")
 	profServiceName     = flag.String("gcp_prof_service_name", "", "Service name for the Go profiler")
 	enableOpenTelemetry = flag.Bool("enable_opentelemetry", false, "Enable OpenTelemetry")
+	metricsAddress      = flag.String("metric_address", ":8079", "OpenTelemetry prometheus endpoint address")
```
mickmis: For clarity mention that this is the address to listen to, not an external endpoint to be reached. And also by default do not listen to all interfaces?

```suggestion
	metricsListeningAddress      = flag.String("metric_address", "localhost:8079", "OpenTelemetry prometheus endpoint address")
```

mickmis (approved):
LGTM modulo minor comments

cmds/core-service/main.go
the-glu: For the listening address, I would listen on all address, why would we limit it to localhost?

It will be ran usually in docker container and it's disabled by default

mickmis: To avoid accidentally exposing the endpoint by mistake?

cmds/core-service/otel.go
```diff
@@ -57,3 +73,29 @@ func newTracerProvider(ctx context.Context) (*trace.TracerProvider, error) {
...
+
+	return provider, nil
+
+}
+
+func serveMetrics(address string) {
+	http.Handle("/metrics", promhttp.Handler())
+	err := http.ListenAndServe(address, nil)
+	if err != nil {
+		fmt.Printf("error serving http: %v", err)
```
mickmis: Why not use the logger here?

pkg/api/ridv1/server.gen.go
the-glu: Initially I wanted to use the same version as traces, 1.37 but I did a typo. Updated everything (including traces) to 1.40.

cmds/core-service/main.go
the-glu: Since it's not enabled by default and it's don't contains dangerous information, it's not really needed no? I don't see the point of having an unusable value (because with that everyone will need to change it to make it usable).

cmds/core-service/main.go
```diff
@@ -47,11 +47,12 @@ var (
...
-	logLevel            = flag.String("log_level", logging.DefaultLevel.String(), "The log level")
-	dumpRequests        = flag.Bool("dump_requests", false, "Log full HTTP request and response (note: will dump sensitive information to logs; intended only for debugging and/or development)")
-	profServiceName     = flag.String("gcp_prof_service_name", "", "Service name for the Go profiler")
-	enableOpenTelemetry = flag.Bool("enable_opentelemetry", false, "Enable OpenTelemetry")
+	logFormat               = flag.String("log_format", logging.DefaultFormat, "The log format in {json, console}")
+	logLevel                = flag.String("log_level", logging.DefaultLevel.String(), "The log level")
+	dumpRequests            = flag.Bool("dump_requests", false, "Log full HTTP request and response (note: will dump sensitive information to logs; intended only for debugging and/or development)")
+	profServiceName         = flag.String("gcp_prof_service_name", "", "Service name for the Go profiler")
+	enableOpenTelemetry     = flag.Bool("enable_opentelemetry", false, "Enable OpenTelemetry")
+	metricsListeningAddress = flag.String("metric_address", ":8079", "OpenTelemetry prometheus endpoint address")
```
mickmis:
```suggestion
	metricsListeningAddress = flag.String("metrics_listening_address", ":8079", "OpenTelemetry prometheus endpoint address")
```

the-glu: Other listening address ('addr') dont have the 'listening' part, is it really ok not to be consistent there?

cmds/core-service/main.go
mickmis: We definitely do not want this to be exposed publicly accidentally. I find that right now as a user it is too easy to do that. As an alternative maybe:
- put a warning in the help command of `enable_opentelemetry` that as a consequence endpoints will be exposed
- add logs when exposing the endpoints that the endpoints are going to be exposed on the address

cmds/core-service/main.go
mickmis: If the aim is consistency, then use `metrics_addr` and do mention it is the binding port to use in the description?

the-glu: Yes done, I reworked base description as well

cmds/core-service/main.go
the-glu: Should be ok

PR #1455, opened 2026-05-06
[tooling]: Generate all openapi-to-go files; make setAuthError generated only when needed

the-glu:
This PR follow #1454 because generated files changed.

I noticed when working on previous PR that some files generated by `openapi-to-go` have not been updated correctly (`make apis` vs `make dss_api`).

This:

* Make `openapi-to-go` working with dummy-oauth
  * There are no authorization and requiered structs for `setAuthError`)
  * `setAuthError` is only created when needed
* Update example's go.mod
* Generate everything

Follow up PR will enforce up-to-date generated files in CI.

PR #1456, opened 2026-05-06
 [tooling]: Enforce in CI generation of gen.go files 

the-glu:
This PR follows #1455 

It does enforce that *.gen.go files have been generated and don't change.

Minor side fix on docker run command, that set `-it` for no reason and that make it failing in ci. Changed to `--rm` because locally it's building up useless containers quickly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,14 @@ jobs:
       run: make go-lint
     - name: Terraform lint
       run: make terraform-lint
+    - name: Ensure files don't change when generating apis
+      run: |
+        git diff --exit-code -- '*.gen.go'  # Should not have been changed by previous steps
+        make apis
+        git diff --exit-code -- '*.gen.go'  # Should not have been changed my make apis
+    - name: Ensure files don't change when running hygiene
+      run: |
+        git diff --exit-code
 
   dss-tests:
     name: DSS tests
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ openapi-to-go-server:
 	docker image build -t interuss/openapi-to-go-server ./interfaces/openapi-to-go-server
 
 dss_apis: openapi-to-go-server
-	docker container run -u "$(USER_GROUP)" -it \
+	docker container run -u "$(USER_GROUP)" --rm \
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -156,3 +156,29 @@ jobs:
         submodules: true
     - name: Check migrations
       run: ./test/migrations/up_and_down.sh
+
+  generated-files:
```
mickmis: Sounds like this belongs to repository hygiene. Also put the script in the makefile, and delete the hashfiles?

mickmis (approved):
LGTM modulo comment

.github/workflows/ci.yml
the-glu: I wanted to avoid having it in the Makefile, because it's going to generate the files, and therefor change the state of the repo/files with a check. Is it really ok to do so?

(And locally / in the Makefile, running the check will fix it, so that not really a check ^^')

mickmis: Well, if there is a change, we are a in a dirty state and should fail. BTW we could make that explicit by not using hashes but checking if there is any local change (e.g. `git diff --exit-code`).
So what about doing:
```
make apis
git diff --exit-code
```

mickmis: And also move it to repo hygiene (i.e. a step in `hygiene-tests`)

the-glu: I did it but

* It will check more than generated files
* It may be influenced by previous steps generating some change for some reasons

so I don't think that the "correctest" approach.

mickmis: True that.
But actually as a whole I'd expect the previous steps to not pollute the local repo. Checking that nothing has changed at the end could be a feature IMO :) Should be a bit more explicit though. 

What about:
- Out of explicitness, in this step only check for generated files: `git diff  --exit-code  -- '*.gen.go'`
- Add a separate step with just `git diff  --exit-code` that checks that no other previous step has changed anything previously

the-glu: Seems better yes, I changed it slightly:

1. Check that generated files didn't change (dues to linting)
2. Make apis
3. Check that generated files didn't change (if that fail here, it's api)
4. Check that files didn't change in another steps

What do you think?

mickmis: Yep great, thanks

PR #1458, opened 2026-05-06
Upgrade to go 1.26.2

the-glu:
Upgrade to latest go version


mickmis (approved):
LGTM modulo CI.
I don't know if that is the culprit, but upgrade of golint-cli might be necessary.

the-glu:
> LGTM modulo CI. I don't know if that is the culprit, but upgrade of golint-cli might be necessary.

Yes, fixed with an upgrade. Initially, GitHub Actions were broken, so I thought the error was due to that.

PR #1460, opened 2026-05-08
[Raft] Add etcd raft dependencies

MariemBaccari:
Adds Raft related dependencies (library, wal, snap and rafthttp) under `pkg/raftstore/consensus`.

One of the imports requires the version 0.5.0 of `github.com/jonboulle/clockwork`, leading to an upgrade. This has caused a few unit tests to break, more specifically due to [this change](https://github.com/jonboulle/clockwork/commit/1dd6ef9296f1eea351758381aa7acaaff6d24c13). 

Implements #1699


mickmis (approved):
LGTM

PR #1464, opened 2026-05-13
[Raft] Add Raft transport

MariemBaccari:
Adds the Raft transport implementation based on [rafthttp](https://pkg.go.dev/go.etcd.io/etcd/server/v3/etcdserver/api/rafthttp). 
It also fixes some import issues. 

At this point, it is possible to start a cluster of connected nodes using:
```
go run ./cmds/core-service \                         
  --store_type=raft \
  --raft_node_id=1 --raft_peers="1=http://localhost:9021,2=http://localhost:9022,3=http://localhost:9023" \
  --accepted_jwt_audiences=dss --public_key_files=build/test-certs/auth2.pem
```

`rafthttp` is used by etcd's own implementation and is optimized for Raft.
The alternative to this is to implement a grpc based transport from scratch. With the prototype naive implementation, the overall performance was not affected by the type of transport used so we prefer the more practical choice here. This decision can be revisited later if needed since etcd/raft is completely transport agnostic. 

Implements #1699

### DIFF
--- a/pkg/raftstore/params/params.go
+++ b/pkg/raftstore/params/params.go
@@ -1,0 +1,66 @@
+package params
+
+import (
+	"flag"
+	"net/url"
+	"strconv"
+	"strings"
+
+	"github.com/interuss/stacktrace"
+)
+
+type (
+	// ConnectParameters bundles up parameters used for connecting nodes in a raftstore cluster.
+	ConnectParameters struct {
+		ID    uint64
+		Peers string
+	}
+)
+
+// PeerMap parses the Peers string into a map of node ID to peer URL.
+func (c ConnectParameters) PeerMap() (map[uint64]*url.URL, error) {
+	peers := make(map[uint64]*url.URL)
+
+	for entry := range strings.SplitSeq(c.Peers, ",") {
+		parts := strings.SplitN(entry, "=", 2)
+		if len(parts) != 2 {
+			return nil, stacktrace.NewError("invalid peer entry %s: must be in format nodeID=peerURL", entry)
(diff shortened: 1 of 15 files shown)


pkg/raftstore/params/params.go
```diff
@@ -0,0 +1,60 @@
...
+type (
+	// ConnectParameters bundles up parameters used for connecting nodes in a raftstore cluster.
+	ConnectParameters struct {
+		ID    uint64
+		Peers string
+	}
+)
+
+// PeerMap parses the Peers string into a map of node ID to peer URL.
+func (c ConnectParameters) PeerMap() (map[uint64]string, error) {
```
the-glu: nil: parse URLs there instead of using strings ?

pkg/raftstore/store.go
```diff
@@ -2,14 +2,39 @@ package raftstore
...
-	newRepo func() R
+	newRepo   func() R
+	consensus *consensus.Consensus
 }
 
-func Init[R any](newRepo func() R) *Store[R] {
-	return &Store[R]{newRepo: newRepo}
+func Init[R any](logger *zap.Logger, newRepo func() R) (*Store[R], error) {
+	// scd, rid and aux will share the same consensus instance, so we initialize it once.
+	if sharedConsensus == nil {
```
the-glu: nit: use sync.once

the-glu (commented):
Looks good, a few nits :)

the-glu (approved):

barroco (approved):

PR #1469, opened 2026-05-19
[doc] Move cleanup part from README to documentation (partially merged in #1491)

the-glu:
Improve documentation by moving and expending on cleanup's README into documentation.

Partially LLM-assisted ^^'

This has been mostly merged in #1491 as it was chained.


docs/operations/cleanup.md
```diff
@@ -0,0 +1,179 @@
...
+
+To mitigate this:
+
+- Run the cleanup during low-intensity periods (e.g. at night).
+- Clean up iteratively, starting with a lower TTL and progressively increasing it.
+If this becomes a recurring issue, batching removals could be considered as a future improvement.
+
+## Usage
+
+Extract from `db-manager evict --help`:
```
BenjaminPelletier:
```suggestion
Extracted from `db-manager evict --help`:
```

docs/operations/cleanup.md
```diff
@@ -0,0 +1,179 @@
+# Database cleanup
+
+With time, data will accumulate overtime in the database, leading to lot of old entries and lower performances.
```
BenjaminPelletier:
```suggestion
Data will accumulate over time in the database if client USSs do not remove their expired entities, and this can lead to lower performance due to quantity of stale entities.  For this reason, InterUSS recommends periodically cleaning up no-longer-relevant entities if USS clients do not always clean up after themselves.
```

docs/operations/cleanup.md
```diff
@@ -0,0 +1,179 @@
...
+By default, the tool only lists expired entities. Deletion is opt-in via the `--delete` flag.
+
+!!! warning
+    Using this tool incorrectly may result in **loss of data**. Before deleting anything:
+
+    - Always review and validate the list of entities identified as expired (run the tool without `--delete` first).
+    - Ensure a backup of the data is available.
+    - Double-check the TTL values passed to `--rid_ttl` and `--scd_ttl`.
+
+Expiration of entities is preferably determined through their end times. When end times are not available, the last update times are used instead.
```
BenjaminPelletier: I think end times should never not be available -- although the SCD schema seems to not have end_time as NOT NULL, it probably should be NOT NULL as [no unspecified end times are allowed](https://github.com/interuss/dss/blob/35bdeebdf56997bc3d69d17495a81720408956d1/pkg/scd/operational_intents_handler.go#L480)

```suggestion
Expiration of entities is preferably determined through their end times. In the unusual event that an end time is not available, the last update time is used instead.
```

docs/operations/cleanup.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Why and when to run the cleanup
+
+Expired entities (operational intents past their end time, stale subscriptions, ISAs whose lifetime has elapsed) accumulate over time in the datastore. While the DSS keeps functioning correctly with stale rows present, letting them pile up has real costs:
```
BenjaminPelletier:
```suggestion
Expired entities (operational intents past their end time, stale subscriptions, ISAs whose lifetime has elapsed) can accumulate over time in the datastore if not cleaned up by clients. While the DSS keeps functioning correctly with some stale rows present, excessive accumulation can impact:
```

docs/operations/cleanup.md
```diff
@@ -0,0 +1,179 @@
...
+
+- Storage growth: unbounded table size on CockroachDB / Yugabytes.
```
BenjaminPelletier:
```suggestion
- Storage growth: unbounded storage usage in data stores (e.g., CockroachDB / YugabyteDB).
```

docs/operations/cleanup.md
```diff
@@ -0,0 +1,179 @@
...
+- Query performance: indexes get larger, range scans on `operational_intents` / `subscriptions` / `identification_service_areas` degrade.
+
+There is no single correct interval or TTL. Reasonable values depend on context and must be defined per DSS pool, taking into account: traffic volume and entity churn, regulatory or contractual data-retention requirements applicable to your jurisdiction, the storage capacity of the datastore cluster, and how long you may legitimately need to query historical entities for analysis or audit.
```
BenjaminPelletier:
```suggestion
There is no single correct interval or TTL. Reasonable values depend on context and must be defined per DSS pool, taking into account: traffic volume and entity churn, regulatory or contractual data-retention requirements applicable to your jurisdiction, the storage capacity of the datastore cluster, and how long clients may legitimately need to query historical entities.
```

docs/operations/cleanup.md
```diff
@@ -0,0 +1,179 @@
...
+
+The defaults shipped with the deployment tooling (`30m` TTL on RID running every 30 min, `2688h` ≈ 56-day TTL on SCD running nightly when enabled) are starting points, not recommendations. Validate them against the criteria above and adjust then in production. When in doubt, start with conservative (longer) TTLs and tighten them as you gain confidence.
```
BenjaminPelletier:
```suggestion
The defaults shipped with the deployment tooling (`30m` TTL on RID running every 30 min, `2688h` ≈ 56-day TTL on SCD running nightly when enabled) are starting points, not recommendations. Validate production values against the criteria above.
```

docs/operations/cleanup.md
```diff
@@ -0,0 +1,179 @@
...
+
+Notes:
+
+- By default, expired entities are only listed - `--delete` is required to actually remove them.
+- `--rid_ttl` and `--scd_ttl` accept durations formatted as [Go `time.Duration` strings](https://pkg.go.dev/time#ParseDuration), e.g. `24h`.
+- The datastore connection flags match those of the `core-service` command.
+
+## Regular cleanup
+
+Beyond running `db-manager evict` manually, the DSS deployment tooling can schedule the cleanup as a recurring Kubernetes `CronJob`. Three deployment paths expose the same set of evict knobs, but will always run with the  `--delete` flat set.
```
BenjaminPelletier:
```suggestion
Beyond running `db-manager evict` manually, the DSS deployment tooling can schedule the cleanup as a recurring Kubernetes `CronJob`. Three deployment paths expose the same set of evict knobs, but will always run with the  `--delete` flag set.
```

docs/operations/cleanup.md
```diff
@@ -0,0 +1,179 @@
...
+
+Shared default: RID cleanup is enabled by default (`*/30 * * * *`, `ttl = 30m`); SCD cleanup is disabled by default (suggested schedule `0 2 * * *`, `ttl = 2688h` - i.e. 2 x 56 days).
```
BenjaminPelletier, line 98: @mickmis could you double-check that my suggestion is accurate?  I think the reason the defaults are the way they are is that each DSS instance is supposed to be responsible for cleaning up its own RID stuff (and deletion is limited to entities created by the DSS instance doing the deletion, as determined by locality), but SCD cleanup is a global activity and therefore DSS instance operators would want to at least coordinate to avoid cleaning up more than necessary (or perhaps designated a single USS as responsible for cleanup).

```suggestion
Shared default: RID cleanup is locality-limited and therefore is enabled by default (`*/30 * * * *`, `ttl = 30m`); SCD cleanup is not locality-limited and therefore disabled by default (suggested schedule `0 2 * * *`, `ttl = 2688h` - i.e. 2 x 56 days).
```

docs/operations/cleanup.md
```diff
@@ -0,0 +1,179 @@
...
+| `evict_scd_subscriptions`       | `true`           |
+| `evict_enable_rid_cron`         | `true`           |
+| `evict_rid_schedule`            | `"*/30 * * * *"` |
+| `evict_rid_ttl`                 | `"30m"`          |
+| `evict_rid_isas`                | `true`           |
+| `evict_rid_subscriptions`       | `true`           |
+
+## Examples
+
+The examples below assume a DSS running locally via the `run_locally.sh` script].
```
BenjaminPelletier:
```suggestion
The examples below assume a DSS running locally via the `run_locally.sh` script.
```

BenjaminPelletier (commented):
Thanks, I think this is very timely as I expect there will be a lot of demand for this documentation in the near future.  I have a few substantive tweaks and a number of typo-level suggestions.

docs/operations/cleanup.md
the-glu: Yes, that coming from old documentation. Should we track it as an issue to have time not nullables?

BenjaminPelletier: Maybe as a P3.  I think the important thing here is noting that it would be unusual/unexpected.

the-glu: Done in https://github.com/interuss/dss/issues/1492

docs/operations/cleanup.md
mickmis: Your suggestion is correct, there is no `writer` column for SCD entities in order to store the locality, while there is for RID entities.

Do note that it is a possible scenario that some RID entities are inserted in the DB without any locality (either because of an older DSS version not supporting this, or because the locality not or mis-configured), in which case some entities might never get deleted. It could be worth a line here?

the-glu: The cleanup tool allow for an empty locality and will correctly clean up then no?

As configured, each DSS instance has it own cron, meaning old ones or one without a locality will just cleanup those without locality, so that probably fine no?

mickmis: > The cleanup tool allow for an empty locality and will correctly clean up then no?

Yes, but you need to call it explicitly without any locality set (or an empty one). If there are any leftover entities without a locality but that all DSS instances are only cleaning up entities with their locality set, the leftover ones might remain indefinitely.

the-glu: Yes ok, but than that not different than a DSS changing locality: there will be leftover if no one run cleanup manually, that probably what we want to document no?

the-glu: I updated the documentation with:

- Rational of default
- Warning about locality changes


https://github.com/interuss/dss/compare/bd8a00f77e2c4aaef1737f89aa39e95209eda3e6..6cfa425115d56ba412c2279e3bf45ef1fdecd0e6

mickmis: >  there will be leftover if no one run cleanup manually, that probably what we want to document no?

Yup

mickmis: Maybe as well a note specific to empty locality, that can happen easily during a migration from no locality feature to locality is present

the-glu: Done

docs/operations/cleanup.md
```diff
@@ -46,6 +46,12 @@ To mitigate this:
 - Clean up iteratively, starting with a lower TTL and progressively increasing it.
 If this becomes a recurring issue, batching removals could be considered as a future improvement.
 
+## Changes in locally
```
mickmis:
```suggestion
## Changes in locality
```

docs/operations/cleanup.md
```diff
@@ -46,6 +46,12 @@ To mitigate this:
 - Clean up iteratively, starting with a lower TTL and progressively increasing it.
 If this becomes a recurring issue, batching removals could be considered as a future improvement.
 
+## Changes in locality
+
+There may be cases where a DSS instance changes its locality. A common scenario is that locality was not previously mandatory, though regular updates may also occur.
```
the-glu:
```suggestion
There may be cases where a DSS instance changes its locality. This commonly occurs if locality wasn't previously required, though routine updates can also trigger a change.
```

mickmis (approved):

Issue #1463, opened 2026-05-12
[raft][gate 0] Raftstore implementation


MariemBaccari:
## Context 
A Raft-based alternative for replication is currently being developed. This issue aims to present the proposed architecture and implementation plan. 

### Consensus layer (under `pkg/raftstore/consensus`)
This is the Raft engine which wraps the etcd Raft library and includes transport and Raft log persistence. 
`Consensus` runs a `handleReady()` goroutine which processes the `node.Ready()` updates provided by etcd Raft. On every iteration, it persists entries and snapshots to the WAL, sends the required messages (Heartbeats, Votes etc. as decided by the underlying etcd Raft) to peers, applies the committed entries to the state machine then calls `node.Advance()` to communicate to the library that it is done processing the update. The general workflow of the goroutine is as follows:
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/877a18fc-3582-44be-b755-2193cc4131ce" />
`Storage` handles persisting hard state, entries and snapshots by wrapping etcd's WAL implementation. 

### Generic Store (`pkg/raftstore/store.go)`
The generic raftstore implements the `store.Store[R]` interface and connects the consensus layer with the specific DBs layers (currently only scd). The raftstore Transact() implementation builds a proposal from the serialized op data attached to the context, registers the action as a pending transaction then calls `consensus.Propose` and blocks. When the proposal comes back as a committed entry, the pending transaction is executed. Since the other nodes don't have access to the action, they use per-operation appliers to apply the entry. This means that the current design implies duplicated logic between the API handlers and these appliers.

## Raft implementation plan 
#1451: Adds the raftstore flag and store stubs 
#1460: Adds Raft related dependencies
#1464: Adds Raft transport
#1470: Adds Raft persistence
#1473: Consensus Layer - Initializes and adds consensus configuration
#1474: Consensus Layer -  Adds Raft updates consumption
#1525: Adds the generic raftstore
#1550: Use a consensus instance per store (rid, scd, aux)
#1577: Adapts the `Transact` signature to support raft.

### Memstore
The `memstore` is the in-memory store used by the `raftstore`, it is implemented in the following PRs:
#1527: Adds the base memstore 
#1529 and #1530: Add the aux memstore
#1534 and #1539: Add the rid memstore 
#1542: Adds the scd memstore

## Future PRs: 
- #1528 Implements the aux Raftstore 
- rid Raftstore 
- scd Raftstore

PR #1470, opened 2026-05-20
[Raft] Add Raft persistence

MariemBaccari:
Adds WAL and snapshot persistency and management for the raftstore. We embed `*Raft.MemoryStorage` in the `storage` instead of setting it as a field in `Consensus` to avoid scattering the storage management.

Implements #1699

### DIFF
--- a/pkg/raftstore/consensus/consensus.go
+++ b/pkg/raftstore/consensus/consensus.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 
+	"github.com/interuss/dss/pkg/logging"
 	"github.com/interuss/stacktrace"
 	"go.etcd.io/etcd/client/pkg/v3/types"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/rafthttp"
@@ -16,11 +17,14 @@ import (
 	"go.uber.org/zap"
 )
 
-const defaultClusterID uint64 = 1
+const (
+	defaultClusterID uint64 = 1
+)
 
 type Consensus struct {
-	node        raft.Node
-	raftStorage *raft.MemoryStorage
+	logger *zap.Logger
+
+	node raft.Node
 
 	transport *rafthttp.Transport
 	server    *http.Server
@@ -29,12 +33,20 @@ type Consensus struct {
 	errorC  chan error
 }
 
-func NewConsensus(logger *zap.Logger, nodeID uint64, peers map[uint64]*url.URL) (*Consensus, error) {
-	consensus := &Consensus{
-		errorC: make(chan error, 1),
+func NewConsensus(ctx context.Context, logger *zap.Logger, nodeID uint64, peers map[uint64]*url.URL, dataDir string, snapshotCatchupEntries uint64) (*Consensus, error) {
+	storage, _, err := newStorage(ctx, logger.With(zap.String("component", "storage")), dataDir, nodeID, snapshotCatchupEntries)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "failed to initialize storage")
 	}
 
-	err := consensus.initTransport(logger.With(zap.String("component", "transport")), nodeID, defaultClusterID, peers)
+	consensus := &Consensus{
+		logger: logging.WithValuesFromContext(ctx, logger),
+
+		storage: storage,
+		errorC:  make(chan error, 1),
+	}
+
+	err = consensus.initTransport(ctx, nodeID, defaultClusterID, peers)
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "failed to initialize transport")
 	}
@@ -42,16 +54,16 @@ func NewConsensus(logger *zap.Logger, nodeID uint64, peers map[uint64]*url.URL) 
 	return consensus, nil
 }
 
-func (c *Consensus) initTransport(logger *zap.Logger, nodeID uint64, clusterID uint64, peers map[uint64]*url.URL) error {
+func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID uint64, peers map[uint64]*url.URL) error {
 	nodeIDStr := fmt.Sprintf("%d", nodeID)
 
 	transport := &rafthttp.Transport{
-		Logger:      logger,
+		Logger:      logging.WithValuesFromContext(ctx, c.logger.With(zap.String("component", "transport"))),
 		ID:          types.ID(nodeID),
 		ClusterID:   types.ID(clusterID),
 		Raft:        c,
 		ServerStats: v2stats.NewServerStats(nodeIDStr, nodeIDStr),
-		LeaderStats: v2stats.NewLeaderStats(logger, nodeIDStr),
+		LeaderStats: v2stats.NewLeaderStats(c.logger, nodeIDStr),
 		ErrorC:      c.errorC,
 	}
 
@@ -82,7 +94,7 @@ func (c *Consensus) initTransport(logger *zap.Logger, nodeID uint64, clusterID u
 	go func() {
 		err := c.server.ListenAndServe()
 		if err != nil && !errors.Is(err, http.ErrServerClosed) {
-			logger.Error("http server error", zap.Error(err))
+			c.logger.Error("http server error", zap.Error(err))
 			c.errorC <- err
 		}
 	}()
@@ -109,4 +121,9 @@ func (c *Consensus) ReportUnreachable(id uint64) {
 // ReportSnapshot implements the rafthttp.Raft interface.
 func (c *Consensus) ReportSnapshot(id uint64, status raft.SnapshotStatus) {
 	c.node.ReportSnapshot(id, status)
+}
+
+// RegisterStore allows registering a snapshot provider function for a specific store
+func (c *Consensus) RegisterStore(name string, provider snapshotProvider) {
+	c.storage.registerSnapshotProvider(name, provider)
 }
--- a/pkg/raftstore/consensus/storage.go
+++ b/pkg/raftstore/consensus/storage.go
@@ -1,12 +1,268 @@
 package consensus
 
 import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/interuss/dss/pkg/logging"
+	"github.com/interuss/stacktrace"
+	"go.etcd.io/etcd/client/pkg/v3/fileutil"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/snap"
 	"go.etcd.io/etcd/server/v3/storage/wal"
+	"go.etcd.io/etcd/server/v3/storage/wal/walpb"
+	"go.etcd.io/raft/v3"
+	"go.etcd.io/raft/v3/raftpb"
+	"go.uber.org/zap"
 )
 
-// storage persists the Raft log and snapshots.
+// snapshotProvider is a function that returns the snapshot data to be included in the Raft snapshot.
+// We use a snapshotProvider from each component (scd, rid and aux).
+type snapshotProvider func() ([]byte, error)
+
+// storage persists the Raft log and snapshots and manages the raft in-memory storage.
 type storage struct {
-	wal         *wal.WAL
-	snapshotter *snap.Snapshotter
+	logger *zap.Logger
+
+	*raft.MemoryStorage
+
+	wal *wal.WAL
+
+	snapper   *snap.Snapshotter
+	providers map[string]snapshotProvider
+
+	snapshotCatchUpEntries uint64
+}
+
+// newStorage initializes the storage by loading the latest snapshot and wal entries from the disk
+// and applies them to the Raft memory storage.
+// It returns the initialized storage, a boolean indicating whether the storage was pre-existent or an error.
+func newStorage(ctx context.Context, logger *zap.Logger, dataDir string, nodeID uint64, snapshotCatchUpEntries uint64) (*storage, bool, error) {
+	logger = logging.WithValuesFromContext(ctx, logger)
+
+	// load the latest snapshot
+	snapshotPath := path.Join(dataDir, fmt.Sprintf("snapshot_%d", nodeID))
+	if !fileutil.Exist(snapshotPath) {
+		err := os.MkdirAll(snapshotPath, 0o750)
+		if err != nil {
+			return nil, false, stacktrace.Propagate(err, "failed to create directory for snapshot storage at: %s", snapshotPath)
+		}
+	}
+
+	walPath := path.Join(dataDir, fmt.Sprintf("wal_%d", nodeID))
+
+	snapper := snap.New(logger, snapshotPath)
+
+	var (
+		w        *wal.WAL
+		err      error
+		snapshot *raftpb.Snapshot
+	)
+	ok := wal.Exist(walPath)
+	if !ok {
+		if err = os.MkdirAll(walPath, 0o750); err != nil {
+			return nil, false, stacktrace.Propagate(err, "failed to create directory for wal storage at: %s", walPath)
+		}
+		w, err = wal.Create(logger, walPath, nil)
+		if err != nil {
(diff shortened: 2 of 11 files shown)


pkg/raftstore/consensus/storage.go
```diff
@@ -1,12 +1,216 @@
...
+
+	*raft.MemoryStorage
+
+	wal *wal.WAL
+
+	snapper   *snap.Snapshotter
+	providers map[string]snapshotProvider
+}
+
+func newstorage(logger *zap.Logger, dataDir string, nodeID uint64) (*storage, error) {
```
the-glu:
```suggestion
func newStorage(logger *zap.Logger, dataDir string, nodeID uint64) (*storage, error) {
```

(nit)

the-glu (approved):
LGTM

pkg/raftstore/consensus/storage.go
```diff
@@ -1,12 +1,245 @@
...
+	providers map[string]snapshotProvider
+}
+
+func newStorage(logger *zap.Logger, dataDir string, nodeID uint64) (*storage, bool, error) {
+	// load the latest snapshot
+	snapshotPath := path.Join(dataDir, fmt.Sprintf("snapshot_%d", nodeID))
+	if !fileutil.Exist(snapshotPath) {
+		err := os.MkdirAll(snapshotPath, 0o750)
+		if err != nil {
+			return nil, false, stacktrace.Propagate(err, "failed to create directory `%s`for snapshot storage", snapshotPath)
```
barroco:
```suggestion
			return nil, false, stacktrace.Propagate(err, "failed to create directory for snapshot storage: %s", snapshotPath)
```

pkg/raftstore/consensus/storage.go
```diff
@@ -1,12 +1,245 @@
...
+
+	*raft.MemoryStorage
+
+	wal *wal.WAL
+
+	snapper   *snap.Snapshotter
+	providers map[string]snapshotProvider
+}
+
+func newStorage(logger *zap.Logger, dataDir string, nodeID uint64) (*storage, bool, error) {
```
barroco: Describe return values, especially the second item (bool)

pkg/raftstore/consensus/storage.go
```diff
@@ -1,12 +1,245 @@
...
 	"go.etcd.io/etcd/server/v3/etcdserver/api/snap"
 	"go.etcd.io/etcd/server/v3/storage/wal"
+	"go.etcd.io/etcd/server/v3/storage/wal/walpb"
+	"go.etcd.io/raft/v3"
+	"go.etcd.io/raft/v3/raftpb"
+	"go.uber.org/zap"
 )
 
-// storage persists the Raft log and snapshots.
+const snapshotCatchUpEntriesN uint64 = 10000
```
barroco: How did you pick this value ?

pkg/raftstore/consensus/storage.go
```diff
@@ -1,12 +1,245 @@
...
+		return stacktrace.Propagate(err, "failed to create snapshot from raft memory storage")
+	}
+
+	err = s.save(snap)
+	if err != nil {
+		return stacktrace.Propagate(err, "failed to save snapshot")
+	}
+
+	compactIndex := uint64(1)
+	if appliedIndex > snapshotCatchUpEntriesN {
```
barroco: Can you please confirm that appliedIndex resets itself after compaction ?

pkg/raftstore/params/params.go
```diff
@@ -58,6 +61,7 @@ var (
 func init() {
 	flag.Uint64Var(&connectParameters.ID, "raft_node_id", 0, "raft node ID for this instance (must be non-zero and unique within the cluster)")
 	flag.StringVar(&connectParameters.Peers, "raft_peers", "", `comma-separated "nodeID=peerURL" pairs for all cluster members, including the current node, e.g. "1=http://node1:9021,2=http://node2:9021,3=http://node3:9021"`)
+	flag.StringVar(&connectParameters.DataDir, "raft_data_directory", defaultDataDir, "directory for raft data (snapshot and WAL storage)")
```
barroco:
```suggestion
	flag.StringVar(&connectParameters.DataDir, "raft_datadir", defaultDataDir, "directory for raft data (snapshot and WAL storage)")
```

pkg/raftstore/consensus/storage.go
barroco: Please be consistent with error descriptions. Example:
`failed to [...]: %s` where %s is a path.

pkg/raftstore/consensus/storage.go
```diff
@@ -1,12 +1,245 @@
...
+	err = s.wal.SaveSnapshot(walpb.Snapshot{
+		Index:     snapshot.Metadata.Index,
+		Term:      snapshot.Metadata.Term,
+		ConfState: &snapshot.Metadata.ConfState,
+	})
+	if err != nil {
+		return stacktrace.Propagate(err, "failed to save snapshot to wal")
+	}
+
+	return s.wal.ReleaseLockTo(snapshot.Metadata.Index)
```
barroco, line 177: Please add a comment on where the lock was acquired

barroco (changes requested):

barroco:
Thank you @MariemBaccari, please find some suggestions inline.

pkg/raftstore/consensus/storage.go
mickmis: +1
And does the lock need to be released when an error is returned?

pkg/raftstore/consensus/consensus.go
```diff
@@ -29,12 +32,20 @@ type Consensus struct {
...
 	consensus := &Consensus{
-		errorC: make(chan error, 1),
+		logger: logger,
+
+		storage: storage,
+		errorC:  make(chan error, 1),
 	}
 
-	err := consensus.initTransport(logger.With(zap.String("component", "transport")), nodeID, defaultClusterID, peers)
+	err = consensus.initTransport(logger.With(zap.String("component", "transport")), nodeID, defaultClusterID, peers)
```
mickmis: I suggest passing around the context rather than the logger: sometimes we need the context of the runtime, and also we can reconstruct the logger using 'logger.WithValuesFromCOntext()`

pkg/raftstore/params/params.go
mickmis: Whenever we write stuff to the local filesystem that becomes a something the operations need to deal with. So somehow, somewhere, we need documentation about how to handle the data in that directory. Is it temp data that can be trashed? If trashed how long does it take to reconstruct it? If not temp, does it need to be backed up? What happens if it is lost? How large can it become? Can it be hot swapped? Is there any impact from the underlying storage (e.g. if it is too slow?) etc.

pkg/raftstore/consensus/storage.go
```diff
@@ -1,12 +1,245 @@
...
+
+	compactIndex := uint64(1)
+	if appliedIndex > snapshotCatchUpEntriesN {
+		compactIndex = appliedIndex - snapshotCatchUpEntriesN
+	}
+
+	err = s.Compact(compactIndex)
+	if err != nil && !errors.Is(err, raft.ErrCompacted) {
+		return stacktrace.Propagate(err, "failed to compact raft memory storage at index %d", compactIndex)
+	}
```
mickmis: Not sure how that's supposed to be called, but given that is an action having some external effect, might be a good idea to factor this away in separate function and have a log entry when the compaction is done (or was already done previously).

pkg/raftstore/params/params.go
mickmis: e.g. I suspect `snapshotCatchUpEntriesN` has an impact on the size of file on filesytem: have this configurable and document it

pkg/raftstore/consensus/storage.go
MariemBaccari: These are os level `flock` locks on WAL segment files that are acquired internally in the WAL implementation. `ReleaseLockTo` is used to release the locks for the entries that are already covered by the snapshot and, thus, the files are not needed anymore.
We shouldn't release the locks when an error is returned since the files are still needed if we fail to save the snapshot. 
I realized, while looking into etcd's source code, that this call is actually useless in the current state of our implementation. Etcd uses this as a way to mark files that can be cleaned up (separately). I will keep it for now and add a comment to the function and the github issue to implement the cleanup later.

pkg/raftstore/consensus/storage.go
MariemBaccari: For most config values (including this one), I went with the etcd defaults. The plan is to measure the parameters' impact on performance and fine tune the values once we have a working prototype.

pkg/raftstore/consensus/storage.go
MariemBaccari: `appliedIndex` will keep increasing actually. The compaction is there to discard the old log entries from the raft in-memory storage. We just keep a margin of `defaultSnapshotCatchUpEntries` to allow slow followers to catchup on entries without having to send them a full snapshot.

pkg/raftstore/consensus/storage.go
barroco: Sounds good, can you please group those so it is easy to adjust them. I foresee that we will want to adjust those values depending on the type of deployment.

barroco: Please add a comment to explain what this variable is used for and mention that it is the default etcd value for the moment. We want to minimize the number of hidden assumptions.

MariemBaccari: I moved it to `params.go` as a flag and specified that it's the etcd default value. I will do the same to the remaining values in the subsequent PR.

pkg/raftstore/params/params.go
MariemBaccari: I clarified the `raft_datadir` flag description, do you think that's sufficient for the moment ? I also added a cleanup task in #1463. I think the implementation of that will come with more documentation regarding the growth of the data directory.

pkg/raftstore/consensus/storage.go
```diff
@@ -1,12 +1,263 @@
...
+	// load the wal entries
+	var w *wal.WAL
+	ok := wal.Exist(walPath)
+	if !ok {
+		err := os.MkdirAll(walPath, 0o750)
+		if err != nil {
+			return nil, false, stacktrace.Propagate(err, "failed to create directory for wal storage at: %s", walPath)
+		}
+
+		w, err := wal.Create(logger, walPath, nil)
```
mickmis: `w` is redefined here. If that is intended, why declaring `var w` above?

pkg/raftstore/consensus/storage.go
mickmis: Please also include this rationale in the comment in code :)

pkg/raftstore/consensus/storage.go
```diff
@@ -1,12 +1,263 @@
...
+		wal: w,
+
+		snapper:                snapper,
+		providers:              make(map[string]snapshotProvider),
+		snapshotCatchUpEntries: snapshotCatchUpEntries,
+	}, ok, nil
+}
+
+func loadSnapshot(logger *zap.Logger, walPath string, snapshotter *snap.Snapshotter) (*raftpb.Snapshot, error) {
+	if !wal.Exist(walPath) {
```
mickmis: This check is done again right after the only call to `loadSnapshot`: looks like the control flow could be clarified a bit. Is the intent to load a snapshot only if the wal does not exist? If so, what about removing this here and calling `loadSnapshot` in an else of `if !ok`?

pkg/raftstore/consensus/storage.go
```diff
@@ -1,12 +1,263 @@
...
+	// TODO - add a separate goroutine to clean up old files
+	return s.wal.ReleaseLockTo(snapshot.Metadata.Index)
+}
+
+func (s *storage) registerSnapshotProvider(name string, provider func() ([]byte, error)) {
+	s.providers[name] = provider
+}
+
+// getSnapshot calls all registered snapshot providers and combines their data into a single snapshot.
+func (s *storage) getSnapshot() ([]byte, error) {
```
mickmis, line 185: Looks like this fits `json.Marshaler` interface: https://pkg.go.dev/encoding/json#Marshaler

pkg/raftstore/consensus/storage.go
```diff
@@ -1,12 +1,263 @@
...
+			return stacktrace.Propagate(err, "failed to save snapshot")
+		}
+	}
+
+	err := s.wal.Save(hardState, entries)
+	if err != nil {
+		return stacktrace.Propagate(err, "failed to save WAL entries")
+	}
+
+	if !raft.IsEmptySnap(snapshot) {
```
mickmis, line 256: Control flow: is it actually necessary to check for this condition twice?

mickmis (commented):
Not sure if already suggested by @barroco, but using https://pkg.go.dev/go.etcd.io/raft/v3@v3.6.0/rafttest for testing would be quite important (not necessarily in this PR, but important to do in general)

mickmis (approved):

pkg/raftstore/params/params.go
mickmis: > I clarified the `raft_datadir` flag description, do you think that's sufficient for the moment ? I also added a cleanup task in #1463. I think the implementation of that will come with more documentation regarding the growth of the data directory.

This should be part of the user documentation - so not in the code necessarily (although code comment is useful too). I'd say as long as it is tracked that it is to be included in the documentation we are OK. I can't find on #1463 though?

pkg/raftstore/consensus/storage.go
```diff
@@ -1,12 +1,263 @@
...
+	snapper   *snap.Snapshotter
+	providers map[string]snapshotProvider
+
+	snapshotCatchUpEntries uint64
+}
+
+// newStorage initializes the storage by loading the latest snapshot and wal entries from the disk
+// and applies them to the Raft memory storage.
+// It returns the initialized storage, a boolean indicating whether the storage was pre-existent or an error.
+func newStorage(ctx context.Context, logger *zap.Logger, dataDir string, nodeID uint64, snapshotCatchUpEntries uint64) (*storage, bool, error) {
```
mickmis, line 43: And an additional clarification: is this storage OK in a cluster environment where there are multiple DSS pods? Is this the intent of the `nodeID` or not at all?

pkg/raftstore/params/params.go
MariemBaccari: I had added the "Old files cleanup" task to "future changes". I just specified the user documentation part now.

pkg/raftstore/consensus/storage.go
MariemBaccari: Thanks for catching this, I did not intend to redefine it.

pkg/raftstore/consensus/storage.go
MariemBaccari: The snapshot is loaded only if the wal does already exist but yes it's better to move the load in an else, the `Exist` call is redundant. Will fix this.

pkg/raftstore/consensus/storage.go
barroco: Thanks

pkg/raftstore/consensus/storage.go
MariemBaccari: That's true but I think we should keep this method signature as it clearly indicates the specific purpose of getting a snapshot of the storage. The underlying implementation can always change and just happens to be a json marshalling for the moment.

pkg/raftstore/consensus/storage.go
MariemBaccari: Since we cannot change the ordering of saving the snapshot -> saving the entries -> applying, I think we have to check for this condition twice.

pkg/raftstore/consensus/storage.go
MariemBaccari: Yes, that is the intent of `nodeID` :) I can run multiple nodes locally safely.

barroco (approved):

PR #1473, opened 2026-05-20
[Raft] Initialization and configuration of consensus

MariemBaccari:
To keep it readable, it only contains part of the consensus logic with some stub functions to be implemented in a follow-up PR. 
It defines the default configuration for Raft with similar values as etcd (to be tuned later).
It also completes the initialization of consensus and the `handleReady()` goroutine. 

Implements #1699

### DIFF
--- a/pkg/raftstore/consensus/consensus.go
+++ b/pkg/raftstore/consensus/consensus.go
@@ -6,8 +6,10 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/interuss/dss/pkg/logging"
+	params "github.com/interuss/dss/pkg/raftstore/params"
 	"github.com/interuss/stacktrace"
 	"go.etcd.io/etcd/client/pkg/v3/types"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/rafthttp"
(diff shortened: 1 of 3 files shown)


pkg/raftstore/store.go
```diff
@@ -40,6 +40,11 @@ func Init[R any](logger *zap.Logger, newRepo func() R) (*Store[R], error) {
 		return nil, sharedConsensusErr
 	}
 
+	// TODO: implement
+	sharedConsensus.RegisterStore("provider", func() ([]byte, error) {
```
the-glu, line 44: Just to understand, there will be one store registered per store type right? (rid/scd/aux)

MariemBaccari: exactly

pkg/raftstore/consensus/config.go
```diff
@@ -0,0 +1,23 @@
+package consensus
+
+import (
+	"go.etcd.io/raft/v3"
+)
+
+const (
+	defaultElectionTick    = 20
```
barroco: Please document the rationale and units for all default values

pkg/raftstore/consensus/consensus.go
```diff
@@ -18,41 +19,92 @@ import (
 
 const (
 	defaultClusterID uint64 = 1
+	tickInterval            = 100 * time.Millisecond
```
barroco: Would it make sense to put it in the config ? I would expect those parameters to be important to optimize the performance in various scenario and having them at the same place would make easier to make them configurable in the future.

pkg/raftstore/consensus/consensus.go
```diff
@@ -16,32 +17,94 @@ import (
...
-	node        raft.Node
-	raftStorage *raft.MemoryStorage
+	logger *zap.Logger
+
+	id   uint64
+	node raft.Node
 
-	transport *rafthttp.Transport
-	server    *http.Server
+	removedPeers map[uint64]bool
```
barroco: This doesn't seem to be used anywhere else in the code at the moment, can you please remove it and introduce it when it will be implemented.

pkg/raftstore/params/params.go
```diff
@@ -28,11 +37,19 @@ type (
 		// a full snapshot from the leader. It must not be deleted while the node is running or
 		// across restarts unless the node is being permanently shut down.
 		// If the directory is lost, the node will recover by receiving a snapshot from the leader.
-		DataDir string
+		DataDir   string
+		ClusterID uint64
```
barroco, line 48: Please, clearly document the variables and their impact

barroco (commented):
Partial review

pkg/raftstore/consensus/consensus.go
```diff
@@ -17,43 +19,86 @@ import (
...
+func NewConsensus(ctx context.Context, logger *zap.Logger, peers map[uint64]*url.URL, connectParams params.ConnectParameters) (*Consensus, error) {
+	storage, old, err := newStorage(ctx, logger.With(zap.String("component", "storage")), connectParams.DataDir, connectParams.ID, connectParams.SnapshotCatchupEntries)
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "failed to initialize storage")
 	}
 
+	var node raft.Node
+	config := connectParams.RaftConfig(storage)
+	if old {
+		node = raft.RestartNode(config)
```
barroco, line 53: Add a log line to indicate to the user the type of start, including important information such as binding ip and port.

pkg/raftstore/consensus/consensus.go
```diff
@@ -17,43 +19,86 @@ import (
...
+	} else {
+		node = raft.StartNode(config, peersList(peers))
+	}
+
 	consensus := &Consensus{
 		logger: logging.WithValuesFromContext(ctx, logger),
 
+		id:   connectParams.ID,
```
barroco: If I understand correctly, this ID is the node ID. Can you please rename accordingly if it is the case ?

pkg/raftstore/consensus/consensus.go
```diff
@@ -17,43 +19,86 @@ import (
...
+	consensus.snapshotIndex = snap.Metadata.Index
+	consensus.appliedIndex = snap.Metadata.Index
+
+	go func() {
+		err := consensus.handleReady(connectParams.TickInterval)
+		if err != nil {
+			consensus.logger.Error("handleReady exited with error, shutting down consensus", zap.Error(err))
+		}
+
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
barroco: Move 5*time.Second to a constant.

barroco: Why 5 seconds ?

pkg/raftstore/consensus/consensus.go
```diff
@@ -17,43 +19,86 @@ import (
...
+		defer cancel()
+		if shutdownErr := consensus.server.Shutdown(shutdownCtx); shutdownErr != nil {
+			consensus.logger.Error("failed to shutdown http server", zap.Error(shutdownErr))
+		}
+
+		consensus.transport.Stop()
+		consensus.node.Stop()
```
barroco, line 98: Please ensure that a log line is printed to inform the user that the node has stopped.

pkg/raftstore/consensus/consensus.go
```diff
@@ -95,14 +140,107 @@ func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID
...
+			}
+
+			err := c.storage.handleReceivedState(rd.Snapshot, rd.HardState, rd.Entries)
+			if err != nil {
+				return stacktrace.Propagate(err, "failed to handle received snapshot")
+			}
+
+			if !raft.IsEmptySnap(rd.Snapshot) {
+				if rd.Snapshot.Metadata.Index <= c.appliedIndex {
+					return stacktrace.NewError("snapshot index %d is not greater than applied index %d", rd.Snapshot.Metadata.Index, c.appliedIndex)
```
barroco:
```suggestion
					return stacktrace.NewError("snapshot index %d shall be greater than current applied index %d", rd.Snapshot.Metadata.Index, c.appliedIndex)
```

pkg/raftstore/consensus/consensus.go
```diff
@@ -95,14 +140,107 @@ func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID
...
+func (c *Consensus) entriesToApply(entries []raftpb.Entry) ([]raftpb.Entry, error) {
+	if len(entries) == 0 {
+		return entries, nil
+	}
+
+	result := make([]raftpb.Entry, 0)
+
+	firstIdx := entries[0].Index
+	if firstIdx > c.appliedIndex+1 {
+		return nil, stacktrace.NewError("first index of committed entry[%d] should <= progress.appliedIndex[%d]+1", firstIdx, c.appliedIndex)
```
barroco:
```suggestion
		return nil, stacktrace.NewError("unexpected gap: first committed entry index %d > applied index %d + 1", firstIdx, c.appliedIndex)
```

pkg/raftstore/consensus/consensus.go
```diff
@@ -95,14 +140,107 @@ func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID
...
+	}
+
+	if c.appliedIndex-firstIdx+1 < uint64(len(entries)) {
```
barroco:
```suggestion
    // Skip entries that have already been applied.
	if skip := c.appliedIndex + 1 - firstIdx; skip < uint64(len(entries)) {
		result = entries[skip:]
	}
```

pkg/raftstore/consensus/consensus.go
```diff
@@ -95,14 +140,107 @@ func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID
...
+		result = entries[c.appliedIndex-firstIdx+1:]
+	}
+
+	return result, nil
+}
+
+// checkUpdateConfState checks if any of the messages to be sent contain a snapshot
```
barroco: The word `check` is misleading since the update is applied. 

Is it intended to make a full copy of the message array ? Would a update in place be possible instead of copying the values ?

barroco (commented):
@MariemBaccari thanks a lot for this PR. You will find additional comments and suggestions. Don't hesitate if you have any questions.

pkg/raftstore/consensus/consensus.go
MariemBaccari: I chose it arbitrarily, do you have tips for assigning an appropriate value here ?

pkg/raftstore/consensus/consensus.go
MariemBaccari: Thanks I did not notice the missing copy, I will fix and replace it with an in-place update 👌

pkg/raftstore/consensus/consensus.go
barroco: The intention here is to make sure processes can shutdown gracefully before killing them, so the system is left in a consistent state. Therefore, we will want responses to requests sent by the node to have time to be received and processed.

barroco (approved):

PR #1474, opened 2026-05-20
[Raft] Consensus entries publishing and proposal 

MariemBaccari:
Completes the consensus business logic implementation from #1473 by adding the entries publishing logic and the `Execute` method which will be used by the `Store` to commit entries to Raft. 

We defined a Raft proposal as:
```
type Proposal struct {
	ID          string    `json:"id"`
	NodeID      uint64    `json:"node_id"`
	Timestamp   time.Time `json:"timestamp"`
	RequestType string    `json:"request_type"`
	Value       []byte    `json:"value"`
	// ReadOnly proposals do not modify the state machine and,
	// therefore, do not need to be applied by nodes who did not initiate them.
	// TODO: This is a temporary solution. In the future, we will use ReadIndex
	// for read-only operations without needing to propose them to Raft.
}
```

- `ID` is a globally unique identifier (UUID) to keep track of the proposal when it comes back as a committed entry. 
- `NodeID` the Raft node ID of the current DSS node.
- `Timestamp` is defined by the proposing node and ensures time-dependent operations execute deterministically across nodes and replays .
- `RequestType` is the type of the request (e.g. DeleteSubscription). 
- `Value` is the proposal payload (i.e. a serialized representation of an ASTM API operation). 
- `ReadOnly` proposals are currently processed the same as ReadWrite operations. They go through Raft as a proposal to ensure serializability. In the future, we will use `ReadIndex` to maintain these guarantees while avoiding the consensus round.  

Implements #1699

### DIFF
--- a/pkg/raftstore/consensus/consensus.go
+++ b/pkg/raftstore/consensus/consensus.go
@@ -2,6 +2,7 @@ package consensus
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
@@ -34,6 +35,10 @@ type Consensus struct {
 
 	once            sync.Once
 	shutdownTimeout time.Duration
+	serverErrC      chan error    // http server errors
+	stopC           chan struct{} // when closed, signals shutdown to the raft updates consumer goroutine
+
+	tracker *proposalsTracker
 
 	confState     raftpb.ConfState
 	snapshotIndex uint64
@@ -76,6 +81,10 @@ func NewConsensus(ctx context.Context, logger *zap.Logger, connectParams params.
 		commitC: commitC,
 
 		shutdownTimeout: 2 * connectParams.ElectionInterval(),
+		serverErrC:      make(chan error, 1),
+		stopC:           make(chan struct{}),
+
+		tracker: newProposalsTracker(),
 	}
 
 	err = consensus.initTransport(ctx, connectParams.NodeID, connectParams.ClusterID, peers)
@@ -92,20 +101,20 @@ func NewConsensus(ctx context.Context, logger *zap.Logger, connectParams params.
 	consensus.snapshotIndex = snap.Metadata.Index
 	consensus.appliedIndex = snap.Metadata.Index
 
-	go func() {
-		err := consensus.handleReady(connectParams.TickInterval)
-		if err != nil {
-			consensus.logger.Error("handleReady exited with error, shutting down consensus", zap.Error(err))
-		}
+	if len(snap.Data) > 0 {
+		consensus.commitC <- EntryCommit{SnapshotData: snap.Data}
+	}
 
-		consensus.Stop(context.Background())
-	}()
+	consensus.startRaftUpdatesConsumer(connectParams.TickInterval, connectParams.SnapshotIntervalEntries)
 
 	return consensus, nil
 }
 
 func (c *Consensus) Stop(ctx context.Context) {
 	c.once.Do(func() {
+		c.logger.Info("stopping consensus")
+		close(c.stopC)
+
 		shutdownCtx, cancel := context.WithTimeout(context.Background(), c.shutdownTimeout)
 		defer cancel()
 		if shutdownErr := c.server.Shutdown(shutdownCtx); shutdownErr != nil {
@@ -119,6 +128,36 @@ func (c *Consensus) Stop(ctx context.Context) {
 		c.node.Stop()
 		c.logger.Info("raft node stopped")
 	})
+}
+
+// HandleClientRequest blocks until the proposal is committed and applied / dropped or until ctx is cancelled.
+func (c *Consensus) HandleClientRequest(ctx context.Context, requestType string, value any, readOnly bool) (any, error) {
+	proposal, err := c.newProposal(ctx, requestType, value, readOnly)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "failed to create proposal")
+	}
+
+	buf, err := json.Marshal(proposal)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "failed to marshal proposal")
+	}
+
+	applied := c.tracker.track(proposal.ID)
+
+	err = c.node.Propose(ctx, buf)
+	if err != nil {
+		c.tracker.untrack(proposal.ID, ProposalResult{Error: err})
+		return nil, stacktrace.Propagate(err, "failed to propose value to Raft")
+	}
+
+	select {
+	case res := <-applied:
+		return res.Result, res.Error
+
+	case <-ctx.Done():
+		c.tracker.untrack(proposal.ID, ProposalResult{Error: ctx.Err()})
+		return nil, ctx.Err()
+	}
 }
 
 func peersList(peers map[uint64]*url.URL) []raft.Peer {
@@ -139,7 +178,9 @@ func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID 
 		Raft:        c,
 		ServerStats: v2stats.NewServerStats(nodeIDStr, nodeIDStr),
 		LeaderStats: v2stats.NewLeaderStats(c.logger, nodeIDStr),
-		ErrorC:      make(chan error),
+		// ErrorC is buffered to make sure rafthttp can send errors even if the raft updates consumer is not listening yet.
+		// The buffer size of 1 is sufficient because the transport will only send one error before stopping.
+		ErrorC: make(chan error, 1),
 	}
 
 	err := transport.Start()
@@ -161,6 +202,8 @@ func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID 
 		return stacktrace.NewError("node ID %d not found in peers map", nodeID)
 	}
 
+	c.transport = transport
+
 	c.server = &http.Server{
 		Addr:    listeningAddr,
 		Handler: transport.Handler(),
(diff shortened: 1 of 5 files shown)


the-glu (approved):
LGTM

pkg/raftstore/consensus/consensus.go
```diff
@@ -28,7 +31,10 @@ type Consensus struct {
 	transport *rafthttp.Transport
 	server    *http.Server
 
-	storage *storage
+	storage   *storage
+	commitChs map[string]chan EntryCommit
```
barroco: commitC would be better aligned with other variable names pattern

pkg/raftstore/consensus/proposal.go
```diff
@@ -0,0 +1,65 @@
...
+	applied := make(chan ProposalResult, 1)
+	p.pending[id] = applied
+	return applied
+}
+
+func (p *proposalsTracker) untrack(id string, result ProposalResult) {
+	p.Lock()
+	defer p.Unlock()
+
+	applied := p.pending[id]
```
barroco: There is a risk here that the channel doesn't exist anymore if proposals are concurrently untracked. Writting to a nil channel will block the routine forever.
```suggestion
    applied, ok := p.pending[id]
    if !ok {
        return
    }
```

pkg/raftstore/consensus/consensus.go
```diff
@@ -213,16 +246,135 @@ func (c *Consensus) handleReady(tickInterval time.Duration) error {
...
+
+	//if readOnly proposal and we did not initiate it, skip it (noop)
+	if prop.ReadOnly && !c.tracker.isPending(prop.ID) {
+		return nil
+	}
+
+	applyDoneC := make(chan ProposalResult, 1)
+	wg.Go(func() {
+		res := <-applyDoneC
+		if c.tracker.isPending(prop.ID) {
```
barroco: There is a risk of race condition here since the lock is acquired and released to check the status, and again on the next line, inside untrack, acquired again and released. The problem is that the value could change in between, can you please make sure this operation is atomic.

barroco (commented):
Partital review. Please consider the comments

pkg/raftstore/consensus/consensus.go
MariemBaccari: Thanks for catching this, your previous fix to `untrack` actually makes the `isPending` call redundant so the operation is now atomic.

pkg/raftstore/consensus/consensus.go
```diff
@@ -274,6 +423,10 @@ func (c *Consensus) ReportSnapshot(id uint64, status raft.SnapshotStatus) {
 }
 
 // RegisterStore allows registering a snapshot provider function for a specific store
-func (c *Consensus) RegisterStore(name string, provider snapshotProvider) {
+// and returns the channel on which committed entries for that store will be sent.
+func (c *Consensus) RegisterStore(name string, provider snapshotProvider) chan EntryCommit {
 	c.storage.registerSnapshotProvider(name, provider)
+	ch := make(chan EntryCommit)
+	c.commitC[name] = ch
```
barroco: This shall be defined before the handleReady loop process a message. Looks like handleReady is initialized in the constructor NewConsensus - Can you please move to a Start method the part of the goroutine so we can ensure there is an explicit sequence between the initialization of the channel and its usage. It is especially important for a restart which would immediately receive messages.

pkg/raftstore/consensus/consensus.go
barroco: It looks like c.transport shall be defined before starting the go routine.

barroco (commented):
Partial review 2

pkg/raftstore/consensus/consensus.go
MariemBaccari: I moved the `handleReady` startup to a `start` method. I think the smoothest way to ensure the sequencing is to allow consensus to auto-start instead of initializing and starting from the store.
The fix I implemented passes the `expectedStoreCount` to consensus and have the `RegisterStore` call `go c.start()` once the count is reached.

Your comment also made me realize that the store (consumer) should provide the channel to the consensus instance (writer) to avoid situations where we write to a channel without consumers yet. I fixed that too :)

barroco: Is there a case where we could have multiple store registration before start ?

MariemBaccari: All the stores (scd, rid and aux) share one consensus instance. So consensus should only start after the last store is registered.

barroco: Thank you for the proposal, may I suggest to replace the expectedStoreCount by a `started` variable which prevent to registration of new stores and multiple starts instead ?
That would allow the following semantic:
1. NewConsensus()
2. RegisterStore() x n
3. Start()

barroco: Hey @MariemBaccari ,
As discussed, the sharedConsensus seems to complexify a lot the initialization and the interface of the store. In addition, there is no point to combine all service's stores in one raft group.
With a better understanding now, I think we want to revisit #1473 to have one consensus per store (simple version). That should simplify the initialization code. Can you please open another PR to adjust it?

barroco: addressed in #1550

pkg/raftstore/consensus/consensus.go
```diff
@@ -139,7 +178,7 @@ func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID
 		Raft:        c,
 		ServerStats: v2stats.NewServerStats(nodeIDStr, nodeIDStr),
 		LeaderStats: v2stats.NewLeaderStats(c.logger, nodeIDStr),
-		ErrorC:      make(chan error),
+		ErrorC:      make(chan error, 1),
```
barroco: Can you please verify that a buffer of 1 will not block any process forever. I think it is ok only if publication to this channel is under our control, however, it would not be if the transport library is using it.

pkg/raftstore/consensus/proposal.go
```diff
@@ -1,7 +1,88 @@
...
 
 	SnapshotData []byte
 }
+
+type Proposal struct {
+	ID          string    `json:"id"`
+	Timestamp   time.Time `json:"timestamp"`
+	RequestType string    `json:"request_type"`
+	Value       []byte    `json:"value"`
+	ReadOnly    bool      `json:"read_only"`
```
barroco: Please add a comment to describe the impact

pkg/raftstore/consensus/consensus.go
```diff
@@ -121,6 +130,36 @@ func (c *Consensus) Stop(ctx context.Context) {
 	})
 }
 
+// Execute blocks until the proposal is committed and applied / dropped or until ctx is cancelled.
+func (c *Consensus) Execute(ctx context.Context, requestType string, value any, readOnly bool) (any, error) {
```
barroco: To clarify, would it make sense to name this method : `HandleClientRequest` ?

pkg/raftstore/consensus/consensus.go
```diff
@@ -174,12 +215,11 @@ func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID
 		}
 	}()
 
-	c.transport = transport
 	return nil
 }
 
 // handleReady processes the Ready channel of the Raft node and applies committed entries to the state machine
-func (c *Consensus) handleReady(tickInterval time.Duration) error {
+func (c *Consensus) handleReady(tickInterval time.Duration, snapshotInterval uint64) error {
```
barroco: To clarify, would it make sense to name this method : `startRaftUpdatesConsumer` ?

pkg/raftstore/consensus/consensus.go
```diff
@@ -92,8 +97,12 @@ func NewConsensus(ctx context.Context, logger *zap.Logger, connectParams params.
 	consensus.snapshotIndex = snap.Metadata.Index
 	consensus.appliedIndex = snap.Metadata.Index
 
+	if len(snap.Data) > 0 {
+		consensus.commitC <- EntryCommit{SnapshotData: snap.Data}
+	}
+
 	go func() {
```
barroco: Move the fork inside startRaftUpdatesConsumer

pkg/raftstore/consensus/consensus.go
```diff
@@ -229,11 +269,104 @@ func (c *Consensus) handleReady(tickInterval time.Duration) error {
...
+		}
+
+		c.snapshotIndex = c.appliedIndex
+	}
+
+	return nil
+}
+
+// processNormalEntry passes the proposal to the store and waits for the result to be returned before untracking it.
+func (c *Consensus) processNormalEntry(data []byte, wg *sync.WaitGroup) error {
```
barroco: Please rename to `submitNormalEntryToStorage`

pkg/raftstore/consensus/consensus.go
```diff
@@ -229,11 +269,104 @@ func (c *Consensus) handleReady(tickInterval time.Duration) error {
...
+	if err != nil {
+		return stacktrace.Propagate(err, "failed to unmarshal committed proposal")
+	}
+
+	//if readOnly proposal and we did not initiate it, skip it (noop)
+	if prop.ReadOnly && !c.tracker.isPending(prop.ID) {
+		return nil
+	}
+
+	applyDoneC := make(chan ProposalResult, 1)
```
barroco, line 353: Add a comment: Ensure tracker is properly clean up.

pkg/raftstore/consensus/consensus.go
```diff
@@ -229,11 +269,104 @@ func (c *Consensus) handleReady(tickInterval time.Duration) error {
...
+		c.tracker.untrack(prop.ID, <-applyDoneC)
+	})
+
+	c.commitC <- EntryCommit{Prop: prop, Done: applyDoneC}
+	return nil
+}
+
+// raftpb.ConfChange is still used internally by Raft, we just need to apply the change to the node.
+// Changes requested by clients are processed by processConfigChangeV2Entry.
+func (c *Consensus) processConfigChangeEntry(data []byte) error {
```
barroco: Please clarify: applyConfigChange

pkg/raftstore/consensus/consensus.go
```diff
@@ -229,11 +269,104 @@ func (c *Consensus) handleReady(tickInterval time.Duration) error {
...
+	err := cc.Unmarshal(data)
+	if err != nil {
+		return stacktrace.Propagate(err, "failed to unmarshal config change data")
+	}
+
+	c.confState = *c.node.ApplyConfChange(cc)
 	return nil
 }
 
+func (c *Consensus) processConfigChangeV2Entry(data []byte) (bool, error) {
```
barroco: Please clarify: applyConfigChangeV2

pkg/raftstore/consensus/consensus.go
```diff
@@ -229,11 +269,104 @@ func (c *Consensus) handleReady(tickInterval time.Duration) error {
...
+	var cc raftpb.ConfChangeV2
+	err := cc.Unmarshal(data)
+	if err != nil {
+		return false, stacktrace.Propagate(err, "failed to unmarshal config change data")
+	}
+
+	c.confState = *c.node.ApplyConfChange(cc)
+
+	// TODO - implement config changes when triggered by a proposal
```
barroco:
```suggestion
	// TODO - implement config changes when triggered by a client
```

barroco: Please create a document to write key development definitions. Let's make the distinction between Peers and Client where Peers are other Raft nodes and Client is user agents calling the DSS API.

barroco (commented):
Partial review 3

pkg/raftstore/consensus/consensus.go
MariemBaccari: Thanks for pointing this out. `rafthttp` does write to this channel internally. I added a new `serverErrC` channel to capture the http server specific error and avoid mixing two error sources. 
I kept the buffer to make sure that `rafthttp` can forward an error even when we are not listening yet. We are also only expecting one error before stopping the goroutine, hence the 1 buffer.

pkg/raftstore/consensus/consensus.go
```diff
@@ -170,70 +211,179 @@ func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID
...
-		select {
-		case <-ticker.C:
-			c.node.Tick()
-		case err := <-c.transport.ErrorC:
-			return stacktrace.Propagate(err, "transport error")
-		case rd, ok := <-c.node.Ready():
-			if !ok {
-				return stacktrace.NewError("could not read from Ready(), shutting down handler")
-			}
+// startRaftUpdatesConsumer starts a goroutine thatprocesses the Ready channel of the Raft node and applies committed entries to the state machine
```
barroco: nit

```suggestion
// startRaftUpdatesConsumer starts a goroutine that processes the Ready channel of the Raft node and applies committed entries to the state machine
```

pkg/raftstore/consensus/consensus.go
MariemBaccari: I will open a separate PR for the doc 👍

pkg/raftstore/consensus/consensus.go
```diff
@@ -92,20 +101,20 @@ func NewConsensus(ctx context.Context, logger *zap.Logger, connectParams params.
...
-		consensus.Stop(context.Background())
-	}()
+	consensus.startRaftUpdatesConsumer(connectParams.TickInterval, connectParams.SnapshotIntervalEntries)
 
 	return consensus, nil
 }
 
 func (c *Consensus) Stop(ctx context.Context) {
 	c.once.Do(func() {
+		c.logger.Info("stopping consensus")
```
barroco, line 115: Can you please confirm that the context contains the service name and that it is visible in the logs to be able to distinguish the consensus of interest ?

pkg/raftstore/consensus/consensus.go
barroco: Please add your explanation as a comment.

pkg/raftstore/consensus/consensus.go
```diff
@@ -170,70 +211,179 @@ func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID
...
+				return
+			case <-ticker.C:
+				c.node.Tick()
+			case err := <-c.transport.ErrorC:
+				c.logger.Error("transport error", zap.Error(err))
+				return
+			case err := <-c.serverErrC:
+				c.logger.Error("http server error", zap.Error(err))
+				return
+			case rd, ok := <-c.node.Ready():
```
barroco: What does `rd` mean ?

barroco (commented):
Partial review 4

pkg/raftstore/consensus/consensus.go
MariemBaccari: It is short for `Ready` (Raft updates that are ready to be consumed) which is an etcd defined struct. I renamed the variable to `raftUpdates` following our discussion.

pkg/raftstore/consensus/consensus.go
MariemBaccari: I just updated that by setting a `service` string in the logger when initializing each raftstore. However, this will not be persisted by the internal Raft library logs. That will be a TODO following #1537.

barroco: Sounds good.

barroco (approved):

PR #1475, opened 2026-05-22
[scd] Remove redundant fetch cells call

MariemBaccari:
This PR removes a redundant call to `fetchCellsForSubscription` and, in `UpsertSubscription`, it removes the `newSubscription.Cells` update since cells are directly fetched from the DB by `fetchSubscription`.

Test: I used the `check_scd_write.sh` and `check_scd_read.sh` scripts with intermediate logging before and after the additional cells fetch to confirm the redundancy.

### DIFF
--- a/pkg/scd/store/sqlstore/subscriptions.go
+++ b/pkg/scd/store/sqlstore/subscriptions.go
@@ -56,45 +56,6 @@ func init() {
 	)
 }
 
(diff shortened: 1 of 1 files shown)


mickmis (approved):
LGTM - good catch! Maybe we will have an improvement in performances too.

PR #1483, opened 2026-05-26
[fix] Add missing locality arg to dss-rid-evict Tanka config

### LINKED ISSUE #1476: dss-rid-evict does not clean entities created by the core-service
**Describe the bug**
We have observed the `dss-rid-evict` job logging a "no RID entity older than ..." message and not cleaning old entities, even when they exist.

This is because the included [Tanka config for `dss-rid-evict`](https://github.com/interuss/dss/blob/04d0f49c7f9778bc7266b6ed863c1a3dc7f07fc6/deploy/services/tanka/evict.libsonnet#L37) does not include the `locality` arg, as [specified by the `evict` command](https://github.com/interuss/dss/blob/04d0f49c7f9778bc7266b6ed863c1a3dc7f07fc6/cmds/db-manager/cleanup/evict.go#L36). This arg *is* [passed to the `core-service`](https://github.com/interuss/dss/blob/04d0f49c7f9778bc7266b6ed863c1a3dc7f07fc6/deploy/services/tanka/core-service.libsonnet#L113) however, which results in the `dss-rid-evict` command not cleaning any entities created by its associated `core-service` instance.

**Expected behavior**
The `dss-rid-evict` job should delete entries which were created by its co-deployed `core-service` instance.

ashishjsharda:
## Problem
The `dss-rid-evict` cron job fails to clean up RID entities despite them existing. It logs "no RID entity older than ..." because `ListExpiredISAs` and `ListExpiredSubscriptions` in `evict.go` (lines 110, 128) filter by `--locality`, which defaults to `""` when not passed  matching nothing.

## Fix
Added `locality: metadata.locality` to the `RIDEvict` args_ block in `evict.libsonnet`, mirroring how `core-service.libsonnet:113` already passes it.

## Verification
- `make evaluate-tanka` (CI) will confirm `--locality` now appears in the   rendered `dss-rid-evict` CronJob manifest
- `make evict-locally` (CI) will confirm end-to-end eviction works

Fixes #1476

### DIFF
--- a/deploy/services/tanka/evict.libsonnet
+++ b/deploy/services/tanka/evict.libsonnet
@@ -55,6 +55,7 @@ local datastoreparameters = import 'datastoreparameters.libsonnet';
                       rid_isa: metadata.evict.rid.ISAs,
                       rid_sub: metadata.evict.rid.subscriptions,
                       rid_ttl: metadata.evict.rid.ttl,
+                      locality: metadata.locality,
                   } + datastoreparameters.all(metadata),
                   volumeMounts: volumes.all(metadata).schemaMounts,
                 }],


the-glu:
Hello @ashishjsharda 

Thank for the PR, I'm not sure why github show a diff in your PR, but I believe this has already been fixed in #1477 that is already merged, or am I missing something?

ashishjsharda:
> Hello @ashishjsharda
> 
> Thank for the PR, I'm not sure why github show a diff in your PR, but I believe this has already been fixed in #1477 that is already merged, or am I missing something?

Correct ,I missed the in-flight PR. #1477 already covers this. Closing this one.Thanks for flagging.

PR #1484, opened 2026-05-26
[terraform] Add option to make GKE cluster deletable

the-glu:
Add an option to ease debugging of google cluster by making them deletable directly from terraform.


deploy/infrastructure/utils/definitions/google_deletable.tf
barroco, line 1: This file seems to be a left over

barroco (approved):

PR #1485, opened 2026-05-26
[geo] normalize s2 loop

RustedBones:
Always normalize s2 loop instead of inverting loop when computed area is above threshold


mickmis (approved):
LGTM

PR #1487, opened 2026-05-26
[tanka/helm] Actually delete entries in cronjob evict tool

### LINKED ISSUE #1486: Make Evict Tool "delete" arg configurable via Tanka/Helm
**Describe the bug**
There is no way to set the ["delete" arg](https://github.com/interuss/dss/blob/04d0f49c7f9778bc7266b6ed863c1a3dc7f07fc6/cmds/db-manager/cleanup/evict.go#L35) for the DSS Evict scripts via the [Tanka/Helm](https://github.com/interuss/dss/blob/04d0f49c7f9778bc7266b6ed863c1a3dc7f07fc6/deploy/services/tanka/evict.libsonnet#L7) configs resulting in a "no entity was deleted, run the command again with the `--delete` flag to do so" log when configured jobs run.

**Expected behavior**
DSS Hosts should be able to configure evict jobs to actually delete entities.

the-glu:
Seems the delete flag was missing from the evict tool cronjob.

It's doesn't really make sense to have them not deleting entries, so no configuration flag has been added (they can be disabled anyway).

Fix #1486

(Not tested locally, but change is simple)


barroco (approved):

barroco (commented):
Please add a note in release notes.

the-glu:
> Please add a note in release notes.

I updated the existing line, since it's the second fix to make evict job working

barroco (approved):

PR #1489, opened 2026-05-27
[tanka/helm] Remove 'name' anotation in AWS load balancer

### LINKED ISSUE #1189: Deploy multiple DSS using terraform and helm in single VPC

Attempting to deploy two DSS instances to a single AWS VPC is successful at the terraform deployment level, but fails quietly when attempting to deploy AWS LoadBalancers through a Kubernetes LoadBalancer Service object.

These LoadBalancer Services should be configurable so that a name clash is avoidable, even if they aren't set by default.

I've provided a PR with a sample solution: https://github.com/interuss/dss/pull/1188

The clashing components `cockroachdb external node service` and `dss gateway external service` should have at least partially configurable names so that a deploying user can alter the names after seeing the clash OR they should be unique to begin with.

An alternative is a user manually altering the resources locally and then performing their deployment. But it requires a deeper understanding of the helm chart, and implies that it *ought* not be allowed, but can be manually worked around.

the-glu:
The 'name' annotation enforce the name on AWS load balancer, but that create issues when deploying multiple DSS K8S clusters in the same AWS region.

This PR just remove the annotation, letting AWS doing the naming.

Existing LBs won't be renamed.

Names automatically generated by AWS are still explicit:

image

This also fix #1189 (but with a different solution, it remove the option to configure the name).


barroco:
Please add a note in the release notes.

the-glu:
> Please add a note in the release notes.

Done

barroco (approved):

PR #1490, opened 2026-05-27
[geo] Leverage s2.CrossSign for loop validation

### LINKED ISSUE #1480: [rid] searching ISA with small polygon crashes DSS with OOM
When searching for ISA with an empty area, the DSS does not respond and eventually crashes wit OOM

**To Reproduce**

1. start DSS
2. call ```rid/v2/dss/identification_service_areas?area=0.000,0.000,0.000,0.005,-0.005,-0.005,0.000,0.005```

**Expected behavior**

DSS shoud either reject the area or respond with intersecting ISAs

**Additional context**
DSS loops in the `s2.Covering`. This looks related to https://github.com/golang/geo/issues/103

RustedBones:
Fix https://github.com/interuss/dss/issues/1480

Use `s2.CrossingSign` to detect intersection and handle case where point in on different edges (`MaybeCross`)

### DIFF
--- a/pkg/geo/s2.go
+++ b/pkg/geo/s2.go
@@ -89,49 +89,6 @@ func loopAreaKm2(loop *s2.Loop) float64 {
 	return (loop.Area() * earthAreaKm2) / (4.0 * math.Pi)
 }
 
-// chordSegmentsIntersect determines if two chord segments (segment 1 from p1a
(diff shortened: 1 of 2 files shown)


mickmis (approved):
LGTM - nice simplifications

PR #1491, opened 2026-05-27
[doc] Add a deployment checklist

the-glu:
This PR follows #1469 

It does add a deployment checklist, with the idea of listing each important steps for a production (or anything related, like a qualification, etc..) setup.

Checkbox are checkable and stored in the brower's stroage, providing (light) support of a real to-do list if needed.

It's partially redundant with some page (e.g. home, operations), but the idea is to centralize everything with pointers for non-local deployments.


docs/production_checklist.md
```diff
@@ -0,0 +1,24 @@
+# Production Checklist
+
+This checklist outlines the major decisions and steps required to deploy a production DSS instance.
+
+## Preparation
+
+* [ ] Review the [architecture requirements](architecture.md).
+* [ ] Decide on the datastore you will use (CockroachDB or YugabyteDB). **All participants in a DSS Pool must use the same datastore**, so plan accordingly.
+* [ ] Decide how and where you will deploy your DSS instances:
+    * This repository provides Terraform configurations for [Amazon Web Services (EKS)](infrastructure/aws.md) and [Google Cloud (GKE)](infrastructure/google.md) to deploy a Kubernetes cluster.
```
BenjaminPelletier:
```suggestion
    * This repository provides Terraform configurations for [Amazon Web Services (EKS)](infrastructure/aws.md) and [Google Cloud (GKE)](infrastructure/google.md) to deploy a Kubernetes cluster (the infrastructure into which the Services will be deployed).
```

docs/production_checklist.md
```diff
@@ -0,0 +1,24 @@
...
+    * This repository provides [Tanka](https://github.com/interuss/dss/blob/master/deploy/services/tanka/) files and [Helm Charts](https://github.com/interuss/dss/blob/master/deploy/services/helm-charts/dss) to be used with a Kubernetes cluster. Terraform will automatically generate these configurations if needed.
```
BenjaminPelletier:
```suggestion
    * This repository provides [Tanka](https://github.com/interuss/dss/blob/master/deploy/services/tanka/) files and [Helm Charts](https://github.com/interuss/dss/blob/master/deploy/services/helm-charts/dss) to be used to deploy Services into a Kubernetes cluster. Terraform will automatically generate these configurations if needed.
```

docs/production_checklist.md
```diff
@@ -0,0 +1,24 @@
...
+    * You may also choose to deploy manually or use custom configuration tools.
+* [ ] Prepare enough resources for the services.
```
BenjaminPelletier: nit

```suggestion
* [ ] Prepare sufficient resources for the services.
```

docs/production_checklist.md
```diff
@@ -0,0 +1,24 @@
...
+    * In particular, review the [CockroachDB recommendations](https://www.cockroachlabs.com/docs/v24.1/recommended-production-settings#cloud-specific-recommendations) and [YugabyteDB recommendations](https://docs.yugabyte.com/stable/deploy/checklist/#public-clouds); the datastore will consume the majority of the resources.
+    * Example sizing is also describled in [architecture](architecture.md#sizing).
+
+
+## Deployment
+
+* [ ] Deploy the DSS instance by following the guides based on your previous infrastructure choices.
+* [ ] If needed, pool your DSS instance. Guides are available for [CockroachDB](operations/pooling-crdb.md) and [YugabyteDB](operations/pooling.md).
```
BenjaminPelletier: Pooling happens as part of following the per-infrastructure-choice deployment guide [according to the "operations" pooling documentation](https://interuss.github.io/dss/latest/operations/pooling-crdb/#establishing-a-pool-with-first-instance) -- for instance, steps 6 and 8 in [the manual deployment documentation](https://interuss.github.io/dss/latest/infrastructure/google-manual/#deploying-a-dss-instance-via-kubernetes).  It looks like the [GCP](https://interuss.github.io/dss/dev/infrastructure/google/) and [AWS](https://interuss.github.io/dss/dev/infrastructure/aws/) deployment documentation are probably weak on making this point clear; I would expect them to say something like steps 6 and 8 in the manual documentation.  The important thing for this PR though is that I don't think "pool your DSS instance" is a separate step from "Deploy the DSS instance".

BenjaminPelletier (commented):
Thanks, this looks good and I think is another valuable piece of documentation that will see a lot of interest in the very near future.  In addition to a few comments below, I think we should change the term "Production".  Multiple groups have the practice of deploying software to multiple environments, at a minimum including actual production serving end users and a shared qualification environment where software prereleases are qualified.  The deployment into the qualification environment is fully relevant to all of this documentation, yet it is not production.  Perhaps "Deployment checklist"?  I know that isn't perfect because we also call running containers on a local machine a "local deployment".  Nonetheless, I think I still lean toward "Deployment checklist" as accurate enough while being maximally obvious when it correctly applies.  If there are concerns, possibly "Cloud deployment checklist" is more accurate as long as the USS isn't deploying on premises.

docs/production_checklist.md
the-glu: Yes the documentation is not clear on that.

For me there are two points (or subpoints): you deploy it, by spawning services and, optionally, you configure it correctly in a pool. First step you can probably do it alone and it may work directly in the non-pooled case, second step involve "special work" that need other DSS participants.

I also think that pointing to the pooling part is important, as it's a 'core' operation (plus others part are not clear (yet)).

I moved it to a sub-step in this document + tweaked wording, would that work?

the-glu:
@BenjaminPelletier Ready for a second pass of review, I reworded "Production" to "Deployment checklist" in general + added a menu entry as well

docs/production_checklist.md
BenjaminPelletier: > For me there are two points (or subpoints): you deploy it, by spawning services and, optionally, you configure it correctly in a pool. First step you can probably do it alone and it may work directly in the non-pooled case, second step involve "special work" that need other DSS participants.

I think the change in the PR is good, but I don't think this is an accurate description.  The way to pool is to make certain choices during the deployment process, not after (there are not two separate steps -- the special work must happen during deployment).  These choices must all have been made before invoking the command to spawn services (e.g., `tk apply`).  When generating certificates, the other members of the pool must be known.  If someone generates certs to spawn services and only then goes to pool, they will need to redo the certs.  The should_init setting is true in examples.  If deployment is performed with should_init, the database will need to be entirely torn down and redone to join a pool.

BenjaminPelletier (approved):

PR #1502, opened 2026-06-02
[metrics] Add RID metrics: subscriptions and isa count

the-glu:
This PR add two RID metrics: subscriptions and isa count

Contrib to #1257

### DIFF
--- a/cmds/core-service/main.go
+++ b/cmds/core-service/main.go
@@ -36,6 +36,8 @@ import (
 	"github.com/interuss/dss/pkg/versioning"
 	"github.com/interuss/stacktrace"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/metric"
 	"go.uber.org/zap"
 )
 
@@ -117,6 +119,14 @@ func createRIDServers(ctx context.Context, locality string, logger *zap.Logger) 
 		return nil, nil, stacktrace.Propagate(err, "Unable to interact with store")
 	}
 
+	if *enableOpenTelemetry {
+		err = registerRIDMetrics(ctx, ridStore)
+
+		if err != nil {
+			return nil, nil, stacktrace.Propagate(err, "Unable to setup metrics")
+		}
+	}
+
 	app := application.NewFromTransactor(ridStore, logger)
 	return &rid_v1.Server{
 			App:               app,
@@ -141,6 +151,42 @@ func createSCDServer(ctx context.Context, logger *zap.Logger) (*scd.Server, erro
 		DSSReportHandler:  &scd.JSONLoggingReceivedReportHandler{ReportLogger: logger},
 		AllowHTTPBaseUrls: *allowHTTPBaseUrls,
 	}, nil
+}
+
+func registerRIDMetrics(ctx context.Context, store rids.Store) error {
+
+	meter := otel.Meter("rid")
+
+	_, err := meter.Int64ObservableUpDownCounter(
+		"rid_subscriptions_total",
+		metric.WithDescription("Number of rid subscriptions"),
+		metric.WithInt64Callback(newCachedObservation(func(ctx context.Context) (int64, error) {
+			repo, err := store.Interact(ctx)
+			if err != nil {
+				return 0, stacktrace.Propagate(err, "Unable to interact with store")
+			}
+			count, err := repo.CountSubscriptions(ctx)
+			return count, err
+		})),
+	)
+	if err != nil {
+		return err
+	}
+
+	_, err = meter.Int64ObservableUpDownCounter(
+		"rid_identification_service_areas_total",
+		metric.WithDescription("Number of rid ISAs"),
+		metric.WithInt64Callback(newCachedObservation(func(ctx context.Context) (int64, error) {
+			repo, err := store.Interact(ctx)
+			if err != nil {
+				return 0, stacktrace.Propagate(err, "Unable to interact with store")
+			}
+			count, err := repo.CountISAs(ctx)
+			return count, err
+		})),
+	)
+
+	return err
 }
 
 // RunHTTPServer starts the DSS HTTP server.
--- a/cmds/core-service/otel.go
+++ b/cmds/core-service/otel.go
@@ -4,12 +4,15 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"sync"
+	"time"
 
 	"github.com/interuss/dss/pkg/logging"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
 	"go.opentelemetry.io/otel/exporters/prometheus"
+	ometric "go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/sdk/metric"
 	"go.opentelemetry.io/otel/sdk/resource"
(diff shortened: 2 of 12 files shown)


mickmis (commented):
Having a metric actively pulled from somewhere external to the runtime and doing a DB query sounds a bit dangerous. What about decoupling the DB polling from the the observe callback? And have a routine periodically querying those stats. We already have some form of 'DB' ping that could be put in common.

the-glu:
> Having a metric actively pulled from somewhere external to the runtime and doing a DB query sounds a bit dangerous. What about decoupling the DB polling from the the observe callback? And have a routine periodically querying those stats. We already have some form of 'DB' ping that could be put in common.

I don't think that an issue? It's very simple, readonly calls, using index / simple count only (and probably cached), and this allows calls to fetch as fast or as slow as needed in term of update interval from the monitoring system, without enforcing an update interval.

Causing performances issues would mean DDOS on purpose, on an private, separate endpoint, and this can already be done on others, public by design.

the-glu:
I added a small cache (1s, not configurable) who should prevent  unexpected cases.

cmds/core-service/main.go
```diff
@@ -143,6 +153,42 @@ func createSCDServer(ctx context.Context, logger *zap.Logger) (*scd.Server, erro
 	}, nil
 }
 
+func registerRIDMetrics(ctx context.Context, store rids.Store) error {
+
+	meter := otel.Meter("rid")
+
+	_, err := meter.Int64ObservableGauge(
```
mickmis: Wouldn't `UpDownCounter` be semantically more correct here?

cmds/core-service/otel.go
```diff
@@ -104,3 +107,34 @@ func serveMetrics(ctx context.Context, listeningAddress string) {
...
+	last      int64
+	fetchedAt time.Time
+	ttl       time.Duration
+	fetch     func(context.Context) (int64, error)
+}
+
+func (c *cachedGauge) Observe(ctx context.Context, o ometric.Int64Observer) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if time.Since(c.fetchedAt) < c.ttl {
```
mickmis, line 123: Does this actually work? At creation with `fetchedAt` a zero value, `time.Since(c.fetchedAt)` would return a negative duration which would always be `< c.ttl` thus always return `c.last`, or am I missing something?

cmds/core-service/otel.go
```diff
@@ -104,3 +107,34 @@ func serveMetrics(ctx context.Context, listeningAddress string) {
...
+	mu        sync.Mutex
+	last      int64
+	fetchedAt time.Time
+	ttl       time.Duration
+	fetch     func(context.Context) (int64, error)
+}
+
+func (c *cachedGauge) Observe(ctx context.Context, o ometric.Int64Observer) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
mickmis, line 122: The doc mentions the async meters should not take an indefinite amount of time, which is the case with such a lock. Not sure how to handle that though. I think that at least there should be a select to listen for the `ctx.Done()` as specified by the doc.

- Remove them and accept there may be multiple concurrent calls to fetch? But that would partially defeat the purpose of having this mechanism. 
- Use `TryLock` to return the cached value to concurrent callers? Ugly and incorrect: a later call might return stale value leading to false readings.
- Don't bother doing all that because that overcomplicates things? Maybe so in the end.

WDYT?

cmds/core-service/main.go
```diff
@@ -143,6 +153,42 @@ func createSCDServer(ctx context.Context, logger *zap.Logger) (*scd.Server, erro
...
+		"rid_subscriptions_count",
+		metric.WithDescription("Number of rid subscriptions"),
+		metric.WithInt64Callback(newCachedGauge(func(ctx context.Context) (int64, error) {
+			repo, err := store.Interact(ctx)
+			if err != nil {
+				return 0, stacktrace.Propagate(err, "Unable to interact with store")
+			}
+			count, err := repo.CountSubscriptions(ctx)
+			return int64(count), err
```
mickmis: Make the DB calls return an `int64`? At least for CRDB (and I suspect for YBDB too), it returns an int64

cmds/core-service/otel.go
the-glu: I think we can just ignore it, the goal is anyways to limit metric usage, to make it wait seems filling the purpose.

cmds/core-service/otel.go
the-glu: I did work when I tested, if c.fetchedAt is zero, `time.Since(c.fetchedAt)` is just big no?

cmds/core-service/main.go
the-glu: not 100% sure about semantic details by reading the doc since both seems fine to be for that case, I switched them ;)

cmds/core-service/otel.go
the-glu: 2562047h47m16.854775807s

it's indeed big.

cmds/core-service/main.go
the-glu: Done. If we each even int32 we will have others problems :P

cmds/core-service/otel.go
the-glu: (Also: it's equivalent to one concurrent request at the same time for metrics in term of waiting time, who is I think more that acceptable in term of waiting time)

mickmis: ah but wait - the collection is actually controlled from within the runtime, not from the outside. I had assumed the opposite. If that is correct we should be able to ditch all that and just configure the otel collection correctly?

mickmis: LLM-suggested example, that gives some pointers:
```go
package main

import (
	"context"
	"time"

	"go.opentelemetry.io/otel"
	"go.opentelemetry.io/otel/exporters/stdout/stdoutmetric"
	"go.opentelemetry.io/otel/sdk/metric"
)

func main() {
	ctx := context.Background()

	// 1. Create a Metric Exporter (e.g., standard output)
	exporter, err := stdoutmetric.New()
	if err != nil {
		panic(err)
	}

	// 2. Create a Periodic Reader, passing the exporter and setting the interval
	reader := metric.NewPeriodicReader(
		exporter,
		metric.WithInterval(15*time.Second),
	)

	// 3. Register the Reader to the MeterProvider
	provider := metric.NewMeterProvider(metric.WithReader(reader))
	otel.SetMeterProvider(provider)

	// Application logic goes here...
	select {} // Block to keep the app running
}
```

the-glu: What?

The prometheus is working in a pull manner, we don't push metric to it? NewPeriodicReader make no sense there?

mickmis: Discussed offline: this does not apply as a 'Prometheus' reader is used in our case. Let's just go as is for now.

cmds/core-service/otel.go
mickmis: ah yeah I got it reversed indeed.

cmds/core-service/main.go
```diff
@@ -143,6 +153,42 @@ func createSCDServer(ctx context.Context, logger *zap.Logger) (*scd.Server, erro
 	}, nil
 }
 
+func registerRIDMetrics(ctx context.Context, store rids.Store) error {
+
+	meter := otel.Meter("rid")
+
+	_, err := meter.Int64ObservableUpDownCounter(
+		"rid_subscriptions_count",
```
mickmis: Looking at https://opentelemetry.io/docs/specs/semconv/general/naming/#do-not-pluralize-updowncounter-names:
```suggestion
		"subscription.count",
```
(assuming the `rid` from the meter is prefixed automatically? if not `rid.subscription.count`)

mickmis (approved):
LGTM modulo naming to follow semantic conventions.

cmds/core-service/main.go
the-glu: I followed Prometheus ones (underscores / plural), as the goal is to generate prometheus metric: https://prometheus.io/docs/practices/naming/ .

(Rid is needed)

Count should probably be total however.

mickmis: ah indeed, alright

cmds/core-service/otel.go
```diff
@@ -104,3 +107,34 @@ func serveMetrics(ctx context.Context, listeningAddress string) {
...
+	c.last = v
+	c.fetchedAt = time.Now()
+	o.Observe(v)
+	return nil
+}
+
+func newCachedGauge(fetch func(context.Context) (int64, error)) ometric.Int64Callback {
+	g := &cachedGauge{ttl: time.Second, fetch: fetch}
+	return g.Observe
+}
```
mickmis: Rename 'gauge' -> 'updowncounter'

the-glu: I switched to cachedObservation as it's generic

PR #1503, opened 2026-06-02
[metrics] Add SCD metrics: subscriptions, constraints and operational intents count

the-glu:
Follow #1502 

Add SCD metrics.

Contrib to #1257


mickmis (approved):
LGTM

PR #1504, opened 2026-06-02
[opentelemetry] Split enable_opentelemetry into enable_metrics and enable_tracing

the-glu:
This PR follow #1503 

It split the `enable_opentelemetry` flag into two dedicated sub flags.

Preparing deployments files to add DSS metrics, it doesn't make sense to enable full tracing just for 'normal' Prometheus metrics, especially as it show warning if partially configured.

No backward compatibly added, as the flag is very new, needed manual work to be used and only tracing has been released, who shall only be used for debugging.

### DIFF
--- a/NEXT_RELEASE_NOTES.md
+++ b/NEXT_RELEASE_NOTES.md
@@ -46,6 +46,8 @@ The release notes should contain at least the following sections:
 
 ## Optional migration tasks
 
+* `enable_opentelemetry` has been splited into two flags: `enable_metrics` and `enable_tracing`
+
 ## Important information
 
 * Fixed a bug where the `evict` command ignored entries without a locality. If your DSS instance does not have a locality set, the next `evict` run may be slow while it processes the backlog of old entries.
--- a/NEXT_RELEASE_NOTES.md
+++ b/NEXT_RELEASE_NOTES.md
@@ -46,7 +46,7 @@ The release notes should contain at least the following sections:
 
 ## Optional migration tasks
 
-* `enable_opentelemetry` has been splited into two flags: `enable_metrics` and `enable_tracing`
+* `enable_opentelemetry` has been split into two flags: `enable_metrics` and `enable_tracing`
 
 ## Important information
 
--- a/cmds/core-service/otel.go
+++ b/cmds/core-service/otel.go
@@ -23,28 +23,42 @@ import (
 
 // setupOTelSDK bootstraps the OpenTelemetry pipeline.
 // If it does not return an error, make sure to call shutdown for proper cleanup.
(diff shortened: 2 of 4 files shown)


docs/operations/monitoring.md
```diff
@@ -137,15 +137,17 @@ You can enable it on the DSS server to get:
 * Tracing for all queries
 * A Prometheus endpoint with some metrics
 
-Currently, this setting is not yet available in Terraform, Helm or Tanka.
+Currently, thoses settings are not yet available in Terraform, Helm or Tanka.
```
mickmis, line 140: those

cmds/core-service/otel.go
```diff
@@ -23,28 +23,43 @@ import (
...
+		if err != nil {
+			return nil, err
+		}
+		otel.SetMeterProvider(meterProvider)
 	}
-	otel.SetMeterProvider(meterProvider)
 
 	shutdown := func(ctx context.Context) error {
-		return errors.Join(tracerProvider.Shutdown(ctx), meterProvider.Shutdown(ctx))
+		if tracerProvider != nil && meterProvider != nil {
```
mickmis: nit: 
```go
		var err error
		if tracerProvider != nil {
			err = errors.Join(err, tracerProvider.Shutdown(ctx))
		}
		if meterProvider != nil {
			err = errors.Join(err, meterProvider.Shutdown(ctx))
		}
		return err
```

NEXT_RELEASE_NOTES.md
```diff
@@ -44,6 +44,8 @@ The release notes should contain at least the following sections:
 
 ## Mandatory migration tasks
 
+* `enable_opentelemetry` has been splited into two flags: `enable_metrics` and `enable_tracing`
```
mickmis: To be updated according to feedback about retrocompatiblity: move in optional migration task the fact of renaming parameter.

mickmis (approved):
> No backward compatibly added, as the flag is very new, needed manual work to be used and only tracing has been released, who shall only be used for debugging.

If it's been released we should have backward compatibility: even if within our development timeline this is a short window of time, who knows how much time there will be between version upgrades for the users.
Given that only the tracing has been released: just have 'enable_telemetry' be an alias of 'enable_tracing' and add a deprecation notice in the usage of the flag?

Otherwise LGTM

NEXT_RELEASE_NOTES.md
```diff
@@ -46,6 +46,8 @@ The release notes should contain at least the following sections:
 
 ## Optional migration tasks
 
+* `enable_opentelemetry` has been splited into two flags: `enable_metrics` and `enable_tracing`
```
mickmis:
```suggestion
* `enable_opentelemetry` has been split into two flags: `enable_metrics` and `enable_tracing`
```

PR #1505, opened 2026-06-02
 [monitoring] Normalize grafana version to 13.0 

the-glu:
This PR follow #1504 

It set grafana version to 13.0 in both case:

- For helm chart, it was defaulting to 12.X, but this was too old for a new dashboard
- For tanka, it was using :latest, witch may create issue

This is now normalized to 13.0 (latest as the time of this PR).


mickmis (approved):
Reviewed 7d6b22a07bb6c16106609a27ab5fb673fc19ad74
LGTM

PR #1506, opened 2026-06-03
[deploy] Add enable_metric flag for DSS + fetch pod metrics in prometheus

the-glu:
This PR add the `enable_metric` flag in general deploy stack (terraform, helm, tanka).

It set the flag + add annotation for prometheus collection.
Promethus pod collection (instead of just services) as been enabled, as each DSS instance own metrics need to be collected, not as a whole for a service (e.g.: http metrics are individual).

Follow #1505

### DIFF
--- a/deploy/infrastructure/dependencies/terraform-commons-dss/helm.tf
+++ b/deploy/infrastructure/dependencies/terraform-commons-dss/helm.tf
@@ -68,6 +68,7 @@ resource "local_file" "helm_chart_values" {
         publicEndpoint      = "https://${var.app_hostname}"
         enableScd           = var.enable_scd
         enableScdGlobalLock = var.enable_scd_global_lock
+        enableDssMetrics    = var.enable_dss_metrics
         locality            = "zone=${var.locality}"
 
         evict = {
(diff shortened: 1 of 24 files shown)


mickmis (approved):
Reviewed 371fb3fd945285964e4a4ba378c6a387a5c3e535
LGTM, but why not adding tracing too?

the-glu:
> Reviewed [371fb3f](https://github.com/interuss/dss/commit/371fb3fd945285964e4a4ba378c6a387a5c3e535) LGTM, but why not adding tracing too?

To keep PR smalls ;)

Tracing is a bit heavier (and need a new 'consumer'), so it worth it dedicated options :)

mickmis:
> > Reviewed [371fb3f](https://github.com/interuss/dss/commit/371fb3fd945285964e4a4ba378c6a387a5c3e535) LGTM, but why not adding tracing too?
> 
> To keep PR smalls ;)
> 
> Tracing is a bit heavier (and need a new 'consumer'), so it worth it dedicated options :)

Alright, I understood wrongly that the plan was to not add it for tracing.

PR #1508, opened 2026-06-03
[auth] Don't crash on tokens with missing 'exp' field

the-glu:
This PR fix an issue with token that are missing the 'exp' field.

`VerifyExpiresAt` call return was not checked, making it useless.

Since we do our own check (with 1 hour of margin), call has been removed and check for ExpiresAt added in condition.

Added relevant tests for all missing properties in a valid token as well.

This is not a security issue as it's "just" crashing the thread doing the check, without answering.

### DIFF
--- a/pkg/auth/claims/claims.go
+++ b/pkg/auth/claims/claims.go
@@ -13,7 +13,7 @@ import (
 
 var (
 	errMissingOrEmptySubject = errors.New("missing or empty subject")
-	errTokenExpireTooFar     = errors.New("token expiration time is too far in the furture, Max token duration is 1 Hour")
+	errTokenExpireTooFar     = errors.New("token expiration time is too far in the furture or not set, Max token duration is 1 Hour")
 	errMissingIssuer         = errors.New("missing Issuer URI")
 	// Now allows test to override with specific time values
 	Now = time.Now
@@ -101,9 +101,7 @@ func (c *Claims) Valid() error {
 	}
 	now := Now()
 
-	c.VerifyExpiresAt(now, true)
-
-	if c.ExpiresAt.After(now.Add(time.Hour)) {
+	if c.ExpiresAt == nil || c.ExpiresAt.After(now.Add(time.Hour)) {
 		return errTokenExpireTooFar
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/auth/claims/claims.go
```diff
@@ -101,9 +101,7 @@ func (c *Claims) Valid() error {
 	}
 	now := Now()
 
-	c.VerifyExpiresAt(now, true)
-
-	if c.ExpiresAt.After(now.Add(time.Hour)) {
+	if c.ExpiresAt == nil || c.ExpiresAt.After(now.Add(time.Hour)) {
```
mickmis: Why not (the lib function protects for nil values):

```suggestion
	if !c.VerifyExpiresAt(now.Add(time.Hour), true) {
```

mickmis (approved):

pkg/auth/claims/claims.go
the-glu: Why not yes, I just took existing test ^^'

the-glu: Actually the fix is wrong. We need to test both case, I'm not sure why test passed, let me check

the-glu: Ok:

Expiration is tested somewhere else.

Here we want to test that expiration is not in the future AND is set, so we cannot do that with the function, we need the inverse condition (we want it to be invalid in 1h and set).

Reverted to orignal + added a 'exp in the far future' test :)

mickmis: > Here we want to test that expiration is not in the future AND is set, so we cannot do that with the function, we need the inverse condition (we want it to be invalid in 1h and set).

From what I understood that's the purpose of the `req` argument? If req is true and the expiration unset, it will return false. Not hugely important though, so up to you.

the-glu: The point of that test (line 104) is to check that 1) it's set and 2) it's not valid in the far future

`VerifyExpiresAt` is made to check if it's set, we still before it's expiration date.

I wouldn't abuse it for our check who is different from the standard.

(We could test that VerifyExpiresAt(+1h, false) doesn't return true, but that double negations everywhere (when not set, it will be valid in 1h and therefor that an issue; when set and valid in 1h that an issue) and don't protect us for change in the underlying check).

Issue #1513, opened 2026-06-08
Log timed-out operations


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
Per [discussion in 1509](https://github.com/interuss/dss/issues/1509#issuecomment-4635248089), it seems that the DSS may not log any information about an operation whose context was canceled due to timeout.  This is exactly the time when information about what happened would be most useful as the most common impactful problem being experienced by DSS users currently is operation timeouts.

**Describe the solution you'd like**
At a minimum (resolves bug-level problem): log the occurrence of a timeout with as much of the same level of detail as a normally-completed operation as possible.

Even better: include all additional information easily available regarding the nature of the timeout (e.g., what was happening when the timeout occurred, number of transactions attempted/retried during the attempt, etc).

Ideal: immediately log whatever is immediately available, then enqueue a mini-investigation that will happen in the background.  The mini-investigation could query the database for diagnostic information related to the event and perform other lookups, producing a mini report for that timeout occurrence that could be retrieved when the mini-investigation was complete.  Investigations could be bounded by having a max queue size (e.g., 3 simultaneous investigations) and bumping the last investigation queued every time a new investigation was queued -- this would ensure the first and last timeout occurrences in a burst would be investigated.  Investigation reports could also be bounded by deleting old reports after the number of reports exceeded a certain value.  Investigations could be short-circuited when the problem occurs due to known conditions: e.g., "There were 247 operations submitted in a 1-second period; this operation failed because the capacity of the system was exceeded so I don't need to proceed any further with this investigation by querying the database or any other steps."  Retrieving the results of these mini-investigations should be easy for the user so that InterUSS can request any users experiencing timeouts provide these investigation reports.  The content of these reports should allow InterUSS to diagnose the problem and recommend solutions.

the-glu:
Minimum done in https://github.com/interuss/dss/pull/1515

Better / Ideal still need to be done

BenjaminPelletier:
Reducing to P2 with completion of minimum

ashishjsharda:
Thanks for #1515, the ctx_err field is great and the timeout handler swap fixes real handlers that ignore context cancellation.

One thing I ran into while poking at this: ctx_err still doesn't get set on a server-side timeout. http.TimeoutHandler builds its deadline context inside its own ServeHTTP and rebinds the request locally:

```go
ctx, cancelCtx = context.WithTimeout(r.Context(), h.dt)
r = r.WithContext(ctx)   // local to this function
[...]

PR #1514, opened 2026-06-09
[core] Fix timeout argument

the-glu:
Working on #1513 , I noticed that server timeout flag have a space inside, making it very impratical to use ( `-'server timeout' 1s` need to be used).

This PR add an underscore and make it backward compatible.


cmds/core-service/main.go
```diff
@@ -332,6 +333,12 @@ func RunHTTPServer(ctx context.Context, ctxCanceler func(), address, locality st
 		multiRouter.Routers = append(multiRouter.Routers, &scdV1Router)
 	}
 
+	if *legacyTimeout != 0 {
```
mickmis: - use flag.Lookup instead
- so that the default can remain the same as before
- put this in an `init()`

cmds/core-service/main.go
```diff
@@ -46,7 +46,8 @@ var (
 	enableSCD         = flag.Bool("enable_scd", false, "Enables the Strategic Conflict Detection API")
 	allowHTTPBaseUrls = flag.Bool("allow_http_base_urls", false, "Enables http scheme for Strategic Conflict Detection API")
 	enableHTTP        = flag.Bool("enable_http", false, "DEPRECATED (replaced by allow_http_base_urls): Enables http scheme for Strategic Conflict Detection API")
-	timeout           = flag.Duration("server timeout", 10*time.Second, "Default timeout for server calls")
+	legacyTimeout     = flag.Duration("server timeout", 0, "Default timeout for server calls (legacy)")
```
mickmis:
```suggestion
	legacyTimeout     = flag.Duration("server timeout", 10*time.Second, "DEPRECATED (replaced by server_timeout) Default timeout for server calls")
```

mickmis (approved):

cmds/core-service/main.go
the-glu: I used `flag.Visit` in main()

PR #1516, opened 2026-06-09
[release scripts] Ignore yugabyte certificate init errors

the-glu:
Workspaces may not be fully clean-up, but that ok to reuse old certificate. Since yugabyte certificate manager's init command force it to be empty, this ignore errors.


mickmis (approved):
LGTM

Issue #813, opened 2022-09-20
Fix database indices


BenjaminPelletier:
Currently, the indices we are using for many of our CockroachDB database tables are suboptimal; we should update our indices to optimize performance, storage, and probability of success.

Some indices are not used in any queries.  Only one index is used when SELECTing data, but sometimes we have indices on each of the columns used in the WHERE clause.  It is [an anti-pattern](https://www.cockroachlabs.com/docs/v22.1/schema-design-indexes#best-practices) to have an index on a sequential key like timestamps.

It is likely this will improve #742

PR #1517, opened 2026-06-10
[crdb] Remove useless indexes

the-glu:
This PR contrib to general performance and to #813 by removing useless indexes in CRDB.

Cells index or primary keys should be used looking at queries, making those indexes useless.

ends_at / updated_at are not removed to use indexes when using evict tool.
owner on `rid.subscriptions` is used in `MaxSubscriptionCountInCellsByOwner`

Notice that not the case for yugabyte and nothing changed there.

### DIFF
--- a/NEXT_RELEASE_NOTES.md
+++ b/NEXT_RELEASE_NOTES.md
@@ -47,6 +47,11 @@ The release notes should contain at least the following sections:
 ## Optional migration tasks
 
 * `enable_opentelemetry` has been split into two flags: `enable_metrics` and `enable_tracing`
+* A new pair of migrations for CockroachDB [has been added](https://github.com/interuss/dss/pull/1517) for RID (4.1.0) and SCD (3.4.0) to remove useless indexes.
+  * The DSS will work regardless of the migration status.
+  * This migration will improve performance. If that is not the case for a specific scenario that was missed, please downgrade to the previous version and open an issue.
+  * These indexes are still useful for Yugabyte, so no such migration exists for it.
+  * Details about migrations are available [there](https://interuss.github.io/dss/latest/operations/database-migrations/)
 
 ## Important information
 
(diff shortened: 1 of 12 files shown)


barroco:
@the-glu, thank you for this. Can you please provide evidence of no degradation ?

the-glu:
> @the-glu, thank you for this. Can you please provide evidence of no degradation ?

Performances tests I ran (locus ones) show no changes, but they are really synthetic. 

This also really depends on the planner / statistics / data in the table, so it's mostly based on query analysis that measurable data.

barroco:
Please update the NEXT_RELEASE_NOTES

BenjaminPelletier:
This looks reasonable to me, but as @barroco notes, users with existing deployments will need/want to do something because of this change so we need something in NEXT_RELEASE_NOTES.

Beyond this PR, we should make sure it's clear how to perform a DB migration and what is and isn't required.  Specifically, one user group thought they needed to synchronize the actions of all pool participants to a narrow time window to update both the schema and software version, while I think we have a design intent to avoid that need (instead allowing a leisurely schema update on the previous software version, then leisurely transitioning of each DSS instance to the new software version, or perhaps in the reverse order).

the-glu:
Release notes updated

NEXT_RELEASE_NOTES.md
```diff
@@ -47,6 +47,10 @@ The release notes should contain at least the following sections:
 ## Optional migration tasks
 
 * `enable_opentelemetry` has been split into two flags: `enable_metrics` and `enable_tracing`
+* A new pair of migrations for CockroachDB [has been added](https://github.com/interuss/dss/pull/1517) for RID (4.1.0) and SCD (3.4.0) to remove useless indexes.
+  * The DSS will work regardless of the migration status.
+  * This migration will improve performance. If that is not the case for a specific scenario that was missed, please downgrade to the previous version and open an issue.
```
barroco, line 52: I think we should document how to perform an upgrade / downgrade. Would you mind documenting it ?

barroco: Please make a link to the change in #1547

the-glu: Done, notice the link 404 until next release

barroco (approved):

PR #1518, opened 2026-06-10
[scd] Update and fetch notification_index in one query

the-glu:
This PR do a small optimization in SCD when incrementing notification index, by querying and updating them in one query instead of doing that with two.

It also do filtering on sql side instead of manual filtering on go side.

`FlightsInSub.py`, ran locally in a single node crdb show a small improvement:

image

Worst cases where round-trip is higher should probably show better improvement.

Small behavior change: number of updated subcriptions / returned is not limited anymore. We can add that back but it feels wrong.

### DIFF
--- a/pkg/scd/store/sqlstore/subscriptions.go
+++ b/pkg/scd/store/sqlstore/subscriptions.go
@@ -293,50 +293,67 @@ func (c *repo) SearchSubscriptions(ctx context.Context, v4d *dssmodels.Volume4D)
 	return subscriptions, nil
 }
 
-// Implements scd.repos.Subscription.IncrementNotificationIndices
-func (c *repo) IncrementNotificationIndices(ctx context.Context, subscriptionIds []dssmodels.ID) ([]int, error) {
-	var updateQuery = `
-			UPDATE scd_subscriptions
-			SET notification_index = notification_index + 1
-			WHERE id = ANY($1)
-			RETURNING id, notification_index`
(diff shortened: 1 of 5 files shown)


pkg/scd/store/sqlstore/subscriptions.go
```diff
@@ -293,50 +293,67 @@ func (c *repo) SearchSubscriptions(ctx context.Context, v4d *dssmodels.Volume4D)
...
-		if err != nil {
-			return nil, stacktrace.Propagate(err, "Error scanning notification index row")
-		}
-		indices[dssmodels.ID(id)] = notificationIndex
+		return nil, stacktrace.Propagate(err, "Could not calculate spatial covering")
 	}
-	if err := rows.Err(); err != nil {
-		return nil, stacktrace.Propagate(err, "Error in rows query result")
+	if len(cells) == 0 {
+		return nil, nil
```
BenjaminPelletier, line 321: Wouldn't it be an error to update subscriptions in a null area?

BenjaminPelletier (approved):

pkg/scd/store/sqlstore/subscriptions.go
the-glu: On the application layer I think you cannot have subscriptions with a null area yes, but on the storage side asking to increase count over a empty area is just a no-op, that not really an error (like before, you could ask it to update counts of an empty subscription list)

PR #1525, opened 2026-06-15
[Raft] Add generic raftstore

MariemBaccari:
Implements the generic raftstore which processes and applies committed entries. 
`Transact` calls `s.Consensus.HandleClientRequest` which blocks until the proposal is committed and applied. 
This ensures that we don't return a response before the request is fully processed. 

Implements #1700

### DIFF
--- a/pkg/raftstore/consensus/consensus.go
+++ b/pkg/raftstore/consensus/consensus.go
@@ -110,9 +110,9 @@ func NewConsensus(ctx context.Context, logger *zap.Logger, connectParams params.
 	return consensus, nil
 }
 
-// TODO: ctx is currently ignored
+// TODO: ctx is currently ignored (see issue: https://github.com/interuss/dss/issues/1610)
 func (c *Consensus) Stop(ctx context.Context) {
-	// TODO: remove once
+	// TODO: remove once (see issue: https://github.com/interuss/dss/issues/1610)
 	c.once.Do(func() {
 		c.logger.Info("stopping consensus")
 		close(c.stopC)
@@ -222,6 +222,7 @@ func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID 
 func (c *Consensus) startRaftUpdatesConsumer(tickInterval time.Duration, snapshotInterval uint64) {
 	go func() {
 		// TODO: this shouldn't be triggered from inside the consensus instance, removing it will allow removing the once.
+		// (see issue: https://github.com/interuss/dss/issues/1610)
 		defer c.Stop(context.Background())
 
 		ticker := time.NewTicker(tickInterval)
--- a/pkg/raftstore/consensus/consensus.go
+++ b/pkg/raftstore/consensus/consensus.go
@@ -131,7 +131,7 @@ func (c *Consensus) Stop(ctx context.Context) {
 }
 
 // HandleClientRequest blocks until the proposal is committed and applied / dropped or until ctx is cancelled.
-func (c *Consensus) HandleClientRequest(ctx context.Context, requestType string, value any, readOnly bool) (any, error) {
+func (c *Consensus) HandleClientRequest(ctx context.Context, requestType string, value []byte, readOnly bool) (any, error) {
 	proposal, err := c.newProposal(ctx, requestType, value, readOnly)
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "failed to create proposal")
--- a/pkg/raftstore/consensus/consensus.go
+++ b/pkg/raftstore/consensus/consensus.go
@@ -110,7 +110,9 @@ func NewConsensus(ctx context.Context, logger *zap.Logger, connectParams params.
 	return consensus, nil
 }
 
+// TODO: ctx is currently ignored
 func (c *Consensus) Stop(ctx context.Context) {
+	// TODO: remove once
 	c.once.Do(func() {
 		c.logger.Info("stopping consensus")
 		close(c.stopC)
@@ -223,6 +225,7 @@ func (c *Consensus) initTransport(ctx context.Context, nodeID uint64, clusterID 
 // startRaftUpdatesConsumer starts a goroutine that processes the Ready channel of the Raft node and applies committed entries to the state machine
 func (c *Consensus) startRaftUpdatesConsumer(tickInterval time.Duration, snapshotInterval uint64) {
 	go func() {
+		// TODO: this shouldn't be triggered from inside the consensus instance, removing it will allow removing the once.
 		defer c.Stop(context.Background())
 
 		ticker := time.NewTicker(tickInterval)
--- a/pkg/raftstore/consensus/consensus.go
+++ b/pkg/raftstore/consensus/consensus.go
@@ -134,11 +134,7 @@ func (c *Consensus) Stop(ctx context.Context) {
 
 // HandleClientRequest blocks until the proposal is committed and applied / dropped or until ctx is cancelled.
 func (c *Consensus) HandleClientRequest(ctx context.Context, requestType string, value []byte, readOnly bool) (any, error) {
-	proposal, err := c.newProposal(ctx, requestType, value, readOnly)
-	if err != nil {
-		return nil, stacktrace.Propagate(err, "failed to create proposal")
-	}
-
+	proposal := c.newProposal(ctx, requestType, value, readOnly)
 	buf, err := json.Marshal(proposal)
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "failed to marshal proposal")
--- a/pkg/raftstore/consensus/proposal.go
+++ b/pkg/raftstore/consensus/proposal.go
@@ -29,7 +29,7 @@ type Proposal struct {
 	ReadOnly bool `json:"read_only"`
 }
 
-func (c *Consensus) newProposal(ctx context.Context, requestType string, value []byte, readOnly bool) (Proposal, error) {
+func (c *Consensus) newProposal(ctx context.Context, requestType string, value []byte, readOnly bool) Proposal {
 	timestamp := timestamp.MustGetRequestTimestamp(ctx)
 
 	return Proposal{
@@ -39,7 +39,7 @@ func (c *Consensus) newProposal(ctx context.Context, requestType string, value [
 		RequestType: requestType,
 		Value:       value,
 		ReadOnly:    readOnly,
-	}, nil
+	}
 }
 
 type ProposalResult struct {
--- a/pkg/raftstore/consensus/proposal.go
+++ b/pkg/raftstore/consensus/proposal.go
@@ -2,11 +2,11 @@ package consensus
 
 import (
 	"context"
-	"encoding/json"
 	"sync"
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/interuss/dss/pkg/timestamp"
 	"github.com/interuss/stacktrace"
 )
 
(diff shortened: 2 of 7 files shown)


pkg/timestamp/timestamp.go
```diff
@@ -0,0 +1,32 @@
...
+
+// WithTimestamp returns a new context with the given timestamp.
+func WithTimestamp(ctx context.Context, t time.Time) context.Context {
+	return context.WithValue(ctx, timestampKey{}, t)
+}
+
+// Middleware is an HTTP middleware that adds a timestamp to the request context.
+func Middleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := WithTimestamp(r.Context(), time.Now().UTC())
```
barroco: What kind of precision do we need ?

barroco (dismissed):
Partial review 1: Please create a separate PR with all changes which touch the current sqlstore implementation.

pkg/timestamp/timestamp.go
```diff
@@ -0,0 +1,32 @@
...
+	t, ok := ctx.Value(timestampKey{}).(time.Time)
+	if !ok {
+		return time.Time{}
+	}
+
+	return t
+}
+
+// WithTimestamp returns a new context with the given timestamp.
+func WithTimestamp(ctx context.Context, t time.Time) context.Context {
```
barroco: What's this timestamp used for ? We need to be more specific.

pkg/timestamp/timestamp.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+	"net/http"
+	"time"
+)
+
+type timestampKey struct{}
+
+// NowFromContext returns the timestamp from the context, or zero if not present.
+func NowFromContext(ctx context.Context) time.Time {
```
barroco: I don't think Now is specific enough. When is it captured ? Is it the Request start ?

pkg/timestamp/timestamp.go
```diff
@@ -0,0 +1,32 @@
...
+	t, ok := ctx.Value(timestampKey{}).(time.Time)
+	if !ok {
+		return time.Time{}
```
barroco: I don't think we should fallback to now. An error should be triggered if no timestamp has been registered.

barroco (commented):
Partial review 2

pkg/timestamp/timestamp.go
MariemBaccari: Since the `sqlstore` uses `time.Now` as well, I assumed there is no need to specify a precision.

pkg/timestamp/timestamp.go
MariemBaccari: The timestamp is set by the dss node that receives the client query and passes it down with the proposal so that operations that are dependent on `time.Now` (such as `AdjustTimeRange`, `validateAndReturnOIRUpsertParams` etc.) can be executed deterministically (across nodes / restarts / catchups). 

I clarified the `WithTimestamp` comment. What do you think ?
I also removed the `UTC()` since the `sqlstore` just uses `time.Now()`.

pkg/timestamp/timestamp.go
MariemBaccari: I agree, it's fixed now.

pkg/timestamp/timestamp.go
MariemBaccari: I removed the `Now` part. The call would look like: `timestamp.FromContext`, is that better or should we make it something like `timestamp.QueryTimestampFromContext` ? I'm afraid this option might be too verbose. 

The timestamp is captured by an HTTP middleware when the query arrives.

pkg/timestamp/timestamp.go
MariemBaccari: Actually, I moved the comment to `FromContext` since it explains how the timestamp is set. Let me know if that works for you.

pkg/raftstore/consensus/proposal.go
```diff
@@ -30,17 +30,16 @@ type Proposal struct {
 	ReadOnly bool `json:"read_only"`
 }
 
-func (c *Consensus) newProposal(ctx context.Context, requestType string, payload any, readOnly bool) (Proposal, error) {
-	// TODO - Fetch timestamp from context
-	value, err := json.Marshal(payload)
-	if err != nil {
-		return Proposal{}, stacktrace.Propagate(err, "failed to serialize proposal payload")
+func (c *Consensus) newProposal(ctx context.Context, requestType string, value []byte, readOnly bool) (Proposal, error) {
+	timestamp, err := timestamp.RequestTimestampFromContext(ctx)
```
mickmis: Check https://github.com/interuss/dss/pull/1535#discussion_r3683039842, once that is merged rebase and use the `MustXX`.

pkg/raftstore/consensus/proposal.go
```diff
@@ -30,17 +30,16 @@ type Proposal struct {
...
+	if err != nil || timestamp.IsZero() {
+		return Proposal{}, stacktrace.Propagate(err, "failed to get timestamp from context")
 	}
 
 	return Proposal{
 		ID:          uuid.NewString(),
 		NodeID:      c.nodeID,
-		Timestamp:   time.Now().UTC(),
+		Timestamp:   timestamp,
```
mickmis: `timestamp` is not UTC

pkg/raftstore/store.go
```diff
@@ -2,46 +2,109 @@ package raftstore
...
 	"github.com/interuss/dss/pkg/raftstore/consensus"
 	raftparams "github.com/interuss/dss/pkg/raftstore/params"
 	"github.com/interuss/dss/pkg/store"
+	"github.com/interuss/dss/pkg/timestamp"
 	"github.com/interuss/stacktrace"
 	"go.uber.org/zap"
 )
 
+type RaftRepo[R any] interface {
+	GetRepo() R
```
mickmis, line 16: Not sure I understand why `GetRepo()` is needed? This makes the struct return itself. Why not taking the struct directly in the first place?

pkg/raftstore/store.go
```diff
@@ -2,46 +2,109 @@ package raftstore
...
+	// Apply is called on every committed entry. The proposal must be applied atomically.
+	Apply(ctx context.Context, proposal consensus.Proposal) (any, error)
+	GetSnapshot() ([]byte, error)
+	RestoreFromSnapshot(data []byte) error
```
mickmis, line 28: Document expectations for implementers.

pkg/raftstore/store.go
```diff
@@ -2,46 +2,109 @@ package raftstore
...
 	"github.com/interuss/dss/pkg/store"
+	"github.com/interuss/dss/pkg/timestamp"
 	"github.com/interuss/stacktrace"
 	"go.uber.org/zap"
 )
 
+type RaftRepo[R any] interface {
+	GetRepo() R
+	// Apply is called on every committed entry. The proposal must be applied atomically.
+	Apply(ctx context.Context, proposal consensus.Proposal) (any, error)
```
mickmis, line 20: - document what is returned
- can we avoid the `any`?

pkg/raftstore/store.go
```diff
@@ -2,46 +2,109 @@ package raftstore
...
-	consensus *consensus.Consensus
+	logger *zap.Logger
+
+	raftRepo RaftRepo[R]
+	cancel   context.CancelFunc
+	registry map[string]store.OperationHandler[R]
+
+	Consensus *consensus.Consensus
+
+	wg sync.WaitGroup
```
mickmis: nit: a wait group is slightly overkill for that use case, a simple channel would be plenty enough to do that

pkg/raftstore/store.go
```diff
@@ -2,46 +2,109 @@ package raftstore
...
-// Interact returns a repository that can be used to query the store without proposing a Raft entry.
+// Interact returns the underlying Raft repo which, for every operation, will propose it to Raft and return the results.
 func (s *Store[R]) Interact(_ context.Context) (R, error) {
-	return s.newRepo(), nil
+	return s.raftRepo.GetRepo(), nil
 }
 
-// Close shuts down the consensus instance.
+// Close shuts down the consensus instance and processCommits loop.
 func (s *Store[R]) Close() error {
```
mickmis, line 89: IMO some things are not quite right with the handling of the life cycle in `Consensus` and `Store`.
- I don't think `Close` is ever actually called at service termination
- It should have a context of its own so that we can do some cleanups while the caller controls the timeout to do so
- `Consensus` should be passed this context, and not `context.Background`
- and I see that `Stop` does not even use the context
- and it has a `Once` which is a smell
- and it tries to perform its own cleanup with a `defer Stop` which does not sound right

In general, at the top-level we'd want to manage the life cycle of this store like we do for the HTTP server:
https://github.com/interuss/dss/blob/8bc5c1916d4976692fe730f0ce4b6475ec102809/cmds/core-service/main.go#L354-L398

But actually none of those are really addressable in this PR...
If you agree with the above, could you make sure this is tracked? (TODOs in code + issue)

pkg/raftstore/store.go
MariemBaccari: The `any` mirrors store.OperationHandler.Execute. The concrete type will be recovered using store.TransactWithResult by the handlers (such as `(s *Server) DeleteSubscription`). Before that, we don't have access to the concrete type since `Apply` is a single method that applies all types of operations. 
I don't see a way to avoid it, especially since the layers below also return an any (#1577) and there is no common type or interface to the result types. If you have a suggestion for this let me know :)

pkg/raftstore/store.go
MariemBaccari: I agree, I just opened #1610 and added the TODOs.

pkg/raftstore/store.go
MariemBaccari: I attempted to remove it in [this commit](https://github.com/interuss/dss/pull/1525/commits/605098da3740ad3a5e74139f86ebef941a20bbf8) but at the cost of a not-so-pretty type assertion `repo, ok := r.(R)`. Let me know what you think / if you have a better or more idiomatic alternative.

mickmis: OK I now see the issue, indeed the assertion is not nice, your previous solution is better.
Alternatively we could also have `Init` accept 2 arguments (`MemRepo` and `R`) but that's roughly equivalent, so up to you.

Actually it would have been nice to be able to do the following, however that's not legal in Go:
```go
type MemRepo[R any] interface {
	T
	GetSnapshot() ([]byte, error)
	RestoreFromSnapshot([]byte) error
	Checkpoint()
	Restore()
}

File: pkg/raftstore/store.go (hunk 9 above)
Comment (mickmis) [reply pkg/raftstore/store.go:20]: Right! It is a similar exchange we had about OperationHandler, actually. I feel like we could do something better to avoid loosing the typing in the middle of all that, yet I'm not managing to find how. So let's go as is.

File: pkg/raftstore/consensus/consensus.go [hunk 12]
```diff
@@ -110,7 +110,9 @@ func NewConsensus(ctx context.Context, logger *zap.Logger, connectParams params.
 	return consensus, nil
 }
 
+// TODO: ctx is currently ignored
```
mickmis: Link the issue created in those TODOs

mickmis (approved):
LGTM modulo comment + revert or implementation of alternative for `MemStore`

mickmis (approved):

Issue #1526, opened 2026-06-15
[raft] Internal storage choice


the-glu:
In the context of the Raft store implementation described in #1525, we need a way to store in-memory objects (such as subscriptions, operational intents, etc.).

A quick analysis revealed various solutions:

* Direct maps/objects
* [Go-memdb](https://github.com/hashicorp/go-memdb#go-memdb)
* [buntdb](https://github.com/tidwall/buntdb)
* [sqlite](https://sqlite.org/)

These solutions offer different tradeoffs: using direct maps means we need to manually maintain our own indexes/structures, whereas a "real" database allows us to use standard queries.

Performance is our primary concern, as we want to apply changes coming from Raft as quickly as possible, and the objects involved are relatively simple.

To evaluate this, a small prototype simulating subscriptions was developed:
```go
type Subscription struct {
	ID                string
	Manager           string
	Version           int64
	NotificationIndex int
	StartTime         time.Time
	EndTime           time.Time
	USSBaseURL        string
	Cells             []uint64
}

```

with the following operations:

```go
Add(s *Subscription) error
GetByID(id string) (*Subscription, error)
Update(s *Subscription) error
Delete(id string) error
SearchByCells(cells []uint64) ([]*Subscription, error)
IncrementByCells(cells []uint64) (int, error)
SearchExpired(now time.Time) ([]*Subscription, error)

```

Performance was tested with 100, 1,000, 10,000, 100,000, and 1,000,000 entries.

Results are:

<img width="2100" height="1540" alt="Image" src="https://github.com/user-attachments/assets/29e271da-1614-408a-b7b2-20767102c4d5" />
<img width="2100" height="1540" alt="Image" src="https://github.com/user-attachments/assets/5b5645f2-0f97-4ee6-aa50-650006a8edd9" />
<img width="1820" height="700" alt="Image" src="https://github.com/user-attachments/assets/3bb87a15-729b-4511-b40b-a30e29e016ea" />

The results show that:

* Simple memory maps are almost always faster and use less memory
* The only slower case is searching for expired entries, as `end_date` is not indexed - at least in this test

Based on this analysis, I propose using simple maps for Raft internal storage instead of the listed frameworks/libraries.

Note that this underlying storage can be changed at any time, and different implementations can even coexist within a DSS pool - this only concerns the way in-memory information is stored.

PR #1527, opened 2026-06-15
 [raft] Add base memstore 

the-glu:

It does add the base for the 'in memory' store to be used by raft as underlying storage.

Rationale for direct, in memory structure is in #1526 

In theory, this store could be used e.g. for testing as 'regular' storage with small tweaks (like implementation of transact and being thread-safe), but for now it's goal is only to be used by raft.

Follow-up PRs will come with actual implementation.

### DIFF
--- a/pkg/memstore/store.go
+++ b/pkg/memstore/store.go
@@ -1,0 +1,63 @@
+package memstore
+
+// Memstore is a special kind of store:
+// Store instances store data in memory. There is no persistent storage.
+// Store instances are a singleton.
+// Repository usage is not thread-safe.
+// It's used by raftstore for projected storage.
+
+import (
+	"context"
+	"sync"
+
+	dsserr "github.com/interuss/dss/pkg/errors"
+	"github.com/interuss/dss/pkg/logging"
+	"github.com/interuss/stacktrace"
+	"go.uber.org/zap"
+)
+
+type MemRepo[R any] interface {
+	GetRepo() R
+}
+
+type Store[R any] struct {
+	logger *zap.Logger
+
+	name    string
+	memRepo MemRepo[R]
+}
+
+var (
+	stores   = map[string]any{}
+	storesMu sync.Mutex
+)
+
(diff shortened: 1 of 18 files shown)


pkg/timestamp/timestamp.go
barroco, line 1: Split in another PR please

pkg/store/params/params.go
```diff
@@ -22,6 +23,7 @@ var (
 )
 
 func init() {
+	// NB: Memstore not available there on purpose.
```
barroco: please explain why

pkg/memstore/store.go
```diff
@@ -0,0 +1,65 @@
+package memstore
+
+// Memstore are a special kind of store:
```
barroco: `is`

pkg/memstore/store.go
```diff
@@ -0,0 +1,65 @@
...
+// Store instances store data in memory. There is no persistent storage.
+// Store instances are a singleton.
+// Repository usage is not thread-safe.
+//
+// As of now, they are made to be used by raftstorage.
+// Adaptations could be done to use them directly in the future.
```
barroco: Remove this line.

pkg/memstore/store.go
```diff
@@ -0,0 +1,65 @@
+package memstore
+
+// Memstore are a special kind of store:
+// Store instances store data in memory. There is no persistent storage.
+// Store instances are a singleton.
+// Repository usage is not thread-safe.
+//
+// As of now, they are made to be used by raftstorage.
```
barroco: Just state the current state.

pkg/timestamp/timestamp.go
the-glu: That already coming from another PR/commit

pkg/memstore/store.go
the-glu: Yes that the current state, what would you change?

barroco (approved):

MariemBaccari:
Implements #1702

PR #1528, opened 2026-06-15
[raft/aux] Add aux Raftstore

MariemBaccari:
Implements the aux Raftstore. 
Removes the registry as it is not needed by aux which only calls `Interact`.
Implements the aux repo which calls the consensus instance for each type of operation (except `GetDSSAirspaceRepresentationID` which is local). 

Implements #1700


pkg/aux_/store/raftstore/store.go
```diff
@@ -2,25 +2,63 @@ package raftstore
...
 	"github.com/interuss/dss/pkg/raftstore"
 	"github.com/interuss/dss/pkg/raftstore/consensus"
 	"github.com/interuss/stacktrace"
 	"go.uber.org/zap"
 )
 
+const storeID = "aux_"
+
+const (
+	reqSaveOwnMetadata = "save_own_metadata"
```
the-glu: should it be an enum ?

and just to be sure: no need to prefix it ?

MariemBaccari: Yes, will make it an enum thanks 👍 
Not sure what we would need to prefix it with ? Something to detect the readonly status ?

the-glu (approved):
95eb8336d9f5da9abd558cbbe3432925ccedd0c0 LGTM

pkg/aux_/store/raftstore/dss.go
```diff
@@ -2,25 +2,85 @@ package raftstore
...
 
-func (r *repo) GetDSSMetadata(_ context.Context) ([]*auxmodels.DSSMetadata, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "GetDSSMetadata not implemented for raftstore")
+func (r *repo) SaveOwnMetadata(ctx context.Context, locality string, publicEndpoint string) error {
+	if locality == "" {
+		return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Locality not set")
+	}
+	if publicEndpoint == "" {
+		return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Public endpoint not set")
+	}
```
mickmis: Deduplicate this validation (push it in the caller, or factor away in a separate function, or add a validation middleware, or else) (also from memstore, and also for other methods in here).

pkg/aux_/store/raftstore/dss.go
```diff
@@ -2,25 +2,85 @@ package raftstore
...
+		Locality:       locality,
+		PublicEndpoint: publicEndpoint,
+	}
+
+	buf, err := json.Marshal(payload)
+	if err != nil {
+		return stacktrace.Propagate(err, "failed to marshal payload")
+	}
+
+	_, err = r.consensus.HandleClientRequest(ctx, string(saveOwnMetadata), buf, false)
```
mickmis: (not for this PR) shouldn't `requestType` be a type exposed alongside `HandleClientRequest`?

pkg/aux_/store/raftstore/store.go
```diff
@@ -2,38 +2,92 @@ package raftstore
...
-func (r *repo) RestoreFromSnapshot([]byte) error {
-	return stacktrace.NewErrorWithCode(dsserr.NotImplemented, "not implemented yet")
+func (r *repo) RestoreFromSnapshot(data []byte) error {
+	return r.memStore.RestoreFromSnapshot(data)
 }
 
-func (r *repo) Apply(_ context.Context, _ consensus.Proposal) (any, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "not implemented yet")
+func (r *repo) Apply(ctx context.Context, proposal consensus.Proposal) (any, error) {
+	memRepo, err := r.memStore.Interact(ctx)
```
mickmis: nit: calling `Interact` here looks like just an unnecessary layer of indirection, since it just returns `GetRepo`: call `GetRepo()` directly?

mickmis: Actually that is making me wonder: do we actually need the memstore abstraction? Don't we only need the memrepo? AKAIU the memstore / memrepo will exclusively be invoked by the raftstore, and never directly by the API layer.
cc @the-glu

pkg/raftstore/params/params.go
```diff
@@ -150,3 +150,7 @@ func GetConnectParameters(subfolder string) (ConnectParameters, error) {
 	p.DataDir = filepath.Join(connectParameters.DataDir, subfolder)
 	return p, nil
 }
+
+func GetClusterID() uint64 {
```
mickmis: Why not calling `GetConnectParameters` instead?

pkg/aux_/store/raftstore/dss.go
```diff
@@ -2,25 +2,85 @@ package raftstore
...
+	"github.com/interuss/dss/pkg/timestamp"
 	"github.com/interuss/stacktrace"
 )
 
-// SaveOwnMetadata returns nil instead of dsserr.NotImplemented because it is needed to allow the server to startup.
-func (r *repo) SaveOwnMetadata(_ context.Context, locality string, publicEndpoint string) error {
-	return nil
+type saveOwnMetadataPayload struct {
+	Locality       string
+	PublicEndpoint string
```
mickmis: Out of clarity, please put the json tags on the fields. This makes it explicit this struct is used for that purpose, and also makes it not necessary to expose them.

pkg/aux_/store/raftstore/dss.go
```diff
@@ -2,25 +2,85 @@ package raftstore
...
 }
 
-func (r *repo) RecordHeartbeat(_ context.Context, heartbeat auxmodels.Heartbeat) error {
-	return stacktrace.NewErrorWithCode(dsserr.NotImplemented, "RecordHeartbeat not implemented for raftstore")
+func (r *repo) GetDSSMetadata(ctx context.Context) ([]*auxmodels.DSSMetadata, error) {
+	result, err := r.consensus.HandleClientRequest(ctx, string(getDSSMetadata), nil, true)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "failed to propose %s", getDSSMetadata)
+	}
+	if result == nil {
```
mickmis: Is this necessary? The `nil` returned from `HandleClientRequest` is actually typed.

pkg/aux_/store/raftstore/dss.go
```diff
@@ -2,25 +2,85 @@ package raftstore
...
+		return nil, nil
+	}
+
+	if _, ok := result.([]*auxmodels.DSSMetadata); !ok {
+		return nil, stacktrace.NewError("unexpected result type: %T", result)
```
mickmis: nit

```suggestion
	if res, ok := result.([]*auxmodels.DSSMetadata); ok {
		return res, nil
```

mickmis (approved):
Overall LGTM, but several minor comments

pkg/aux_/store/raftstore/store.go
the-glu: I don't think it cost much to keep the abstraction and to keep everything with the same "layering", so that was on purpose yes. 

We probably want at some point make the memstore testable directly, or test the raftstore with others kinds of store.

pkg/aux_/store/raftstore/dss.go
the-glu: Remarks in case of that change it, but we may want to use something else than JSON for future optimizations there.

MariemBaccari: I will add the json tags for clarity (they can be removed later). The fields still need to be exported for json marshalling tho. The tags only affect the name.

pkg/aux_/store/raftstore/dss.go
MariemBaccari: I agree. Since this is a small change and this is the first use of request types, I stil included it in this PR if that's okay with you.

MariemBaccari: I agree. Since this is a small change and our first use of requestType, I included it in this PR if that's okay with you.

pkg/aux_/store/raftstore/dss.go
```diff
@@ -2,25 +2,80 @@ package raftstore
...
+	}
+
+	if heartbeat.Timestamp == nil {
+		now := timestamp.MustGetRequestTimestamp(ctx)
+		heartbeat.Timestamp = &now
+	}
+
+	if heartbeat.NextHeartbeatExpectedBefore != nil && heartbeat.NextHeartbeatExpectedBefore.Before(*heartbeat.Timestamp) {
+		return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Cannot expect the timestamp of the next heartbeat before the timestamp of the new heartbeat")
+	}
```
mickmis: Dedup those too

mickmis (approved):
just one last thing

pkg/aux_/store/raftstore/dss.go
MariemBaccari: I just pushed that. We now use `now := timestamp.MustGetRequestTimestamp(ctx)` for everything including the sqlstore but this should be ok right ?

pkg/aux_/pool_participants.go
```diff
@@ -105,6 +106,25 @@ func (a *Server) PutDSSInstancesHeartbeat(ctx context.Context, req *restapi.PutD
...
+	if heartbeat.Locality == "" {
+		resp.Response400 = &restapi.ErrorResponse{Message: dsserr.Handle(ctx, stacktrace.NewErrorWithCode(dsserr.BadRequest, "Locality not set"))}
+		return resp
+	}
+	if heartbeat.Source == "" {
+		resp.Response400 = &restapi.ErrorResponse{Message: dsserr.Handle(ctx, stacktrace.NewErrorWithCode(dsserr.BadRequest, "Source not set"))}
+		return resp
+	}
+
+	if heartbeat.Timestamp == nil {
```
mickmis: Actually put that as an `else` of  `if req.Timestamp != nil`?

pkg/aux_/pool_participants.go
```diff
@@ -105,6 +106,25 @@ func (a *Server) PutDSSInstancesHeartbeat(ctx context.Context, req *restapi.PutD
...
+		now := timestamp.MustGetRequestTimestamp(ctx)
+		heartbeat.Timestamp = &now
+	}
+
+	if heartbeat.NextHeartbeatExpectedBefore != nil && heartbeat.NextHeartbeatExpectedBefore.Before(*heartbeat.Timestamp) {
```
mickmis: Actually put that into `if req.NextHeartbeatExpectedBefore != nil` and drop the first condition?

pkg/aux_/pool_participants.go
```diff
@@ -105,6 +106,25 @@ func (a *Server) PutDSSInstancesHeartbeat(ctx context.Context, req *restapi.PutD
 		heartbeat.NextHeartbeatExpectedBefore = &ts
 	}
 
+	if heartbeat.Locality == "" {
+		resp.Response400 = &restapi.ErrorResponse{Message: dsserr.Handle(ctx, stacktrace.NewErrorWithCode(dsserr.BadRequest, "Locality not set"))}
+		return resp
+	}
+	if heartbeat.Source == "" {
```
mickmis: Actually combine that with `if req.Source == nil`?

pkg/aux_/pool_participants.go
```diff
@@ -105,6 +106,25 @@ func (a *Server) PutDSSInstancesHeartbeat(ctx context.Context, req *restapi.PutD
 		heartbeat.NextHeartbeatExpectedBefore = &ts
 	}
 
+	if heartbeat.Locality == "" {
```
mickmis: This is actually set on the server, by configuration (`a.Locality`). And is now validated by `createAuxServer` to not be empty: this can be dropped.

mickmis (approved):
- I have more suggestions opportunities for simplifications / clarity
- In addition: by removing the validation from the store itself, we are now expecting the caller to provide correct parameters, otherwise unexpected behavior might occur. I think that is alright since anyway the store interface the the logic calling it are tightly coupled, however we should ensure a future dev knows what are valid parameters. I propose to add this to the documentation of the interface methods (https://github.com/interuss/dss/blob/2bf6aa7a410f76c590f78c70be324352c91a72a4/pkg/aux_/repos/repos.go).

mickmis (approved):

PR #1529, opened 2026-06-15
 [raft/memstore] Add aux memstore 

the-glu:

---

It adds the memory implementation for the aux store.

Snapshots, usage of now from context, and checkpoints are done in the next PRs.

### DIFF
--- a/pkg/aux_/store/memstore/dss.go
+++ b/pkg/aux_/store/memstore/dss.go
@@ -2,24 +2,85 @@ package memstore
 
 import (
 	"context"
+	"database/sql"
+	"time"
 
 	auxmodels "github.com/interuss/dss/pkg/aux_/models"
 	dsserr "github.com/interuss/dss/pkg/errors"
 	"github.com/interuss/stacktrace"
 )
 
-func (r *repo) SaveOwnMetadata(_ context.Context, locality string, publicEndpoint string) error {
-	return stacktrace.NewErrorWithCode(dsserr.NotImplemented, "SaveOwnMetadata not implemented for memstore")
+func (r *repo) SaveOwnMetadata(_ context.Context, loc string, publicEndpoint string) error {
+	if loc == "" {
+		return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Locality not set")
+	}
+	if publicEndpoint == "" {
+		return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Public endpoint not set")
+	}
+
+	r.participants[locality(loc)] = &participant{
+		publicEndpoint: publicEndpoint,
+		updatedAt:      time.Now(),
+	}
+	return nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/aux_/store/memstore/dss.go
```diff
@@ -2,24 +2,85 @@ package memstore
...
+	if locality == "" {
+		return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Locality not set")
+	}
+	if publicEndpoint == "" {
+		return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Public endpoint not set")
+	}
+
+	r.participants[locality] = &participant{
+		publicEndpoint: publicEndpoint,
+		updatedAt:      time.Now(),
```
MariemBaccari, line 23: You should now be able to access the proposal timestamp from the context using `timestamp.NowFromContext`. 
It is set in the generic raft store right before `Apply` is called. 
This is valid for rid as well.

the-glu: Yes, done in https://github.com/interuss/dss/pull/1535

pkg/aux_/store/memstore/store.go
```diff
@@ -2,17 +2,41 @@ package memstore
...
+	heartbeats map[heartbeatKey]auxmodels.Heartbeat
+}
+
+type participant struct {
+	publicEndpoint string
+	updatedAt      time.Time
+}
+
+type heartbeatKey struct {
+	locality string
```
mickmis: nit: consider adding a `type locality string` given it is also a key in the `participants` map.

pkg/aux_/store/memstore/dss.go
```diff
@@ -2,24 +2,85 @@ package memstore
...
+		return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Cannot expect the timestamp of the next heartbeat before the timestamp of the new heartbeat")
+	}
+
+	r.heartbeats[heartbeatKey{locality: heartbeat.Locality, source: heartbeat.Source}] = heartbeat
+	return nil
 }
 
 func (r *repo) GetDSSAirspaceRepresentationID(_ context.Context) (string, error) {
-	return "", stacktrace.NewErrorWithCode(dsserr.NotImplemented, "GetDSSAirspaceRepresentationID not implemented for memstore")
+	return "", stacktrace.NewErrorWithCode(dsserr.NotImplemented, "GetDSSAirspaceRepresentationID not implementable for memstore")
```
mickmis, line 85: Wouldn't an ID generated at init fulfill this role for the memstore?

mickmis (approved):
LGTM

pkg/aux_/store/memstore/dss.go
the-glu: We could have a random ID there, but would that make sense ? 

Raft have to implement it down form of ID (probably the cluster ID), so we don't really need it anyways.

MariemBaccari:
Implements #1702

PR #1530, opened 2026-06-15
[raft/memstore] Add snapshop capability to aux memstore

the-glu:

---

It adds `GetSnapshot` / `RestoreFromSnapshot` capability to memstore + aux implementation, to be used in raftstore for snapshots.

Encoding is done using `gob` (which should be faster than json for a lot of data) + `state` has been moved to a specific struct for direct encoding / decoding.

A version has been added to the snapshot. Incompatible changes that require "migration" of data will be handled there in the future, based on the version field.

### DIFF
--- a/pkg/aux_/store/memstore/store.go
+++ b/pkg/aux_/store/memstore/store.go
@@ -14,27 +14,33 @@ type locality string
 
 // repo is a full implementation of aux_.repos.Repository for memory-based storage.
 type repo struct {
-	// participants holds pool participants metadata, keyed by locality.
-	participants map[locality]*participant
-	// heartbeats holds the latest heartbeat per (locality, source).
-	heartbeats map[heartbeatKey]auxmodels.Heartbeat
+	state state
+}
+
+// state is the serializable in-memory state.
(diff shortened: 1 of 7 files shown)


pkg/aux_/store/memstore/store.go
```diff
@@ -12,27 +12,37 @@ import (
...
+// state is the serializable in-memory state.
+type state struct {
+	// Participants holds pool participants metadata, keyed by locality.
+	Participants map[string]*participant
+	// Heartbeats holds the latest heartbeat per (locality, source).
+	Heartbeats map[heartbeatKey]auxmodels.Heartbeat
 	// participants holds pool participants metadata, keyed by locality.
 	participants map[string]*participant
 	// heartbeats holds the latest heartbeat per (locality, source).
 	heartbeats map[heartbeatKey]auxmodels.Heartbeat
```
mickmis: Remove?

mickmis (approved):
LGTM modulo comment

MariemBaccari:
Implements #1702

PR #1532, opened 2026-06-16
Add TLS encryption

mahdi-orbitalize:
**What I did**:
- `consensus.go` was updated to enable client-side (handled by `rafthttp` if we provide a `transport.TLSInfo` when constructing a `rafthttp.transport`) and server-side (handled by extracting a `tls.Config` from the `transport.TLSInfo` and passing it to the server at construction *plus* using  `ListenAndServeTLS` instead of `ListenAndServe`)
- `gen-cert.sh` was introduced to make certificates (with the number of nodes as a parameter). Note that it's made to work for local setups (i.e. address is `localhost` or `127.0.0.1`). I also provided example certificates to save the trouble of generating them.
- `params.go` was updated to support passing the paths to the TLS certificates as a parameter. It also enforces the use of `HTTPS` instead of `HTTP` for the peers' URLs. Also, more tests were added to `params_test.go` to test the new features.
- `start_cluster.sh` was created to quickly test a 3-node setup with TLS and easily inspect logs.

**How I tested it:**
- Verified that logs where similar when trying with/without encryption (and that it fails when it's supposed to fail)

### DIFF
--- a/pkg/raftstore/consensus/consensus.go
+++ b/pkg/raftstore/consensus/consensus.go
@@ -6,8 +6,10 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/interuss/dss/pkg/logging"
+	params "github.com/interuss/dss/pkg/raftstore/params"
 	"github.com/interuss/stacktrace"
 	"go.etcd.io/etcd/client/pkg/v3/types"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/rafthttp"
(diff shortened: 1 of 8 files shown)


build/test-certs/raft-certs/ca.crt
```diff
@@ -0,0 +1,27 @@
+-----BEGIN CERTIFICATE-----
```
the-glu: Please make local certificates ignored and don't push them in the repository, as they may be sensitive, even for testing

pkg/raftstore/params/params.go
```diff
@@ -64,22 +53,68 @@ func (c ConnectParameters) PeerMap() (map[uint64]*url.URL, error) {
...
-	flag.StringVar(&connectParameters.Peers, "raft_peers", "", `comma-separated "nodeID=peerURL" pairs for all cluster members, including the current node, e.g. "1=http://node1:9021,2=http://node2:9021,3=http://node3:9021"`)
-	flag.StringVar(&connectParameters.DataDir, "raft_datadir", defaultDataDir, "directory for raft data (WAL segments and snapshots), required for restarts. These should not be deleted while the node is running or across restarts unless the node is being permanently shut down.")
-
-	flag.Uint64Var(&connectParameters.SnapshotCatchupEntries, "raft_snapshot_catchup_entries", defaultSnapshotCatchupEntries, "number of entries for a slow follower to catch-up after compacting")
+	flag.StringVar(&connectParameters.Peers, "raft_peers", "", `comma-separated "nodeID=peerURL" pairs for all cluster members, including the current node, e.g. "1=https://node1:9021,2=https://node2:9021,3=https://node3:9021"`)
+	flag.StringVar(&connectParameters.TLS, "raft_tls", "", `TLS certificates, format: ca=/path/to/ca.crt,cert=/path/to/node.crt,key=/path/to/node.key"`)
```
the-glu: can you make it 3 different flags ? that would be cleaner, easier to set and easier to prase

the-glu (changes requested):
Could you squash your changes in one commit, leaving only the one from the base PR and the one from your PR ? That would avoid mixing changes

pkg/raftstore/consensus/consensus.go
```diff
@@ -33,41 +32,50 @@ type Consensus struct {
...
+		return stacktrace.Propagate(err, "failed to parse TLS certificates")
+	}
+
+	tlsInfo := transport.TLSInfo{
+		TrustedCAFile: tlsCerts.CAFile,
+		CertFile:      tlsCerts.CertFile,
+		KeyFile:       tlsCerts.KeyFile,
+	}
+
+	cfg, err := tlsInfo.ServerConfig()
```
the-glu: can you make RequireAndVerifyClientCert explicit ?

It's hidden in there with a if on TrustedCAFile, however as it's very important I would enforce it there.

Side nit: cfg is only used bellow, I would keep the variable close to its usage

pkg/raftstore/params/params.go
```diff
@@ -64,22 +99,55 @@ func (c ConnectParameters) PeerMap() (map[uint64]*url.URL, error) {
...
+	flag.IntVar(&connectParameters.HeartbeatTick, "raft_heartbeat_tick", defaultHeartbeatTick,
+		"Ticks between leader heartbeats. Effective interval = raft_heartbeat_tick x raft_tick_interval. Must be less than raft_election_tick. Lower values detect follower loss faster but increase network traffic.")
+	flag.Uint64Var(&connectParameters.MaxSizePerMsg, "raft_max_size_per_msg", defaultMaxSizePerMsg, "Maximum bytes in a single Raft message sent to a peer. Smaller values lower the recovery cost but increase the number of messages sent, affecting throughput during replication.")
+	flag.IntVar(&connectParameters.MaxInflightMsgs, "raft_max_inflight_msgs", defaultMaxInflightMsgs, "Maximum number of in-flight Raft messages during optimistic replication phase. This should be set to avoid overflowing the transport layer sending buffer.")
+
+	flag.StringVar(&connectParameters.CAFile, "raft_tls_ca", "", `CA certificate, format: ca=/path/to/ca.crt`)
+	flag.StringVar(&connectParameters.CertFile, "raft_tls_crt", "", `node's certificate, format: ca=/path/to/node.crt`)
+	flag.StringVar(&connectParameters.KeyFile, "raft_tls_key", "", `node's private key, format: key=/path/to/node.key`)
```
the-glu: Description need an update (no `xy=`)

the-glu (changes requested):
Can you make it optional, with a 

`raft_insecure`

flag, default to false?

If set to true, don't require https + log a warning about beeing run in an insecure mode. If there are certificate, don't verify them (no CA check / trust blindly client certificates) . If there is no certificate, don't enable https at all.

It will be useful for local deployments and testing (to validate that test fail if mis configured).

Thanks

mahdi-orbitalize:
> Can you make it optional, with a
> 
> `raft_insecure`
> 
> flag, default to false?
> 
> If set to true, don't require https + log a warning about beeing run in an insecure mode. If there are certificate, don't verify them (no CA check / trust blindly client certificates) . If there is no certificate, don't enable https at all.
> 
> It will be useful for local deployments and testing (to validate that test fail if mis configured).
> 
> Thanks

Adressed in 64f3f9a90c1a204174e92dea90e04bf456726a52

Issue #1492, opened 2026-05-28
Datastore schema allows start/end time to be null, but not on the application layer


the-glu:
Initially reported in https://github.com/interuss/dss/pull/1469#discussion_r3312685459_

It seems there is a discrepancy between database constrains and application constrains.

Checking why and, is possible, enforcing that on the DB layer would improve robustness.

PR #1534, opened 2026-06-16
 [raft/memstore] Add rid memstore 

the-glu:

---

It adds rid memstore, with snapshots.

Tests (minus snapshots & transaction & initialization) are identical to SQL tests and form a big part of this PR. They can be checked locally with e.g. `diff pkg/rid/store/memstore/identification_service_area_test.go pkg/rid/store/sqlstore/identification_service_area_test.go`

'Strange' behavior with limits has been implemented as-is.

Cell index/optimization is reserved for a future PR to keep this one as simple as possible.

### DIFF
--- a/pkg/rid/store/memstore/identification_service_area.go
+++ b/pkg/rid/store/memstore/identification_service_area.go
@@ -2,6 +2,7 @@ package memstore
 
 import (
 	"context"
+	"slices"
 	"time"
 
 	"github.com/golang/geo/s2"
@@ -11,30 +12,116 @@ import (
 	"github.com/interuss/stacktrace"
 )
 
-func (r *repo) GetISA(_ context.Context, id dssmodels.ID, forUpdate bool) (*ridmodels.IdentificationServiceArea, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "GetISA not implemented for memstore")
+func isaRecordFromModel(isa *ridmodels.IdentificationServiceArea, updatedAt time.Time) *isaRecord {
+	return &isaRecord{
+		ID:         isa.ID,
+		URL:        isa.URL,
+		Owner:      isa.Owner,
+		Cells:      slices.Clone(isa.Cells),
+		StartTime:  clonePtr(isa.StartTime),
+		EndTime:    clonePtr(isa.EndTime),
+		AltitudeHi: clonePtr(isa.AltitudeHi),
+		AltitudeLo: clonePtr(isa.AltitudeLo),
+		Writer:     isa.Writer,
+		UpdatedAt:  updatedAt,
+	}
 }
 
-func (r *repo) DeleteISA(_ context.Context, isa *ridmodels.IdentificationServiceArea) (*ridmodels.IdentificationServiceArea, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "DeleteISA not implemented for memstore")
+// toModel rebuilds the ISA model
+func (rec *isaRecord) toModel() *ridmodels.IdentificationServiceArea {
+	return &ridmodels.IdentificationServiceArea{
+		ID:         rec.ID,
+		URL:        rec.URL,
+		Owner:      rec.Owner,
+		Cells:      slices.Clone(rec.Cells),
+		StartTime:  clonePtr(rec.StartTime),
+		EndTime:    clonePtr(rec.EndTime),
+		Version:    dssmodels.VersionFromTime(rec.UpdatedAt),
+		AltitudeHi: clonePtr(rec.AltitudeHi),
+		AltitudeLo: clonePtr(rec.AltitudeLo),
+		Writer:     rec.Writer,
+	}
+}
+
+func (r *repo) GetISA(_ context.Context, id dssmodels.ID, _ bool) (*ridmodels.IdentificationServiceArea, error) {
+	rec, ok := r.state.ISAs[id]
+	if !ok {
+		return nil, nil
+	}
+	return rec.toModel(), nil
 }
 
 func (r *repo) InsertISA(_ context.Context, isa *ridmodels.IdentificationServiceArea) (*ridmodels.IdentificationServiceArea, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "InsertISA not implemented for memstore")
+	if err := validateWriteData(isa.Cells, isa.StartTime, isa.EndTime); err != nil {
+		return nil, err
+	}
+	if _, ok := r.state.ISAs[isa.ID]; ok {
+		return nil, stacktrace.NewError("ISA with id %s already exists", isa.ID)
+	}
+	rec := isaRecordFromModel(isa, r.clock.Now())
+	r.state.ISAs[isa.ID] = rec
+	return rec.toModel(), nil
 }
 
 func (r *repo) UpdateISA(_ context.Context, isa *ridmodels.IdentificationServiceArea) (*ridmodels.IdentificationServiceArea, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "UpdateISA not implemented for memstore")
+	if err := validateWriteData(isa.Cells, isa.StartTime, isa.EndTime); err != nil {
+		return nil, err
+	}
+	prev, ok := findForWrite(r.state.ISAs, isa.ID, isa.Version)
+	if !ok {
+		return nil, nil
+	}
+	rec := isaRecordFromModel(isa, r.clock.Now())
+	rec.Owner = prev.Owner // It's not possible to update the owner of an ISA, this ensure it's to changed to a new value.
+	r.state.ISAs[isa.ID] = rec
+	return rec.toModel(), nil
+}
+
+func (r *repo) DeleteISA(_ context.Context, isa *ridmodels.IdentificationServiceArea) (*ridmodels.IdentificationServiceArea, error) {
+	rec, ok := findForWrite(r.state.ISAs, isa.ID, isa.Version)
+	if !ok {
+		return nil, nil
+	}
+	out := rec.toModel()
+	delete(r.state.ISAs, isa.ID)
+	return out, nil
 }
 
 func (r *repo) SearchISAs(_ context.Context, cells s2.CellUnion, earliest *time.Time, latest *time.Time) ([]*ridmodels.IdentificationServiceArea, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "SearchISAs not implemented for memstore")
+	if len(cells) == 0 {
+		return nil, stacktrace.NewErrorWithCode(dsserr.BadRequest, "Missing cell IDs for query")
+	}
+	if earliest == nil {
+		return nil, stacktrace.NewError("Earliest start time is missing")
+	}
+
+	want := cellSet(cells)
+	var out []*ridmodels.IdentificationServiceArea
+	for _, rec := range r.state.ISAs {
+		// ends_at >= earliest
+		if rec.EndTime == nil || rec.EndTime.Before(*earliest) { // TODO: Don't allow endtime to be null, see #1492
+			continue
+		}
+		// COALESCE(starts_at <= latest, true)
+		if latest != nil && rec.StartTime != nil && rec.StartTime.After(*latest) { // TODO: Don't allow startup to be null, see #1492
+			continue
+		}
+		if !overlaps(rec.Cells, want) {
+			continue
+		}
+		out = append(out, rec.toModel())
+
+		if len(out) > dssmodels.MaxResultLimit { // This mimics sqlstore behaviour, but it's not very good.
+			break
+		}
+	}
+	return out, nil
 }
 
 func (r *repo) ListExpiredISAs(_ context.Context, writer string, threshold time.Time) ([]*ridmodels.IdentificationServiceArea, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "ListExpiredISAs not implemented for memstore")
+	return listExpired[ridmodels.IdentificationServiceArea](r.state.ISAs, writer, threshold, dssmodels.MaxResultLimit), nil
 }
 
 func (r *repo) CountISAs(_ context.Context) (int64, error) {
-	return 0, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "CountISAs not implemented for memstore")
+	return int64(len(r.state.ISAs)), nil
 }
--- a/pkg/rid/store/memstore/store.go
+++ b/pkg/rid/store/memstore/store.go
@@ -2,17 +2,176 @@ package memstore
 
 import (
 	"context"
+	"time"
 
+	"github.com/golang/geo/s2"
+	"github.com/interuss/dss/pkg/geo"
 	"github.com/interuss/dss/pkg/memstore"
+	dssmodels "github.com/interuss/dss/pkg/models"
 	"github.com/interuss/dss/pkg/rid/repos"
+	"github.com/interuss/stacktrace"
+	"github.com/jonboulle/clockwork"
 	"go.uber.org/zap"
 )
 
 // repo is a full implementation of rid.repos.Repository for memory-based storage.
-type repo struct{}
+type repo struct {
+	state state
+	clock clockwork.Clock
+}
+
+// state is the serializable in-memory state.
+type state struct {
+	// ISAs holds the stored ISAs keyed by ID.
+	ISAs map[dssmodels.ID]*isaRecord
+	// Subscriptions holds the stored subscriptions keyed by ID.
+	Subscriptions map[dssmodels.ID]*subscriptionRecord
+}
+
+// isaRecord is the gob-serializable representation of an ISA. It intentionally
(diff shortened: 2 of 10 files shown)


pkg/rid/store/memstore/store.go
```diff
@@ -2,17 +2,137 @@ package memstore
...
+	return &v
+}
+
+func cloneFloat32(f *float32) *float32 {
+	if f == nil {
+		return nil
+	}
+	v := *f
+	return &v
+}
```
mickmis:
```suggestion
func clonePtr[T any](v *T) *T {
	if v == nil {
		return nil
	}
	return new(*v)
}
```

pkg/rid/store/memstore/store.go
```diff
@@ -2,17 +2,137 @@ package memstore
...
+	}
+	return false
+}
+
+func cloneCells(cells s2.CellUnion) s2.CellUnion {
+	if cells == nil {
+		return nil
+	}
+	return append(s2.CellUnion(nil), cells...)
+}
```
mickmis: Use `slices.Clone` instead?

pkg/rid/store/memstore/store.go
```diff
@@ -2,17 +2,137 @@ package memstore
...
+// overlaps reports whether any cell is present in set (equivalent to the SQL
+// "cells && $x" array-overlap operator).
+func overlaps(cells s2.CellUnion, set map[s2.CellID]struct{}) bool {
+	for _, c := range cells {
+		if _, ok := set[c]; ok {
+			return true
+		}
+	}
+	return false
+}
```
mickmis, line 115: Just use `slices.Contains` and invoke it for each element of the initial `cells` slice?

pkg/rid/store/memstore/identification_service_area.go
```diff
@@ -11,30 +11,141 @@ import (
...
+		// COALESCE(starts_at <= latest, true)
+		if latest != nil && rec.StartTime != nil && rec.StartTime.After(*latest) {
+			continue
+		}
+		if !overlaps(rec.Cells, want) {
+			continue
+		}
+		out = append(out, rec.toModel())
+
+		if len(out) > dssmodels.MaxResultLimit { // This miminc sqlstore behaviour, but it's not very good.
```
mickmis:
```suggestion
		if len(out) > dssmodels.MaxResultLimit { // This mimics sqlstore behaviour, but it's not very good.
```

pkg/rid/store/memstore/identification_service_area.go
```diff
@@ -11,30 +11,141 @@ import (
...
+	}
+	if earliest == nil {
+		return nil, stacktrace.NewError("Earliest start time is missing")
+	}
+
+	want := cellSet(cells)
+	var out []*ridmodels.IdentificationServiceArea
+	for _, rec := range r.state.ISAs {
+		// ends_at >= earliest
+		if rec.EndTime == nil || rec.EndTime.Before(*earliest) {
```
mickmis: Not sure why we allow nil end time in the first place in the SQL store ...
However if such a thing was to happen that's equivalent to +inf end time so I think we'd want:
```suggestion
		if rec.EndTime != nil && rec.EndTime.Before(*earliest) {
```

mickmis: + why no check on `earliest` while there is on `latest`?

pkg/rid/store/memstore/identification_service_area.go
```diff
@@ -11,30 +11,141 @@ import (
...
+	}
+	return out, nil
 }
 
 func (r *repo) ListExpiredISAs(_ context.Context, writer string, threshold time.Time) ([]*ridmodels.IdentificationServiceArea, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "ListExpiredISAs not implemented for memstore")
+	var out []*ridmodels.IdentificationServiceArea
+	for _, rec := range r.state.ISAs {
+		// ends_at <= threshold
+		if rec.EndTime == nil || rec.EndTime.After(threshold) {
```
mickmis: Same as previous comment

pkg/rid/store/memstore/identification_service_area.go
```diff
@@ -11,30 +11,141 @@ import (
...
+		if writer == "" {
+			if rec.Writer != "" {
+				continue
+			}
+		} else if rec.Writer != writer {
+			continue
+		}
+		out = append(out, rec.toModel())
+
+		if len(out) > dssmodels.MaxResultLimit { // This miminc sqlstore behaviour, but it's not very good.
```
mickmis: typo

pkg/rid/store/memstore/subscriptions.go
```diff
@@ -11,42 +11,203 @@ import (
...
+	}
+	prev, ok := r.state.Subscriptions[s.ID]
+	if !ok {
+		return nil, nil
+	}
+	if !dssmodels.VersionFromTime(prev.UpdatedAt).Matches(s.Version) {
+		return nil, nil
+	}
+	rec := subRecordFromModel(s, r.clock.Now())
+	rec.Owner = prev.Owner
```
mickmis: ?

pkg/rid/store/memstore/subscriptions.go
```diff
@@ -11,42 +11,203 @@ import (
...
+	for _, rec := range r.state.Subscriptions {
+		if rec.EndTime == nil || rec.EndTime.Before(now) {
+			continue
+		}
+		if !overlaps(rec.Cells, want) {
+			continue
+		}
+		out = append(out, rec.toModel())
+
+		if len(out) > dssmodels.MaxResultLimit { // This miminc sqlstore behaviour, but it's not very good.
```
mickmis: well I don't need to put any more comments about those do I :D

pkg/rid/store/memstore/subscriptions.go
```diff
@@ -11,42 +11,203 @@ import (
...
 func (r *repo) SearchSubscriptions(_ context.Context, cells s2.CellUnion) ([]*ridmodels.Subscription, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "SearchSubscriptions not implemented for memstore")
+	if len(cells) == 0 {
+		return nil, stacktrace.NewErrorWithCode(dsserr.BadRequest, "no location provided")
+	}
+	now := r.clock.Now()
+	want := cellSet(cells)
+	var out []*ridmodels.Subscription
+	for _, rec := range r.state.Subscriptions {
+		if rec.EndTime == nil || rec.EndTime.Before(now) {
```
mickmis: Same as for ISA

pkg/rid/store/memstore/subscriptions.go
```diff
@@ -11,42 +11,203 @@ import (
...
+		return nil, stacktrace.NewErrorWithCode(dsserr.BadRequest, "no location provided")
+	}
+	now := r.clock.Now()
+	want := cellSet(cells)
+	var out []*ridmodels.Subscription
+	for _, rec := range r.state.Subscriptions {
+		if rec.Owner != owner {
+			continue
+		}
+		if rec.EndTime == nil || rec.EndTime.Before(now) {
```
mickmis: Same here

pkg/rid/store/memstore/subscriptions.go
```diff
@@ -11,42 +11,203 @@ import (
...
+		out = append(out, rec.toModel())
+
+		if len(out) > dssmodels.MaxResultLimit { // This miminc sqlstore behaviour, but it's not very good.
+			break
+		}
+	}
+	return out, nil
 }
 
 func (r *repo) SearchSubscriptionsByOwner(_ context.Context, cells s2.CellUnion, owner dssmodels.Owner) ([]*ridmodels.Subscription, error) {
```
mickmis, line 120: Deduplicate with `SearchSubscription`

pkg/rid/store/memstore/subscriptions.go
```diff
@@ -11,42 +11,203 @@ import (
...
 
+// UpdateNotificationIdxsInCells increments the notification index for each
+// subscription in the given cells.
 func (r *repo) UpdateNotificationIdxsInCells(_ context.Context, cells s2.CellUnion) ([]*ridmodels.Subscription, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "UpdateNotificationIdxsInCells not implemented for memstore")
+	now := r.clock.Now()
+	want := cellSet(cells)
+	var out []*ridmodels.Subscription
+	for _, rec := range r.state.Subscriptions {
+		if rec.EndTime == nil || rec.EndTime.Before(now) {
```
mickmis: Same here

pkg/rid/store/memstore/subscriptions.go
```diff
@@ -11,42 +11,203 @@ import (
...
+		if len(out) > dssmodels.MaxResultLimit { // This miminc sqlstore behaviour, but it's not very good.
+			break
+		}
+	}
+	return out, nil
 }
 
+// UpdateNotificationIdxsInCells increments the notification index for each
+// subscription in the given cells.
 func (r *repo) UpdateNotificationIdxsInCells(_ context.Context, cells s2.CellUnion) ([]*ridmodels.Subscription, error) {
```
mickmis, line 141: Deduplicate with `SearchSubscription`

pkg/rid/store/memstore/subscriptions.go
```diff
@@ -11,42 +11,203 @@ import (
...
+		if !overlaps(rec.Cells, want) {
+			continue
+		}
+		rec.NotificationIndex++
+		out = append(out, rec.toModel())
+	}
+	return out, nil
 }
 
 func (r *repo) MaxSubscriptionCountInCellsByOwner(_ context.Context, cells s2.CellUnion, owner dssmodels.Owner) (int, error) {
```
mickmis, line 150: Deduplicate with `SearchSubscriptionsByOwner`

pkg/rid/store/memstore/subscriptions.go
```diff
@@ -11,42 +11,203 @@ import (
...
-	return 0, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "MaxSubscriptionCountInCellsByOwner not implemented for memstore")
+	now := r.clock.Now()
+	want := cellSet(cells)
+	counts := make(map[s2.CellID]int, len(cells))
+	for _, rec := range r.state.Subscriptions {
+		if rec.Owner != owner {
+			continue
+		}
+		if rec.EndTime == nil || rec.EndTime.Before(now) {
```
mickmis: Same here

pkg/rid/store/memstore/subscriptions.go
```diff
@@ -11,42 +11,203 @@ import (
...
+			continue
+		}
+		for _, c := range rec.Cells {
+			if _, ok := want[c]; ok {
+				counts[c]++
+			}
+		}
+	}
+	best := 0
+	for _, n := range counts {
```
mickmis: Use slices.Max?

pkg/rid/store/memstore/subscriptions.go
```diff
@@ -11,42 +11,203 @@ import (
...
+	}
+	return best, nil
 }
 
 func (r *repo) ListExpiredSubscriptions(_ context.Context, writer string, threshold time.Time) ([]*ridmodels.Subscription, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "ListExpiredSubscriptions not implemented for memstore")
+	var out []*ridmodels.Subscription
+	for _, rec := range r.state.Subscriptions {
+		// ends_at <= threshold
+		if rec.EndTime == nil || rec.EndTime.After(threshold) {
```
mickmis: Same here

mickmis (commented):
- None of those are concurrency-safe, that is expected, right? Maybe clarify that in the doc of the memstore.
- Is it planned that this is validated somehow by running the prober test suite against it in the CI? It not, it shoudl

the-glu:
> * None of those are concurrency-safe, that is expected, right? Maybe clarify that in the doc of the memstore.

This was stated in a previous PR and is in the code here: https://github.com/interuss/dss/blob/master/pkg/memstore/store.go#L3

Is that enough?

>     * Is it planned that this is validated somehow by running the prober test suite against it in the CI? It not, it shoudl

It's not expected for the memstore directly to works again the prober test suite: it mean to be internal only and not used directly via HTTPs call (especially since it's not concurrency-safe, on purpose).

Indirect validation (as underlying storage for the raft implementation) is however done in https://github.com/interuss/dss/pull/1538 (when everything is implemented, so tests pass).

mickmis:
> This was stated in a previous PR and is in the code here: > https://github.com/interuss/dss/blob/master/pkg/memstore/store.go#L3
> Is that enough?

OK was not aware of that. Sounds good to me, with just one thing: could you move on `Store` and/or `Init` so that it is caught by godoc.

> It's not expected for the memstore directly to works again the prober test suite: it mean to be internal only and not used directly via HTTPs call (especially since it's not concurrency-safe, on purpose).
> 
> Indirect validation (as underlying storage for the raft implementation) is however done in https://github.com/interuss/dss/pull/1538 (when everything is implemented, so tests pass).

:+1:

pkg/rid/store/memstore/store.go
```diff
@@ -2,17 +2,137 @@ package memstore
...
+	NotificationIndex int
+	Owner             dssmodels.Owner
+	Cells             s2.CellUnion
+	StartTime         *time.Time
+	EndTime           *time.Time
+	AltitudeHi        *float32
+	AltitudeLo        *float32
+	Writer            string
+	UpdatedAt         time.Time
+}
```
mickmis, line 60: Coming back to this after reviewing #1539: This is really just duplications of the corresponding models, except for `UpdatedAt`. Just use the corresponding models, and add a `Clone()` method to them? Embed or add as field the corresponding model.

pkg/rid/store/memstore/identification_service_area.go
the-glu: For the strange endtime behavior:

Yes, however there the idea was just to be sqlstore-equivalent, with it defaults.

Fixing this behaviour has been done in separates PR to limit change in small units: #1543 for RID.

It's also tracked there: https://github.com/interuss/dss/issues/1492

I propose to keep it wrong there and do the explicit change in it PR.

the-glu: > why no check on earliest while there is on latest?

I'm not sure I understand the question, witch check?

Idea there is to be equivalent to

```
		isasInCellsQuery = fmt.Sprintf(`
			SELECT
				%s
			FROM
				identification_service_areas
			WHERE
				ends_at >= $1
			AND
				COALESCE(starts_at <= $2, true)
			AND
				cells && $3
			LIMIT $4`, isaFields)
```

pkg/rid/store/memstore/subscriptions.go
the-glu: You cannot change the owner of a subscription, so this enforce it (subscriptionFields vs updateSubscriptionFields).

This is done silently as in sql part.

I added a comment.

pkg/rid/store/memstore/store.go
the-glu: I'm not sure it's a good idea: I think we want internal object to be independent from external ones and not mix them together, should we want to do changes. 

Here, `Version` don't make sense on the memstore layer, and we don't want it to be available / part of the code wrongly thing it's usable because it's present in the struct.

Removing possible nil value on dates as well ( #1543 ) wouldn't be possible on the memstore without changing the model, forcing changes on e.g. sqlstore or don't enforing it with typing.

pkg/rid/store/memstore/store.go
the-glu: That would make it O(n*m) instead of O(n+m) no ?

pkg/rid/store/memstore/identification_service_area.go
mickmis: > I'm not sure I understand the question, witch check?

Nevermind, the check is actually done above before the loop.

mickmis: And FTR for the other thing: discussed offline and OK as is + addition of TODO

pkg/rid/store/memstore/subscriptions.go
mickmis: Ah I missed that the fields in sqlstore were not including the owner. Looks good.

pkg/rid/store/memstore/store.go
mickmis: Discussed offline: OK as is

pkg/rid/store/memstore/subscriptions.go
```diff
@@ -11,42 +14,160 @@ import (
...
+		return nil, nil
+	}
+	out := rec.toModel()
+	delete(r.state.Subscriptions, s.ID)
+	return out, nil
+}
+
+// liveSubscriptionsInCells yields the non-expired subscriptions touching cells,
+// optionally restricted to a single owner.
+func (r *repo) liveSubscriptionsInCells(cells s2.CellUnion, owner *dssmodels.Owner) iter.Seq[*subscriptionRecord] {
```
mickmis, line 95: nice

mickmis (approved):
LGTM

MariemBaccari:
Implements #1702

PR #1535, opened 2026-06-16
 [raft/memstore] Use now from context 

the-glu:

---

It uses the helper from timestamp to use now from context, if available, to ensure consistent time in the raft context.

### DIFF
--- a/pkg/aux_/store/memstore/dss.go
+++ b/pkg/aux_/store/memstore/dss.go
@@ -3,14 +3,14 @@ package memstore
 import (
 	"context"
 	"database/sql"
-	"time"
 
 	auxmodels "github.com/interuss/dss/pkg/aux_/models"
 	dsserr "github.com/interuss/dss/pkg/errors"
+	"github.com/interuss/dss/pkg/timestamp"
 	"github.com/interuss/stacktrace"
 )
 
-func (r *repo) SaveOwnMetadata(_ context.Context, loc string, publicEndpoint string) error {
+func (r *repo) SaveOwnMetadata(ctx context.Context, loc string, publicEndpoint string) error {
 	if loc == "" {
 		return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Locality not set")
 	}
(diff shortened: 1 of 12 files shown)


MariemBaccari (approved):
LGTM

pkg/aux_/store/memstore/dss.go
```diff
@@ -3,24 +3,30 @@ package memstore
...
+	now, err := timestamp.RequestTimestampFromContext(ctx)
+
+	if err != nil {
+		return err
+	}
+
 	r.state.Participants[locality] = &participant{
 		PublicEndpoint: publicEndpoint,
-		UpdatedAt:      time.Now().UTC(),
+		UpdatedAt:      now,
```
mickmis: `timestamp` package does not specify `.UTC()`.

pkg/aux_/store/memstore/dss.go
```diff
@@ -69,7 +75,12 @@ func (r *repo) RecordHeartbeat(_ context.Context, heartbeat auxmodels.Heartbeat)
 	}
 
 	if heartbeat.Timestamp == nil {
-		now := time.Now().UTC()
+
+		now, err := timestamp.RequestTimestampFromContext(ctx)
```
mickmis: `timestamp` package does not specify `.UTC()`.

mickmis (approved):
- if available, add linter guarding against use of `time.Now()`, at least in those directories as a start?

Otherwise LGTM modulo comments about `UTC`.

pkg/aux_/store/memstore/dss.go
```diff
@@ -3,24 +3,30 @@ package memstore
...
-func (r *repo) SaveOwnMetadata(_ context.Context, loc string, publicEndpoint string) error {
+func (r *repo) SaveOwnMetadata(ctx context.Context, loc string, publicEndpoint string) error {
 	if loc == "" {
 		return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Locality not set")
 	}
 	if publicEndpoint == "" {
 		return stacktrace.NewErrorWithCode(dsserr.BadRequest, "Public endpoint not set")
 	}
 
+	now, err := timestamp.RequestTimestampFromContext(ctx)
```
mickmis: Looking at the `timestamp` package, it looks like whenever we want to use it, not having it available would mean it is a programming error. WDYT about adding:
```go
func MustGetRequestTimestamp(ctx context) time.Time { ... }
```
Which panics if not present or invalid?
This would IMO improve readability.

mickmis (approved):
LGTM
I have a suggestion, LMK what you think of it and if you agree whether you would do it in this PR or a follow-up one.

pkg/aux_/store/memstore/dss.go
the-glu: yes I do agree that would be more readable, it felt wrong when merging the PR.

Will fix it there :)

mickmis (approved):
LGTM

MariemBaccari:
Implements #1702

PR #1536, opened 2026-06-16
[raft] Make raft runnable-locally

the-glu:
This PR follows #1691

It adds raft as a profile for `make start-locally`

tested with `COMPOSE_PROFILES=with-raft make start-locally`

### DIFF
--- a/build/dev/docker-compose_dss.yaml
+++ b/build/dev/docker-compose_dss.yaml
@@ -111,6 +111,7 @@ services:
       - $PWD/../test-certs:/var/test-certs:ro
       - $PWD/startup/core_service.sh:/startup/core_service.sh:ro
       - $PWD/startup/coverdata:/startup/coverdata:rw  # we will save coverage info here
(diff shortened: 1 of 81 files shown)


MariemBaccari (approved):

mickmis (commented):
IIUC this enables only running a single raft node? But with the default config (2 USS nodes), those will not be pooled?

the-glu:
> IIUC this enables only running a single raft node? But with the default config (2 USS nodes), those will not be pooled?

Hum I'm not 100% sure because this crossed when the default changed, but that on the monitoring side that you can have multiples USS, not here no? I don't see it in files there but on the other repo yes.

mickmis (approved):
> Hum I'm not 100% sure because this crossed when the default changed, but that on the monitoring side that you can have multiples USS, not here no? I don't see it in files there but on the other repo yes.

You are right I got mixed-up with the other deployment :)

MariemBaccari:
Implements #1703

PR #1538, opened 2026-06-16
 [raft] Add tests in CI 

the-glu:
This PR follows #1536.

Adds automatic CI tests using raft as a datastore.

Evict tests are not enabled yet, as evict is still a TODO for raft.


mickmis (approved):
LGTM, but consider setting up a job matrix to deduplicate CRDB-YBDB-raft.

MariemBaccari:
Implements #1703

PR #1539, opened 2026-06-16
[raft/memstore] Add Checkpoint/Restore

the-glu:

---

It adds a Checkpoint/Restore to every memstore, to be used by raft in `Transact()` operations, allowing it to quickly restore the memstore back to the start of a transaction should an error occur, without the need for complex revert operations.

### DIFF
--- a/pkg/aux_/store/memstore/store.go
+++ b/pkg/aux_/store/memstore/store.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"time"
 
-	auxmodels "github.com/interuss/dss/pkg/aux_/models"
 	"github.com/interuss/dss/pkg/aux_/repos"
 	"github.com/interuss/dss/pkg/memstore"
 	"go.uber.org/zap"
@@ -14,7 +13,8 @@ type locality string
 
 // repo is a full implementation of aux_.repos.Repository for memory-based storage.
 type repo struct {
-	state state
+	state      state
+	checkpoint state
 }
 
 // state is the serializable in-memory state.
@@ -22,7 +22,7 @@ type state struct {
 	// Participants holds pool participants metadata, keyed by locality.
 	Participants map[locality]*participant
 	// Heartbeats holds the latest heartbeat per (locality, source).
-	Heartbeats map[heartbeatKey]auxmodels.Heartbeat
+	Heartbeats map[heartbeatKey]*heartbeat
 }
 
 type participant struct {
@@ -35,12 +35,23 @@ type heartbeatKey struct {
 	Source   string
 }
 
+type heartbeat struct {
+	Timestamp                   *time.Time
+	NextHeartbeatExpectedBefore *time.Time
+	Reporter                    string
+}
+
 func newRepo() *repo {
+
+	state := state{
+		Participants: map[locality]*participant{},
+		Heartbeats:   map[heartbeatKey]*heartbeat{},
+	}
+
 	return &repo{
-		state: state{
-			Participants: map[locality]*participant{},
-			Heartbeats:   map[heartbeatKey]auxmodels.Heartbeat{},
-		}}
+		state:      state,
+		checkpoint: state.clone(),
+	}
 }
 
 func Init(ctx context.Context, logger *zap.Logger) (*memstore.Store[repos.Repository], error) {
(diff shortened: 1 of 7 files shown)


pkg/memstore/store.go
```diff
@@ -21,6 +21,8 @@ type MemRepo[R any] interface {
 	GetRepo() R
 	GetSnapshot() ([]byte, error)
 	RestoreFromSnapshot([]byte) error
+	Checkpoint() any
+	Restore(any) error
```
mickmis: Document expectations of methods for implementers.

mickmis: About signature: AFAIU the methods do not have to return the actual data, it can remain within the struct. Instead: 
- return an ID if multiple checkpoints must be handled, or better if possible
- do not return an ID and just have `Checkpoint()` and `Rollback()` (but that works only if we are guaranteed to have one checkpoint to handle at a given time, but it should be the case since there is no concurrency on the memstore?)

This also avoid dealing with the `any` type.

pkg/rid/store/memstore/store.go
```diff
@@ -134,3 +134,37 @@ func cloneFloat32(f *float32) *float32 {
 	v := *f
 	return &v
 }
+
+// clone returns a copy of s with independent maps and records. Cell slices and
+// time pointers are shared, as they are never mutated in place.
+func (s state) clone() state {
```
mickmis, line 204: - Even with the dereferencing trick for cloning the struct, it's still going to be shallow, so here e.g. the inner `cells` will not be cloned correctly because it is a slice, nor any of the pointers. Add to `isaRecord` and `subscriptionRecord` a `clone` method. See https://github.com/interuss/dss/pull/1534/changes/aa29f16c9c39b69a4d60df4e858bd1b27ac06c9d#r3672814423 too.
- nit: as a first step you could use `maps.Clone` and then iterate to deep clone the values

pkg/scd/store/memstore/store.go
```diff
@@ -16,3 +17,11 @@ func Init(ctx context.Context, logger *zap.Logger) (*memstore.Store[repos.Reposi
 }
 
 func (r *repo) GetRepo() repos.Repository { return r }
+
+func (r *repo) Checkpoint() any {
+	return nil
```
mickmis: panic instead?

pkg/aux_/store/memstore/store.go
```diff
@@ -50,3 +51,33 @@ func Init(ctx context.Context, logger *zap.Logger) (*memstore.Store[repos.Reposi
 }
 
 func (r *repo) GetRepo() repos.Repository { return r }
+
+// clone returns a copy of s with independent maps and participant records.
+func (s state) clone() state {
```
mickmis, line 83: Same as other comment: problematic shallow copy (heartbeat contains pointer). I suggest adding a clone method to all (even participants which technically don't need it so far).

mickmis (changes requested):

pkg/rid/store/memstore/store.go
the-glu: Yes. That "on purpose" as stated in the comment:

We never mutated those in place (they are inner objects) and we want to be as fast as possible, hence the partial clone.

That a bit of a trade off between speed and "correct clone | correct usage of internal data", but that probably acceptable no? Especially if it's tested (not saying that 100% the case yet however).

pkg/memstore/store.go
the-glu: Switched to an internal, single checkpoint.

pkg/rid/store/memstore/store.go
the-glu: Switched to full clones() everywhere. For some reasons Heartbeats struct in aux was global not like the reset, so it has been moved like the others to add a clone()

pkg/rid/store/memstore/store.go
```diff
@@ -60,10 +62,12 @@ type subscriptionRecord struct {
 func newRepo() *repo {
 	r := &repo{}
 	r.resetState()
+	r.Checkpoint()
```
mickmis: nit: conceptually shouldn't that be part of `resetState`?

mickmis (approved):
LGTM modulo CI

pkg/rid/store/memstore/store.go
the-glu: Yes

MariemBaccari:
Implements #1702

PR #1542, opened 2026-06-17
[raft/memstore] Add scd memstore

the-glu:

---

It adds SCD memstore.

The PR is quite big, as it's the whole SCD memstore. However, individual files can be checked separately.

There were no existing tests compared to RID, except for expired operational intents / subscriptions. Those two have been copied, and the rest is new.

### DIFF
--- a/pkg/scd/store/memstore/availability.go
+++ b/pkg/scd/store/memstore/availability.go
@@ -3,16 +3,36 @@ package memstore
 import (
 	"context"
 
-	dsserr "github.com/interuss/dss/pkg/errors"
 	dssmodels "github.com/interuss/dss/pkg/models"
 	scdmodels "github.com/interuss/dss/pkg/scd/models"
-	"github.com/interuss/stacktrace"
+	"github.com/interuss/dss/pkg/timestamp"
+	"github.com/jackc/pgx/v5"
 )
 
-func (r *repo) GetUssAvailability(_ context.Context, id dssmodels.Manager) (*scdmodels.UssAvailabilityStatus, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "GetUssAvailability not implemented for memstore")
+func (rec *availabilityRecord) toModel() *scdmodels.UssAvailabilityStatus {
+	return &scdmodels.UssAvailabilityStatus{
+		Uss:          rec.Uss,
+		Availability: rec.Availability,
+		Version:      scdmodels.NewOVNFromTime(rec.UpdatedAt, rec.Uss.String()),
+	}
 }
 
-func (r *repo) UpsertUssAvailability(_ context.Context, ussa *scdmodels.UssAvailabilityStatus) (*scdmodels.UssAvailabilityStatus, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "UpsertUssAvailability not implemented for memstore")
+func (r *repo) GetUssAvailability(_ context.Context, id dssmodels.Manager) (*scdmodels.UssAvailabilityStatus, error) {
+	rec, ok := r.state.Availabilities[id]
+	if !ok {
+		return nil, pgx.ErrNoRows // TODO: #1608
+	}
+	return rec.toModel(), nil
+}
+
+func (r *repo) UpsertUssAvailability(ctx context.Context, s *scdmodels.UssAvailabilityStatus) (*scdmodels.UssAvailabilityStatus, error) {
+	now := timestamp.MustGetRequestTimestamp(ctx)
+
+	rec := &availabilityRecord{
+		Uss:          s.Uss,
+		Availability: s.Availability,
+		UpdatedAt:    now,
+	}
+	r.state.Availabilities[s.Uss] = rec
+	return rec.toModel(), nil
 }
--- a/pkg/scd/store/memstore/availability_test.go
+++ b/pkg/scd/store/memstore/availability_test.go
@@ -1,0 +1,31 @@
+package memstore
+
+import (
+	"testing"
+
+	scdmodels "github.com/interuss/dss/pkg/scd/models"
+	"github.com/jackc/pgx/v5"
+	"github.com/stretchr/testify/require"
+)
+
+func TestUssAvailabilityUpsertGet(t *testing.T) {
+	ctx := writeCtx()
+	r := setUpStore(t)
+
+	got, err := r.UpsertUssAvailability(ctx, sampleAvailability())
+	require.NoError(t, err)
+	require.Equal(t, manager, got.Uss)
+	require.Equal(t, scdmodels.UssAvailabilityStateNormal, got.Availability)
+	require.Equal(t, scdmodels.OVN("HXjEfPAc0lkinCf0ejtSGiGPE4o2Qogm-iXAGPG-QNo_"), got.Version)
+
+	fetched, err := r.GetUssAvailability(ctx, manager)
+	require.NoError(t, err)
+	require.Equal(t, got.Version, fetched.Version)
+	require.Equal(t, scdmodels.UssAvailabilityStateNormal, fetched.Availability)
+}
+
+func TestGetUssAvailabilityMissingReturnsErrNoRows(t *testing.T) {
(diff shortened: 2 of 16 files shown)


pkg/scd/store/memstore/availability.go
```diff
@@ -3,16 +3,36 @@ package memstore
...
+		Version:      scdmodels.NewOVNFromTime(rec.UpdatedAt, rec.Uss.String()),
+	}
+}
+
+// GetUssAvailability implements scd.repos.UssAvailability.GetUssAvailability.
 func (r *repo) GetUssAvailability(_ context.Context, id dssmodels.Manager) (*scdmodels.UssAvailabilityStatus, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "GetUssAvailability not implemented for memstore")
+	rec, ok := r.state.Availabilities[id]
+	if !ok {
+		return nil, pgx.ErrNoRows
```
MariemBaccari: Why do we return `pgx.ErrNoRows` instead of empty or a custom error and add this check in the handler ? Having pgx leak in the memstore feels odd.

the-glu: That was to avoid changes in callers, since they use that to detect no row, but that should be improved yes. I propose to do that in a follow up PR.

the-glu: #1608

pkg/scd/store/memstore/store.go
```diff
@@ -2,25 +2,270 @@ package memstore
...
+	if start != nil && v4d.EndTime != nil && start.After(*v4d.EndTime) { // TODO: Don't allow startup to be null, see #1492
+		return false
+	}
+	if end != nil && v4d.StartTime != nil && end.Before(*v4d.StartTime) { // TODO: Don't allow endtime to be null, see #1492
+		return false
+	}
+	return true
+}
+
+func clonePtr[T any](v *T) *T {
```
the-glu: question: we should probably move common function to a utils package right? Or should every memstore kept it owns helpers ?

the-glu:
@mickmis That one should be ready for a review

pkg/scd/store/memstore/availability_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+func TestUssAvailabilityUpsertGet(t *testing.T) {
+	ctx := writeCtx()
+	r := setUpStore(t)
+
+	got, err := r.UpsertUssAvailability(ctx, sampleAvailability())
+	require.NoError(t, err)
+	require.Equal(t, manager, got.Uss)
+	require.Equal(t, scdmodels.UssAvailabilityStateNormal, got.Availability)
+	require.NotEmpty(t, got.Version)
```
mickmis: Check actual value? Hardcoding expectation here is fine IMO

pkg/scd/store/memstore/availability_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	fetched, err := r.GetUssAvailability(ctx, manager)
```
mickmis, line 21: Version not tested

pkg/scd/store/memstore/store.go
mickmis: Agreed, it could simply be in `pkg/memstore`

pkg/scd/store/memstore/constraints.go
```diff
@@ -2,29 +2,96 @@ package memstore
...
+	if !ok {
+		return nil, pgx.ErrNoRows // TODO: #1608
+	}
+	return rec.toModel(), nil
 }
 
-func (r *repo) UpsertConstraint(_ context.Context, constraint *scdmodels.Constraint) (*scdmodels.Constraint, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "UpsertConstraint not implemented for memstore")
+func (r *repo) UpsertConstraint(ctx context.Context, s *scdmodels.Constraint) (*scdmodels.Constraint, error) {
+	if _, err := dsssql.CellUnionToCellIdsWithValidation(s.Cells); err != nil {
```
mickmis, line 66: Just validate with `geo.ValidateCell` then?

pkg/scd/store/memstore/constraints_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+func TestConstraintUpsertGetDelete(t *testing.T) {
+	ctx := writeCtx()
+	r := setUpStore(t)
+
+	got, err := r.UpsertConstraint(ctx, sampleConstraint())
+	require.NoError(t, err)
+	require.Equal(t, constraintId, got.ID)
+	require.Equal(t, manager, got.Manager)
+	require.NotEmpty(t, got.OVN)
```
mickmis: Test actual value

pkg/scd/store/memstore/subscriptions.go
```diff
@@ -2,47 +2,148 @@ package memstore
...
 
 	"github.com/golang/geo/s2"
-	dsserr "github.com/interuss/dss/pkg/errors"
 	dssmodels "github.com/interuss/dss/pkg/models"
 	scdmodels "github.com/interuss/dss/pkg/scd/models"
+	"github.com/interuss/dss/pkg/timestamp"
 	"github.com/interuss/stacktrace"
 )
 
+func (rec *subscriptionRecord) toModel() *scdmodels.Subscription {
```
mickmis, line 17: Reviewing this I realize our DSS implementation completely ignores altitudes for subscriptions. Our model does contain altitude interval, but it is never actually used. I don't think it is actually in breach of the standard, but that's still surprising. A clarifying note here regarding why there is no altitude in the record model would make sense to avoid any future confusion. I will also clarify at the weekly whether that is intentional or not.

pkg/scd/store/memstore/subscriptions.go
```diff
@@ -2,47 +2,148 @@ package memstore
 
 import (
```
mickmis, line 3: Clarify somewhere this intentionally (and why) does not have the behavior of the time-based notification index

mickmis (approved):
LGTM modulo minor comments

pkg/scd/store/memstore/availability_test.go
the-glu: I changed, but it's a bit weak (changing computation mean changing it, witch is not great)

pkg/scd/store/memstore/availability_test.go
the-glu: Not sure about this one, it's tested two lignes bellow?

pkg/scd/store/memstore/constraints.go
the-glu: It's a small helper on the loop + convert, is it worth it to change it back?

pkg/scd/store/memstore/availability_test.go
mickmis: hum yes

pkg/scd/store/memstore/availability_test.go
mickmis: I think it's a good thing actually? IMO it is very acceptable, desirable even, to have to change the value here if we change how we compute the OVN, because it is part of the behavior of the function we are unit testing.

pkg/scd/store/memstore/constraints.go
mickmis: nit: I'd factor away the loop in `utils.go` and use it here. But is it worth it? Meh up to you.

mickmis (approved):

pkg/scd/store/memstore/availability_test.go
the-glu: Depend on what you want to test. I'm more to test "something is returned" and "it's the same thing that is returned later", not "the value is generated the way I expected because I froze time/random", that not really IMO the goal of the test. The requirement is to return a value with some characteristics, not a value we choose and enforce because of it's characteristics.

Should the version be slightly different and the function implementing change, changing all tests to conform to the new value (and probably by taking the test message to get the new one) don't add much.

(But IMO/nits ^^')

MariemBaccari:
Implements #1702

PR #1543, opened 2026-06-17
[memstore/rid] Don't allow nil value for start/end time

the-glu:
This PR follows #1538

It removes some legacy code in RID coming from sqlstore. Since ISA/Subscriptions cannot have nil start/end time, this removes this possibility in the memstore layer.

Tests have been adapted as well.

Contributes to #1492

### DIFF
--- a/pkg/rid/store/memstore/store.go
+++ b/pkg/rid/store/memstore/store.go
@@ -1,0 +1,18 @@
+package memstore
+
+import (
+	"context"
+
+	"github.com/interuss/dss/pkg/memstore"
+	"github.com/interuss/dss/pkg/rid/repos"
+	"go.uber.org/zap"
+)
+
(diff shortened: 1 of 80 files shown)


pkg/rid/store/memstore/store.go
```diff
@@ -119,11 +125,8 @@ func cloneCells(cells s2.CellUnion) s2.CellUnion {
 	return append(s2.CellUnion(nil), cells...)
 }
 
-func cloneTime(t *time.Time) *time.Time {
-	if t == nil {
-		return nil
-	}
-	v := *t
+func timePtr(t time.Time) *time.Time {
```
mickmis, line 128: Since 1.26 this should not be required anymore, just calling `new(t)` should be doing just that.

mickmis (approved):
Please track in #1492 for which stores this is implemented.
LGTM, with one minor improvement.

PR #1550, opened 2026-06-19
[Raft] Use separate consensus instances per store

MariemBaccari:
Following [this discussion](https://github.com/interuss/dss/pull/1474#discussion_r3437139283), this PR implements the use of separate consensus instances per store. This works better than the alternative of using a single consensus group for all services since some can be disabled. This also distributes the necessary leadership work across potentially different nodes.

Implements #1699

### DIFF
--- a/pkg/aux_/store/raftstore/params/params.go
+++ b/pkg/aux_/store/raftstore/params/params.go
@@ -1,0 +1,29 @@
+package params
+
+import (
+	"flag"
+
+	raftparams "github.com/interuss/dss/pkg/raftstore/params"
+	"github.com/interuss/stacktrace"
+)
+
+const peersFlag = "aux_raft_peers"
+
+var peers string
+
+func init() {
+	flag.StringVar(&peers, peersFlag, "", `Comma-separated "nodeID=peerURL" pairs for the aux store, e.g. "1=http://node1:9031,2=http://node2:9031,3=http://node3:9031"`)
+}
+
+func GetConnectParameters() (raftparams.ConnectParameters, error) {
+	if peers == "" {
+		return raftparams.ConnectParameters{}, stacktrace.NewError("--%s is required", peersFlag)
+	}
+
+	p, err := raftparams.GetConnectParameters("aux")
+	if err != nil {
+		return raftparams.ConnectParameters{}, err
+	}
+	p.Peers = peers
+	return p, nil
+}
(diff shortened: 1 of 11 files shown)


pkg/aux_/store/raftstore/params/params.go
```diff
@@ -0,0 +1,107 @@
...
+
+	raftparams "github.com/interuss/dss/pkg/raftstore/params"
+	"github.com/interuss/stacktrace"
+)
+
+const (
+	defaultRaftDataDir   = "raft_data_aux"
+	defaultRaftClusterID = 3
+
+	flagNodeID                  = "aux_raft_node_id"
```
the-glu: Do we need fully splited instances? Shouldn't we share a little bit some parameters?

Like node id / datadir especially, + ev. parameters teakings (everything starting at flagSnapshotCatchupEntries), shouldn't we just have common values?

(With datadir using subfolders)

Having to set everything * number of datastore seems tedious :/

barroco: Agree, at the moment, there is no use case where a DSS instance will be differently configured in a pool in terms of raft configuration.

MariemBaccari: Thanks for the feedback, I reverted these back to shared parameters that lead to splitted instances.

pkg/aux_/store/raftstore/params/params.go
```diff
@@ -0,0 +1,29 @@
...
+	raftparams "github.com/interuss/dss/pkg/raftstore/params"
+	"github.com/interuss/stacktrace"
+)
+
+const peersFlag = "aux_raft_peers"
+
+var peers string
+
+func init() {
+	flag.StringVar(&peers, peersFlag, "", `Comma-separated "nodeID=peerURL" pairs for the aux store, e.g. "1=http://node1:9031,2=http://node2:9031,3=http://node3:9031"`)
```
barroco, line 15: Is it possible to use a different path since it http based transport ? That would have the benefit to keep the same base url / port for services and simplify configuration

MariemBaccari: I gave this a try but it looks like `rafthttp` does not accept URLs with a path (such as `http://localhost:9013/aux`)

barroco: Ok, let's keep it on different ports and different peers for the moment. We can revisit this later if needed. Thanks

barroco (approved):

PR #1552, opened 2026-06-19
[rid] Extract polygon conversion from geo module

RustedBones:
`AreaToCellIDs` defined in `geo` module is specific to the `rid` for the `area` parameter conversion.

Define a `FromGeoPolygonString` in `rid` common api to convert the area to an internal polygon representation.

### DIFF
--- a/pkg/rid/models/api/common/conversions.go
+++ b/pkg/rid/models/api/common/conversions.go
@@ -1,0 +1,73 @@
+package common
+
+import (
+	"bufio"
+	"bytes"
+	"strconv"
+	"strings"
+
+	"github.com/interuss/dss/pkg/geo"
+	dssmodels "github.com/interuss/dss/pkg/models"
+	"github.com/interuss/stacktrace"
+)
+
+func splitAtComma(data []byte, atEOF bool) (int, []byte, error) {
+	if atEOF && len(data) == 0 {
+		return 0, nil, nil
+	}
+
+	if i := bytes.IndexByte(data, ','); i >= 0 {
+		return i + 1, data[:i], nil
+	}
+
+	if atEOF {
+		return len(data), data, nil
+	}
+
+	return 0, nil, nil
+}
+
+// AreaFromRest parses "area" in the format 'lat0,lon0,lat1,lon1,...'
+// and returns the resulting GeoPolygon, or else:
+// * ErrOddNumberOfCoordinatesInAreaString
+// * ErrNotEnoughPointsInPolygon
+// * ErrBadCoordSet
+//
+// TODO(tvoss):
+// * Agree and implement a maximum number of points in area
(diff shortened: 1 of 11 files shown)


pkg/rid/server/v1/isa_handler.go
```diff
@@ -240,12 +240,16 @@ func (s *Server) SearchIdentificationServiceAreas(ctx context.Context, req *rest
...
-			return restapi.SearchIdentificationServiceAreasResponseSet{Response413: &restapi.ErrorResponse{
-				Message: dsserr.Handle(ctx, stacktrace.Propagate(err, "Invalid area"))}}
-		}
+		return restapi.SearchIdentificationServiceAreasResponseSet{Response400: &restapi.ErrorResponse{
+			Message: dsserr.Handle(ctx, stacktrace.PropagateWithCode(err, dsserr.BadRequest, "Invalid area"))}}
+	}
+	cu, err := p.CalculateCovering()
+	if errors.Is(err, geo.ErrAreaTooLarge) {
+		return restapi.SearchIdentificationServiceAreasResponseSet{Response413: &restapi.ErrorResponse{
+			Message: dsserr.Handle(ctx, stacktrace.Propagate(err, "Invalid area"))}}
```
BenjaminPelletier: We can be more helpful here (same comment in other similar situations):


```suggestion
			Message: dsserr.Handle(ctx, stacktrace.Propagate(err, "Requested area is too large"))}}
```

BenjaminPelletier (approved):
LGTM after enhancing a few error messages (see comment)

pkg/rid/models/api/common/conversions.go
```diff
@@ -0,0 +1,73 @@
...
+
+// AreaFromRest parses "area" in the format 'lat0,lon0,lat1,lon1,...'
+// and returns the resulting GeoPolygon, or else:
+// * ErrOddNumberOfCoordinatesInAreaString
+// * ErrNotEnoughPointsInPolygon
+// * ErrBadCoordSet
+//
+// TODO(tvoss):
+// * Agree and implement a maximum number of points in area
+func FromGeoPolygonSring(area string) (*dssmodels.GeoPolygon, error) {
```
mickmis:
```suggestion
func FromGeoPolygonString(area string) (*dssmodels.GeoPolygon, error) {
```

pkg/rid/models/api/common/conversions.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	if atEOF {
+		return len(data), data, nil
+	}
+
+	return 0, nil, nil
+}
+
+// AreaFromRest parses "area" in the format 'lat0,lon0,lat1,lon1,...'
```
mickmis: doc comment not matching

Issue #1541, opened 2026-06-17
Add flag to eliminate notification index contention


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
When ASTM F3411 (and later F3548) were written, there was a desire to include a field in notification messages sent due to subscriptions to make the subscriber aware of the sequence of notifications.  That is, when USS1 changed the airspace, the DSS would acknowledge the change and also instruct USS1 to notify any subscribers (e.g., USS2 due to a pre-existing subscription) of the change they just made, including a notification_index value that USS1 would deliver to USS2 along with the notification.  The intent of this field was to somewhat reduce the time between a notification being lost or received out of order (and therefore USS2 not having current airspace information) and the USS detecting the information loss and proactively recovering.  However, this is merely a small optimization because USS2 will always discover the desynchronization next time they attempt to change the airspace and should generally be able to fairly easily recover at that point.

The problem with this mechanism is a sequential notification index necessarily needs to be synchronized with airspace changes, so a subscription effectively serializes all airspace activity within it.  Moreover, all subscriptions in the area need to be locked, so to change a small section of airspace, all of the airspace contained in the union of all subscriptions that intersect that small section needs to be locked until the airspace change is complete (so all of their notification indices can be incremented exactly one for that airspace change and the appropriate index can be returned from the DSS response to be included in the notification).  To restate differently: sequential notification_index forces maximum concurrency to 1 over an area that is the union of all subscriptions intersecting any particular airspace change.

[...]

PR #1558, opened 2026-06-23
[dss] Add flag to enable time-based notification index

the-glu:
This PR contribute to #1541 by adding a flag to use current time for notification indexes.

When active, this will also remove the explicit lock on subscriptions.

A few remarks:

- The field in the table is not removed and still fetched / inserted, it's just left to default value. This shouldn't have a meaningful impact on performance and avoid maintaining a double query system / schema for all queries.
- Subscription table still need to be read to fetch and return the list of endpoint to notify. This mean the potential for contention still exists, as in CRDB serializable mode, inserting a new subscription in a parallel query should make the transaction to be restarted.
- Please assess performance before using the flag for your use case
- Doesn't include changes on infrastructure, that will come in a follow-up PR.
- Passage of flag like that starts to be tedious (especially as there are others PR that also add more flags). I will propose in a follow up PR a shared object and update of the aux endpoint to be normalized. For now the option is not exposed.

### DIFF
--- a/cmds/core-service/main.go
+++ b/cmds/core-service/main.go
@@ -66,7 +66,8 @@ var (
 	keyRefreshTimeout = flag.Duration("key_refresh_timeout", 1*time.Minute, "Timeout for refreshing keys for JWT verification")
 	jwtAudiences      = flag.String("accepted_jwt_audiences", "", "comma-separated acceptable JWT `aud` claims")
 
-	scdGlobalLock = flag.Bool("enable_scd_global_lock", false, "Experimental: Use a global lock when working with SCD subscriptions. Reduce global throughput but improve throughput with lot of subscriptions in the same areas.")
+	scdGlobalLock                     = flag.Bool("enable_scd_global_lock", false, "Experimental: Use a global lock when working with SCD subscriptions. Reduce global throughput but improve throughput with lot of subscriptions in the same areas.")
+	timeBasedNotificationIndexEnabled = flag.Bool("enable_time_based_notification_index", false, "Use a time based notification index when working with RID and SCD subscriptions.")
 )
 
 func createKeyResolver() (auth.KeyResolver, error) {
@@ -112,7 +113,7 @@ func createAuxServer(ctx context.Context, locality string, publicEndpoint string
 
 func createRIDServers(ctx context.Context, locality string, logger *zap.Logger) (*rid_v1.Server, *rid_v2.Server, error) {
 
-	ridStore, err := rids.Init(ctx, logger, true)
+	ridStore, err := rids.Init(ctx, logger, true, *timeBasedNotificationIndexEnabled)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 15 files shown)


cmds/core-service/main.go
```diff
@@ -66,7 +66,8 @@ var (
 	keyRefreshTimeout = flag.Duration("key_refresh_timeout", 1*time.Minute, "Timeout for refreshing keys for JWT verification")
 	jwtAudiences      = flag.String("accepted_jwt_audiences", "", "comma-separated acceptable JWT `aud` claims")
 
-	scdGlobalLock = flag.Bool("enable_scd_global_lock", false, "Experimental: Use a global lock when working with SCD subscriptions. Reduce global throughput but improve throughput with lot of subscriptions in the same areas.")
+	scdGlobalLock              = flag.Bool("enable_scd_global_lock", false, "Experimental: Use a global lock when working with SCD subscriptions. Reduce global throughput but improve throughput with lot of subscriptions in the same areas.")
+	timeBasedNotificationIndex = flag.Bool("enable_time_based_notification_index", false, "Use a time based notification index when working with RID and SCD subscriptions.")
```
barroco:
```suggestion
	timeBasedNotificationIndex = flag.Bool("enable_time_based_notification_index", false, "Use a time-based notification index when working with RID and SCD subscriptions.")
```

pkg/scd/operational_intents_handler.go
```diff
@@ -103,9 +103,11 @@ func (a *Server) DeleteOperationalIntentReference(ctx context.Context, req *rest
 			subscriptionIds = append(subscriptionIds, *old.SubscriptionID)
 		}
 
-		err = r.LockSubscriptionsOnCells(ctx, old.Cells, subscriptionIds, old.StartTime, old.EndTime)
-		if err != nil {
-			return stacktrace.Propagate(err, "Unable to acquire lock")
+		if !a.TimeBasedNotificationIndex {
```
barroco, line 106: I just ran a test with 21 nodes and the baseline netem profile "delay 36ms 40ms 50% distribution paretonormal loss 0.25% 15%" without lock (including datastore_max_open_conns=30 as described in #1583):

Image

To validate that the lock had to be skipped, the following graph shows the difference with keeping the locking:

Screenshot 2026-06-24 at 16 12 54

Keeping the lock creates useless contention since the lock [was made to anticipate the locks aquisition](https://github.com/interuss/dss/pull/1004) required by the increment.

For completeness, this is the test when time-based notification index is disabled which is very similar to keeping the lock as expected:

Screenshot 2026-06-24 at 16 31 57


<details>

<summary>Test setup</summary>

**Local Environment**: 96 cores machine - 192GB RAM

```
export NUM_USS=21
export NUM_NODES=1
export DB_TYPE=crdb
export INTRA_USS_NETEM_CONF="delay 600us 40us 25% distribution normal loss 0.0005%"
export INTER_USS_NETEM_CONF="delay 36ms 40ms 50% distribution paretonormal loss 0.25% 15%"
<edit monitoring/build/dev/startup/core_service.sh to specify -datastore_max_open_conns=30>
```
Note that the cluster was initialized to minimize cold start under bad conditions and restarted with the NETEM configuration. Cockroach under such configuration struggles to serve the UI. This observation will be discussed in a separate issue.

**Test Runner**
```
AUTH_SPEC="DummyOAuth(http://localhost:8085/token,locust)" uv run --index https://pypi.org/simple locust -f ./monitoring/loadtest/locust_files/FlightsInSub.py -H http://localhost:8010 --uss-base-url http://testdummy.interuss.org/uss --cluster-count 11 --base-lat 34 --base-lng -118 --area-radius 15182 --max-flight-distance 10000
```

**Latency beween nodes**

Image

</details>

barroco (commented):
LGTM. Please fix the ci.

pkg/scd/operational_intents_handler.go
the-glu: Thanks for the test, CI should be ok now :)

barroco (approved):

PR #1561, opened 2026-06-24
[doc] Add page about latencies

the-glu:
This PR adds a documentation page about latency, showing how it can affect performance.

Various benchmarks done for performance issues show that a "better" deployment, if possible, will help, and that there are probably non-achievable thresholds of performance in some configurations: resources like CPU and RAM aren't the only factors at play.

It includes tests done to show how latency impacts performance, either artificially or with real-life deployments.


docs/architecture/latency.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Time to answer a request
+
+When a call is made to the DSS, the core service queries the database. The
+database is not a single node: it is a cluster spread across DSS participants.
+To stay strongly consistent, a write must reach a quorum of nodes before it is
+acknowledged, and a read may also need to contact other nodes depending on where
+the data lives.
+
+Each of these internal hops adds the round-trip latency between nodes. A single
```
barroco:
```suggestion
Each of these internal hops adds to the round-trip latency between nodes. A single
```

docs/architecture/latency.md
```diff
@@ -0,0 +1,253 @@
+# Latency
+
+## Introduction
+
```
mickmis:
```suggestion
```

docs/architecture/latency.md
```diff
@@ -0,0 +1,253 @@
...
+but have very low redundancy, for example if the rack loses power. On the
+opposite side of the options, servers spread throughout the world will offer
+maximum redundancy, for example against country-level issues, but will be very
+slow. The goal is a balance: geographically close enough for low latency,
+diverse enough for redundancy.
+
+Think and plan for the failures you want to handle, as a pool, and then spread
+nodes across failure domains accordingly for resilience.
+
+## Typical scenarios examples
```
mickmis: Consider moving this subsection into a subpage instead.

docs/architecture/latency.md
```diff
@@ -0,0 +1,253 @@
...
+once a synchronized operation is forced (adding a subscription), the gap
+explodes with distance. In US-West it's invisible, coast-to-coast q/s falls ~40%
+with 50th latencies around ~750ms, and across the Atlantic the system becomes
+nearly unusable: average q/s down to ~6 (−69%), latencies capped at 10000ms, and
+failures appearing only there (up to 0.5/s).
+
+This test was done on endpoints performing relatively simple queries. More
+complex ones, like SCD are even more affected.
+
+## Conclusion
```
mickmis, line 279: I don't think this belongs here. I suggest removing this subsection and move the content bringing value to the introduction and/or recommendations subsections.

docs/architecture/latency.md
```diff
@@ -0,0 +1,253 @@
...
+This matters for high-volume cases such as many subscriptions or large query
+results: the more data must be transferred and synchronized between distant
+nodes, the more latency limits how much can realistically be moved in a given
+time window. There is a practical ceiling on how much can be transferred before
+responsiveness degrades.
+
+![](../assets/latency_impact.png)
+/// caption
+**Simulated** throughput vs
+latency on a virtual link, practical throughput is even lower.
```
mickmis: Even if this is just an illustration, do put details on what this represents exactly? I.e. queries on DSS, no. of nodes, etc.

docs/architecture/latency.md
```diff
@@ -0,0 +1,253 @@
...
+### Example of latency vs performance
+
+The following graph shows, in a controlled environment, the impact of latency on
+simple requests.
+
+Tests have been done with DSS version v0.22.0, CockroachDB, 3 USS with one node,
+on a virtual machine with ample resources.
+
+Latency is injected with
+`tc qdisc add dev eth0 root netem delay Xms 2ms distribution paretonormal`, with
```
mickmis: Provide more context about what this means (a link to the doc is probably enough).

docs/architecture/latency.md
```diff
@@ -0,0 +1,253 @@
...
+evolve with time and even placement of your virtual machine could randomly
+impact latency and performances.
+
+All examples are run with version v0.22.0, one node per location, with virtual
+machines with ample resources to focus on showing latency variations.
+CockroachDB encryption was off to simplify tests.
+
+Locust instances are always located next to the DSS being tested, on the same
+machine. The test ran is
+`docker run -e AUTH_SPEC="DummyOAuth(http://172.17.0.1:8085/token,localhost)"  -p 8089:8089 -v .:/app/ interuss/monitoring-dev uv run locust -f loadtest/locust_files/ISA.py -H http://172.17.0.1 -u 10 --uss-base-url http://dss.localututm`.
```
mickmis: Put this in a block rather than inline for readability.

mickmis (approved):
LGTM modulo minor comments

docs/architecture/latency.md
the-glu: Fully removed, content is indeed already elsewhere

docs/architecture/latency.md
the-glu: The 'raw' throughput in Mbps, in term of data size (it's on the axis label), I'm not sure what details to add in the caption, what did you had in mind?

mickmis: A description of what was tested: what queries / test scenario was run on what kind of setup.

mickmis (approved):

docs/architecture/latency-examples.md
```diff
@@ -0,0 +1,148 @@
+# Latency: Typical scenarios examples
+
+This section showcases sample deployments, their latencies, and their results
+on a specific test to serve as real-life reference examples. The importance of
+latency is described in [this section](https://www.google.com/search?q=latency.md).
```
barroco: Link looks wrong.

docs/architecture/latency-examples.md
barroco, line 1: Edit request: Please focus on the latency impact and limitation it creates as well as emphasize on the test conditions which are stable and not optimized, which is irrelevant for this document

barroco (commented):
Please see comments.

docs/architecture/latency-examples.md
```diff
@@ -0,0 +1,193 @@
...
+
+Locust instances are always located next to the DSS being tested, on the same
+machine. The test ran is:
+
+```docker run -e AUTH_SPEC="DummyOAuth(http://172.17.0.1:8085/token,localhost)"  -p 8089:8089 -v .:/app/ interuss/monitoring-dev uv run locust -f loadtest/locust_files/ISA.py -H http://172.17.0.1 -u 10 --uss-base-url http://dss.localututm```.
+
+The test has been chosen to be light and to be able to run it across a wide
+number of configurations, with the possibility of adding one subscription to
+force 'synchronization' between nodes. Others, more complex tests (like
+FlightsInSubs which create one implicit subscription for each call) generate too
```
barroco: Add a link to the test.

docs/architecture/latency-examples.md
```diff
@@ -0,0 +1,193 @@
+# Latency: Typical scenarios examples
```
barroco: Looking at the new architecture, please move it back to the end of Latency so we keep the information together for the moment.

docs/architecture/latency.md
```diff
@@ -0,0 +1,101 @@
...
+
+![](../assets/latency_rid.png)
+/// caption
+Results of testing RID ISA calls with
+various latencies.
+///
+
+This is just an example of one call, but it shows that even with a simple
+operation, there is a limit on how fast queries can be processed, no matter
+optimizations performed.
```
barroco: I would remove the last part. Compression or batching may be options no ?

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+
+![](../assets/generated/latency_tta.png)
+/// caption
+A generic request. Each
+arrow adds cumulative latency
+///
+
+### Bandwidth
+
+Latency and bandwidth interact. With high latency, each operation (assuming no parallelism) takes longer
```
BenjaminPelletier, line 34: Could you help me understand why we would assume no parallelism?  Of course latency and bandwidth are effectively different expressions of the same quantity when there is no (or fixed) parallelism, but it seems like that is not a very interesting nor relevant case.  With parallelism (which represents real-world conditions), it seems like we could have any combination of {high and low} {latency and bandwidth} (for instance, carrier pigeons with micro SD cards would be extremely high latency but also high bandwidth).  I am confused why this section seems to say high latency necessarily leads to lower bandwidth/throughput -- what is the mechanism by which high latency necessarily lowers throughput?

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+to complete, so fewer operations finish per second, which lowers effective
+throughput even when raw bandwidth is available. Synchronization traffic between
+nodes also competes for that bandwidth.
+
+This matters for high-volume cases such as many subscriptions or large query
+results: the more data must be transferred and synchronized between distant
+nodes, the more latency limits how much can realistically be moved in a given
+time window. There is a practical ceiling on how much can be transferred before
```
BenjaminPelletier, line 42: Same question here as above: why does latency limit how much (how much what?) can be realistically moved in a given time window?  For instance, geostationary satellites (not Starlink) have massive latency, but also pretty massive bandwidth -- they can move a lot of data in a given time window even though they have very poor latency.

As written, it seems like this paragraph of explanation would also apply to Raftstore -- it is subject to the same latency as CRDB when deployed into the same substrate, yet Raftstore achieves much higher bandwidth.  So, it seems like the dominant factor in achievable bandwidth isn't latency.

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+responsiveness degrades.
+
+![](../assets/latency_impact.png)
+/// caption
+**Simulated** throughput vs
+latency on a virtual link of a generic TCP connexion captured with [iperf3](https://software.es.net/iperf/), practical throughput is even lower.
+///
```
BenjaminPelletier, line 49: Is this a graph of [the BD product](https://en.wikipedia.org/wiki/Bandwidth-delay_product) using the 65k vanilla TCP limit?  If so, is this limit actually relevant to our use case?  I.e., is 100 Mbps insufficient to achieve order of 100 QPS?  If not, what is the cause of the falloff?

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+
+Queries are done against all DSS, with 3 processes in parallel (for a total
+of 9) and at the same time to remove as many variations as possible. Notice
+there is still some expected variance, the goal here is to show global trends.
+
+![](../assets/latency_rid.png)
+/// caption
+Results of testing RID ISA calls with
+various latencies.
+///
```
BenjaminPelletier, line 76: The time to respond trend here makes perfect sense to me: when it takes longer to exchange information between internal nodes, of course the overall request will take longer to fulfill.  In fact, we can even impute an estimate for the number of node traversals each request takes: 45ms node latency increases request latency by about 2200ms, putting the number of node-edge traversals per request at around 50.

However, I don't understand why there should fundamentally be any decrease in throughput due to latency.  It certainly could be that queries interfere with each other and therefore have to wait, so contention + latency = decrease in throughput, but above we write that the queries were chosen not to create congestion issues which I take as intending to rule out this mechanism.  There is a BD limit for TCP communication, but it seems like lots of little hops between nodes rather than one big hop should rule this out as a mechanism as well.  What is our theory for why throughput on this graph drops with latency?

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+
+This is just an example of one call, but it shows that even with a simple
+operation, there is a incompressible limit on how fast queries can be processed.
```
BenjaminPelletier, line 79: I think we need to be precise here: what does "how fast queries can be processed" mean?  Certainly latency controls how fast **a query** can be processed (and therefore how fast queries can be processed), but I don't see the incompressibility of how fast a **batch** of queries can be processed due to latency.  If we mean the former:


```suggestion
operation, there is a incompressible limit on how fast a query can be processed.
```

...but I don't see how that is very interesting since the thing that is really mysterious here to me is how latency leads to an incompressible limit on bandwidth.  And, if latency actually leads to an incompressible limit on bandwidth, what is the mechanism and how to Raftstore overcome it?

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+
+    Results have to be interpreted relative to each other in various situations,
+    not as the performance to be expected from the DSS.
+
+    The performance baseline will be described in another document.
+
+Locust instances are always located next to the DSS being tested, on the same
+machine. The test ran is:
+
+```docker run -e AUTH_SPEC="DummyOAuth(http://172.17.0.1:8085/token,localhost)"  -p 8089:8089 -v .:/app/ interuss/monitoring-dev uv run locust -f loadtest/locust_files/ISA.py -H http://172.17.0.1 -u 10 --uss-base-url http://dss.localututm```.
```
BenjaminPelletier, line 100: nit: Why `monitoring-dev`?  Isn't this an actual monitoring tool rather than a tool for development of monitoring tools?

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+force 'synchronization' between nodes. Others, more complex tests (like
+FlightsInSubs which create one implicit subscription for each call) generate too
+many contentions with version v0.22.0 to be meaningful.
+
+### Constraints: identifying the latency impact with a stable environment
+
+These examples measure the impact of **latency alone**. Every environment
+parameter is held constant across the three scenarios; the only variable is
+geographic distance, and therefore inter-node latency. Any difference in the
+Any difference in the results is attributable to that latency as much as
```
BenjaminPelletier:
```suggestion
results is attributable to that latency as much as
```

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+practically possible.
+
+Held constant:
+
+* **DSS version**: v0.22.0 in every run.
+* **Topology**: one node per location, three nodes across two providers.
+* **Machine resources**: ample CPU and memory; compute is never the bottleneck.
+* **CockroachDB encryption**: off.
+* **Client placement**: Locust runs on the same machine as the DSS under test,
+  so client↔DSS latency is zero. Only *inter-node* latency varies.
```
BenjaminPelletier:
```suggestion
  so client↔DSS latency is negligible. Only *inter-node* latency varies.
```

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+* **Load profile**: the same light test, fixed at 10 users (`-u 10`), with no
+  parallelization: each user sends one request, waits for the full response,
+  then sends the next.
+* **Datastore connections**: left to default value (4).
+
+The serial load is what makes latency directly visible in throughput: each
+user's rate is capped by the round-trip time, so once latency dominates, total
+throughput falls as roughly `users / latency`. Read the q/s columns as a readout
+of latency, not of DSS capacity.
```
BenjaminPelletier, line 132: This makes a lot of sense to me: adding latency causes each individual query to take longer, but throughput is only affected if parallelism is capped.  It seems like (to a first approximation) throughput should be able to scale linearly with parallelism (regardless of latency) until the throughput limit is hit, and the throughput limit is entirely separate from latency.  This seems in contradiction to many places above, like the "Bandwidth" section.

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+
+Latency measured by CockroachDB is as follows:
+
+![](../assets/generated/latency_west.png)
+
+Results after 2 minutes are:
+
+| Node | q/s   | 50th latency | 95th latency |
+| ---- | ----- | ------------ | ------------ |
+| N1   | 19.35 | 2ms          | 11ms         |
```
BenjaminPelletier, line 148: I think this document is intended as a description of a consistent phenomenon/behavior rather than a specific test report, so we should limit the information displayed to that which illustrates the point and is expected to be reasonably reproducible under comparable conditions and procedure.  So, for instance, it seems like the QPS should be reported with zero decimals (19 q/s, 19 q/s, 18 q/s) -- otherwise, the reader may be confused looking for the semantic significance of the difference between 19.35 and 19.01.

Also, there seems to be a substantial difference in 95th percentile latency -- is this significant?  If so, let's narrate that significance.  If not, let's omit it to avoid distracting the reader from the significant information, or else add narration explaining why the difference isn't significant.

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+With one subscription added in the database, forcing a synchronized update:
+
+| Node | q/s   | 50th latency | 95th latency | Failures/s |
+| ---- | ----- | ------------ | ------------ | ---------- |
+| N1   | 19.1  | 2ms          | 17ms         | 0          |
+| N2   | 18.81 | 3ms          | 14ms         | 0          |
+| N3   | 17.53 | 16ms         | 120ms        | 0          |
+
+While this shows a slight decrease in performance (approximately 2 percent for
+q/s and 50th latency), the extra, synchronized step needed is almost invisible:
```
BenjaminPelletier, line 166: Do we think 2 percent for QPS and 50th percentile latency is a true difference that would persist at approximately that level if we were to repeat the exact same experiment?  If the resolving power of this experiment is that fine, why does N3 latency decrease by 6%?

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+| ---- | ---- | ------------ | ------------ | ---------- |
+| N1   | 17.6 | 2ms          | 92ms         | 0          |
+| N2   | 4.2  | 2200ms       | 2500ms       | 0          |
+| N3   | 12.3 | 54ms         | 100ms        | 0          |
+
+That's a 40% loss in performances, 24000% increased latency for the 50th
+percentile and more than 6000% of increase for the 95% percentile, compared to
+the same step in previous scenario.
+
+The impact is way more visible there - and it is pure latency, not capacity: on
```
BenjaminPelletier, line 204: This is confusing: the table shows a substantial drop in throughput, yet the narration is that the impact is pure latency.  I don't disagree with the explanation, but I think that means we are not presenting the data as clearly as practical.  If QPS and latency are directly linked through an arithmetic relationship, there is certainly no reason to show both.  If QPS and latency are sometimes linked through an arithmetic relationship and sometimes not, it seems like we should be able to come up with a different metric that is orthogonal and report that instead of one of the sometimes-linked metrics.

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+
+Latency measured by CockroachDB is as follows:
+
+![](../assets/generated/latency_europe.png)
+
+Results after 2 minutes are:
+
+| Node | q/s  | 50th latency | 95th latency |
+| ---- | ---- | ------------ | ------------ |
+| N1   | 18.4 | 2ms          | 180ms        |
```
BenjaminPelletier, line 222: Do you know how 2ms can be possible?  It seems like a database operation cannot be complete until a majority of the Raft group has adopted it, and the operation cannot have been adopted by any other node in 2ms when communication requires a minimum of 30ms.

docs/architecture/latency.md
```diff
@@ -0,0 +1,293 @@
...
+and more than 1000% of increase for the 95% percentile. Without a synchronized
+step, latency is still low enough that the client cadence, not latency, sets the
+rate - hence the small q/s loss.
+
+With one subscription added in the database, forcing a synchronized update:
+
+| Node | q/s  | 50th latency | 95th latency | Failures/s |
+| ---- | ---- | ------------ | ------------ | ---------- |
+| N1   | 17.6 | 2ms          | 92ms         | 0          |
+| N2   | 4.2  | 2200ms       | 2500ms       | 0          |
```
BenjaminPelletier, line 197: This seems extremely surprising; what mechanism would explain N2 latency increasing to 34x while N3 only increased to 1.5x?  It seems like that is the most significant feature of this data and would need to be explained before we could draw any other conclusion -- averaging 34x and 1.5x to get 24x seems to be ignoring whether the 1.5-34x range is reasonable (and therefore whether an average is appropriate).

BenjaminPelletier (commented):
Thanks for doing these experiments -- I think the results are going to be very important to define our system performance boundaries.

docs/architecture/latency.md
the-glu: Yes, there is always some parallelism and it can be increased up to certain limits, but the idea is to measure the impact in a fixed setting.

Also parallelism has it's own limits, it can be increased to a certain point, but not to infinity, and the same, relative graph will occur. (and practically that the Bandwidth-delay product you quoted bellow).

docs/architecture/latency.md
the-glu: I had issues running locust with the non-dev image, but maybe that something not actual anymore?

docs/architecture/latency.md
the-glu: > I think we need to be precise here: what does "how fast queries can be processed" mean? Certainly latency controls how fast **a query** can be processed (and therefore how fast queries can be processed), but I don't see the incompressibility of how fast a **batch** of queries can be processed due to latency. 

Yes, that more for a single query, but also no, because you cannot have infinie-sized batch of queries to have unlimited queries processing. Simple limits would be how much you can keep in memory, how much you receive from clients (and how long do you wait to regroup them), or even number of open connection that will at least be limited to at most 65k.

> ...but I don't see how that is very interesting since the thing that is really mysterious here to me is how latency leads to an incompressible limit on bandwidth.

Maybe the bandwidth part is a bit confusing: the section above was meant to show 'data transfer limits' and here I don't think 'bandwidth' in term of data / s is reverent.

> And, if latency actually leads to an incompressible limit on bandwidth, what is the mechanism and how to Raftstore overcome it?

The main difference is in the number of 'synchronization step'. In a very simplified way, raft only need one round trip exchange  per "operation|http query", whereas in cockroachdb do that per "sql query" and they can me multiple ones.

docs/architecture/latency.md
the-glu: But parallelism is capped no? I don't see how you can achieve infinite parallelism.

docs/architecture/latency.md
the-glu: > otherwise, the reader may be confused looking for the semantic significance of the difference between 19.35 and 19.01.

Yes ok, I will do a pass of rounding.

> Also, there seems to be a substantial difference in 95th percentile latency -- is this significant? If so, let's narrate that significance. If not, let's omit it to avoid distracting the reader from the significant information, or else add narration explaining why the difference isn't significant.

We should probably narrate it, but one important part of the 95th is to check if some queries timeout when it reach 10000ms.

docs/architecture/latency.md
the-glu: I think so yes.

For N3 specifically (and in others situation bellow):

N3 is more sensitive to more queries, as it need to synchronize with it peers in the other provider, where the leader (N1) is.

docs/architecture/latency.md
the-glu: Yes: there are two latency: the one between the nodes and the ones of request.

It need to be rewritten (specially since the number is the other latency in the same sentence).

docs/architecture/latency.md
the-glu: That a good question. N1 is definitely the leaded in that case, and I think the reasonable explanation is that it's doing optimizations / caching, knowing it is the leader and don't wait for adoption?

I can investigate more if you want, I never noticed that strange, but I think that consistent in general with what I saw.

docs/architecture/latency.md
the-glu: I'm wondering if I made a typo when recording the 2200ms (220ms?), because that would feet the *4 ratio between N2 and N3, so I simulated the same setup locally (with modified scripts - having the option at some point would be great ^^').

I'm getting the same kind of results (explainable by the fact that it's on my machine, latencies are simulated, etc..):

N1 at high Q/S (around 17), ~35ms for 50th latency, ~350ms for 95th
N2 is basically not being able to follow, with high oscillations, between 0 and 8 q/s, latencies going into timeouts.
N3 oscillating, but around 12 q/s as well. 55ms for 50th latency, between 500 and 1000ms for 95th


image

image

image

(Note: Graphs are cut to the end of the test - I refreshed them to be sure to have correct value but not directly after - leaving a big empty space)

Network matrix (N2/N3 swapped dues to initialization order but I stay consistent everywhere else):

image


When I tested (I lost the graph but it was not that bad), N2 was probably on the edge of being still functional, with high but still working latencies.

(That just a feeling because I don't have data, but I also suspect cockroach's range are spread differently, leading to worse performances, especially since now N1 have one RTT of latency) 

 ---

Now for the potential explanation: N2 has 2 time the latency to the leader (N1) than N3. I would think that it's making it less 'capable' of making it transactions passing, as they are longer. N1 mainly have to talk to it self and can process/apply transaction very quickly, whereas N2 has to wait for a reply and have a very higher penalty to retry them. Seeing that was probably just bellow the limit, I think the 2200ms is possible.

the-glu: Here is also the diff I used to simulate that: [forcedlat.patch](https://github.com/user-attachments/files/29794163/forcedlat.patch)

docs/architecture/latency.md
the-glu: For the BD: Yes, that basically the point of this section. We should probably link the page.

It's relevant, because depending on data moved, you will start to hit those limits, especially with raft.

In a 9 node setup, with payload side of 35Kb (observed with lot of OVNs), that 220Mbits/s to send in total to others nodes.

docs/architecture/latency.md
the-glu: > Same question here as above: why does latency limit how much (how much what?) can be realistically moved in a given time window? For instance, geostationary satellites (not Starlink) have massive latency, but also pretty massive bandwidth -- they can move a lot of data in a given time window even though they have very poor latency.

Dues to the Bandwidth-delay product quoted bellow. That not the same setup as starlink, we're using single machine, with tcp, somewhat limited buffers, etc. And that also to be seeing in a "fixed settings", even with higher parallelism, you will hit thoses kind of limits.
 
> As written, it seems like this paragraph of explanation would also apply to Raftstore -- it is subject to the same latency as CRDB when deployed into the same substrate, yet Raftstore achieves much higher bandwidth. So, it seems like the dominant factor in achievable bandwidth isn't latency.

That not something we measured, we measured q/s on the client side, not actual bandwidth they achieve on their internal links.

docs/architecture/latency.md
the-glu: >  It certainly could be that queries interfere with each other and therefore have to wait, so contention + latency = decrease in throughput, but above we write that the queries were chosen not to create congestion issues which I take as intending to rule out this mechanism. 

Yes, there are no contention. However, there are always some penalty for sync / data to be transferred around to get results and that have some impact.

> There is a BD limit for TCP communication, but it seems like lots of little hops between nodes rather than one big hop should rule this out as a mechanism as well. 

Directly in that case, yes, we can probably exclude DB.

> What is our theory for why throughput on this graph drops with latency?

On top of previous part (needing to get data), since we're in a "fix number of queries in //", there will always be limit dues to the fixed number of queries that can be processed at the time.

Assuming 9 processes doing queries, if one query individually take 10 time more, you will still get 10 time less q/s globally.

the-glu:
@BenjaminPelletier Thanks for the review, I tried to give more details/context, let me know if it's more clear; we can then make improvement to the document to make it better.

PR #1564, opened 2026-06-25
[raft/snapshot] Fix snapshot

MariemBaccari:
Makes the store reload a snapshot when restarting and makes the presence of a snapshot in `EntryCommit` explicit.


MariemBaccari:
Closing temporarily

PR #1566, opened 2026-06-25
[aux] Add a dedicated endpoint for global options

the-glu:
Following #1565 , this PR add a dedicated config endpoint for all future flags.


interfaces/aux_/aux_.yaml
```diff
@@ -95,12 +95,15 @@ components:
           items:
             type: string
 
-    SCDLockModeResponse:
+    GlobalOptionsResponse:
       type: object
       properties:
         global_lock:
```
barroco:
```suggestion
        scd_global_lock:
```
Since it is doesn't cover rid

barroco (approved):

PR #1567, opened 2026-06-25
[http] Limit body size to 4Mb

the-glu:
Limit body size early in the HTTP handler chain to 1Mb.

### DIFF
--- a/cmds/core-service/main.go
+++ b/cmds/core-service/main.go
@@ -340,6 +340,7 @@ func RunHTTPServer(ctx context.Context, ctxCanceler func(), address, locality st
 	handler = authorizer.TokenMiddleware(handler)
 	handler = http.TimeoutHandler(handler, *timeout, "request timeout")
 	handler = logging.HTTPMiddleware(logger, *dumpRequests, handler)
+	handler = maxBodySizeMiddleware(1<<22, handler) // 4 MB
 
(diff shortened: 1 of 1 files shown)


cmds/core-service/main.go
```diff
@@ -340,6 +340,7 @@ func RunHTTPServer(ctx context.Context, ctxCanceler func(), address, locality st
 	handler = authorizer.TokenMiddleware(handler)
 	handler = http.TimeoutHandler(handler, *timeout, "request timeout")
 	handler = logging.HTTPMiddleware(logger, *dumpRequests, handler)
+	handler = maxBodySizeMiddleware(1<<20, handler) // 1 MB
```
barroco: As discussed separately: 
ASTM F3548-21 limits vertex count to 10,000 and each vertex can be about 50 bytes: "lat": -123.45670000000001, "lng": 12.34500000000001 which is 500 KB -- we wouldn't want it much smaller to make sure technically-compliant-but-large request bodies are not falsely rejected.

barroco (approved):
As discussed separately:
ASTM F3548-21 limits vertex count to 10,000 and each vertex can be about 50 bytes: "lat": -123.45670000000001, "lng": 12.34500000000001 which is 500 KB -- we wouldn't want it much smaller to make sure technically-compliant-but-large request bodies are not falsely rejected.

PR #1569, opened 2026-06-25
[terraform] Increase GKE disk size

the-glu:
The base disk image of google kubernetes is too small, leading to random DiskPresureError and make it hard to deploy service.

This bump it by 10G to ensure some margin (only 2.5G with current version of OS/k8s is left.)

### DIFF
--- a/deploy/infrastructure/dependencies/terraform-google-kubernetes/cluster.tf
+++ b/deploy/infrastructure/dependencies/terraform-google-kubernetes/cluster.tf
@@ -26,7 +26,7 @@ resource "google_container_node_pool" "dss_pool" {
 
   node_config {
     machine_type = var.google_machine_type
-    disk_size_gb = 15 # Kubernetes PVC will create distinct disks.
+    disk_size_gb = 25 # Kubernetes PVC will create distinct disks.
 
     # TODO: Use non-default service account with IAM roles
     oauth_scopes = [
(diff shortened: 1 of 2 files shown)


deploy/infrastructure/dependencies/terraform-google-kubernetes/cluster.tf
```diff
@@ -26,7 +26,7 @@ resource "google_container_node_pool" "dss_pool" {
 
   node_config {
     machine_type = var.google_machine_type
-    disk_size_gb = 15 # Kubernetes PVC will create distinct disks.
+    disk_size_gb = 25 # Kubernetes PVC will create distinct disks.
```
barroco, line 29: This is a breaking change, please document migration instructions for existing clusters and add a note in the release notes.

What would you think about ignoring the change for existing cluster ?https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes

the-glu: Yes, done.

Notice that upgrade in place is working, I tested on a test cluster, no need for manual steps outside a terraform apply.

barroco (approved):

wing-utm-sharing-airspace:
We ran into this just today.

the-glu:
> We ran into this just today.

With this PR? Was the 25G not enough?

wing-utm-sharing-airspace:
Sorry, I should have been clearer. We ran into the problem you identified "leading to random DiskPresureError and make it hard to deploy service." Moving to 25GB stopped the random disk pressures and evictions!

PR #1571, opened 2026-06-25
[rid] Improve UpdateNotificationIdxsInCells by locking row in order

the-glu:
When trying to use the latest monitoring image version - which includes a tougher test on RID - Yugabyte is unable to pass the test.

This PR implements improve the UpdateNotificationIdxsInCells by first locking row ordered by PK, limiting deadlock. This improve performance and, combined with others fixes make it passing.

See #1570 as well.

Performance on Cockroach is hard to determine since we don't have load test on RID.

Using a modified version of RID that insert a subscription show no changes (single node, no latency):

image


barroco:
Measurement with 60ms latency.

v0.23.0-rc1
Screenshot 2026-06-25 at 18 54 36

This PR
image

We can observe an improvement in terms performance (QPS and response times).

barroco (approved):

PR #1573, opened 2026-06-25
[raft/config] Control config changes

MariemBaccari:
Adds a `raftctl` cli that can add / update and remove nodes from the raft cluster.

### DIFF
--- a/pkg/raftstore/admin/admin.go
+++ b/pkg/raftstore/admin/admin.go
@@ -1,0 +1,153 @@
+package admin
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"net/http"
+	"sync"
+
+	"github.com/interuss/dss/pkg/logging"
+	"github.com/interuss/stacktrace"
+	"go.etcd.io/raft/v3/raftpb"
+	"go.uber.org/zap"
+)
+
+var adminAddr string
+
+const raftAdminAddrFlag = "raft_admin_addr"
+
+func init() {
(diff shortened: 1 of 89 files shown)


pkg/raftstore/admin/admin.go
```diff
@@ -0,0 +1,153 @@
...
+	"go.etcd.io/raft/v3/raftpb"
+	"go.uber.org/zap"
+)
+
+var adminAddr string
+
+const raftAdminAddrFlag = "raft_admin_addr"
+
+func init() {
+	flag.StringVar(&adminAddr, raftAdminAddrFlag, "", "If set and the raft store backend is in use, determines the address and port for the raft admin endpoint (POST /admin/members), used to add/remove/update a node across all of the rid/scd/aux raft clusters in a single request. Disabled if empty.")
```
the-glu, line 23: It should be enabled by default no?

pkg/raftstore/admin/admin.go
```diff
@@ -0,0 +1,153 @@
...
+		if adminAddr == "" {
+			logger.Fatal(fmt.Sprintf("--%s is required", raftAdminAddrFlag))
+		}
+
+		adminServer := &AdminServer{
+			logger:   logging.WithValuesFromContext(ctx, logger),
+			registry: globalRegistry,
+		}
+
+		adminServer.httpServer = &http.Server{
```
the-glu, line 54: It should have a form of authentification no? Even with listen address restricted to 127.0.0.1 that not safe enough.

the-glu (commented):
Partial review

MariemBaccari:
Closing temporarily

PR #1574, opened 2026-06-26
[raft] Change transact signature

MariemBaccari:
Changes the `Transact` signature to support `Raftstore` parameters.

### DIFF
--- a/pkg/store/store.go
+++ b/pkg/store/store.go
@@ -7,6 +7,12 @@ import (
 	"github.com/interuss/stacktrace"
 )
 
+// Request carries the information needed by the Raftstore to handle a request.
+type Request struct {
+	RequestType string
+	Payload     any
+}
+
 // store.Store is the generic means to access and interact with any type of data backing the DSS
 // may ever use, by obtaining a means to perform R-specific (repo type) operations.
 type Store[R any] interface {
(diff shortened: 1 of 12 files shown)


pkg/store/store.go
```diff
@@ -16,7 +16,9 @@ type Store[R any] interface {
 	Interact(context.Context) (R, error)
 	// Attempt to apply the operations in f to the R Repo it is supplied.  All operations performed
 	// on the R Repo by f will be applied or rejected atomically.
-	Transact(ctx context.Context, f func(context.Context, R) error) error
+	// requestType and payload are used by the Raftstore to build the proposal.
+	// The returned any is the proposal result.
+	Transact(ctx context.Context, requestType string, payload any, f func(context.Context, R) error) (any, error)
```
barroco: As discussed, in order to avoid leaking into the f function, we have to pass the request to the transact method to allow the raft store to process it.
Would it be possible to group requestType and payload into a Request struct ?

MariemBaccari: Done 👌

pkg/store/store.go
```diff
@@ -16,7 +22,9 @@ type Store[R any] interface {
 	Interact(context.Context) (R, error)
 	// Attempt to apply the operations in f to the R Repo it is supplied.  All operations performed
 	// on the R Repo by f will be applied or rejected atomically.
```
barroco: Please update this description to scope the usage for sqlstore

MariemBaccari: I updated it to ignore the specific store details. What do you think ?

MariemBaccari:
Replaced by #1577

PR #1575, opened 2026-06-26
[raft/store] Add timestamp middleware

MariemBaccari:
Adds the timestamp middleware that will be used by the Raftstore to apply time-dependent operations deterministically.

Implements #1700

### DIFF
--- a/pkg/timestamp/timestamp.go
+++ b/pkg/timestamp/timestamp.go
@@ -1,0 +1,43 @@
+package timestamp
+
+import (
+	"context"
+	"net/http"
+	"time"
+
+	"github.com/interuss/stacktrace"
+)
+
+type timestampKey struct{}
+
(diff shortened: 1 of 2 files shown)


pkg/timestamp/timestamp.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+	"net/http"
+	"time"
+
+	"github.com/interuss/stacktrace"
+)
+
+type timestampKey struct{}
+
+// RequestTimestampFromContext returns the request timestamp from the context, or an error if not present or zero.
```
barroco: It is not going to return 0 correct ?

MariemBaccari: Exactly, if timestamp is 0 it returns an error.

barroco (approved):

PR #1581, opened 2026-06-30
[raft/aux] Pass down locality

MariemBaccari:
Passes down locality from main to the stores and through the Raft proposal. This is needed by the aux implementation.


MariemBaccari:
Not needed if aux keeps using `Interact`.

PR #1582, opened 2026-06-30
[raft/aux] Implement aux Raftstore

MariemBaccari:
- Moves the business logic of the request handlers to `pkg/aux_/actions`.
- The handlers now build a `store.ActionAdapter` then use `a.Store.TransactResult`.
- Implements the aux Raftstore.


MariemBaccari:
duplicate of #1528

PR #1584, opened 2026-07-01
[scd] Move scd conversions to scd package

RustedBones:
Move scd conversions from `pkg/models` to `pkg/scd`


mickmis (approved):
LGTM modulo solving conflicts

PR #1597, opened 2026-07-13
[raft/rid] Extract delete subscription

MariemBaccari:
Extracts the `DeleteSubscription` business logic into the registry so it can later be executed by the Raftstore. This will allow us to remove the `Application` layer.

Implements #1701


mickmis (approved):
(3fb09e3e09a7c92d0574c6bbc1d113491b210ec0) LGTM

PR #1600, opened 2026-07-14
[dss] Only search for operational intents if requested in PutSubscription

wing-utm-sharing-airspace:
Just a minor copy/paste to avoid one database call if a client did not request `notify_for_operational_intents`.


barroco (approved):
Thank you!

PR #1604, opened 2026-07-26
Change AWS storage class from gp2 to gp3

wing-utm-sharing-airspace:
`gp2` is not just slower, but also currently more expensive than `gp3`

### DIFF
--- a/docs/infrastructure/google-manual.md
+++ b/docs/infrastructure/google-manual.md
@@ -326,7 +326,7 @@ your chosen provider.
         provider has some default storage classes that should work:
           - Google Cloud: `standard`
           - Azure: `default`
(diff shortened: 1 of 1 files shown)


mickmis (approved):
LGTM if is for @the-glu

the-glu:
> LGTM if is for @the-glu

Yes, however they are a few extra references to gp2 in the code, can you do a replace all for existing values in examples / config + replace description in https://github.com/interuss/dss/blob/aae231fd93b23adf700dac5e4a7731ff8ae714e1/deploy/infrastructure/utils/definitions/aws_kubernetes_storage_class.tf#L10 and regenerate variable ?

PR #1606, opened 2026-07-29
Replace `gp2` with `gp3`

wing-utm-sharing-airspace:
`gp2` currently is slower and more expensive.


deploy/infrastructure/utils/variables.py
```diff
@@ -17,7 +17,7 @@
 GENERATED_VARIABLES_FILENAME = "variables.gen.tf"
 INTERNAL_VARIABLES_FILENAME = "variables_internal.tf"
 GENERATED_COMMENT = """
-# This file has been automatically generated by /deploy/infrastructure/utils/generate_terraform_variables.py.
+# This file has been automatically generated by /deploy/infrastructure/utils/variables.py.
```
the-glu: Why the change? Isen't `generate_terraform_variables` the script to ran?

deploy/infrastructure/utils/variables.py
```diff
@@ -163,7 +163,7 @@ def parse_definition(variable_name: str, tf_definition: str) -> Tuple[str, str,
         )
 
     declared_var_name = list(variables[0].keys())[0]
-    if declared_var_name != variable_name:
+    if declared_var_name.strip('"') != variable_name:
```
the-glu: Is this really correct? The same should be the same including spaces no?

deploy/infrastructure/modules/terraform-google-dss/TFVARS.gen.md
```diff
@@ -15,13 +15,16 @@ The following table describes the variables of this terraform module.
             <th>Description</th>
         </thead>
         <tbody><tr>
-                <td>app_hostname (<code>string</code>)</td>
-                <td><p>Fully-qualified domain name of your HTTPS Gateway ingress endpoint.</p>
-<p>Example: <code>dss.example.com</code></p>
+                <td>app_hostname (<code>rin</code>)</td>
+                <td><p>&quot;&lt;&lt;-EOT
```
the-glu: The diff is strange, why EOT have been added everywhere?

the-glu (commented):
gp2/gp3 changes looks good, I'm not sure about script generation changes and the diff. Maybe consider a separate PR to split concerns?

deploy/infrastructure/utils/variables.py
wing-utm-sharing-airspace: There doesn't seem to be a `generate_terraform_variables.py`. As in an explicit python script

deploy/infrastructure/utils/variables.py
wing-utm-sharing-airspace: So, when I ran it, I got an error that showed `declared_var_name` with quotes. Would it be possible for you to check?

deploy/infrastructure/utils/variables.py
wing-utm-sharing-airspace: What I mean is, I ran the variables.py script, which is what the generate_terraform_variables bash script runs, right?

the-glu: So what I always do is run `./generate_terraform_variables.sh` in the folder, who run variables.py. 

This will run it with explicit version of the package needed with the docker image.

deploy/infrastructure/utils/variables.py
the-glu: When running generation on your branch with `./generate_terraform_variables.sh`, it does seems to be working for me.

deploy/infrastructure/modules/terraform-google-dss/TFVARS.gen.md
the-glu: Following that, when I ran `./generate_terraform_variables.sh` locally, this seems to make it switch back to the master version:

image

deploy/infrastructure/utils/variables.py
wing-utm-sharing-airspace: I appreciate the explanation. I misunderstood the comment in the py file. I should have read the official documentation :)

deploy/infrastructure/utils/variables.py
wing-utm-sharing-airspace: Thank you for taking the time to verify. TIL how to generate the files correctly.

deploy/infrastructure/modules/terraform-google-dss/TFVARS.gen.md
wing-utm-sharing-airspace: Ran the tool as you instructed this time. Hopefully this is not an issue anymore.

the-glu (approved):
LGTM now, thanks!

deploy/infrastructure/utils/variables.py
the-glu: No worries. I checked since it probably mean latest version of the library introduced some change and created #1609 to upgrade it.

mickmis (approved):
LGTM with @the-glu approval

PR #1607, opened 2026-07-30
[tooling] Enable more go linters

the-glu:
Working on #1535, I noticed that not a lot of go linters where active, and some can be useful.

This PR activate a set of them, and do some minor fixes to pass liniting.

It's mostly unused code / variable and 'enforced' framework usage for tests


mickmis (approved):
LGTM

PR #1609, opened 2026-07-31
[infrasctructure/utils] Upgrade variable generator to latest versions

the-glu:
Upgrade variable generator to latest version of libraries with fixes.

Small change in generated file, but that legit (sub types are now properly displayed without ${})


mickmis (approved):
LGTM

Issue #482, opened 2021-02-26
Automatically configure CRDB replication


BenjaminPelletier:
CockroachDB stores data in ranges.  Each range is a Raft consensus group composed of a certain number of replicas, each replica on a different node.  The default replica count is [5 for system ranges](https://www.cockroachlabs.com/docs/stable/configure-replication-zones.html#create-a-replication-zone-for-a-system-range) (ranges that contain data about the CockroachDB database cluster) and [3 for data ranges](https://www.cockroachlabs.com/docs/stable/demo-replication-and-rebalancing.html#:~:text=CockroachDB%20replicates%20each%20range).  A range is functional if and only if a strict majority of its replicas are available ("quorum is met").  If a majority of any range's replicas permanently fail, that range is permanently lost unless a forensic recovery technique is applied to restore data from one of the minority replicas.

The default deployment configuration for a DSS pool is a minimum of 3 DSS instances, each DSS instance with 3 CRDB nodes (though [this may change](https://github.com/interuss/dss/issues/481)).  **We want to guarantee survival of the pool and all data in the pool when a minority of DSS instances are lost.**  The default configuration of 3 replicas that may be freely assigned to any node in the cluster does not achieve this objective as 2-3 of of those replicas may reside on the same DSS instance (thus causing a loss of quorum if that DSS instance goes down).

If we do not configure CRDB to ensure that every DSS instance receives a replica of every range, then we must [increase the number of replicas for all ranges](https://www.cockroachlabs.com/docs/stable/configure-replication-zones.html) (system and data) to 7.  Or, more generally, to survive the loss of a minority of DSS instances with N DSS instances, the number of replicas for all ranges must be set to 2 * 3 * floor(N / 2) + 1.  Alternately, we could [configure CRDB replication](https://www.cockroachlabs.com/docs/stable/configure-replication-zones.html) so that every DSS instance stores exactly one replica of each range.

barroco:
Changing to P3 since Yugabyte is now the recommended data store. Replication factor is properly set to the number of master nodes.

PR #1613, opened 2026-08-05
[docs] Add CRDB survivability configuration

mickmis:
This relates to https://github.com/interuss/monitoring/pull/1551 and *might* be enough to address #482 if we decide pursuing automatic configuration is not relevant anymore given the current CRDB situation.

### DIFF
--- a/docs/operations/pooling-crdb.md
+++ b/docs/operations/pooling-crdb.md
@@ -124,7 +124,8 @@ following those instructions.
   `--certs-dir` and `--ca-key` flags.
     - Do not specify `--insecure`
 - The ordering of the `--locality` flag keys must be the same across all
-  CockroachDB nodes in the cluster.
+  CockroachDB nodes in the cluster. See [CockroachDB Pool Survivability](survivability-crdb.md)
+  for details on setting up these values and configuring replication zone constraints.
 - All DSS instances in the same cluster must point their ntpd at the same NTP
   Servers.
   [CockroachDB recommends](https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html#considerations)
@@ -209,6 +210,11 @@ to scale down the Statefulset and that removes the last node first (ex:
 
     1. If there are unhealthy nodes please investigate and fix them so that the
        ranges can return to a healthy state
+
+1. Identify whether the replication zone constraints should be adapted
+   accordingly. This might be the case e.g. when all nodes of an USS are leaving
+   the pool. See [CockroachDB Pool Survivability](survivability-crdb.md) for
+   details on setting up these values.
 
 1. Identify the node id we intend to decommission from the previous commands
    then decommission them. The following command assumes that `cockroachdb-0` is
--- a/docs/operations/survivability-crdb.md
+++ b/docs/operations/survivability-crdb.md
@@ -15,7 +15,22 @@ To achieve this, we:
 
 ## 1. Setting Node Locality
 
-When starting each CockroachDB node, you must configure its `--locality` flag to specify which USS and node it represents. The locality must use first at least the `uss` key, and optionally second the `node` key.
+When starting each CockroachDB node, you must configure its `--locality` flag to specify which USS and node it represents. For the purpose of ensuring survivability of the cluster, it is important that all USSes align of the structure of their locality settings.
+We recommend that the locality use first at least the `uss` key, and optionally second the `node` key.
+Because CockroachDB will distribute range replicas using values of locality keys, the order of the keys defining the hierarchy to do so, if there is some existing locality set preventing this to be achieved, it is important that:
+- all USSes define a specific key whose value is specific to them (e.g. `uss=uss1`);
+- and that the number of total replicas set is covers at least the total number of combination of keys of to the USS key.
+
+Example: with three USSs each with two nodes define the following localities:
+- `region=east,uss=uss1,node=uss1_node1`
+- `region=west,uss=uss1,node=uss1_node2`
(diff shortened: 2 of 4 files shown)


docs/operations/pooling-crdb.md
the-glu, line 189: Should a comment in adding/leaving be added to point to the doc since user may want to update the value in those process ?

mickmis: Yep good point

docs/operations/survivability-crdb.md
```diff
@@ -0,0 +1,67 @@
...
+
+To achieve this, we:
+1. Start each CockroachDB node with a specific locality embedding its USS and node identifier.
+2. Configure replication zones using the CockroachDB `ALTER RANGE` SQL command to enforce placement constraints across the pool.
+
+---
+
+## 1. Setting Node Locality
+
+When starting each CockroachDB node, you must configure its `--locality` flag to specify which USS and node it represents. The locality must use first at least the `uss` key, and optionally second the `node` key.
```
barroco: InterUSS recommends those keys, however, users may already be using other keys (ie region, zone). 
Can you please add a note on what they should do  if it was the case ?

docs/operations/survivability-crdb.md
```diff
@@ -0,0 +1,67 @@
...
+
+```sql
+ALTER RANGE default CONFIGURE ZONE USING
+  num_replicas = 3,
+  constraints = '{"+uss=uss1": 1, "+uss=uss2": 1, "+uss=uss3": 1}';
+```
+
+### Explanation of the Parameters:
+* `num_replicas = 3`: Tells CockroachDB to keep 3 copies of each range.
+* `constraints`: A JSON object specifying per-replica constraints. Do note that specifying this might not always be necessary depending on your deployment.
```
barroco: Please, be more precise on the case when it would not be necessary.

docs/operations/survivability-crdb.md
mickmis:
```suggestion
When starting each CockroachDB node, you must configure its `--locality` flag to specify which USS and node it represents. For the purpose of ensuring survivability of the cluster, it is important that all USSes align of the structure of their locality settings.
We recommend that the locality use first at least the `uss` key, and optionally second the `node` key.
Because CockroachDB will distribute range replicas using values of locality keys, the order of the keys defining the hierarchy to do so, if there is some existing locality set preventing this to be achieved, it is important that:
- all USSes define a specific key whose value is specific to them (e.g. `uss=uss1`);
- and that the number of total replicas set is covers at least the total number of combination of keys of to the USS key.

Example: with three USSs each with two nodes define the following localities:
- `region=east,uss=uss1,node=uss1_node1`
- `region=west,uss=uss1,node=uss1_node2`
- `region=east,uss=uss2,node=uss2_node1`
- `region=west,uss=uss2,node=uss2_node2`
- `region=east,uss=uss3,node=uss3_node3`
- `region=west,uss=uss3,node=uss3_node3`
Then, the total number of replicas should be set to at least 6 (2 `region` * 3 `uss`).

If, for some reason specific to your deployment, this is not possible to achieve, you will need to configure specific range constraints on your deployment to ensure survivability. This is however not within the scope of this documentation.

File: docs/operations/survivability-crdb.md [hunk 3]
```diff
@@ -0,0 +1,67 @@
...
+
+## 2. Configuring Replication Constraints (`ALTER RANGE`)
+
+By default, CockroachDB automatically distributes replicas to optimize resource usage and load. To enforce the "one replica per USS" survivability rule, you must manually define replication zone constraints using the `ALTER RANGE default CONFIGURE ZONE` SQL command.
+
+The `default` range is the cluster-wide catch-all. Any database or table created within the DSS (including the RID and SCD tables) that does not have its own specific zone configuration will inherit these default placement constraints.
+
+### The `ALTER RANGE` SQL Statement
+
+To configure a 3-replica cluster where exactly one replica lives on each of the three USSs:
```
mickmis:
```suggestion
To configure a 3-replica cluster where exactly one replica lives on each of the three USSs, assuming locality is set on each node as described above:
```

docs/operations/survivability-crdb.md
```diff
@@ -0,0 +1,67 @@
...
+
+```sql
+ALTER RANGE default CONFIGURE ZONE USING
+  num_replicas = 3,
+  constraints = '{"+uss=uss1": 1, "+uss=uss2": 1, "+uss=uss3": 1}';
```
mickmis:
```suggestion
  num_replicas = 3;
```

docs/operations/survivability-crdb.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+### Explanation of the Parameters:
+* `num_replicas = 3`: Tells CockroachDB to keep 3 copies of each range.
+* `constraints`: A JSON object specifying per-replica constraints. Do note that specifying this might not always be necessary depending on your deployment.
+    * `"+uss=uss1": 1` tells CockroachDB that exactly one replica must be located on nodes matching the locality `uss=uss1`.
```
mickmis:
```suggestion

Do note that outside of the default case described in this documentation, you will need to adjust the above SQL query by configuring different ranges and/or setting specific constraints. This is however not within the scope of this documentation.
```

docs/operations/survivability-crdb.md
mickmis: Clarified the scope of the doc so that:
- it covers only the default case
- and what is important if they are outside the default case, however they are on their own in that case

PR #1616, opened 2026-08-07
[rid] Fix query when using enableTimeBasedNotificationIndex

### LINKED ISSUE #1615: Missing `FROM` in SQL query causes ISA creation to fail
**Describe the bug**
We rolled out v23-rc2 in our preprod environment (with `enableTimeBasedNotificationIndex: true`) and all `CreateIdentificationServiceArea` started failing with `column cells does not exist` . Turns out, CRDB was thinking that the query was valid even though it was missing a `FROM`

https://github.com/interuss/dss/commit/6d0a7ad39a46cb635574c7ce7f3fc5623a2ce167#diff-07ffdc6adf630459f054c69d5a55e2b2e7f4707c352d9411aefb9654c89fed6c

`
Caused by: Error in query: 
            SELECT
                id, owner, url, notification_index, cells, starts_at, ends_at, writer, updated_at
            subscriptions
            WHERE
                cells && $1
                AND ends_at >= $2
 --- at /app/pkg/rid/store/sqlstore/subscriptions.go:27 (repo.process) ---
Caused by: ERROR: column "cells" does not exist (SQLSTATE 42703)", "ts":"2026-08-07T03:00:14.476Z"}
`

**To Reproduce**
Turn on the `enableTimeBasedNotificationIndex` flag and try creating ISA

As soon as we turned off the flag, the errors vanished.

the-glu:
Fix #1615

### DIFF
--- a/pkg/rid/store/sqlstore/subscriptions.go
+++ b/pkg/rid/store/sqlstore/subscriptions.go
@@ -225,6 +225,7 @@ func (r *repo) UpdateNotificationIdxsInCells(ctx context.Context, cells s2.CellU
 		updateQuery = fmt.Sprintf(`
 			SELECT
                 %s
+            FROM
             subscriptions
 			WHERE
 				cells && $1


BenjaminPelletier:
I think this is good, but have we verified that it works?  Normally CI success would be good evidence that it works, but it seems like the root problem of #1615 is that our CI doesn't cover non-default flags, so have we positively verified that this change fixes the issue and doesn't break anything else?

the-glu:
Yes, via https://github.com/interuss/dss/pull/1618 this works (and that commit is needed to make the CI passing).

#1618 will also prevent such cases in the future.

(Notice this has been validated via the prober suite only, as the qualifier need changes to accept this option).

BenjaminPelletier (approved):

PR #1618, opened 2026-08-07
[tooling] Test special flags with prober & qualify

the-glu:
This switch to a job matrix, to test various combination of flags & datastore, ensuring full coverage of possibilities automatically.

Base job with coverage (crdb, no flags) is kept as is.

This PR require the follows PRs to be merged to pass:

* #1617 
* #1616 
* https://github.com/interuss/dss/pull/1619
* https://github.com/interuss/monitoring/pull/1618
  * This will require a new release of monitoring tools. As of now I used a custom image build on that PR.

### DIFF
--- a/build/dev/probe_locally.sh
+++ b/build/dev/probe_locally.sh
@@ -54,6 +54,11 @@ RESULTFILE="$(pwd)/e2e_test_result"
 touch "${RESULTFILE}"
 cat /dev/null > "${RESULTFILE}"
 
+PROBER_EXTRA_ARGS=()
+if [[ "${CORE_SERVICE_EXTRA_FLAGS:-}" == *"-enable_time_based_notification_index"* ]]; then
+	PROBER_EXTRA_ARGS+=(--scd-time-based-notification-index true)
+fi
+
 if ! docker run --rm --link "$OAUTH_CONTAINER":oauth \
 	--link "$CORE_SERVICE_CONTAINER":core-service \
 	--network dss_sandbox-default \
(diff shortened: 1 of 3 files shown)


BenjaminPelletier:
I think is is definitely what we need to address the root cause of #1615 (lack of CI coverage).  However, why does the CI pass when #1616 hasn't been merged?  It seems like catching that failure would be a primary objective of expanded test coverage.

the-glu:
> I think is is definitely what we need to address the root cause of #1615 (lack of CI coverage). However, why does the CI pass when #1616 hasn't been merged? It seems like catching that failure would be a primary objective of expanded test coverage.

This PR includes all fixes listed in the description, to ensure we get a working version at the end. They are multiples fixes needed since there are multiples failures reasons, but this PR represent the "final" version that should works.

I tested manually at least locally that https://github.com/interuss/dss/pull/1616 is needed.

(Hence the draft as well, to avoid merging fixes that need to be in their own PRs)

BenjaminPelletier:
In the future, it seems like the better path could be to create this PR with failing CI since demonstrating the CI fails for a known failure is very valuable, plus we get the non-merge guarantee without having to mark it as draft (which normally indicates not ready for review so reviewers may not even read the description).

BenjaminPelletier (approved):

build/dev/probe_locally.sh
```diff
@@ -54,12 +54,17 @@ RESULTFILE="$(pwd)/e2e_test_result"
...
+	PROBER_EXTRA_ARGS+=(--scd-time-based-notification-index true)
+fi
+
 if ! docker run --rm --link "$OAUTH_CONTAINER":oauth \
 	--link "$CORE_SERVICE_CONTAINER":core-service \
 	--network dss_sandbox-default \
 	-v "${RESULTFILE}:/app/test_result" \
 	-w /app/monitoring/prober \
-	interuss/monitoring:v0.31.0 \
+    interuss/monitoring:v0.31.0 \
```
the-glu: note: new image requiered there.

the-glu:
> In the future, it seems like the better path could be to create this PR with failing CI since demonstrating the CI fails for a known failure is very valuable, plus we get the non-merge guarantee without having to mark it as draft (which normally indicates not ready for review so reviewers may not even read the description).

True - the PR is now is in that mode. There also have been a successful run just before in history if needed.

BenjaminPelletier (dismissed):
Switching approval pending monitoring release to have the PR pass CI again

mickmis (approved):
LGTM with passing CI

PR #1620, opened 2026-08-10
[test] Add coverage for isYugabyteRetryable; fix TestNewRSAAuthClient on Windows

ashishjsharda:
`pkg/sqlstore/store.go` had no test coverage for `isYugabyteRetryable`, the function that determines whether a Yugabyte transaction error should trigger a retry. As a pure function over `pgconn.PgError` codes, it can be fully covered without a live database connection, so this adds a table-driven test for it.

This also fixes `TestNewRSAAuthClient`, which called `os.CreateTemp("/tmp", ...)` and consequently failed on Windows, where no `/tmp` directory exists. Replaced with `t.TempDir()`, which is portable and self-cleaning.

## Test plan

- `go test ./pkg/sqlstore/... ./pkg/auth/...` passes
- New `isYugabyteRetryable` cases cover both retryable Postgres error codes (`40001`, `40P01`), a non-retryable code, a wrapped error, a non-`PgError`, and `nil`

### DIFF
--- a/pkg/auth/auth_test.go
+++ b/pkg/auth/auth_test.go
@@ -97,7 +97,7 @@ func TestNewRSAAuthClient(t *testing.T) {
 	ctx, cancel := context.WithCancel(t.Context())
 	defer cancel()
 
-	tmpfile, err := os.CreateTemp("/tmp", "bad.pem")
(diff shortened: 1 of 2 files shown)


mickmis (approved):
LGTM, will just edit the PR title.

PR #1622, opened 2026-08-11
[raft/memstore] Add Get and RestoreFrom Snapshot to memstore

MariemBaccari:
Exposes the `GetSnapshot` and `RestoreFromSnapshot` repo methods through the memstore.

### DIFF
--- a/pkg/memstore/store.go
+++ b/pkg/memstore/store.go
@@ -79,3 +79,11 @@ func (s *Store[R]) Restore() {
 func (s *Store[R]) Close() error {
 	return nil
 }
+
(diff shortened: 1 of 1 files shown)


pkg/memstore/store.go
mickmis, line 37: Maybe embedding would make sense here given that `Store` is binding to all methods of `MemRepo`:
```suggestion
	MemRepo[R]
```

mickmis (approved):
LGTM. Do consider comment for a follow-up PR.

pkg/memstore/store.go
MariemBaccari: #1625

MariemBaccari:
Implements #1702

PR #1623, opened 2026-08-11
[raft/rid] Raftstore skeleton

MariemBaccari:
Adds the RID Raftstore skeleton

Implements #1700

### DIFF
--- a/pkg/rid/store/raftstore/store.go
+++ b/pkg/rid/store/raftstore/store.go
@@ -3,37 +3,70 @@ package raftstore
 import (
 	"context"
 
-	dsserr "github.com/interuss/dss/pkg/errors"
+	"github.com/interuss/dss/pkg/memstore"
 	"github.com/interuss/dss/pkg/raftstore"
 	"github.com/interuss/dss/pkg/raftstore/consensus"
 	"github.com/interuss/dss/pkg/rid/actions"
 	"github.com/interuss/dss/pkg/rid/repos"
+	ridmemstore "github.com/interuss/dss/pkg/rid/store/memstore"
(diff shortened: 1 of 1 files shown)


pkg/rid/store/raftstore/store.go
```diff
@@ -3,37 +3,70 @@ package raftstore
...
+	return r.memStore.RestoreFromSnapshot(data)
 }
 
-func (r *repo) Apply(_ context.Context, _ consensus.Proposal) (any, error) {
-	return nil, stacktrace.NewErrorWithCode(dsserr.NotImplemented, "not implemented yet")
+func (r *repo) Apply(ctx context.Context, proposal consensus.Proposal) (any, error) {
+	switch proposal.RequestType {
+
+	default:
+		handler, ok := actions.Registry[string(proposal.RequestType)]
```
mickmis, line 60: Update registry map key type to use `RequestType`

mickmis (approved):
(661d76e6c273e7889459ff9e5953d135b06f5d96) LGTM

pkg/rid/store/raftstore/store.go
MariemBaccari: Since the registry is defined outside of both raftstore and sqlstore and is used by both, it feels odd leaking `consensus.RequestType` into it. What do you think ?

mickmis: Yes indeed. That actually raises the question of whether there is (and will always be) a 1-1 mapping between DSS operations and raft messages? But it looks like those are different although overlapping sets. So LGTM.

PR #1626, opened 2026-08-11
[store] json encoding generics

MariemBaccari:
Since all requests will be json encoded and decoded for now, this PR adds generic functions that implement this functionality to be used in the `OperationHandler`. They will eventually be replaced by the custom encoding. 

Implements #1701

### DIFF
--- a/pkg/store/store.go
+++ b/pkg/store/store.go
@@ -2,6 +2,7 @@ package store
 
 import (
 	"context"
+	"encoding/json"
 	"io"
 
 	"github.com/interuss/stacktrace"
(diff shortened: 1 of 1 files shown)


pkg/store/store.go
```diff
@@ -31,6 +32,27 @@ type OperationHandler[R any] struct {
...
+func DecodeJSON[T any, PT interface {
+	*T
+	OperationRequest
+}](buf []byte) (OperationRequest, error) {
+	req := new(T)
+	if err := json.Unmarshal(buf, req); err != nil {
+		return nil, stacktrace.Propagate(err, "failed to unmarshal request")
+	}
+	return PT(req), nil
+}
```
mickmis, line 54: Doesn't this work and is simpler?
```go
func DecodeJSON[T OperationRequest](buf []byte) (OperationRequest, error) {
	req := new(T)
	if err := json.Unmarshal(buf, req); err != nil {
		return nil, stacktrace.Propagate(err, "failed to unmarshal request")
	}
	return T(*req), nil
}
```

pkg/store/store.go
```diff
@@ -31,6 +32,27 @@ type OperationHandler[R any] struct {
 	IsReadOnly bool
 }
 
+// EncodeJSON is a general-purpose OperationHandler.Encode that marshals the request as JSON.
+func EncodeJSON(request OperationRequest) ([]byte, error) {
```
mickmis, line 37: Checking how it used in #1597, can't those be actually attached instead to `OperationHandler`, and `T` be an additional type parameter of `OperationHandler`?

mickmis (commented):
Suggestions for simplification. I might not have the overall picture though.

pkg/store/store.go
MariemBaccari: This does work as long as the callers remember to pass a pointer. I think it's an ok change to make since we expect custom encoding / decoding in the near future anyways.

pkg/store/store.go
MariemBaccari: This wouldn't work since the registry is defined as a map:
`var Registry = map[string]dssstore.OperationHandler[repos.Repository]{}`
and we would need to provide a specific request type there. 

However, since we expect to have custom encoding, I think it's okay to leave things as they are now. I extracted the json encoding and decoding methods in this PR to make things simpler for the naive implementation and for the transition to the optimized encoding.

pkg/store/store.go
mickmis: Why does it need to be a pointer?
And if actually needed: add this to the function doc.

pkg/store/store.go
mickmis: In that case please add in both function docs a TODO mentioning that.

mickmis (approved):

pkg/store/store.go
MariemBaccari: Because `OperationID` is always a pointer receiver method: https://github.com/interuss/dss/blob/894284d4f73aa60ad7cfe78dc3f90f7bd92fb70b/pkg/api/ridv2/interface.gen.go

I added this to the doc.

mickmis: Right, but actually the compiler will prevent that in any case (tested that with: https://go.dev/play/p/Lgi5I12sIJG).

PR #1627, opened 2026-08-11
[raft] Pass down locality

MariemBaccari:
Pass down locality through consensus (needed for rid operations) mirroring the timestamp middleware.

Implements #1700

### DIFF
--- a/pkg/locality/locality.go
+++ b/pkg/locality/locality.go
@@ -1,0 +1,36 @@
+package locality
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/interuss/stacktrace"
+)
+
+type localityKey struct{}
+
+// MustGetRequestLocality returns the request locality from the context and panics if it is not
+// present, which is a programming error.
+func MustGetRequestLocality(ctx context.Context) string {
+	locality, ok := ctx.Value(localityKey{}).(string)
+	if !ok {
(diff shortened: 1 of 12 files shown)


pkg/locality/locality.go
mickmis, line 1: About naming in this file: it is go-idiomatic to drop the package name from the entities contained in the package. In addition I suggest following [doc of `context.Value`](https://pkg.go.dev/context#Context) out of consistency. E.g. here:
- `type key struct {}`
- `FromContext(ctx context.Context) (string, error or bool) {}`
- `MustFromContext(ctx context.Context) string {}`
- `NewContext(ctx context.Context, locality string) context.Context {}`
- `Middleware(locality string) func(...) ... {}`

mickmis (commented):
LGTM modulo naming suggestion.
PS: this could be applied too to the `timestamp` package too.

mickmis (approved):
LGTM modulo naming suggestion.
PS: this could be applied too to the `timestamp` package too.

MariemBaccari:
@mickmis I opened #1668 to address the naming comment for both `timestamp` and `locality` at the same time.

mickmis:
Note: ignored CI failure, which was flakyness (PR on top of this one succeeded).

PR #1629, opened 2026-08-12
[terraform/aws] Create gp3 storage class, add release notes

the-glu:
#1606 changed the default storage class for AWS clusters, but the `gp3` storage class is not available by default in EKS clusters.

This PR adds a mechanism to terraform to create it. It can be opted out of, should someone with an existing cluster have created it by other means, e.g. manually.

Release notes have also been updated with information about the potential migration step needed.

This has been tested with a fresh cluster and a PVC that has been successfully provisioned.

### DIFF
--- a/NEXT_RELEASE_NOTES.md
+++ b/NEXT_RELEASE_NOTES.md
@@ -55,6 +55,10 @@ The release notes should contain at least the following sections:
   * Details about migrations are available [there](https://interuss.github.io/dss/latest/operations/database-migrations/)
 * The 3.4.1 (or 1.1.1 for Yugabyte) migration for SCD is mandatory if you plan to use the new SCD hash lock option.
   * This migration can be applied without any impact if you do not plan to use it.
+* The default storage class for AWS changed from `gp2` to `gp3`.
+  * For existing clusters, ensure the value of `aws_kubernetes_storage_class` in your terraform config is set to `gp2` if it was not defined. Switching storage class is not supported.
+  * As `gp3` doesn't exist by default in EKS, terraform will now create the new storage class. If `gp3` already exists in your cluster, set `aws_create_storage_class` to false.
+  * There is a behaviour change between the previous default storage class and the new one: `reclaimPolicy` is set to `Retain` to limit accidental data deletion.
 
 ## Important information
 
--- a/deploy/infrastructure/dependencies/terraform-aws-kubernetes/storage.tf
+++ b/deploy/infrastructure/dependencies/terraform-aws-kubernetes/storage.tf
@@ -1,0 +1,10 @@
+resource "helm_release" "storage-class" {
+  count = var.aws_create_storage_class ? 1 : 0
+
+  chart = "${path.module}/charts/storage-class"
+  name  = "storage-class"
+  wait  = true
+  depends_on = [
+    aws_eks_addon.aws-ebs-csi-driver
+  ]
+}
--- a/deploy/infrastructure/dependencies/terraform-aws-kubernetes/variables.gen.tf
+++ b/deploy/infrastructure/dependencies/terraform-aws-kubernetes/variables.gen.tf
@@ -53,6 +53,20 @@ variable "aws_iam_permissions_boundary" {
 }
 
 
+variable "aws_create_storage_class" {
+  type        = bool
+  description = <<-EOT
+  Create storage class in cluster.
+
+  For now, create a gp3 storage class in the cluster and set it as the default one.
+
+  Example: `true`
+  EOT
+
+  default = true
+}
+
+
 variable "app_hostname" {
   type        = string
   description = <<-EOT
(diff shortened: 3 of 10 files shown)


mickmis (approved):
> This has been tested with a fresh cluster and a PVC that has been successfully provisioned.

Have you tested the behavior on an existing cluster?

the-glu:
> Have you tested the behavior on an existing cluster?

Not explicitly, but as terraform maintain the state, I think it should be fine even if applied later. Especially as cluster from users that already have a configuration / copied example configuration will stay on gp2.

I tried to delete / recreate that part on my test cluster and it worked.

mickmis:
> > Have you tested the behavior on an existing cluster?
> 
> Not explicitly, but as terraform maintain the state, I think it should be fine even if applied later. Especially as cluster from users that already have a configuration / copied example configuration will stay on gp2.
> 
> I tried to delete / recreate that part on my test cluster and it worked.

@barroco does that sound OK to you? I'm not confident enough on those tech to be sure or not. However for sure we should avoid deployment issues when users are upgrading.

deploy/infrastructure/dependencies/terraform-aws-kubernetes/charts/gp3-storage-class/templates/storageclass.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: gp3
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
+provisioner: ebs.csi.aws.com
+volumeBindingMode: WaitForFirstConsumer
+reclaimPolicy: Delete
```
barroco: I understand this is the default. However, I think our users would like to have a more production ready default and setting the reclaimPolicy to `Retain` seems more appropriate and a good improvement. What do you think ?

barroco:
> > > Have you tested the behavior on an existing cluster?
> > 
> > 
> > Not explicitly, but as terraform maintain the state, I think it should be fine even if applied later. Especially as cluster from users that already have a configuration / copied example configuration will stay on gp2.
> > I tried to delete / recreate that part on my test cluster and it worked.
> 
> @barroco does that sound OK to you? I'm not confident enough on those tech to be sure or not. However for sure we should avoid deployment issues when users are upgrading.

Since the change is only creating resources, it would only fail if resources with same names were added manually previously. So I agree with @the-glu. If a user had created those resources, they can use the flag as documented in the release note.

Though, migrating from gp2 to gp3 is out of scope of this change and is left to users to organize. Guidance may be provided but I would expect an issue to be opened specifically to request it.

NEXT_RELEASE_NOTES.md
```diff
@@ -55,6 +55,9 @@ The release notes should contain at least the following sections:
   * Details about migrations are available [there](https://interuss.github.io/dss/latest/operations/database-migrations/)
 * The 3.4.1 (or 1.1.1 for Yugabyte) migration for SCD is mandatory if you plan to use the new SCD hash lock option.
   * This migration can be applied without any impact if you do not plan to use it.
+* The default storage class for AWS changed from `gp2` to `gp3`.
+  * For existing clusters, check the value of `aws_kubernetes_storage_class` in your terraform config and set it to `gp2` if needed to keep the previous behaviour.
```
barroco:
```suggestion
  * For existing clusters, ensure the value of `aws_kubernetes_storage_class` in your terraform config is set to `gp2` if it was not defined. Switching storage class is not supported.
```
Can you please test the behaviour of switch from gp2 to gp3 so we can document it properly? Migration steps are out of scope from my point of view but can be added later.

barroco (changes requested):

NEXT_RELEASE_NOTES.md
the-glu: With a 'gp2' cluster, changing `aws_kubernetes_storage_class` to `gp3` fails as expected:

> for: "STDIN": error when patching "STDIN": StatefulSet.apps "cockroachdb" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'revisionHistoryLimit', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden

deploy/infrastructure/dependencies/terraform-aws-kubernetes/variables.gen.tf
```diff
@@ -53,6 +53,18 @@ variable "aws_iam_permissions_boundary" {
 }
 
 
+variable "aws_create_gp3_storage_class" {
```
barroco:
```suggestion
variable "aws_create_storage_class" {
```
Making it more generic would allow adding other storage type (eg `io2`)
Can you add `io2` as well since it is mentionned in the description. Can be added later.

deploy/infrastructure/dependencies/terraform-aws-kubernetes/storage.tf
```diff
@@ -0,0 +1,10 @@
+resource "helm_release" "gp3-storage-class" {
```
barroco:
```suggestion
resource "helm_release" "default-storage-class" {
```

barroco (approved):

barroco:
Can you please confirm you tested it again ?

the-glu:
> Can you please confirm you tested it again ?

Yes, tested with a new cluster and default config and works as expected:

image

PR #1630, opened 2026-08-12
[deploy] Add timeout option for evict tool to terraform/helm/tanka

### LINKED ISSUE #1628: `timeout` flag propagation for evict tool
Is the `timeout` flag being propagated to the binary? I can't find the flag here https://github.com/interuss/dss/blob/master/deploy/services/tanka/evict.libsonnet

the-glu:
Fix #1628 by adding the timeout flag to the various layers of the deployment stack and updating the documentation.

Tested on both a GKE and an AWS cluster: the flag is correctly set on the evict CronJobs, the jobs run without error, and leaving the value empty results in no flag being passed.


deploy/services/helm-charts/dss/values.schema.json
```diff
@@ -306,6 +306,10 @@
                       "type": "string",
                       "description": "Go duration format (https://pkg.go.dev/time#ParseDuration)"
                     },
+                    "timeout": {
+                      "type": "string",
+                      "description": "Go duration format (https://pkg.go.dev/time#ParseDuration). Empty to use the command default."
```
barroco, line 311: Outside of the scope of this PR, it would be nice to improve the descriptions of all variables to not only describe the format of the values.

barroco (approved):

Issue #1598, opened 2026-07-14
Release Compatibility Documentation


wing-utm-sharing-airspace:
**Feature Request**
Is it possible going forward to document in the release notes of each new candidate what the earliest backwards compatible previous release candidate is? 

Backwards compatible here meaning that two DSS deployments in the same pool could be running these different releases without impacting performance or safety critical functions of the DSS.

If there is general guidance (e.g. all minor version updates within the same major version are compatible, versions are compatible if there is no schema change...) that would also help

**Additional context**
For context, we are not planning to run different DSS versions within the same pool for an extended period of time but this would help DSS hosts come up with a thoughtful rollout plan and cutover period when upgrading to new software releases without affecting production flights.

the-glu:
Adding that information in the release notes (and/or documentation) seems a good idea indeed.

For the current situation / release until it's done, should that help:

I didn't explicitly tested combinations (I will do in the PR with details), but recent releases (at least 0.23.0-rc2 -> 0.20.0) should be compatible with others recent releases. Schema updates only include new functionalities / improvements and are backward compatible with old version of the code. In general, the expected procedure is: upgrade the schema and rollout restart the DSS with new images.

BenjaminPelletier:
I think this is an important element to support nominal operational usage of the DSS.  The user journey I think this documentation would primarily serve is "_I am part of a DSS pool in State X.  We want to change to State Y.  What is the procedure we should follow to get from X to Y?_"  "State X" would be something like all instances are running v0.20.0 with SCD schema v3.2.0 and "State Y" would be something like all instances are running the v0.23.0 with the corresponding recommended SCD schema.

There are many forms the answers to this kind of question could take, but I would envision something like a compatibility table plus explanatory text like "The DSS version is expected to be the same across all instances in general, but a pool should continue to operate normally for periods of up to about a week (actual value TBD) when the pool contains DSS version combinations as indicated in the table:"

| | v5 | v4 | v3 | v2 | v1 |
|---|---|---|---|---|---|
| **v5** | * SCD schema v3<br>* No operations should be performed when --flag1 is inconsistent between instances | * SCD schema v3<br>* --flag1 not set | Not supported | Not supported | Not supported |
| **v4** | | * SCD schema v2-v3<br>* SCD schema v3 recommended | * SCD schema v2-v3 | * SCD schema v2 | Not supported |
| **v3** | | | * SCD schema v2-v3 | * SCD schema v2 | Not supported |
[...]

PR #1631, opened 2026-08-12
[doc] Add notes about compatibility and release process

the-glu:
This PR adds to the documentation a method for performing upgrades, along with a compatibility matrix between recent DSS releases. This contributes to #1598 and is related to #1624 .

We should probably update this page as well as add the matrix to the release. However, we cannot add it to the release documentation, since we need the release itself to run the tests on the official image. Should we try to find a workaround, or assume that master will be up to date rather than the release's documentation?

The release procedure will need to be updated as well, once the previous point is settled.

### DIFF
--- a/docs/operations/upgrades.md
+++ b/docs/operations/upgrades.md
@@ -1,0 +1,101 @@
+# DSS upgrades
+
+This page describes how to upgrade a DSS pool, and which versions may temporarily coexist
+while the upgrade is in progress.
+
+## Upgrade preparation
+
+Before upgrading a pool:
+
+1. Read the release notes of every version between the currently deployed one and the target
+   one, not only those of the target version.
+2. Check the compatibility table below for the source and target versions.
+3. Choose the procedure: a [rolling upgrade](#rolling-upgrade-procedure) if the two versions are compatible; a [Zero-Traffic upgrade](#zero-traffic-upgrade-procedure)
+   otherwise, or if incompatible settings must be changed.
+4. Agree on which DSS instance in the pool applies the database migrations.
+
+
+## Compatibility between releases
+
+Releases are generally backward compatible with at least the previous release (and often earlier releases as well; see below), which allows certain combinations of versions of the DSS to
+run simultaneously during an upgrade.
+
+To limit the impact of a partially upgraded pool, we recommend keeping the window during
+which multiple versions run as short as practically possible.
+
+### Versions compatibility
+
+The following matrix shows what is possible when a user wants to upgrade a pool on version A (rows) to
+version B (columns).
+
+The table always assumes a migration to the latest schema of the target version B prior to DSS version upgrade per "Rolling upgrade procedure" below.  Where this cannot be accomplished (e.g., DSS version X cannot function with the latest schema of DSS version X+1), the transition will be indicated as incompatible.
+
+| A \ B | [v0.20.2](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.20.2) | [v0.21.1](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.21.1) | [v0.22.0](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.22.0) | [v0.23.0-rc3](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.23.0-rc3) |
+|---|---|---|---|---|
+| **[v0.20.2](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.20.2)** | ✅ | ✅<sup>1</sup> | ✅<sup>1</sup> | ✅<sup>1</sup> |
+| **[v0.21.1](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.21.1)** | ⚪ | ✅ | ✅ | ✅ |
+| **[v0.22.0](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.22.0)** | ⚪ | ⚪ | ✅ | ✅ |
+| **[v0.23.0-rc3](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.23.0-rc3)** | ⚪ | ⚪ | ⚪ | ✅ |
+
+
+✅ compatible · ⚠️ degraded, see explanation below · ❌ incompatible · ⚪ not evaluated
+
+<sup>1</sup>Some tests in a multi-version pool may fail due to improvements in the test suite and DSS behavior.
+
+Based on these results, InterUSS suggests that a rolling upgrade may be performed between any
+of the versions listed above.
+
+### Flags compatibility
+
+Feature flags (particularly those detailed on [the performance page](./performances.md)) are
+activated per node, with no global synchronization. Modifying these flags via a progressive
+rollout restart triggers a transitional state with the following impacts:
+
+- Performance degradation: the cluster temporarily operates with mixed locking strategies. This
+  untested configuration is expected to increase resource contention and response latency.
+- Requirement violations: because nodes respond inconsistently during the transition, the system
+  fails to comply with requirements F3548-21-DSS0210 and F3411-22a-DSS0070.
+
+We recommend applying the [Zero-Traffic upgrade procedure](#zero-traffic-upgrade-procedure) for the following flags:
+
+* enable_scd_global_lock
+* enable_scd_hash_lock
+* enable_time_based_notification_index
+
+All flags that are local, including feature flags for monitoring, can be enabled or disabled without any specific precautions.
+
+## Upgrade procedures
+
+### Rolling upgrade procedure
+
+This procedure has a lower impact and can be carried out asynchronously.
+
+Ensure the source and target DSS versions are compatible using the table above. Periods during which
+multiple DSS versions run in a DSS pool shall be as short as practically possible and limited to
+transitions.
+
+The steps are:
+
+1. Perform the database schema migrations to the latest schema versions of the target DSS version.
+2. Upgrade the DSS `core-service` instances, one at a time.
+
+The database migrations are applied once for the whole pool, by the member agreed upon beforehand, and
+before the `core-service` image can be upgraded. Each member can then upgrade their own
+instances at their own pace.
+
+### Zero-Traffic upgrade procedure
+
+This procedure has more impact and requires more synchronization between the members of a DSS
+pool. It is, however, less prone to errors, and it is required to upgrade between incompatible
+releases or to change the value of settings identified above.
+
+The steps are:
+
+1. Stop all regular cronjobs.
+2. Stop all DSS `core-service` instances.
+3. Perform database migrations to the schema versions associated with the target DSS version.
+4. Start the new `core-service` instances.
+5. Resume all regular cronjobs.
+
+The service is unavailable between steps 2 and 4. Every member of the pool must have completed
+step 2 before step 3 can be started.
--- a/docs/operations/.nav.yml
+++ b/docs/operations/.nav.yml
@@ -6,6 +6,7 @@ nav:
   - "Monitoring": monitoring.md
   - "Health checks": healthchecks.md
   - "Migrations": migrations.md
+  - "Upgrades": upgrades.md
   - "Database migrations": database-migrations.md
   - "Performances": performances.md
   - "Cleanup": cleanup.md
--- a/docs/operations/performances.md
+++ b/docs/operations/performances.md
@@ -9,7 +9,7 @@ See the detailed [section about cleanup](cleanup.md).
 ## The SCD global lock option
 
 !!! danger
-     All DSS instances in a DSS pool must use the same value for this option. Mixing will result in dramatically lower performance.
+     All DSS instances in a DSS pool must use the same value for this option. Mixing will result in dramatically lower performance. See the [upgrades](upgrades.md#zero-traffic-upgrade-procedure) page for information on how to enable or disable these flags in an existing deployment.
 
      You can use the `/aux/v1/configuration/global_options` endpoint to retrieve the current value for a specifc DSS instance.
 
@@ -38,7 +38,7 @@ All graphs have been generated with the [loadtest present in the monitoring repo
 ## The SCD hash lock option
 
 !!! danger
-     All DSS instances in a DSS pool must use the same value for this option. Mixing will result in dramatically lower performance.
+     All DSS instances in a DSS pool must use the same value for this option. Mixing will result in dramatically lower performance. See the [upgrades](upgrades.md#zero-traffic-upgrade-procedure) page for information on how to enable or disable these flags in an existing deployment.
 
      You can use the `/aux/v1/configuration/global_options` endpoint to retrieve the current value for a specifc DSS instance.
 
@@ -54,7 +54,7 @@ The number of locks (65535) is a compromise between lock contention (the more lo
 ## The time-based notification index option
 
 !!! danger
-    All DSS instances in a DSS pool must use the same value for this option. Mixing them will result in undetermined behavior.
+    All DSS instances in a DSS pool must use the same value for this option. Mixing them will result in undetermined behavior.  See the [upgrades](upgrades.md#zero-traffic-upgrade-procedure) page for information on how to enable or disable these flags in an existing deployment.
 
      You can use the `/aux/v1/configuration/global_options` endpoint to retrieve the current value for a specifc DSS instance.
 


docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+    * Qualifier: [Request to update USS availability status with empty version fails](https://github.com/interuss/monitoring/blob/f4acda1c9b4717662524c03479075027199d5785/monitoring/uss_qualifier/scenarios/astm/utm/dss/uss_availability_mutation.md#request-to-update-uss-availability-status-with-empty-version-fails-check)
+    * Prober: test_create_constraint_expired
+    * Prober: test_create_constraint
+
+Based on these results, InterUSS suggests that a rolling upgrade may be performed between any
+of the versions listed above.
+
+
+!!! danger
+    Changes to some flags (especially those affecting performance) are not covered by these tests and require a full stop procedure.
```
barroco:
```suggestion "Progressive Rollout Impacts"

Feature flags (particularly those detailed on [the performance page](./performance.md)) are activated per node. There is no global synchronization. Modifying these flags via a progressive rollout restart will trigger a transitional state with the following impacts:

- Performance Degradation: The cluster will temporarily operate with mixed locking strategies. This untested configuration is expected to increase resource contention and response latency.

- Requirement Violations: Because nodes will respond inconsistently during the transition, the system will fail to comply with requirements F3548-21-DSS0210 and F3411-22a-DSS0070.
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+while the upgrade is in progress.
+
+## General procedure
+
+Before upgrading a pool:
+
+1. Read the release notes of every version between the currently deployed one and the target
+   one, not only those of the target version.
+2. Check the compatibility table below for the source and target versions.
+3. Choose the procedure: a rolling upgrade if the two versions are compatible; a full stop
```
barroco: nit: Add link to the full stop procedure

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+    Changes to some flags (especially those affecting performance) are not covered by these tests and require a full stop procedure.
+    Never run multiple instances of the DSS, even at the same version, with incompatible flags.
+
+### Reproducing these results
+
+The matrix is produced by `release/compat/run-compat-matrix.sh`, which runs every ordered pair
+of the versions listed above locally with Docker Compose and reports the outcome of the prober
+and of the USS qualifier.
+
+## Full stop procedure
```
barroco: I don't think "Full stop" is correct.
Something like (`Zero-Traffic Upgrade`) would be best since we don't want users to stop their datastore cluster - just stop external traffic.

docs/operations/upgrades.md
barroco: See below on full stop naming.

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+
+Before upgrading a pool:
+
+1. Read the release notes of every version between the currently deployed one and the target
+   one, not only those of the target version.
+2. Check the compatibility table below for the source and target versions.
+3. Choose the procedure: a rolling upgrade if the two versions are compatible; a full stop
+   otherwise, or if incompatible settings must be changed.
+4. Agree with the other members of the pool on who applies the migrations, and on the window
+   during which the upgrade takes place.
```
barroco:
```suggestion
4. Agree on which DSS instance in the pool applies the migrations.
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+
+
+## Compatibility between releases
+
+Some releases are compatible with each other, which allows different versions of the DSS to
```
barroco:
```suggestion
Releases are generally backward compatible, which allows different versions of the DSS to
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+run simultaneously during an upgrade.
+
+To limit the impact of a partially upgraded pool, we recommend keeping the window during
+which multiple versions run as short as practically possible.
+
+### Test results
+
+Each pair below was tested with two `core-service` instances sharing a single CockroachDB
```
barroco: Let's split the result and the check methodology.

First results and instructions which the user can relate. `I have a pool on version A. I want to migrate to version B`
1. That means switching A and B.
2. Add a note on the expected schema migration 
3. Link to the release notes on verison names
4. Make it explicit flags which were tested.

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Compatibility between releases
+
+Some releases are compatible with each other, which allows different versions of the DSS to
+run simultaneously during an upgrade.
+
+To limit the impact of a partially upgraded pool, we recommend keeping the window during
+which multiple versions run as short as practically possible.
+
+### Test results
```
barroco:
```suggestion
### Versions compatibility
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+| **v0.21.1** | ⚠️<sup>2</sup> | – | ✅ | ✅ |
+| **v0.22.0** | ⚠️<sup>2</sup> | ✅ | – | ✅ |
+| **v0.23.0-rc3** | ⚠️<sup>2</sup> | ✅ | ✅ | – |
+
+Rows: version whose migrations are applied to the datastore.
+Columns: second `core-service` sharing that datastore.
+
+✅ pass · ⚠️ see notes · ❌ failed · ⛔️ instance doesn't start
+
+Important notes:
```
barroco:
```suggestion
Findings:
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Test results
+
+Each pair below was tested with two `core-service` instances sharing a single CockroachDB
+datastore, with the database migrated by the `db-manager` of version A. The prober and the
+USS qualifier were then run against both instances.
+
+Monitoring image used: interuss/monitoring:v0.31.0
+
+| A \ B | v0.20.2 | v0.21.1 | v0.22.0 | v0.23.0-rc3 |
```
barroco: Link to the release notes

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+Columns: second `core-service` sharing that datastore.
+
+✅ pass · ⚠️ see notes · ❌ failed · ⛔️ instance doesn't start
+
+Important notes:
+
+1. v0.21.1 introduced a mandatory migration (the `aux_` database schema) that must be
+   applied before running those versions. An instance of v0.21.1 or later will not start
+   against a datastore migrated only by v0.20.2.
+2. These tests fail because of the monitoring image used, which checks behavior that was
```
barroco:
```suggestion
2. These tests fail because of improvements of the test suite, which checks behavior that was
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+To limit the impact of a partially upgraded pool, we recommend keeping the window during
+which multiple versions run as short as practically possible.
+
+### Test results
+
+Each pair below was tested with two `core-service` instances sharing a single CockroachDB
+datastore, with the database migrated by the `db-manager` of version A. The prober and the
+USS qualifier were then run against both instances.
+
+Monitoring image used: interuss/monitoring:v0.31.0
```
barroco: Let's push down the test methodology.

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+
+| A \ B | v0.20.2 | v0.21.1 | v0.22.0 | v0.23.0-rc3 |
+|---|---|---|---|---|
+| **v0.20.2** | – | ⛔️<sup>1</sup> | ⛔️<sup>1</sup> | ⛔️<sup>1</sup> |
```
barroco: Let's inverse the row order for readability.

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+| **v0.21.1** | ⚠️<sup>2</sup> | – | ✅ | ✅ |
+| **v0.22.0** | ⚠️<sup>2</sup> | ✅ | – | ✅ |
+| **v0.23.0-rc3** | ⚠️<sup>2</sup> | ✅ | ✅ | – |
+
+Rows: version whose migrations are applied to the datastore.
+Columns: second `core-service` sharing that datastore.
+
+✅ pass · ⚠️ see notes · ❌ failed · ⛔️ instance doesn't start
```
barroco:
```suggestion
✅ pass · ⚠️ degraded, see findings below · ❌ incompatible · ⛔️ instance doesn't start
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+Monitoring image used: interuss/monitoring:v0.31.0
+
+| A \ B | v0.20.2 | v0.21.1 | v0.22.0 | v0.23.0-rc3 |
+|---|---|---|---|---|
+| **v0.20.2** | – | ⛔️<sup>1</sup> | ⛔️<sup>1</sup> | ⛔️<sup>1</sup> |
+| **v0.21.1** | ⚠️<sup>2</sup> | – | ✅ | ✅ |
+| **v0.22.0** | ⚠️<sup>2</sup> | ✅ | – | ✅ |
+| **v0.23.0-rc3** | ⚠️<sup>2</sup> | ✅ | ✅ | – |
+
+Rows: version whose migrations are applied to the datastore.
```
barroco: Let's stick to the following case: User wants to migrate from A to B.
So the schema should be the one of B (columns).

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+   only fixed in v0.21.1. Falling back to an older version of the monitoring image (v0.24.0)
+   makes this column pass.
+  The tests that don't pass are:
+    * Qualifier: [Request to update USS availability status with empty version fails](https://github.com/interuss/monitoring/blob/f4acda1c9b4717662524c03479075027199d5785/monitoring/uss_qualifier/scenarios/astm/utm/dss/uss_availability_mutation.md#request-to-update-uss-availability-status-with-empty-version-fails-check)
+    * Prober: test_create_constraint_expired
+    * Prober: test_create_constraint
+
+Based on these results, InterUSS suggests that a rolling upgrade may be performed between any
+of the versions listed above.
+
```
barroco, line 47: Please move the test methodology to a new section / with reproducing tests. - separate from results. Make tested flags explicit.
Since there is a lot of interest for the time-based notification index, we may want to state if there is any impact or not.

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+1. Stop all regular cronjobs.
+2. Stop all DSS `core-service` instances.
+3. Perform the migrations to the target version.
+4. Start the new `core-service` instances.
+5. Resume all regular cronjobs.
+
+The service is unavailable between steps 2 and 4. Every member of the pool must have completed
+step 2 before step 3 is started.
+
+## Rolling upgrade procedure
```
barroco: Move this up since this should be the general approach.

barroco: Create a new chapter for the two procedures: `Upgrade procedures`

docs/operations/upgrades.md
barroco: Move this warning in the new top level: `upgrade procedures`

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+
+This procedure has a lower impact and can be carried out asynchronously.
+
+Ensure the source and target versions are compatible using the table above. It is still
+recommended that all instances of a DSS end up on the same version, and that disparities
```
barroco: I would be more assertive as above: `Running multiple versions on a DSS pool shall be as short as practically possible and limited to transition periods.`

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+between versions be limited as much as practically possible.
+
+The steps are:
+
+1. Perform the migrations to the target version.
+2. Upgrade the DSS `core-service` instances, one at a time.
+
+The migrations are applied once for the whole pool, by the member agreed upon beforehand, and
+before any instance is upgraded. Each member can then upgrade their own instances at their own
```
barroco:
```suggestion
before core-service image can be upgraded. Each member can then upgrade their own instances at their own
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
...
+pace, within the agreed window.
```
barroco:
```suggestion
pace.
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,106 @@
+# DSS upgrades
+
+This page describes how to upgrade a DSS pool, and which versions may temporarily coexist
+while the upgrade is in progress.
+
+## General procedure
```
barroco:
```suggestion
## Upgrade preparation
```

barroco (commented):
Thanks @the-glu for preparing this. See my comments inline.

the-glu:
@barroco Thanks for the review, a improved version is ready for a second pass

docs/operations/upgrades.md
```diff
@@ -0,0 +1,115 @@
...
+Releases are generally backward compatible, which allows different versions of the DSS to
+run simultaneously during an upgrade.
+
+To limit the impact of a partially upgraded pool, we recommend keeping the window during
+which multiple versions run as short as practically possible.
+
+### Versions compatibility
+
+The following matrix shows what is possible when a pool is on version A (rows) and you want to
+migrate to version B (columns).
```
barroco:
```suggestion
The following matrix shows what is possible when a user wants to upgrade a pool on version A (rows) to
version B (columns).
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,115 @@
...
+
+The migrations are applied once for the whole pool, by the member agreed upon beforehand, and
+before the `core-service` image can be upgraded. Each member can then upgrade their own
+instances at their own pace.
+
+### Zero-Traffic upgrade procedure
+
+This procedure has more impact and requires more synchronization between the members of a DSS
+pool. It is, however, less prone to errors, and it is required to upgrade between incompatible
+releases or to change settings in an incompatible way.
```
barroco:
```suggestion
releases or to change settings, which are not compatible.
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,115 @@
...
+The steps are:
+
+1. Stop all regular cronjobs.
+2. Stop all DSS `core-service` instances.
+3. Perform the migrations to the target version.
+4. Start the new `core-service` instances.
+5. Resume all regular cronjobs.
+
+The service is unavailable between steps 2 and 4. Every member of the pool must have completed
+step 2 before step 3 is started.
```
barroco:
```suggestion
step 2 before step 3 can be started.
```

barroco (approved):
@the-glu, many thanks!

docs/operations/upgrades.md
```diff
@@ -0,0 +1,115 @@
...
+
+To limit the impact of a partially upgraded pool, we recommend keeping the window during
+which multiple versions run as short as practically possible.
+
+### Versions compatibility
+
+The following matrix shows what is possible when a user wants to upgrade a pool on version A (rows) to
+version B (columns).
+
+The table always assumes a migration to the latest schema of the target version B.
```
BenjaminPelletier:
```suggestion
The table always assumes a migration to the latest schema of the target version B prior to DSS version upgrade per "Rolling upgrade procedure" below.  Where this cannot be accomplished (e.g., DSS version X cannot function with the latest schema of DSS version X+1), the transition will be indicated as incompatible.
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,115 @@
...
+
+| A \ B | [v0.20.2](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.20.2) | [v0.21.1](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.21.1) | [v0.22.0](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.22.0) | [v0.23.0-rc3](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.23.0-rc3) |
+|---|---|---|---|---|
+| **[v0.20.2](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.20.2)** | – | ⚠️<sup>1</sup> | ⚠️<sup>1</sup> | ⚠️<sup>1</sup> |
+| **[v0.21.1](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.21.1)** | - | – | ✅ | ✅ |
+| **[v0.22.0](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.22.0)** | - | - | – | ✅ |
+| **[v0.23.0-rc3](https://github.com/interuss/dss/releases/tag/interuss%2Fdss%2Fv0.23.0-rc3)** | - | - | - | – |
+
+
+✅ passed · ⚠️ degraded, see findings below · ❌ incompatible · ⛔️ instance doesn't start
```
BenjaminPelletier: Let's make sure all contents of table cells above are covered in the legend.  If we don't want to test downgrades, it seems like the correct value for the cell would be "not evaluated" (perhaps "NE", 🔎, or "?"), but I'm not sure what the dash indicates (especially when the same character is used for the boxes relating version X to X which is a trivial upgrade).

I would also recommend putting green check marks in as many places as applicable.  It seems clear that upgrading a pool to v0.20.2 from v0.20.2 is very easy (nothing needs to be done), so I would expect a green check mark there even though nothing was tested.

Finally, I would word the meanings in terms of actual compatibility rather than the means by which the compatibility was determined.  IIUC, we empirically measure a mixed pool using uss_qualifier and assign a green checkmark when that pool passes uss_qualifier, thus ✅  = passed.  But, this is user documentation and I don't think it matters to users how we generate this table, but rather the information it conveys with regard to its subject (compatibility).  So, I would expect


```suggestion
✅ compatible · ⚠️ degraded, see explanation below · ❌ incompatible
```

I'm not sure what "instance doesn't start" would mean, separate from "incompatible".

Also, I would expect "degraded" to refer to the performance/functionality of the DSS product.  It sounds like there's no degradation when operating a pool with v0.20.2 + other versions, it's just that the `monitoring` versions associated with those releases test different things and therefore have findings on some instances versus others.  That seems like a functionality/features finding, not a version compatibility finding.  I don't think we'd want to combine those two different questions in this table: if an upgrade causes the performance, behavior, or functionality to degrade during the upgrade, that's when I would indicate "degraded".  Otherwise, if the pool performs, behaves, and functions somewhere between the old and new versions throughout the upgrade, I would indicate "compatible".

docs/operations/upgrades.md
```diff
@@ -0,0 +1,115 @@
...
+### Flags compatibility
+
+Feature flags (particularly those detailed on [the performance page](./performances.md)) are
+activated per node, with no global synchronization. Modifying these flags via a progressive
+rollout restart triggers a transitional state with the following impacts:
+
+- Performance degradation: the cluster temporarily operates with mixed locking strategies. This
+  untested configuration is expected to increase resource contention and response latency.
+- Requirement violations: because nodes respond inconsistently during the transition, the system
+  fails to comply with requirements F3548-21-DSS0210 and F3411-22a-DSS0070.
```
BenjaminPelletier, line 57: It seems like we're basically saying that the product does not function correctly during a rolling change of these flag values.  If that's the case, I think we need to go further here and say changing flag values requires a zero-traffic "upgrade".

Given that important information, I think we should link back to the details here from the danger message on the performance page; e.g.:

```
All DSS instances in a DSS pool must use the same value for this option. Mixing them will result in undetermined behavior.  See [upgrades](...) page for information on how to enable or disable these flags in an existing deployment.
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,115 @@
...
+
+Each pair above is tested with two `core-service` instances sharing a single CockroachDB
+datastore, with the database migrated by the `db-manager` of version B. All configuration
+(flags in particular) is left to default values. The prober and the USS qualifier are then run
+against both instances.
+
+### Flags compatibility
+
+Feature flags (particularly those detailed on [the performance page](./performances.md)) are
+activated per node, with no global synchronization. Modifying these flags via a progressive
```
BenjaminPelletier, line 51: Is this true of all feature flags?  Wouldn't it work fine, for instance, for some instances to use a time-based notification index behavior while other instances use the legacy notification index behavior?  It seems like, e.g., -enable_tracing, -enable_metrics, -enable_opentelemetry, etc are all feature flags as well.

docs/operations/upgrades.md
```diff
@@ -0,0 +1,115 @@
...
+The matrix is produced by `release/compat/run-compat-matrix.sh`, which runs every pair of the
+versions listed above locally with Docker Compose and reports the outcome of the prober and of
+the USS qualifier.
+
+Monitoring image used: interuss/monitoring:v0.31.0
+
+Each pair above is tested with two `core-service` instances sharing a single CockroachDB
+datastore, with the database migrated by the `db-manager` of version B. All configuration
+(flags in particular) is left to default values. The prober and the USS qualifier are then run
+against both instances.
```
BenjaminPelletier: It seems like we need instructions to do this and we need to update https://github.com/interuss/dss/tree/master/release to reflect this necessary task when releasing (and, separately, it seems like we need to link to that page from https://github.com/interuss/dss/blob/master/RELEASE.md)

I don't think it will be important to users interested in upgrading how we generated or validated our documentation, just that it is correct.  I think this section can be entirely moved to https://github.com/interuss/dss/tree/master/release and tweaked to read as an explanation for what run-compat-matrix.sh does and why.

docs/operations/upgrades.md
```diff
@@ -0,0 +1,115 @@
...
+- Requirement violations: because nodes respond inconsistently during the transition, the system
+  fails to comply with requirements F3548-21-DSS0210 and F3411-22a-DSS0070.
+
+## Upgrade procedures
+
+### Rolling upgrade procedure
+
+This procedure has a lower impact and can be carried out asynchronously.
+
+Ensure the source and target versions are compatible using the table above. Periods during which
```
BenjaminPelletier: There are a lot of different types of versions floating around; let's be clear:


```suggestion
Ensure the source and target DSS versions are compatible using the table above. Periods during which
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,115 @@
...
+multiple versions run in a DSS pool shall be as short as practically possible and limited to
+transitions.
+
+The steps are:
+
+1. Perform the migrations to the target version.
```
BenjaminPelletier: I think we can be clearer here:

```suggestion
1. Perform the database schema migrations to the latest schema versions of the target DSS version.
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,115 @@
...
+
+The migrations are applied once for the whole pool, by the member agreed upon beforehand, and
+before the `core-service` image can be upgraded. Each member can then upgrade their own
+instances at their own pace.
+
+### Zero-Traffic upgrade procedure
+
+This procedure has more impact and requires more synchronization between the members of a DSS
+pool. It is, however, less prone to errors, and it is required to upgrade between incompatible
+releases or to change settings, which are not compatible.
```
BenjaminPelletier:
```suggestion
releases or to change settings which are not compatible.
```

(the comma implies no settings are compatible)

BenjaminPelletier (approved):
I think this is a great start and fine to add as-is.  I think many of my comments are somewhat important, but they could potentially all be addressed in a followup PR if more convenient.

docs/operations/upgrades.md
the-glu: Yes, there is a linked PR #1624 in draft that helped generated part of this document. When it done I will update it to reflect what we want there, and I expect instructions to be part of that PR as well a this point.

I think what is left there is important for the user: the version used, the datastore, the migrations, the flags used and what tests that have been run, do you sill think it should be removed?

docs/operations/upgrades.md
the-glu: For the time-based, nodes would respond differently based on their flag (either DB index or time-based index on response, update of DB index and nothing), so that fails to comply with F3548-21-DSS0210 and F3411-22a-DSS0070 no?

For the others flags there are indeed fine.

Idea was to be as generic as possible there, but I think we can move to a list of flags and their impact.

docs/operations/upgrades.md
the-glu: That adjust meaning but that look fine.

For the original one: incompatible was: "instances start, but test fail" (in opposition to instance doesn't start at all). I does convoy some information

docs/operations/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+## Upgrade preparation
+
+Before upgrading a pool:
+
+1. Read the release notes of every version between the currently deployed one and the target
+   one, not only those of the target version.
+2. Check the compatibility table below for the source and target versions.
+3. Choose the procedure: a [rolling upgrade](#rolling-upgrade-procedure) if the two versions are compatible; a [Zero-Traffic upgrade](#zero-traffic-upgrade-procedure)
+   otherwise, or if incompatible settings must be changed.
+4. Agree on which DSS instance in the pool applies the migrations.
```
BenjaminPelletier: Changing DSS versions is arguably a migration from one DSS version to another one also, so we can be specific:

```suggestion
4. Agree on which DSS instance in the pool applies the database migrations.
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+
+
+## Compatibility between releases
+
+Releases are generally backward compatible, which allows different versions of the DSS to
```
BenjaminPelletier: They're not fully backwards compatible, right?

```suggestion
Releases are generally backward compatible with at least the previous release (and often earlier releases as well; see below), which allows certain combinations of versions of the DSS to
```

docs/operations/upgrades.md
BenjaminPelletier: The user wants to know if two versions are compatible and it seems like that question is already answered in the table above without this information.  If the answer above depends on flags used, what tests have been run, etc, then it seems like those caveats should be included in the answer somehow; the user shouldn't be given an answer whose interpretation depends on knowing how that documentation was written.  The documentation-generation methodology is relevant to InterUSS contributors/developers, but I don't think it should be relevant to users, and this document is user documentation.

docs/operations/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+
+Observations:
+
+1. These tests may fail because of improvements to the test suite, which checks behavior that was
+   only fixed in v0.21.1. Falling back to an older version of the monitoring image (v0.24.0)
+   makes this row pass.
+   The tests that don't pass are:
+    * Qualifier: [Request to update USS availability status with empty version fails](https://github.com/interuss/monitoring/blob/f4acda1c9b4717662524c03479075027199d5785/monitoring/uss_qualifier/scenarios/astm/utm/dss/uss_availability_mutation.md#request-to-update-uss-availability-status-with-empty-version-fails-check)
+    * Prober: test_create_constraint_expired
+    * Prober: test_create_constraint
```
BenjaminPelletier: In general, I would expect that different DSS versions would fix bugs, adjust behaviors, etc, so it would not be uncommon for the tests for version A to not fully work on version B and potentially the tests for version B to not fully work on version A.  But, I don't think this indicates incompatibility between the actual DSS product at those two versions -- instead, that's an incompatibility between the testing infrastructures associated with those two versions.  On this page, I think the thing users care about is actual DSS product compatibility.

So, I think this information is useful on a contributor/developer-facing page (perhaps the one we'll make documenting how this document needs to be updated as part of the release process), but I think the most we'd need here would be something like:


```suggestion
<sup>1</sup>Some tests in a multi-version pool may fail due to improvements in the test suite and DSS behavior.
```

docs/operations/upgrades.md
BenjaminPelletier: I think we need to go further here and say changing flag values requires a zero-traffic "upgrade".

To be more explicit, I think all of the "Flags compatibility" section up to here should be replaced with something like:

> All DSS instances in a pool must use the same values for the flags listed below.  Changing the value of any of the flags below requires following the zero-traffic upgrade procedure.
>
> * enable_scd_global_lock
> * enable_scd_hash_lock
> * enable_time_based_notification_index

docs/operations/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+* addr
+* server_time
+* public_endpoint
+* log_format
+* log_level
+* dump_requests
+* gcp_prof_service_name
+* enable_opentelemetry
+* enable_metrics
+* enable_tracing
```
BenjaminPelletier: I think we should list the flags that require special handling rather than trying to list the flags that don't require special handling when changing their value, especially since the list here seems to be substantially incomplete.  There are about 56 flags per `docker run --rm interuss/dss:v0.23.0-rc4 core-service -help` and it seems like most of those don't require special handling when changing their value.

docs/operations/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Upgrade procedures
+
+### Rolling upgrade procedure
+
+This procedure has a lower impact and can be carried out asynchronously.
+
+Ensure the source and target DSS versions are compatible using the table above. Periods during which
+multiple versions run in a DSS pool shall be as short as practically possible and limited to
```
BenjaminPelletier:
```suggestion
multiple DSS versions run in a DSS pool shall be as short as practically possible and limited to
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+transitions.
+
+The steps are:
+
+1. Perform the database schema migrations to the latest schema versions of the target DSS version.
+2. Upgrade the DSS `core-service` instances, one at a time.
+
+The migrations are applied once for the whole pool, by the member agreed upon beforehand, and
```
BenjaminPelletier:
```suggestion
The database migrations are applied once for the whole pool, by the member agreed upon beforehand, and
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+before the `core-service` image can be upgraded. Each member can then upgrade their own
+instances at their own pace.
+
+### Zero-Traffic upgrade procedure
+
+This procedure has more impact and requires more synchronization between the members of a DSS
+pool. It is, however, less prone to errors, and it is required to upgrade between incompatible
+releases or to change settings which are not compatible.
```
BenjaminPelletier:
```suggestion
releases or to change the value of settings identified above.
```

docs/operations/upgrades.md
```diff
@@ -0,0 +1,129 @@
...
+
+The steps are:
+
+1. Stop all regular cronjobs.
+2. Stop all DSS `core-service` instances.
+3. Perform the migrations to the target version.
```
BenjaminPelletier:
```suggestion
3. Perform database migrations to the schema versions associated with the target DSS version.
```

docs/operations/upgrades.md
the-glu: I removed the section, but I still think it was somehow relevant. The documentation cannot validate every scenario with all possible combinations. We present the base, common case; if users deviate from that, they should run their own tests. This section explained how to do so.

barroco (approved):
As discussed offline with @BenjaminPelletier, this PR can be merged and additional/unadressed feedback will be addressed in follow up PRs. @the-glu please capture remaining items in issues if any.

Issue #1632, opened 2026-08-12
core-service panic when key refresh is impossible


the-glu:
**Describe the bug**
The core-service panics (and therefore shuts down the service) when it is unable to refresh keys on the jwks_endpoint.

**To Reproduce**
Start the core-service with a jwks_endpoint. Prevent the DSS from accessing this endpoint (e.g. shut down the server).

```
Monitoring: disabled
Debug Mode: off
2026-08-12T14:16:13.656Z        info    core-service/main.go:255        version {"address": ":8082", "version": "0.0.0+3b904b48-dirty"}
2026-08-12T14:16:13.656Z        info    core-service/main.go:256        build   {"address": ":8082", "description": {"Time":"2026-08-07.16:21:39","Commit":"3b904b48-dirty","Host":"buildkitsandbox"}}
2026-08-12T14:16:13.656Z        info    core-service/main.go:257        config  {"address": ":8082", "scd": true}
2026-08-12T14:16:13.656Z        info    core-service/main.go:259        config  {"address": ":8082", "scdGlobalLock": false}
2026-08-12T14:16:13.656Z        info    core-service/main.go:260        config  {"address": ":8082", "scdHashLock": false}
2026-08-12T14:16:13.656Z        info    core-service/main.go:261        config  {"address": ":8082", "timeBasedNotificationIndex": false}
2026-08-12T14:16:13.787Z        info    core-service/main.go:409        Starting DSS HTTP server        {"address": ":8082"}
2026-08-12T14:16:18.679Z        info    logging/http.go:109     GET /healthy HTTP/1.1   {"address": ":8082", "req_dump": "", "resp_dump": "ok", "req_headers": {"Accept":["*/*"],"Connection":["close"],"User-Agent":["Wget"]}, "resp_status_code": 200, "resp_status_text": "OK", "peer_address": "[::1]:44456", "start_time": "2026-08-12T14:16:18.679Z", "duration_ms": 0.1}

(http://172.17.0.1:8000/oauth2/jwks.json on my machine service a key set stopped)

[...]

PR #1634, opened 2026-08-13
[auth] Don't panic on transient failures when refreshing JWKS keys

the-glu:
This PR contributes to #1632 by implementing a retry mechanism for transient JWKS key refresh failures.

The service will still panic when receiving a valid reply where, e.g., the required key is not found.

Implementing it via the `CodeRetryable` system had the positive side effect of applying the same retry behavior on startup in case of issues - similar to the datastore - which makes sense instead of directly crashing, so I kept it.

No documentation changes are included as none currently exist. I will create follow-up PRs for that along with deployment changes.

### DIFF
--- a/cmds/core-service/main.go
+++ b/cmds/core-service/main.go
@@ -23,6 +23,7 @@ import (
 	aux "github.com/interuss/dss/pkg/aux_"
 	auxs "github.com/interuss/dss/pkg/aux_/store"
 	"github.com/interuss/dss/pkg/build"
+	dsserr "github.com/interuss/dss/pkg/errors"
 	"github.com/interuss/dss/pkg/logging"
 	"github.com/interuss/dss/pkg/rid/application"
 	rid_v1 "github.com/interuss/dss/pkg/rid/server/v1"
@@ -30,7 +31,6 @@ import (
 	rids "github.com/interuss/dss/pkg/rid/store"
 	"github.com/interuss/dss/pkg/scd"
 	scds "github.com/interuss/dss/pkg/scd/store"
-	"github.com/interuss/dss/pkg/store"
 	"github.com/interuss/dss/pkg/store/params"
 	"github.com/interuss/dss/pkg/timestamp"
 	"github.com/interuss/dss/pkg/version"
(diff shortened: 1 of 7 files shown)


pkg/errors/errors.go
```diff
@@ -45,6 +45,10 @@ const (
 
 	// NotImplemented is used when a feature needed for the operation has not yet been implemented.
 	NotImplemented
+
+	// CodeRetryable is used when an operation failed for a transient reason and
+	// may succeed if it is attempted again later.
+	CodeRetryable
```
mickmis: 'retryable' is actually not an error, but rather a characteristic of one. Looking at / getting inspired from gRPC status codes, an 'unavailable' error could be used. 
Since all usages of this are updated, could be the opportunity to change that.
WDYT?

pkg/auth/auth.go
mickmis, line 116: Also return code here?
Or if the intent is to not return the code in case of an invalid response, the error above L104 should not embed the code either out of consistency.

cmds/core-service/main.go
```diff
@@ -65,6 +65,7 @@ var (
 	jwksEndpoint      = flag.String("jwks_endpoint", "", "URL pointing to an endpoint serving JWKS")
 	jwksKeyIDs        = flag.String("jwks_key_ids", "", "IDs of a set of key in a JWKS, separated by commas")
 	keyRefreshTimeout = flag.Duration("key_refresh_timeout", 1*time.Minute, "Timeout for refreshing keys for JWT verification")
+	jwksMaxCacheAge   = flag.Duration("jwks_max_cache_age", 1*time.Hour, "Maximum duration during which keys that could not be refreshed are still used before shutting down the service")
```
mickmis: - `jwks_ttl` instead?
- mention 1 hour is a sane default due to https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L92

mickmis (commented):
This is not exactly what is described in https://github.com/interuss/dss/issues/1632#issuecomment-5270291779, but I think that is OK as a first step. To be seen if that warrants a follow-up issue though, notably fro the dynamic adaptation depending on the incoming tokens expiration.

pkg/errors/errors.go
the-glu: Yes, done

pkg/auth/auth.go
the-glu: We are not dealing with an invalid response here, but rather the case of a valid response that is missing the expected key. This is considered a "hard" error not something retryable.

 L104 handles cases where the response is not a valid JWKS, which means the response itself is invalid.
 
 So for me that the correct behavior?

cmds/core-service/main.go
the-glu: #1636 add some general documentation, for the second point would that be enough to put it there (PR may need some improvement for that) ? I would try to avoid lengthy explanations there.

First point done.

the-glu:
>  To be seen if that warrants a follow-up issue though, notably fro the dynamic adaptation depending on the incoming tokens expiration.

Yes, it probably since dynamic adaptation is a bit 'tricky' and is probably worth a dedicated PR

pkg/auth/auth.go
mickmis: Ah - I misread that, you are right. Thought the key IDs were taken from the JWKS response.

mickmis (approved):
LGTM
I will remove however from the description the fact that it fixes the issue and let @BenjaminPelletier assess what to do with it (close it, leave it open, and/or open a follow-up issue).

PR #1635, opened 2026-08-13
[auth] Rename key_refresh_timeout flag to jwks_refresh_interval

the-glu:
This PR follows #1634.

 It normalizes the `key_refresh_timeout` flag to be consistent with other JWKS flags and improves its description.


NEXT_RELEASE_NOTES.md
```diff
@@ -45,6 +45,8 @@ The release notes should contain at least the following sections:
 
 ## Mandatory migration tasks
 
+* The `key_refresh_timeout` flag has been renamed to `jwks_refresh_interval`. The former is still accepted but is deprecated and will be removed in a future release.
```
mickmis: It is actually optional then? 
And to be added as mandatory when we actually remove the flag.

mickmis (approved):
818faef215f973c903a0f9720e638ed5358f978b?w=1#r3774258471
LGTM modulo comment

NEXT_RELEASE_NOTES.md
the-glu: Yes, I would prefer it to be mandatory but that actually optional ;)

PR #1636, opened 2026-08-13
[doc] Add documentation about authentication and keys

the-glu:
This PR follows #1635.

It adds general documentation on handling keys, which was not previously present.


mickmis (approved):
LGTM
Rebase needed though

PR #1644, opened 2026-08-19
[raft/scd] Extract get and query subscription

MariemBaccari:
Extract get and query subscription business logic. 

Implements #1701


mickmis (approved):
(eec05d31d4a3bfd066d1fdb5470bfba62b1293b0) LGTM

PR #1646, opened 2026-08-19
[raft/uss] Extract uss availability

MariemBaccari:
Extract uss availability business logic

Implements #1701

### DIFF
--- a/pkg/scd/actions/availability.go
+++ b/pkg/scd/actions/availability.go
@@ -1,0 +1,104 @@
+package actions
+
+import (
+	"context"
+
+	restapi "github.com/interuss/dss/pkg/api/scdv1"
+	dsserr "github.com/interuss/dss/pkg/errors"
+	dssmodels "github.com/interuss/dss/pkg/models"
+	scdmodels "github.com/interuss/dss/pkg/scd/models"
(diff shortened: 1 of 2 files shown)


pkg/scd/actions/availability.go
```diff
@@ -0,0 +1,109 @@
...
+		IsReadOnly: true,
+	}
+	Registry[restapi.SetUssAvailabilityOperationID] = dssstore.OperationHandler[repos.Repository]{
+		Encode:  dssstore.EncodeJSON,
+		Decode:  dssstore.DecodeJSON[*restapi.SetUssAvailabilityRequest],
+		Execute: ExecuteSetUssAvailability,
+	}
+}
+
+func GetDefaultAvailabilityResponse(id dssmodels.Manager) *restapi.UssAvailabilityStatusResponse {
```
mickmis: Meh: I don't think it's worth exposing that to be used from the handlers.
Just inline it in here since it is used once only. 
And in handlers keep the function or inline too (version can be dropped since it's the zero value).

mickmis (approved):
(fc8653a11d59541ffb10aab8cdb2653fb97316c5) LGTM modulo comment

PR #1647, opened 2026-08-19
[raft/scd] Extract delete constraint

MariemBaccari:
Extract delete constraint business logic.

Implements #1701

### DIFF
--- a/pkg/scd/actions/constraint.go
+++ b/pkg/scd/actions/constraint.go
@@ -1,0 +1,79 @@
+package actions
+
+import (
+	"context"
+
+	"github.com/golang/geo/s2"
+	restapi "github.com/interuss/dss/pkg/api/scdv1"
+	dsserr "github.com/interuss/dss/pkg/errors"
+	dssmodels "github.com/interuss/dss/pkg/models"
+	scdmodels "github.com/interuss/dss/pkg/scd/models"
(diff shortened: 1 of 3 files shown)


pkg/scd/constraints_handler.go
mickmis:
```suggestion
	if req.Ovn == "" {
```

pkg/scd/actions/constraint.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	// Return response to client
+	return &restapi.ChangeConstraintReferenceResponse{
+		ConstraintReference: *old.ToRest(),
+		Subscribers:         makeSubscribersToNotify(subs),
+	}, nil
+}
+
+func makeSubscribersToNotify(subscriptions []*scdmodels.Subscription) []restapi.SubscriberToNotify {
```
mickmis: Used / will be used by both operational intent and constraint: put that in a separate file?

mickmis (approved):
LGTM modulo comments

pkg/scd/actions/constraint.go
MariemBaccari: I moved it to `registry.go` which can contain common stuff including the empty registry and helpers.

PR #1649, opened 2026-08-20
[raft/scd] Extract get and query constraint

MariemBaccari:
Extract get and query constraint business logic.

Implements #1701


mickmis (approved):
(379c1004350d0311eb78f747a4036c264d29fac9) LGTM

PR #1650, opened 2026-08-20
[raft/scd] Extract update and create constraint

MariemBaccari:
Extract update and create constraint business logic.

Implements #1701

### DIFF
--- a/pkg/scd/constraints_handler.go
+++ b/pkg/scd/constraints_handler.go
@@ -2,18 +2,17 @@ package scd
 
 import (
 	"context"
-	"time"
 
-	"github.com/golang/geo/s2"
 	"github.com/interuss/dss/pkg/api"
 	restapi "github.com/interuss/dss/pkg/api/scdv1"
 	dsserr "github.com/interuss/dss/pkg/errors"
 	dssmodels "github.com/interuss/dss/pkg/models"
+	"github.com/interuss/dss/pkg/scd/actions"
 	scdmodels "github.com/interuss/dss/pkg/scd/models"
 	"github.com/interuss/dss/pkg/scd/repos"
 	dssstore "github.com/interuss/dss/pkg/store"
+	"github.com/interuss/dss/pkg/timestamp"
 	"github.com/interuss/stacktrace"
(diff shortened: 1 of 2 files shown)


pkg/scd/constraints_handler.go
```diff
@@ -155,165 +164,26 @@ func (a *Server) UpdateConstraintReference(ctx context.Context, req *restapi.Upd
...
-		scdmodels.WithRequireEndTimeAfter(now),
-	)
-	if err != nil {
-		return nil, stacktrace.Propagate(err, "Invalid extents")
-	}
-
-	valid.cells, err = valid.uExtent.CalculateSpatialCovering()
-	if err != nil {
-		return nil, stacktrace.Propagate(err, "Invalid area")
-	}
```
mickmis, line 313: What's the rationale for not validating here with `UnionVolumes4DFromSCDRest` and `CalculateSpatialCovering` anymore?

IIUC the strategy for validation is:
- keep in REST handlers whatever validation there is, even if it is a side effect of the parsing
- duplicate in `ExecuteXXX` functions what is needed for parsing

Is this correct?

mickmis (approved):
LGTM modulo clarification.
But what this shows is that we may want to come up with something better for  deduplicating validation/parsing of those parameters.
LMK if you agree and have any idea. But in any case for after this PR train is merged
Maybe a `Validate` field on the `OperationHandler`?

pkg/scd/constraints_handler.go
MariemBaccari: Regarding the strategy, I aim to do as much validation as possible on the handler side, especially for bad request cases to avoid doing a consensus round just to end up rejecting the request for such a reason. Indeed, duplication occurs in the operation because the validation also comes with parsing (ID etc.).
For this specific case, not validating the volumes was a mistake on my part (fixed in latest push) as I kept these checks for the subscription side as well. I also merged the validation between the handler and the action / operation  and kept the handler specific part (AllowHTTPBaseUrls check) there.   

I just opened issue #1676 to track this duplication issue.

mickmis:
Note: ignored failed CI check, see #1660

PR #1645, opened 2026-08-19
[raft/scd] Extract create and update subscription

MariemBaccari:
Extract create and update subscription business logic.

Implements #1701

PR #1643, opened 2026-08-19
[raft/scd] Extract delete subscription

MariemBaccari:
Extract delete subscription business logic in scd.

Implements #1701

PR #1642, opened 2026-08-19
[raft/scd] Add scd raftstore skeleton

MariemBaccari:
Adds the SCD raftstore skeleton.

Implements #1700

PR #1651, opened 2026-08-20
[raft/scd] Return result and error in TransactWithResult

MariemBaccari:
Chained PR: https://github.com/interuss/dss/pull/1627 -> https://github.com/interuss/dss/pull/1642 -> https://github.com/interuss/dss/pull/1643 -> https://github.com/interuss/dss/pull/1644 -> https://github.com/interuss/dss/pull/1645 -> https://github.com/interuss/dss/pull/1646 -> https://github.com/interuss/dss/pull/1649 -> https://github.com/interuss/dss/pull/1650 -> https://github.com/interuss/dss/pull/1651 -> https://github.com/interuss/dss/pull/1653 -> https://github.com/interuss/dss/pull/1654 -> https://github.com/interuss/dss/pull/1656 -> https://github.com/interuss/dss/pull/1657 -> https://github.com/interuss/dss/pull/1655 -> https://github.com/interuss/dss/pull/1666 -> https://github.com/interuss/dss/pull/1667 -> https://github.com/interuss/dss/pull/1668 -> https://github.com/interuss/dss/pull/1669

Fix `TransactWithResult` to return both the result and the error. This is needed by `CreateOperationalIntentReference` which returns both an error and a `AirspaceConflictResponse` result.

### DIFF
--- a/cmds/core-service/main.go
+++ b/cmds/core-service/main.go
@@ -24,6 +24,7 @@ import (
 	auxs "github.com/interuss/dss/pkg/aux_/store"
 	"github.com/interuss/dss/pkg/build"
 	dsserr "github.com/interuss/dss/pkg/errors"
+	requestlocality "github.com/interuss/dss/pkg/locality"
 	"github.com/interuss/dss/pkg/logging"
 	"github.com/interuss/dss/pkg/rid/application"
 	rid_v1 "github.com/interuss/dss/pkg/rid/server/v1"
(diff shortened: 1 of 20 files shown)


mickmis (commented):
Have you considered instead introducing a new type containing both return values (ChangeOperationalIntentReferenceResponse and AirspaceConflictResponse)?
Because:
- I'd say it's not very Go-idiomatic to return both a value and an error (although we already do that I agree)
- But worse: the handler (#1655) is going to have to have to cast the same result to different types and this is to be avoided if possible

MariemBaccari:
@mickmis That's a great point ... I will close this PR and add the fix to #1655 instead.

PR #1653, opened 2026-08-20
[raft/scd] Extract get and query opintent

MariemBaccari:
Extract get and query opintent business logic.

Implements #1701


mickmis (approved):
(bcd214873109a66f175be460554cc2659e84692f) LGTM

PR #1655, opened 2026-08-20
[raft/scd] Extract update and create opintent

MariemBaccari:
Extract update and create opintent business logic.

Implements #1701

### DIFF
--- a/pkg/scd/actions/operational_intents.go
+++ b/pkg/scd/actions/operational_intents.go
@@ -14,6 +14,7 @@ import (
 	scdmodels "github.com/interuss/dss/pkg/scd/models"
 	"github.com/interuss/dss/pkg/scd/repos"
 	dssstore "github.com/interuss/dss/pkg/store"
+	"github.com/interuss/dss/pkg/timestamp"
 	"github.com/interuss/stacktrace"
 )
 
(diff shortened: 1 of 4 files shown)


mickmis (commented):
Review waiting on #1651

pkg/scd/operational_intents_handler.go
```diff
@@ -183,195 +191,12 @@ func (a *Server) UpdateOperationalIntentReference(ctx context.Context, req *rest
...
-	// Note: validateAndReturnOIRUpsertParams and checkUpsertPermissionsAndReturnManager could be moved out of this method and only the valid params passed,
-	// but this requires some changes in the caller that go beyond the immediate scope of #1088 and can be done later.
-	validParams, err := actions.ValidateAndReturnOIRUpsertParams(now, entityid, ovn, params, a.AllowHTTPBaseUrls)
-	if err != nil {
-		return nil, nil, stacktrace.PropagateWithCode(err, dsserr.BadRequest, "Failed to validate Operational Intent Reference upsert parameters")
-	}
-	manager, err := actions.CheckUpsertPermissionsAndReturnManager(authorizedManager, validParams.State)
-	if err != nil {
-		return nil, nil, stacktrace.PropagateWithCode(err, dsserr.PermissionDenied, "Caller is not allowed to upsert with the requested state")
-	}
```
mickmis, line 209: - looks like this should be invoked in REST handler for validation
- and could actually be dropped in the `ExecutePutOperationalIntentReference` function: if validated the only thing it returns is `dssmodels.Manager(*authorizedManager.ClientID)`

pkg/scd/actions/operational_intents.go
```diff
@@ -624,3 +635,202 @@ func EnsureSubscriptionCoversOIR(ctx context.Context, r repos.Repository, sub *s
...
+		}
+	}
+
+	var responseConflict *restapi.AirspaceConflictResponse
+	if validParams.State.RequiresKey() {
+		responseConflict, err = validateKeyAndProvideConflictResponse(ctx, repo, manager, validParams, attachedSub)
+		if err != nil {
+			// responseConflict is non-nil here on a dsserr.MissingOVNs error: return it alongside
+			// the error so the handler can still send it to the client. See the doc comment above.
+			return &PutOperationalIntentReferenceResult{Conflict: responseConflict}, stacktrace.PropagateWithCode(err, stacktrace.GetCode(err), "Failed to validate key")
```
mickmis, line 797: This is the only instance of returning an error from the execute function while still returning a 'valid' response. I think this is confusing, unnecessary and we should get rid of it. Of `dsserr.MissingOVNs` too. And in the HTTP handler we can just check for a non-nil `result.Conflict`. WDYT? (does not have to be in this PR)

mickmis (approved):
LGTM modulo comments

pkg/scd/operational_intents_handler.go
MariemBaccari: Addressed in latest push

pkg/scd/actions/operational_intents.go
MariemBaccari: I agree, thanks for suggesting this. I just opened #1690 to address the comment.

PR #1657, opened 2026-08-20
[raft/scd] Move opintent upsert helpers into actions

MariemBaccari:
Move opintent upsert helpers into the actions folder. A following PR will extract the action from the handler. 

Implements #1701


mickmis (approved):
(5e5bb93e7a1748f35cf4d83b8cbbd79d68bef47c) LGTM

PR #1661, opened 2026-08-21
[tooling] Add benchmarker configuration and test instructions

BenjaminPelletier:
This PR adds an initial benchmarker configuration intended to characterize performance of DSS release candidates in a scenario equivalent (in terms of performance implications) to the [Busy S2cell Test](https://github.com/interuss/dss/issues/1541#issuecomment-4846753840).  Running Aug 20 in the evening Pacific on 0.23.0-rc5 deployments yielded:

scalability_curve_crdb

scalability_curve_ybdb

Since the Y axes of these graphs indicate success flights and each flight consists of three queries (create op intent, update op intent, delete op intent), I think CRDB meets the objectives specified for Busy S2cell Test in both the leader instance (AWS in this case) and follower instance (GCP in this case).  However, it looks like the test may need more refinement to clearly identify whether YugabyteDB meets the objectives -- a naive reading of the graph above would indicate it does not.


release/tests/benchmarker/README.md
```diff
@@ -0,0 +1,26 @@
+# benchmarker release tests
+
+This folder contains benchmarker configurations intended for characterizing performance of DSS releases in reference deployments in reference conditions.
+
+To run a configuration, first host a dummy OAuth server locally on port 8085.  From the root of the `monitoring` repo:
+
+```shell
+make start-locally
+```
```
the-glu, line 9: Probably a nit / for later, but would be great to have a script similar to "run-tests.sh" who takes care of everything automatically ('run-tests.sh' already have code to spawn the oauth server without the rest, not having to edit files is always nice, and this would use the same monitoring image as the rest).

BenjaminPelletier: Yes, more automation is definitely good, but have to get there in steps :)

barroco (approved):

PR #1668, opened 2026-08-28
[raftstore] Rename context methods

MariemBaccari:
Following [feedback](https://github.com/interuss/dss/pull/1627#discussion_r3861291010) from #1627, this PR renames the methods of the `timestamp` and `locality` packages following the [doc of context.Value](https://pkg.go.dev/context#Context).

Implements #1700


mickmis (approved):
(e22eb6c3d0b34387f761019a097138a0c6d9d2b2) LGTM

PR #1669, opened 2026-08-28
[raftstore] Embed memstore and use checkpoint

MariemBaccari:
This PR adds the use of the memstore checkpoint to ensure raft operations are atomic and embeds the memstore directly to simplify the raftstore. 

Implements #1700 and #1702

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -16,10 +16,6 @@ const (
 	// larger than the max area allowed. See geo/s2.go.
 	AreaTooLarge = stacktrace.ErrorCode(iota)
 
-	// MissingOVNs is the error to signal that an AirspaceConflictResponse should
-	// be returned rather than the standard error response.
-	MissingOVNs
-
 	// AlreadyExists is used when attempting to create a resource that already
 	// exists.
 	AlreadyExists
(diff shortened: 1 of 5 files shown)


pkg/raftstore/store.go
```diff
@@ -117,7 +124,11 @@ func (s *Store[R]) processCommits(ctx context.Context, commitCh <-chan consensus
 
 			proposalCtx := timestamp.NewContext(ctx, commit.Prop.Timestamp)
 			proposalCtx = locality.NewContext(proposalCtx, commit.Prop.Locality)
+			s.raftRepo.Checkpoint()
 			result, err := s.raftRepo.Apply(proposalCtx, commit.Prop)
+			if err != nil {
```
mickmis, line 116: Some logging advising of the restore could be handy here.

pkg/raftstore/store.go
```diff
@@ -27,6 +27,13 @@ type RaftRepo[R any] interface {
 	// RestoreFromSnapshot replaces all state with the snapshot in data.
 	// data is always the output of a prior GetSnapshot.
 	RestoreFromSnapshot(data []byte) error
+
+	// Checkpoint saves the current state, called before every proposal is applied.
+	Checkpoint()
+
+	// Restore reverts the state to the last Checkpoint, called when Apply returns an error so a
+	// failed proposal cannot leave a partial mutation in place.
+	Restore()
```
mickmis: In the end shouldn't `RaftRepo` embed the `MemRepo` interface? It just has `Apply` in addition.

mickmis (approved):
(5fe0a169283135e56b2cb068c4e06a18c86b0f85) LGTM modulo comments

PR #1672, opened 2026-08-28
[raft/scd] Implement operational intents repo methods

MariemBaccari:
Implements the scd opintents repo methods for the raftstore. The raftstore repo methods issue individual proposals.

Implements #1700

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -16,10 +16,6 @@ const (
 	// larger than the max area allowed. See geo/s2.go.
 	AreaTooLarge = stacktrace.ErrorCode(iota)
 
-	// MissingOVNs is the error to signal that an AirspaceConflictResponse should
-	// be returned rather than the standard error response.
-	MissingOVNs
-
 	// AlreadyExists is used when attempting to create a resource that already
 	// exists.
 	AlreadyExists
--- a/pkg/scd/actions/operational_intents.go
+++ b/pkg/scd/actions/operational_intents.go
@@ -517,7 +517,7 @@ func createAndStoreNewImplicitSubscription(ctx context.Context, r repos.Reposito
 
 // validateKeyAndProvideConflictResponse ensures that the provided key contains all the necessary OVNs relevant for the area covered by the OperationalIntent.
 // - If all required keys are provided, (nil, nil) will be returned.
-// - If keys are missing, the conflict response to be sent back as well as an error with the dsserr.MissingOVNs code will be returned.
+// - If keys are missing, the conflict response will be returned (conflict, nil).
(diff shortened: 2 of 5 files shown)


pkg/scd/store/raftstore/store.go
```diff
@@ -54,6 +54,10 @@ func (r *repo) Apply(ctx context.Context, proposal consensus.Proposal) (any, err
 		string(listExpiredSubscriptions), string(countSubscriptions):
 		return r.applySubscription(ctx, proposal)
 
+	case string(getOperationalIntent), string(deleteOperationalIntent), string(upsertOperationalIntent), string(searchOperationalIntents),
+		string(getDependentOperationalIntents), string(listExpiredOperationalIntents), string(countOperationalIntents):
+		return r.applyOperationalIntent(ctx, proposal)
```
mickmis: nit: consider inlining applyOperationalIntent

pkg/scd/store/raftstore/operational_intents.go
```diff
@@ -2,38 +2,132 @@ package raftstore
...
+	getDependentOperationalIntents consensus.RequestType[[]dssmodels.ID]                 = "getDependentOperationalIntents"
+	listExpiredOperationalIntents  consensus.RequestType[[]*scdmodels.OperationalIntent] = "listExpiredOperationalIntents"
+	countOperationalIntents        consensus.RequestType[int64]                          = "countOperationalIntents"
+)
+
+func (r *repo) GetOperationalIntent(ctx context.Context, id dssmodels.ID) (*scdmodels.OperationalIntent, error) {
+	buf, err := json.Marshal(id)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "failed to marshal payload")
+	}
```
mickmis, line 28: Not for this PR, but with the look of it, looks like we could be pushing the marshalling of the value into `HandleClientRequest` and change the type of `value` to `any`. Except if some operations have more complex things going on?

mickmis (approved):
LGTM

pkg/scd/store/raftstore/operational_intents.go
MariemBaccari: With the gate 1 work (#1709), I will look into payload size optimizations and that should include investigating custom encoding per payload type. I kept the marshalling outside so that consensus can receive an opaque payload like it would with custom encoding. 
I referenced your suggestion in the contributions spreadsheet to keep it in mind in case we don't end up going with custom encoding for some reason. Is it okay if I revisit later ?

mickmis: Sounds good :+1:

PR #1682, opened 2026-09-08
[raft/rid] Extract InsertSubscription and route Create through the store

MariemBaccari:
Extracts insert subscription business logic. 

Implements #1701

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -16,10 +16,6 @@ const (
 	// larger than the max area allowed. See geo/s2.go.
 	AreaTooLarge = stacktrace.ErrorCode(iota)
 
-	// MissingOVNs is the error to signal that an AirspaceConflictResponse should
-	// be returned rather than the standard error response.
-	MissingOVNs
-
 	// AlreadyExists is used when attempting to create a resource that already
 	// exists.
 	AlreadyExists
--- a/pkg/scd/actions/operational_intents.go
+++ b/pkg/scd/actions/operational_intents.go
@@ -517,7 +517,7 @@ func createAndStoreNewImplicitSubscription(ctx context.Context, r repos.Reposito
 
 // validateKeyAndProvideConflictResponse ensures that the provided key contains all the necessary OVNs relevant for the area covered by the OperationalIntent.
 // - If all required keys are provided, (nil, nil) will be returned.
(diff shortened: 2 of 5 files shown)


pkg/rid/operations/subscription.go
```diff
@@ -2,16 +2,59 @@ package operations
...
+	Op         string
+	ID         dssmodels.ID
+	Owner      dssmodels.Owner
+	URL        string
+	Version    *dssmodels.Version
+	Cells      s2.CellUnion
+	StartTime  *time.Time
+	EndTime    *time.Time
+	AltitudeLo *float32
+	AltitudeHi *float32
```
mickmis, line 33: You probably haven't got around to that yet, but this shuold use CellsVolume4D, right?

pkg/rid/operations/subscription.go
```diff
@@ -73,3 +126,57 @@ func executeDeleteSubscription(ctx context.Context, repo repos.Repository, reque
...
+		EndTime:    payload.EndTime,
+		AltitudeLo: payload.AltitudeLo,
+		AltitudeHi: payload.AltitudeHi,
+		Writer:     locality.MustFromContext(ctx),
+	}
+
+	return InsertSubscription(ctx, repo, sub)
+}
+
+func InsertSubscription(ctx context.Context, repo repos.Repository, sub *ridmodels.Subscription) (*ridmodels.Subscription, error) {
```
mickmis, line 151: Why not inlining this function into executeInsertSubscription, and why expose it?

mickmis (approved):
LGTM modulo minor comments

pkg/rid/operations/subscription.go
```diff
@@ -2,16 +2,59 @@ package operations
...
+	Cells      s2.CellUnion
+	StartTime  *time.Time
+	EndTime    *time.Time
+	AltitudeLo *float32
+	AltitudeHi *float32
+}
+
+func (p *putSubscriptionPayload) OperationID() string { return p.Op }
+
+func NewPutSubscriptionPayload(opID string, id dssmodels.ID, owner dssmodels.Owner, url string, version *dssmodels.Version, extents *dssmodels.Volume4D) (*putSubscriptionPayload, error) {
```
mickmis, line 38: Ah and same comment as the previous PRs: do not pass the `opID`, although I'm sure you were going to do that already :)

PR #1685, opened 2026-09-08
[raft/rid] Extract UpdateSubscription and route Update through the store

MariemBaccari:
Extracts update subscription business logic.

Implements #1701

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -16,10 +16,6 @@ const (
 	// larger than the max area allowed. See geo/s2.go.
 	AreaTooLarge = stacktrace.ErrorCode(iota)
 
-	// MissingOVNs is the error to signal that an AirspaceConflictResponse should
-	// be returned rather than the standard error response.
-	MissingOVNs
-
 	// AlreadyExists is used when attempting to create a resource that already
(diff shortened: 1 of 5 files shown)


pkg/rid/server/v2/subscription_handler.go
```diff
@@ -235,21 +234,21 @@ func (s *Server) UpdateSubscription(ctx context.Context, req *restapi.UpdateSubs
...
 	}
-
-	sub := &ridmodels.Subscription{
-		ID:      id,
-		Owner:   dssmodels.Owner(*req.Auth.ClientID),
-		URL:     string(req.Body.UssBaseUrl),
-		Version: version,
-		Writer:  s.Locality,
+	url := string(req.Body.UssBaseUrl)
+	if !s.AllowHTTPBaseUrls {
```
mickmis, line 238: I see this check was not implemented before. Good spot. 
Actually move it into `NewPutSubscriptionPayload`? Like this it is implemented once instead of 4 times, removing the twos from `CreateSubscription`. I think it's fine passing `allowHTTPBaseURLs`  to the function.

mickmis (approved):
LGTM

PR #1690, opened 2026-09-09
[scd] Return conflict without error

MariemBaccari:
Following feedback on #1655, this PR returns only the airspace conflict response and drops the missing OVN error to simplify the upsert opintent path. 

Implements #1701


mickmis (approved):
(0c04bf8cf71b536f3de8ab60a37dd91ce2ae2ff4)
LGTM

PR #1675, opened 2026-08-28
[raft/rid] Implement subscriptions repo methods

MariemBaccari:
Implements the rid subscriptions repo methods for the raftstore. The raftstore repo methods issue individual proposals.

Implements #1700

PR #1691, opened 2026-09-11
[raft] Fix start action missing required context

the-glu:
This PR follow #1675 , it does fix missing context preventing core-service to start.

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -16,10 +16,6 @@ const (
 	// larger than the max area allowed. See geo/s2.go.
 	AreaTooLarge = stacktrace.ErrorCode(iota)
 
-	// MissingOVNs is the error to signal that an AirspaceConflictResponse should
-	// be returned rather than the standard error response.
-	MissingOVNs
-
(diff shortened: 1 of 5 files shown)


MariemBaccari:
Implements #1700

mickmis (approved):
I think we can expect needing this for any call to a repository that is not subsequent to a client request. Here it is at the start of the service, but I think e.g. the cleanup tool will also be problematic here (https://github.com/interuss/dss/blob/b40d5c1c622982ed2267e93ea9f1eb1dce22a4c1/cmds/db-manager/cleanup/evict.go). So please include this one as well.
It may be worth factoring away somewhere a function like `NewServerContext` but I'm not sure so up to you.

Otherwise LGTM

MariemBaccari:
@mickmis The evict tool doesn't currently work with the raftstore (it still uses `dssstore.NewFuncOperation`). Its implementation will be tracked in #1718 so I suggest we include it in a follow-up PR. Moreover, by then, I think the context part will have changed a lot (we will probably remove the seed since we want to pre-compute as much values as we can).

Issue #1701, opened 2026-09-17
[raft][gate 0] Refactor transact and operations business logic


MariemBaccari:
This issue keeps track of the refactoring of the `Transact` method of the store and of the business logic of operations to remove it from the handlers into a dedicated package to ensure that proposals execute in the same way across nodes. More details on this decision can be found in #1577.

PR #1711, opened 2026-09-21
[scd] Make repo search/notify methods cells-native

MariemBaccari:
Following this [discussion](https://github.com/interuss/dss/pull/1670#discussion_r4036802473) under #1670, this PR makes the SCD search and increment repo methods take a cells based volume instead of the current `*dssmodels.Volume4D`. This way, the backends no longer recompute the S2 covering themselves and this is done at the http handler level. Eventually, this will allow us to pre-compute the S2 cells covering before proposing to Raft to avoid floating point arithmetics related to geometry on the apply path. 
The PR also adds a missing `sort` in `precomputedCellGeometry.CalculateCovering()`.

Implements #1701


pkg/models/geo.go
```diff
@@ -75,6 +76,9 @@ func (pcg precomputedCellGeometry) CalculateCovering() (s2.CellUnion, error) {
 		idx++
 	}
 
+	// Map iteration order is random and some CellUnion methods like Contains require a sorted union.
+	sort.Slice(result, func(i, j int) bool { return result[i] < result[j] })
```
mickmis: Sounds like this should be a guarantee of `CalculateCovering`. If you agree can you split that up in a separate PR, implement it in other implementations, and add this to the interface doc?
In addition, this should be covered by unit tests.

pkg/scd/operations/constraint.go
```diff
@@ -124,25 +124,35 @@ func executePutConstraint(ctx context.Context, repo repos.Repository, request ds
...
+			AltitudeHi: validParams.uExtent.SpatialVolume.AltitudeHi,
+		}
 	} else {
 		oldVol4 := &dssmodels.Volume4D{
 			StartTime: old.StartTime,
 			EndTime:   old.EndTime,
 			SpatialVolume: &dssmodels.Volume3D{
 				AltitudeHi: old.AltitudeUpper,
 				AltitudeLo: old.AltitudeLower,
-				Footprint: dssmodels.GeometryFunc(func() (s2.CellUnion, error) {
```
mickmis, line 138: Looks like this `GeometryFunc` is actually not required anymore with the change in this PR, remove it entirely?

pkg/scd/operations/constraint.go
```diff
@@ -124,25 +124,35 @@ func executePutConstraint(ctx context.Context, repo repos.Repository, request ds
 		version = old.Version + 1
 	}
 
-	// Compute total affected Volume4D for notification purposes
-	var notifyVol4 *dssmodels.Volume4D
+	// Compute total affected cellsVolume for notification purposes
+	var notifyCellsVol4D *dssmodels.CellsVolume4D
 	if old == nil {
-		notifyVol4 = validParams.uExtent
```
mickmis, line 130: I think `uExtent` can now be a `CellsVolume4D`. That would simplify things a bit.

pkg/scd/operations/operational_intents.go
```diff
@@ -292,36 +293,46 @@ func validateUpsertRequestAgainstPreviousOIR(
...
+// computeNotificationCellsVolume computes the volume that needs to be queried for subscriptions
 // given the requested extent and the (possibly nil) previous operational intent.
-// The returned volume is either the union of the requested extent and the previous OIR's extent, or just the requested extent
+// The returned volume covers either the union of the requested extent and the previous OIR's extent, or just the requested extent
 // if the previous OIR is nil.
-func computeNotificationVolume(
+func computeNotificationCellsVolume(
 	previousOIR *scdmodels.OperationalIntent,
-	requestedExtent *dssmodels.Volume4D) (*dssmodels.Volume4D, error) {
+	requestedExtent *dssmodels.Volume4D,
```
mickmis: Looks like `requestedExtent` could just instead be a `CellsVolume4D`. And in the caller `validParams.UExtent` is passed: this probably can just be a `CellsVolume4D` too just like for constraints.

pkg/scd/operations/operational_intents.go
```diff
@@ -527,9 +538,17 @@ func validateKeyAndProvideConflictResponse(
...
+		Cells:      params.Cells,
+		StartTime:  params.UExtent.StartTime,
+		EndTime:    params.UExtent.EndTime,
+		AltitudeLo: params.UExtent.SpatialVolume.AltitudeLo,
+		AltitudeHi: params.UExtent.SpatialVolume.AltitudeHi,
+	}
+
 	// Identify OperationalIntents missing from the key
 	var missingOps []*scdmodels.OperationalIntent
-	relevantOps, err := r.SearchOperationalIntents(ctx, params.UExtent)
```
mickmis, line 532: Same comment as the previous, but `UExtent` can be a `CellsVolume4D`.

pkg/scd/operations/subscription.go
mickmis: Consider adding a helper that returns a `CellsVolume4D` to replace `Volume4DFromSCDRest`. I think that many if not all of the usages of this function could be replaced, and if so, remove that function.

pkg/scd/store/memstore/subscriptions.go
```diff
@@ -30,33 +30,27 @@ func (rec *subscriptionRecord) toModel() *scdmodels.Subscription {
...
 }
 
-// subscriptionsInVolume4D yields the subscriptions intersecting v4d.
-func (r *repo) subscriptionsInVolume4D(v4d *dssmodels.Volume4D) (iter.Seq[*subscriptionRecord], error) {
-	want, err := coveringSet(v4d)
-	if err != nil {
-		return nil, err
-	}
+// subscriptionsIncellsVolume yields the subscriptions intersecting cellsVolume.
+func (r *repo) subscriptionsIncellsVolume(cellsVolume *dssmodels.CellsVolume4D) iter.Seq[*subscriptionRecord] {
```
mickmis:
```suggestion
func (r *repo) subscriptionsInCellsVolume4D(cellsVolume *dssmodels.CellsVolume4D) iter.Seq[*subscriptionRecord] {
```

mickmis (approved):
I think it's a great simplification, that can actually be taken a step further, see my comments. At the end of the day I don't think the DSS ever has to process the geometries of the Volume4D directly?
So LGTM modulo comments.

pkg/scd/operations/subscription.go
MariemBaccari: Added the helper which indeed replaces most of the usages. `Volume4DFromSCDRest` will probably be removed in follow-up PRs when the remaining of the usages move to the new paradigm.

pkg/models/geo.go
MariemBaccari: Agreed, opened #1722 to address this :)

pkg/scd/models/operational_intents.go
```diff
@@ -110,17 +106,18 @@ func (o *OperationalIntent) ToRest() *restapi.OperationalIntentReference {
 		UssAvailability: o.UssAvailability.ToRest(),
 	}
 
-	if o.StartTime != nil {
-		result.TimeStart = restapi.Time{
-			Value:  o.StartTime.Format(time.RFC3339Nano),
-			Format: TimeFormatRFC3339,
+	if o.CellsVolume4D != nil {
```
mickmis, line 109: Since CellsVolume4D is embedded, you don't actually need to check for it being nil when accessing its fields. Demo: https://go.dev/play/p/0g1bLtymQNw
Applies to all others too

pkg/scd/operations/constraint.go
```diff
@@ -124,25 +123,10 @@ func executePutConstraint(ctx context.Context, repo repos.Repository, request ds
...
-				Footprint: dssmodels.GeometryFunc(func() (s2.CellUnion, error) {
-					return old.Cells, nil
-				}),
-			}}
-		notifyVol4, err = dssmodels.UnionVolumes4D(validParams.uExtent, oldVol4)
-		if err != nil {
-			return nil, stacktrace.Propagate(err, "Error constructing 4D volumes union")
-		}
+	// Compute total affected cellsVolume for notification purposes
+	notifyCellsVol4D := validParams.volume
```
mickmis, line 127: nice

pkg/scd/operations/operational_intents.go
```diff
@@ -145,18 +145,7 @@ func executeDeleteOperationalIntentReference(ctx context.Context, repo repos.Rep
...
-		SpatialVolume: &dssmodels.Volume3D{
-			AltitudeHi: old.AltitudeUpper,
-			AltitudeLo: old.AltitudeLower,
-			Footprint: dssmodels.GeometryFunc(func() (s2.CellUnion, error) {
-				return old.Cells, nil
-			}),
-		}}
-
-	subsToNotify, err := repo.IncrementNotificationIndicesForOperationalIntents(ctx, notifyVolume)
+	subsToNotify, err := repo.IncrementNotificationIndicesForOperationalIntents(ctx, old.Clone())
```
mickmis, line 148: IS this `Clone()` really necessary? It would be surprising for the underlying implementation to mutate `old`. I would avoid doing that unless warranted (and if it is, it should be explicitly disclosed in the function doc of IncrementNotificationIndicesForOperationalIntents, but I don't see why it should).

pkg/scd/operations/operational_intents.go
```diff
@@ -292,45 +281,27 @@ func validateUpsertRequestAgainstPreviousOIR(
 	return nil
 }
 
-// computeNotificationVolume computes the volume that needs to be queried for subscriptions
+// computeNotificationCellsVolume computes the volume that needs to be queried for subscriptions
 // given the requested extent and the (possibly nil) previous operational intent.
-// The returned volume is either the union of the requested extent and the previous OIR's extent, or just the requested extent
+// The returned volume covers either the union of the requested extent and the previous OIR's extent, or just the requested extent
 // if the previous OIR is nil.
-func computeNotificationVolume(
```
mickmis, line 299: At this point, might as well inline the function? It is called only once.

pkg/models/geo.go
```diff
@@ -170,6 +167,95 @@ func UnionVolumes4D(volumes ...*Volume4D) (*Volume4D, error) {
 	return result, nil
 }
 
+// UnionCellsVolumes4D unions cells volumes and returns a volume that covers all the individual volumes in space (cells) and time.
+func UnionCellsVolumes4D(volumes ...*CellsVolume4D) *CellsVolume4D {
```
mickmis, line 171: This is worthy of some unit tests.

mickmis (approved):
LGTM, just a few minor comments. This is a nice improvement!

pkg/scd/models/operational_intents.go
MariemBaccari: Awesome ! I didn't know about this :)

pkg/scd/operations/operational_intents.go
MariemBaccari: It's not necessary, thanks for catching it

pkg/models/geo.go
MariemBaccari: Added some tests. A lot of them mirror the volume 4d union method whose implementation is very similar. They will be removed in follow-up PRs :)

PR #1712, opened 2026-09-21
[raft/scd] Precompute S2 cells for put subscription

MariemBaccari:
Precomputes the S2 covering for `PutSubscription` once in the http handler and pass down a `putSubscriptionPayload` with the necessary information instead of the full http request.

Implements #1701


pkg/scd/operations/subscription.go
```diff
@@ -44,29 +45,27 @@ func init() {
...
-	}
+type putSubscriptionPayload struct {
+	Op           string
+	Subscription *scdmodels.Subscription
+}
 
+func (p *putSubscriptionPayload) OperationID() string { return p.Op }
+
+// NewPutSubscriptionPayload performs the request validation that can be done ahead of the transaction.
+func NewPutSubscriptionPayload(opID string, subscriptionid restapi.SubscriptionID, manager dssmodels.Manager, version string, params *restapi.PutSubscriptionParameters, allowHTTPBaseUrls bool) (*putSubscriptionPayload, error) {
```
mickmis: I would suggest some decoupling here by not letting the caller define the `opID`. We could have this generic function unexposed and two functions exposed:
```go
func NewCreateSubscriptionPayload(subscriptionid restapi.SubscriptionID, manager dssmodels.Manager, params *restapi.PutSubscriptionParameters, allowHTTPBaseUrls bool) (*putSubscriptionPayload, error) // and no need for version param here 

func NewUpdateSubscriptionPayload(subscriptionid restapi.SubscriptionID, manager dssmodels.Manager, version string, params *restapi.PutSubscriptionParameters, allowHTTPBaseUrls bool) (*putSubscriptionPayload, error)
```

pkg/scd/operations/subscription.go
```diff
@@ -44,29 +45,27 @@ func init() {
...
-
-	switch req := request.(type) {
-	case *restapi.CreateSubscriptionRequest:
-		manager, subscriptionid, params = *req.Auth.ClientID, req.Subscriptionid, req.Body
-	case *restapi.UpdateSubscriptionRequest:
-		manager, subscriptionid, version, params = *req.Auth.ClientID, req.Subscriptionid, req.Version, req.Body
-	default:
-		return nil, stacktrace.NewError("unexpected request type %T for operation %q", request, restapi.CreateSubscriptionOperationID)
-	}
+type putSubscriptionPayload struct {
```
mickmis: Doing so means we will by carrying around the `Op` unnecessarily. Why not simply two structs:
```go
type createSubscriptionPayload struct {
	Subscription *scdmodels.Subscription
}
func (p *createSubscriptionPayload) OperationID() string { return restapi.CreateSubscriptionOperationID }

type updateSubscriptionPayload struct {
	Subscription *scdmodels.Subscription
}
func (p *updateSubscriptionPayload) OperationID() string { return restapi.UpdateSubscriptionOperationID }
```

pkg/scd/operations/subscription.go
mickmis: Also: consider returning just a `OperationRequest`, that's all the caller needs.

mickmis (approved):
LGTM modulo comments

pkg/scd/models/conversions.go
```diff
@@ -75,17 +77,122 @@ func UnionVolumes4DFromSCDRest(vol4s []restapi.Volume4D, validators ...Volume4DV
 	return union, nil
 }
 
-// CellsVolume4DFromSCDRest converts vol4 SCD v1 REST model to a CellsVolume4D.
 func CellsVolume4DFromSCDRest(vol4 *restapi.Volume4D) (*dssmodels.CellsVolume4D, error) {
-	volume, err := Volume4DFromSCDRest(vol4)
+	filter := &dssmodels.CellsVolume4D{}
```
MariemBaccari, line 81: Note that this is copied from the `Volume4DFromSCDRest` implementation which will eventually be deleted in follow-up PRs when we get rid of the Volume4D intermediate in dss models.

mickmis: Temporary duplication is fine, please just mention it with a TODO comment in doc of functions to be removed.

pkg/scd/models/conversions.go
```diff
@@ -75,17 +77,122 @@ func UnionVolumes4DFromSCDRest(vol4s []restapi.Volume4D, validators ...Volume4DV
...
+	case vol3.OutlinePolygon != nil:
+		return polygonCellsFromSCDRest(vol3.OutlinePolygon)
+	case vol3.OutlineCircle != nil:
+		return circleCellsFromSCDRest(vol3.OutlineCircle)
+	default:
+		return nil, geo.ErrMissingFootprint
+	}
+}
+
+func circleCellsFromSCDRest(c *restapi.Circle) (s2.CellUnion, error) {
```
mickmis, line 163: This is duplicating `func (gc *GeoCircle) CalculateCovering() (s2.CellUnion, error)`. Is this intended? Is your goal to also get rid of those? Fine by me, but please just ensure there is a TODO comment on doc of functions to be removed, and also that the cells are normalized (c.f. other PR).

mickmis (approved):
LGTM

pkg/scd/models/conversions.go
MariemBaccari: Yes the goal is to remove them (#1714). I added TODO comments for that. 
The cells returned by `circleCellsFromSCDRest` are actually denormalized for now tho. I will track the normalization issue but it won't be implemented in this chain of PRs yet.

PR #1713, opened 2026-09-21
[raft/scd] Precompute S2 cells for put constraint

MariemBaccari:
Precomputes the S2 covering for `PutConstraint` once in the http handler and passes down a `putConstraintPayload` with the necessary information instead of the full http request.

Implements #1701


pkg/scd/operations/constraint.go
```diff
@@ -75,34 +74,76 @@ func executeGetConstraint(ctx context.Context, repo repos.Repository, request ds
...
-	case *restapi.CreateConstraintReferenceRequest:
-		manager, entityid, params = *req.Auth.ClientID, req.Entityid, req.Body
-	case *restapi.UpdateConstraintReferenceRequest:
-		manager, entityid, ovn, params = *req.Auth.ClientID, req.Entityid, req.Ovn, req.Body
-	default:
-		return nil, stacktrace.NewError("unexpected request type %T for operation %q", request, restapi.CreateConstraintReferenceOperationID)
+func (p *putConstraintPayload) OperationID() string { return p.Op }
+
+// NewPutConstraintPayload performs the request validation that can be done ahead of the transaction.
+func NewPutConstraintPayload(opID string, entityid restapi.EntityID, ovn restapi.EntityOVN, manager dssmodels.Manager, params *restapi.PutConstraintReferenceParameters, allowHTTPBaseUrls bool, now time.Time) (*putConstraintPayload, error) {
```
mickmis: Same comments as for #1712: 
- do not let caller define `opID`
- rather define two structs here instead of `putConstraintPayload`
- that also avoids carrying `Op` unnecessarily
- and define two functions New(Create|Update)ConstraintPayload

pkg/scd/operations/constraint.go
```diff
@@ -75,34 +74,76 @@ func executeGetConstraint(ctx context.Context, repo repos.Repository, request ds
...
+	id, err := dssmodels.IDFromString(string(entityid))
+	if err != nil {
+		return nil, stacktrace.NewError("Invalid ID format: `%s`", entityid)
+	}
+
+	if len(params.UssBaseUrl) == 0 {
+		return nil, stacktrace.NewError("Missing required UssBaseUrl")
+	}
```
mickmis: This can be dropped, ValidateUSSBaseURL will catch that anyway.

mickmis (approved):
LGTM modulo comments

pkg/scd/models/conversions.go
mickmis, line 1: I assume those are also replacing some other functions? 
-> Add TODOs on the functions to be cleaned up

mickmis (approved):

pkg/scd/models/conversions.go
MariemBaccari: Yes !

mickmis (approved):

PR #1714, opened 2026-09-21
[raft/scd] Precompute S2 cells for put operational intent reference

MariemBaccari:
Precomputes the S2 covering for `PutOperationalIntent` once in the http handler and passes down a custom payload with the necessary information instead of the full http request.

Implements #1701


pkg/scd/operations/operational_intents.go
```diff
@@ -497,6 +511,29 @@ func ValidateAndReturnOIRUpsertParams(
...
+type putOIRPayload struct {
+	Op      string
+	Manager dssmodels.Manager
+	Params  *validOIRParams
+}
+
+func (p *putOIRPayload) OperationID() string { return p.Op }
+
+// NewPutOIRPayload performs the request validation that can be done ahead of the transaction.
+func NewPutOIRPayload(opID string, entityid restapi.EntityID, ovn restapi.EntityOVN, params *restapi.PutOperationalIntentReferenceParameters, auth *api.AuthorizationResult, allowHTTPBaseUrls bool, now time.Time) (*putOIRPayload, error) {
```
mickmis: Same comments as for #1712 and #1713 :

- do not let caller define opID
- rather define two structs here instead of putOIRPayload
- that also avoids carrying Op unnecessarily
- and define two functions New(Create|Update)OIRPayload

pkg/scd/operations/operational_intents.go
mickmis: I think we can do without any `Volume4D` here? `previousOIR` only contains the cells anyway.

mickmis (approved):
LGTM modulo comments

pkg/scd/operations/operational_intents.go
```diff
@@ -452,6 +443,72 @@ func ValidateAndReturnOIRUpsertParams(
...
+	manager, validParams, err := newOIRUpsert(entityid, ovn, params, auth, allowHTTPBaseUrls, now)
+	if err != nil {
+		return nil, err
+	}
+	return &updateOIRPayload{Manager: manager, Params: validParams}, nil
+}
+
+// newOIRUpsert performs the request validation that can be done ahead of the transaction.
+func newOIRUpsert(entityid restapi.EntityID, ovn restapi.EntityOVN, params *restapi.PutOperationalIntentReferenceParameters, auth *api.AuthorizationResult, allowHTTPBaseUrls bool, now time.Time) (dssmodels.Manager, *validOIRParams, error) {
+	validParams, err := ValidateAndReturnOIRUpsertParams(now, entityid, ovn, params, allowHTTPBaseUrls)
```
mickmis: Can ValidateAndReturnOIRUpsertParams and CheckUpsertPermissionsAndReturnManager be unexposed now?

mickmis (approved):
LGTM

pkg/scd/operations/operational_intents.go
MariemBaccari: Yes :)

